PR #1, opened 2014-07-28
[WIP] Improve the IDE compatibility

leofeyer:
The PR includes some exemplary changes to improve the IDE compatibility, which @aschempp has shown me at the Austriaday. @contao/developers I don't know if it were worth the effort to add this kind of information everywhere?

During my tests, I have noticed two limitations (there might be more):

```php
 * @return Collection|null The model collection or null
```

This should be replaced with `@return static`, so the IDE can e.g. return `MemberModel` instead of just `Model`. However, this does only work with `@return static`, not with `@return static|null`, although the latter is actually the correct one. Does the IDE compatibility weigh more in this case?

```php
/** @var Database_Result $result */
$result = $db->query("SELECT * FROM tl_content WHERE id=2");

echo $result->headline;
echo $result->text;
```

The IDE will always show the "property accessed via magic method" notice in this case, because it is impossible for the `Database_Result` class to know its properties.


leofeyer:
Regarding the `@return static|null` issue: there is a [ticket](https://youtrack.jetbrains.com/issue/WI-19018) in the PHPStorm tracker dated June 28th, 2013. Even though it has been verified, nothing has changes since then.

I therefore vote +1 for the following two workarounds:

1. Use `@return \Database\Result|object` to suppress the "field accessed via magic method" errors.
2. Use `@return static` instead of `@return static|null` to make PHPStorm recognize the object.

aschempp:
> I therefore vote +1 for the following two workarounds:
> - Use @return \Database\Result|object to suppress the "field accessed via magic method" errors.
> - Use @return static instead of @return static|null to make PHPStorm recognize the object.

+1 from me too

leofeyer:
I found a different fix for the database issue: `class Result extends \stdClass`. This way we can keep the phpDoc comments as is.

Too bad we cannot do the same for the `Template` classes, because they already extend the controller.

aschempp:
I guess we should do this as less as possible anyways. I started using `@property` in models for Isotope to declare the available properties (e.g. https://github.com/isotope/core/blob/master/system/modules/isotope/library/Isotope/Model/Payment.php#L25-L46). This obviously won't work for custom attributes, but it does at least for the core DCA fields.

Ideally, we could also develop a PhpStorm plugin that does support model properties based on the DCA files. I guess that would be technically possible (but challenging) if it's possible to add support for type resolution of a Symfony container ;-)

discordier:
Evil hack, let System extend stdClass?

leofeyer:
@discordier Thought about it, too :)

aschempp:
> Evil hack, let System extend stdClass?

It wouldn't really be correct, because `System` and `Controller` do not allow arbitrary parameters.

ausi:
What if someone wants to check if an object is a plain PHP object with `$object instanceof stdClass`?

ausi:
For example: http://php.net/manual/de/reserved.classes.php#113971

leofeyer:
I found a better solution:

```php
class Template
{
	…

	/**
	 * Create a new template object
	 *
	 * @param string $strTemplate    The template name
	 * @param string $strContentType The content type (defaults to "text/html")
	 *
	 * @return static|object The template object
	 */
	public static function create($strTemplate='', $strContentType='text/html')
	{
		return new static($strTemplate, $strContentType);
	}

	…
```

Now if you use the factory method instead of `new`, PhpStorm will handle the object correctly.

```php
$this->Template = FrontendTemplate::create('nav_default');
```

aschempp:
Well the problem is also that most of the time we use `$this->Template`. So the controller class should probably define this as a property and define the type…

ausi:
As I already mentioned in https://github.com/contao/core-bundle/pull/1#issuecomment-52027843 IMO `@return static|object` is wrong.

Toflar:
Moreover, adding or removing code just for one specific IDE autocompletion feature imho feels really, really, really wrong! That applies for both, a useless factory method as well as extending `stdClass` for no reason.

ausi:
@Toflar +1

leofeyer:
IDE compatibility is the least we can do to make developers life easier, given the fact that a complete refactoring will take years :)

But why don't you wait until I make the PR, then we can discuss further.

leofeyer:
The PR is now available: #7634

As discussed, I have removed the `extends \stdClass` part and I have removed the `Template::create()` factory method.

PR #31, opened 2015-02-23
[WIP] Use Doctrine for database connection

aschempp:
Implements https://github.com/contao/contao/issues/21


leofeyer:
Please add `@todo` comments in the phpDoc or `// FIXME` comments in the code wherever we need to fix or optimize things.

Toflar:
ToDo:

* [ ] Remove doctrine config from `contao/contao` `config.yml` because it's now in the core bundle!

leofeyer:
As discussed in Mumble on April 8th, we should stick with the `fetchAll()` solution for now and try to optimize it later.

leofeyer:
I am working on this PR today and I don't like that we are still supporting passing a custom config array to the `Database::getInstance()` method. Can we change this as follows:

```php
public static function getInstance($strConnection='doctrine.dbal.default_connection')
{
    $this->resConnection = $kernel->getContainer()->get(strConnection);
}
```

leofeyer:
Also, since Doctrine provides the DBAL in the future, there is no need to have distinct driver classes. I'd therefore integrate everything directly into the `Database`, `Database\Statement` and `Database\Result` classes. Any objections?

aschempp:
That would be a BC break for no reason, because the current implementation works…

Toflar:
I'm with @aschempp. No need to do that. I prefer a complete BC break in the future which will be removing `Database` all together :)

leofeyer:
What are you referring to? Integrating stuff into the base classes or passing the connection ID?

leofeyer:
The `Database` class still contains two deprecated methods `executeCached()` and `executeUncached()`. Shall we remove them now or in Contao 5?

aschempp:
I would remove them in Contao 5. I'm sure people are still using executeUncached (I am, too). We should add the deprecated version info though.

leofeyer:
By the way, the implementation in this PR is pretty confusing and does **not** work. Try the following:

```php
$this->Database->isUniqueValue('tl_user', 'username', 'k.jones')
```

I'm reworking it and will create a PR for reviewing.

leofeyer:
See #182.

PR #40, opened 2015-02-24
[RTM] Removed Contao database debug info from web profiler

aschempp:
This is to be merged when the doctrine feature is available (see https://github.com/contao/core-bundle/pull/31)


leofeyer:
Do we still need this? The web profiler already provides the total time and the memory usage.

aschempp:
This is to remove the database info once the Doctrine driver is merged. Memory and runtime has already been removed in #161

PR #44, opened 2015-02-24
Late bootstrap legacy code.

discordier:
Sister PR of contao/contao#55


discordier:
@Toflar I am currently writing the unit tests.
Just to let you know so you won't start doing the same.

Toflar:
:+1:

discordier:
Ok, the crucial parts are done.
The base includes have been moved to core-bundle now and are properly unit tested.

I had to blow up the bootstrap for unit tests with several more classes and change the bootstrap include itself to only load classes if not defined yet as otherwise the thing would simply be impossible to test at all.

@leofeyer, @aschempp, @Toflar please review.

src/EventListener/BootstrapLegacyListener.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Boot the legacy code for all web SAPIs.
+     *
+     * @param GetResponseEvent $event The event object
+     *
+     * @throws RouteNotFoundException When no route could be found for the request.
+     */
+    public function onBootLegacyForRequest(GetResponseEvent $event)
```
aschempp: Afaik in the other listeners we named the method after the event (e.g. `onKernelResponse`).

Toflar: But we have two methods so this is perfectly fine.

discordier: So was I thinking.
We might want to add a phpDoc stating that it will get called on the event `kernel.request` but this might be superfluous information.

If we end up naming the method after the event, we must also change: `onBootLegacyForConsole` to `onConsoleCommand` which I personally think of misleading.

aschempp:
I would also suggest to move the exceptions to real exception classes (because they are also used in non-legacy code) and possibly use `class_alias` for BC.

Toflar:
I like that idea.

discordier:
I already thought about it and would like to have them in a separate namespace but they are legacy code only and I wonder if it will confuse developers.
`UnusedArgumentsException` is only used in `FrontendTemplate` in legacy code and `UnresolvableDependenciesException` is only being catched in `bootstrap.php` but not thrown anymore from anywhere.
Therefore they can savely get phased out within the next major.

The `interfaces.php` could also be split up but we should discuss this first as we should rather focus on implementing Symfony Form based replacements for the `DataContainer` classes than beautifying the legacy stuff.

To put long story short, IMO both things are beyond the scope of this PR.

aschempp:
You're absolutely right, I must have mixed up something. Let's leave it like that, but we should deprecate at least `UnresolvableDependenciesException` with a DocBlock.

discordier:
I agree on deprecating the unused classes but this also does not belong here. ;)

leofeyer:
I have merged the PR into a feature branch, but there are a few things I don't like.

1. I really, really don't want to add all the defensive programming features. It was not possible to define constants or class aliases in Contao 3 and I see no reason why we should add a new feature to the legacy system.

2. Why did you define all the class aliases in the `tests/bootstrap.php` file?

3. Moving the bootstrap logic into the event listener is not yet complete. Can I complete it or is there another PR somewhere? Also, why don't we simply include the bootstrap file in the event listener? Then we could save us the work.

leofeyer:
If you don't mind, I would prefer to rename every occurrence of "legacy" (e.g. "boot Contao" instead of "boot legacy"). Just to get rid of the word "legacy".

discordier:
1. All the defensive means only the constants which might get predefined or are you talking about any other defensive approach? In general, defensive programming is ALWAYS a good idea and we are writing new code here which is booting the legacy code. But as long as everything (in the new code) remains testable I do not stick to having this defensive behaviour.
2. As all the legacy classes are referring to each other via the root namespace, the here loaded classes would not be defined at all as they are only being loaded into the Contao namespace. As we do not provide empty stubs for all legacy classes but only the "root classes", we have to alias them to the root namespace as new code refers to them from `\Contao\XY` where legacy code refers to `\XY`. This now comes into play as we are instantiating the `ModuleLoader` etc. in `bootstrap.php` which need them.
3. The way to go from here is up to you. Either we move the stuff back to bootstrap.php or move the remaining stuff from bootstrap into the class. Feel free to do as you wish.

On your last comment renaming Legacy to Contao... I think this will lead up to confusion.
You have your project `contao/contao` which is called `Contao 4` and in there now is a listener called `bootContao` which does not boot Contao but boots `Contao 3`. The real booting of `Contao 4` happens in the Kernel boot.
Furthermore, when we will phase out Contao 3 compat with Contao 5, the listener will get removed and people will then think there is no way to boot contao anymore? ;)

If you really dislike the word `legacy` (which we are using in all occations referring to Contao 3 so far), it might be better to rename it `bootLegacy` => `bootContaoThree` or something like that.

leofeyer:
Regarding 2) I have just fixed the issue described in the `AutomatorCommand` class by simply changing the `Automator` class:

```php
namespace Contao;

class Automator extends System { // <-- used to be \System
}
```

Since we are not officially supporting the Contao 3 class loader anymore, would that be an option to get rid of the class aliases? Or do you prefer to have the aliases?

leofeyer:
> On your last comment renaming Legacy to Contao... I think this will lead up to confusion.

I agree with you. How about `initializeSystem` then? This would at least use the wording of Contao 3.

leofeyer:
I have created a follow-up PR with my changes: #52

leofeyer:
Merged in a05737631e4627d56513f21110724d320c161b3f now.

PR #45, opened 2015-02-25
Simple security

aschempp:
See https://github.com/contao/contao/issues/44


leofeyer:
Is it correct that the debug bar says "authenticated" although we are anonymous?

aschempp:
yes that's correct, we've discussed that. Compare it to a building access control. Either you have a badge with your name, or you get a badge with "visitor". The access control always "authenticates" you. Your badge just does not have permission to open any doors.

leofeyer:
Makes sense. Does Symfony follow this logic?

aschempp:
Not sure what you mean with *follow this logic*. For regular bundles its a bit different because they do not prepend a configuration. They merely parse your `app/config/config.yml` configuration to set up the container (e.g. create services for stuff you configured).

leofeyer:
What does the Symfony debug console show if you are not authenticated? Does it also show "authenticated" and "anon."?

aschempp:
Well it obviously depends :D

1. If the Security Bundle is not loaded (in AppKernel), there is no information about users
2. If there is no firewall for your current route, you are not authenticated
3. If there is a firewall but you are not logged in, it shows "authenticated" and "anon."

Case 3 exactly applies to us, so we are following Symfony "standards" here.

leofeyer:
Ok, so obviously your answer to my initial question is "yes", which is good :)

PR #46, opened 2015-02-25
Always enable twig bundle, but only load config if bundle is available

aschempp:


discordier:
How come the service kernel.bundles get's retrieved in here but is not available?
Didn't you say in Ahrweiler that it is present?

aschempp:
Not sure what you mean. `kernel.bundles` is an array with key="bundle name" and value="bundle class". I explained that the bundle instances are not available.

aschempp:
If you are referring to the unit tests: I don't know. They are set in the kernel:
https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpKernel/Kernel.php#L579

leofeyer:
Do we really need to wrap all the calls in `if` conditions? We are loading the bundles in the `autoload.json` file, so we can safely assume that the bundles are available, can't we?

aschempp:
It's similar to what Symfony does: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L77

The `autoload.json` does not make sure the bundles are loaded, it just loads them if they are there.

leofeyer:
> The `autoload.json` does not make sure the bundles are loaded

That's true, but Composer ensures that the bundles are there, so they are always loaded :)

aschempp:
I still think its good practice, otherwise the container building will fail, and even our custom error messages wont be rendered.

leofeyer:
Can you please update/rebase this PR so it handles the `security.yml` file? Then I should be able to merge it.

aschempp:
Done, including some fixes.

leofeyer:
How about renaming `prependBundleConfig()` to `prependConfigIfLoaded()`?

aschempp:
hmm, then it should be `prependConfigIfBundleIsLoaded`? :D
Because it's about the given bundle…

leofeyer:
Not quite, because the first argument would be the bundle name:

```php
$this->prependConfigIfLoaded('DoctrineBundle', 'doctrine.yml', $container);
```

But we might as well use `prepareBundleConfig()`.

Toflar:
To me it looks good as it is now. The method is well documented saying "if loaded" and honestly, `prependConfigIfLoaded` looks a bit awkward :D

leofeyer:
Another thing: We are now always loading the web profiler, aren't we? Previously we only loaded it in the `dev` and `test` environments (whereby `test` does not make sense at all).

aschempp:
No, we're only loading it if the bundle is there. And the bundle is only there in `dev`/`test` according to autoload.json.

leofeyer:
I see (now).

PR #50, opened 2015-02-27
Routing fixes

aschempp:
Fixes:
- backend access attempt when not logged in
- preview bar

### DIFF
--- a/contao/classes/BackendUser.php
+++ b/contao/classes/BackendUser.php
@@ -226,30 +226,31 @@ class BackendUser extends \User
 			return true;
 		}
 
-		list($path) = explode('?', \Environment::get('request'), 2);
+		/** @var KernelInterface $kernel */
+		global $kernel;
 
-		if (substr($path, -13) == '/contao/login')
+		$container = $kernel->getContainer();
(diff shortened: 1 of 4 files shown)


contao/classes/BackendUser.php
```diff
@@ -223,22 +226,32 @@ public function authenticate()
...
 
 		// Redirect to the last page visited upon login
-		if (substr($path, -7) == '/contao' || substr($path, -15) == '/contao/preview')
+		if ($route == 'contao_backend' || $route == 'contao_backend_preview')
 		{
-			$strRedirect .= '?referer=' . base64_encode(\Environment::get('request'));
+			$parameters['referer'] = base64_encode(\Environment::get('request'));
 		}
 
-		\Controller::redirect($strRedirect);
```
Toflar: Shouldn't we use a `ResponseException` here?

aschempp: That's not yet available in this branch, the exception handling is a follow-up to this.

Toflar: Agreed, but then you should add a FIXME :)

PR #52, opened 2015-03-02
Follow-up on 'Late bootstrap legacy code. #44'

leofeyer:
This PR includes my changes to #44. Aside the fact that the code has not yet been optimized, it is now in a mergeable state in my opinion. @contao/developers Please review and vote.


aschempp:
Is there a corresponding PR to contao/contao? I guess it needs an update to the initialize.php?

leofeyer:
Yes, contao/contao#55

leofeyer:
However, I have renamed `add_language_to_url` to `prepend_locale`.

contao/helper/exception.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class UnusedArgumentsException extends Exception {}
+
+
+/**
+ * Class UnresolvableDependenciesException
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class UnresolvableDependenciesException extends Exception {}
```
aschempp: We should add `@deprecated` here.

src/EventListener/InitializeSystemListener.php
```diff
@@ -0,0 +1,257 @@
...
+
+        // Override the SwiftMailer defaults
+        \Swift::init(function() {
+            $preferences = \Swift_Preferences::getInstance();
+            $preferences->setTempDir(TL_ROOT . '/system/tmp')->setCacheType('disk');
+            $preferences->setCharset(Config::get('characterSet'));
+        });
+
+        // Define the relative path to the installation (see #5339)
+        if (Config::has('websitePath') && TL_SCRIPT /* FIXME */ != 'contao/install.php') {
```
aschempp: The FIXME should be on the whole if/else block not just this line, because the elseif is also matching a PHP script.

aschempp:
Looks good so far, except that you removed `defineIfNotDefined` ;-)

I also reviewed the option to remove/merge the `ContainerScopeListener`, but I think it should stay standalone as it is not related to the "legacy" system initialization.

contao/helper/exception.php
Toflar: Only if there's an alternative - is there? :)

src/EventListener/InitializeSystemListener.php
```diff
@@ -0,0 +1,257 @@
...
+        if (file_exists(TL_ROOT . '/system/config/initconfig.php')) {
+            include TL_ROOT . '/system/config/initconfig.php';
+        }
+
+        RequestToken::initialize();
+
+        // Check the request token upon POST requests
+        if ($_POST && !RequestToken::validate(Input::post('REQUEST_TOKEN'))) {
+
+            // Force a JavaScript redirect upon Ajax requests (IE requires absolute link)
```
Toflar: Even though this is legacy code I think we should get rid of the `exit;` and either throw a `ResponseException` or a `DieNicelyException`, both upcoming with @discordier's changes. So we could add a FIXME here?

Toflar:
> I also reviewed the option to remove/merge the ContainerScopeListener, but I think it should stay standalone as it is not related to the "legacy" system initialization.

I thought about that too as it is assigned to me in the todo issue. If we get rid of the bootstrap listener we still need the backend and frontend scope so it's okay to me if it's in a separate listener.

contao/helper/exception.php
aschempp: We're not using this anymore, as dependencies are handled by Composer. The class is just there for legacy compatibility if someone would have used it (for god knows why…).

leofeyer: We are in fact still using it. It can only be marked as deprecated once we deprecate the Contao class loader.

src/EventListener/InitializeSystemListener.php
leofeyer: Don't know where you see an `elseif` condition?

leofeyer:
I have added the FIXME so we don't forget to replace `exit;` with an exception. Can I merge this now?

Toflar:
Yes.

src/EventListener/InitializeSystemListener.php
```diff
@@ -0,0 +1,257 @@
...
+            $preferences = \Swift_Preferences::getInstance();
+            $preferences->setTempDir(TL_ROOT . '/system/tmp')->setCacheType('disk');
+            $preferences->setCharset(Config::get('characterSet'));
+        });
+
+        // Define the relative path to the installation (see #5339)
+        if (Config::has('websitePath') && TL_SCRIPT /* FIXME */ != 'contao/install.php') {
+            Environment::set('path', Config::get('websitePath'));
+        } elseif ('BE' === TL_MODE) {
+            Environment::set('path', preg_replace('/\/contao\/[a-z]+\.php$/i', '', Environment::get('scriptName')));
```
discordier: Right here, the preg_replace will never match anymore as no URL will be `/contao/xyz.php` anymore but only `/contao/xyz`.

aschempp: Exactly :-)

src/EventListener/InitializeSystemListener.php
discordier: I am with @Toflar on this one.
We should provide proper ResponseExceptions here.

I also wonder if the whole session handling will interfere with the Symfony session and if we should/could move the RequestToken check "outside" as firewall pass.

aschempp: We are currently not using the Symfony Session (Contao does not enable it at all). Also, the RequestToken class is part of the legacy code, so we would need to re-engineer everything for that. I think we'd rather use Symfony CSRF, but I guess that won't be compatible.

leofeyer: I have added a FIXME.

PR #53, opened 2015-03-04
Added helper methods url() and path() to template class

Toflar:
Tested with a simple `fe_page.html5` in `templates` and the following code:

```php
<?php echo $this->url('contao_share'); ?>
<br>
<?php echo $this->path('contao_share'); ?>
<br>
<?php echo $this->url('_profiler_info', array('about' => 'foobar')); ?>
<br>
<?php echo $this->path('_profiler_info', array('about' => 'foobar')); ?>
```

Returns this:
```
http://localhost/contao4/web/app_dev.php/_contao/share
../_contao/share
http://localhost/contao4/web/app_dev.php/_profiler/info/foobar
../_profiler/info/foobar
```

### DIFF
--- a/contao/library/Contao/Template.php
+++ b/contao/library/Contao/Template.php
@@ -9,6 +9,8 @@
  */
 
 namespace Contao;
+use Symfony\Component\HttpKernel\KernelInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 
 /**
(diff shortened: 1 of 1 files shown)


leofeyer:
That's great, but we have actually agreed on having just one method named `path()` or `route()` (I prefer the latter), which then returns a relative path (the route minus `TL_PATH`).

Toflar:
Well, I still don't agree :) It's better to have both to be compliant with the Twig functions `url()` and `path()`. It also eases documentation. Users will not understand if there are two methods named the same but doing different things.

leofeyer:
Yeah, except that only `url()` is usable in the Contao context, because `path()` returns a relative path which will not work with the `<base>` tag we are using.

Toflar:
Hmm, but the `<base>` tag should contain the base URL that then works together with the relative path?

Toflar:
I mean the `href` attribute should imho contain
```php
$container->get('router')->getContext()->getBaseUrl();
```

leofeyer:
Here you can see the difference:

```php
// Installation path: /var/www/contao-3

// URL: http://localhost/contao-3/
// <base href="http://localhost/contao-3/">

// generateFrontendUrl(): news/james-wilson-returns.html

// Symfony-Router path(): /contao-3/news/james-wilson-returns.html
// Symfony-Router url(): http://localhost/contao-3/news/james-wilson-returns.html
```

So what `Template::route()` has to do is:

```php
$path = $router->generate($routeName);

return str_replace(Environment::get('path') . '/', '', $path);
```

leofeyer:
See #59.

PR #54, opened 2015-03-04
Use route name and request path when initializing system

aschempp:
Now that the Contao system is initialized after routing we can match the route name.

### DIFF
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -77,7 +77,7 @@ class InitializeSystemListener
 
         $this->setConstants($mode, $this->router->generate($routeName, $request->attributes->get('_route_params')));
 
-        $this->boot();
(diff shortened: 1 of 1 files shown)


Toflar:
+1

src/EventListener/InitializeSystemListener.php
```diff
@@ -228,13 +231,17 @@ private function setSwiftMailerDefaults()
...
      * Defines the relative path to the installation (see #5339).
+     *
+     * @param string $routeName The route name
+     * @param string $basePath  The URL base path
      */
-    private function setRelativePath()
+    private function setRelativePath($routeName, $basePath)
     {
-        if (Config::has('websitePath') && '/contao/install' !== substr(TL_SCRIPT, -15)) {
+        // FIXME: why not always use the request path? Security?
```
leofeyer: No, we just didn't have a better way to do it. But now we do :)

PR #55, opened 2015-03-04
Contao resources from DIC

aschempp:
This will make the `ContaoBundleInterface` obsolete and also cache the resource paths and public folders in the DIC.

 - [x] Should return a `Finder` instance to iterate over the folder
 - [x] Rename service to `contao.resource_provider`
 - [x] Rename class to `ResourceProvider`
 - [x] Do not use `realpath()` (see https://github.com/contao/core-bundle/pull/55#issuecomment-77329984)


aschempp:
Be aware that each Contao bundle (e.g. calendar-bundle) needs to be adjusted. They must extend the regular bundle (instead of `ContaoBundle` and instead of the `getContaoResources` method it should contain this:

```php
/**
 * {@inheritdoc}
 */
public function build(ContainerBuilder $container)
{
    $container->addCompilerPass(
        new AddContaoResourcesPass(
            $this->getName(),
            $this->getPath() . '/../contao',
            [$this->getPath() . '/../contao/assets']
        )
    );
}
```

src/Resources/config/services.yml
```diff
@@ -68,6 +68,11 @@ services:
         class: Contao\CoreBundle\Security\User\ContaoUserProvider
         public: false
 
+    contao.resources:
```
Toflar: I think for the sake of code readability the service id should be named `contao.resources_provider`.

src/HttpKernel/Bundle/ResourcesProvider.php
```diff
@@ -0,0 +1,56 @@
...
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\HttpKernel\Bundle;
+
+class ResourcesProvider
```
discordier: IMO the class should have a singular naming `ResourceProvider` no matter how many resources it provides.
I have not seen any plural naming in any symfony project for classes like this and AFAIR we have all agreed to use singular naming throughout Contao code.

Toflar: +1

aschempp:
I have partially rewritten the provider and added a container optimization pass. The paths are no longer verified when adding to the service (which would result in FS checks whenever the service is requested). They are now verified and combined when building the container, also resulting in failing to compile the kernel if an invalid path is specified. This currently actually happens for the listing bundle because it specifies an assets folder that does not exist.

Toflar:
Very good! Remains
* [ ] Class `ResourcesProvider` should be renamed to `ResourceProvider`
* [ ] Service ID `contao.resources` should be renamed to `contao.resource_provider`

leofeyer:
I also don't like the naming, however I don't agree with you, @Toflar. The service provides Contao resources (multiple, not just a single one) and should therefore be called `contao.resources_provider`.

But the service should not have a `getResourcesPath()` method, because that is confusing IMHO.

I also see a huge problem with using `realpath()`, since it will not work with symlinks (if you e.g. use a symlink to add a repository to your vendor folder instead of moving it there). If possible, we should use `file_exists()` instead, which should also handle paths like `/var/www/contao/vendor/core-bundle/src/../contao`.

aschempp:
I have updated the PR.

Regarding `realpath`, according to the documentation is should work fine with symlinks but actually resolve them?

> realpath() expands all symbolic links and resolves references to '/./', '/../' and extra '/' characters in the input

leofeyer:
>  but actually resolve them?

And that's exactly what we don't want. For example:

```
$ ls -la system/modules
drwxr-xr-x   core
lrwxr-xr-x   isotope -> ../../../local/dev/isotope
```

Now if both modules contain DCA files, here's what the output of the resource provider looks like:

```
Array
(
    [0] => /var/www/contao/system/modules/core/dca
    [1] => /var/www/local/dev/isotope/dca
)
```

However, here's what would actually be correct:

```
Array
(
    [0] => /var/www/contao/system/modules/core/dca
    [1] => /var/www/contao/system/modules/isotope/dca
)
```

Therefore you should only use `realpath()` if you really have to. In our case, however, a `file_exists()` should work, too.

leofeyer:
I've been thinking about the PR and I would love to have the following methods in the resource provider:

 * `getDcaFiles()`
 * `getLanguageFiles()`
 * `getConfigFiles()`
 * `getTemplateFiles()`
 * `getPublicFolders()`

Then we could get rid of the `getResourcesPath()` method and we could greatly simplify the legacy code:

```php
// Currently
foreach ($kernel->getContaoBundles() as $bundle)
{
	$strFile = $bundle->getContaoResourcesPath() . '/dca/' . $strName . '.php';

	if (file_exists($strFile))
	{
		$objCacheFile->append(static::readPhpFileWithoutTags($strFile));
	}
}

// New
foreach ($kernel->getContainer()->get('contao.resource_provider')->getDcaFiles() as $strFile)
{
	$objCacheFile->append(static::readPhpFileWithoutTags($strFile));
}
```

What do you think?

leofeyer:
Also, wasn't the plan to store the paths somewhere (DIC or separate cache) so they file system searches do not have to be executed upon every request?

aschempp:
1. what's wrong on the local path?
2. I disagree with the methods. They would disallow to access other folders and "hardcode" on the known once. I think that's bad practice.
3. The resources paths are stored in the DIC, or do you mean the path to templates etc?

leofeyer:
1. Just trust me, it will not work. I have had these kind of problems quite often.
2. We could still keep the base paths, but having the concrete paths cached somewhere will save us from a lot of `file_exists()` calls.
3. No, I mean the resources paths. But now we still have the base paths stored in the DIC, which means we still have to get the resources paths, find the ones which have a `dca/` folder, get the DCA files. If we would store the four known subpaths as well (`config/`, `dca/`, `languages`, `templates/`), we could improve the runtime performance.

leofeyer:
Also, I'd prefer this:

```php
/**
 * {@inheritdoc}
 */
public function build(ContainerBuilder $container)
{
    $container->addCompilerPass(
        new AddContaoResourcesPass($this->getName(), $this->getPath() . '/../contao')
    );
    $container->addCompilerPass(
        new AddPublicFoldersPass($this->getName(), [$this->getPath() . '/../contao/assets'])
    );
}
```

Public folders are completely separate from the Contao resources. The `/web` folder as document root is a new feature which is not deprecated, therefore it should not be in the Contao resources provider at all.

Toflar:
I have some questions regarding both, the implementation and @leofeyer's comments :)

leofeyer:
So ask them :)

Toflar:
Yeah, but on the phone - it’s too complicated to write down :)

leofeyer:
Then we have to wait until Wednesday I guess?

aschempp:
Rebasing done.

contao/controllers/BackendInstall.php
```diff
@@ -731,16 +728,9 @@ protected function enableSafeMode()
...
-			$modules = array();
-
-			foreach ($kernel->getContaoBundles() as $bundle)
-			{
-				$modules[] = $bundle->getName();
-			}
+			$modules = \ModuleLoader::getActive();
 
-			if (count(array_diff($modules, array('ContaoCoreBundle', 'calendar', 'comments', 'devtools', 'faq', 'listing', 'news', 'newsletter', 'repository'))) > 0)
+			if (count(array_diff($modules, array('core', 'calendar', 'comments', 'faq', 'listing', 'news', 'newsletter'))) > 0)
```
leofeyer: This is wrong. We need to use the new bundle names here since they have all been converted.

aschempp: Nope, I basically reverted that to the old code, because `ModuleLoader` will map legacy module names (see https://github.com/contao/core-bundle/pull/55/files#diff-3c973c84d5ed12dfb4bee28f58f15e66R35)

contao/library/Contao/Automator.php
```diff
@@ -607,20 +580,18 @@ public function generateInternalCache()
...
 		global $kernel;
 
+		$resources = $kernel->getContainer()->get('contao.resource_provider');
+
 		// Generate the class/template laoder cache file
 		$objCacheFile = new \File('system/cache/config/autoload.php');
 		$objCacheFile->write('<?php '); // add one space to prevent the "unexpected $end" error
 
-		foreach ($kernel->getContaoBundles() as $bundle)
+		foreach ($resources->findFilesIn('config', 'autoload.php') as $file)
```
leofeyer: This is still too complicated as @Toflar mentioned in today's call. It should be:

```php
global $kernel;

$resources = $kernel->getContainer()->get('contao.resource_provider');

// findIn() should only set the base paths and ignoreDotFiles()
$files = $resources->findIn('config')->files()->name($fileName);
```

Especially, we should not call `followLinks()` by default, because there is no way to revert this.

leofeyer:
Merged in f62f3a109f2d36be74976370bf0ddbdd50412a46.

PR #56, opened 2015-03-05
Added unit tests for non existent load after definitions and circular re...

Toflar:
...ferences of load after definitions

### DIFF
--- a/tests/Autoload/ConfigResolverTest.php
+++ b/tests/Autoload/ConfigResolverTest.php
@@ -67,6 +67,51 @@ class ConfigResolverTest extends TestCase
     }
 
     /**
+     * Test non-existent load after definition is ignored
+     */
+    public function testNonExistentLoadAfterIsIgnored()
+    {
+        $config = $this->getConfig('bundle-name', 'bundle-class')->setLoadAfter(['i-do-not-exist-bundle']);
+        $resolver = new ConfigResolver();
+        $resolver->add($config);
+
+        $this->assertSame(['bundle-name' => 'bundle-class'], $resolver->getBundlesMapForEnvironment('all'));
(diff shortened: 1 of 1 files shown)


leofeyer:
Hm, while applying the CS fixes to your PR I have notices that your `testCircularReference` test is actually wrong. If there are two bundles (e.g. core-bundle and listing-bundle) with the same load-after definition (e.g. integration-bundle), the load-after definition bundle (integration-bundle) has to be loaded **before** the two other bundle and not after them!

Toflar:
Ah, lol - true. Checking!

Toflar:
Ah no, it's correct. The comment might be misleading but the code is correct.

leofeyer:
No way.

```php
// We have 3 bundles so the 3rd must be integration-bundle
$this->assertSame('integration-bundle', $bundlesMapKeys[2]);
```

The integration bundle should be `$bundlesMapKeys[0]`, shouldn't it?

leofeyer:
> Ah no, it's correct. The comment might be misleading but the code is correct.

You are right, the comment is misleading. The test ist just fine.

However, the test seems unnecessary to me, because `testGetBundlesMapForEnvironment()` does test exactly this, too, doesn't it? Maybe we should add the test to the bundles map provider instead?

Toflar:
> However, the test seems unnecessary to me, because testGetBundlesMapForEnvironment() does test exactly this, too, doesn't it? Maybe we should add the test to the bundles map provider instead?

Yeah, obviously it's an additional test - we had 100% coverage before ;)

leofeyer:
Done in 4feed7017dd7ce6e699f685df0bf512c077a155a.

PR #59, opened 2015-03-06
Add the Template::route() method and replace all hardcoded template routes

leofeyer:
This is the follow-up PR to #53 as discussed yesterday. @contao/developers Please review.

### DIFF
--- a/contao/classes/BackendUser.php
+++ b/contao/classes/BackendUser.php
@@ -216,7 +216,7 @@ class BackendUser extends \User
 
 
 	/**
-	 * Redirect to contao/index.php if authentication fails
+	 * Redirect to the login screen if authentication fails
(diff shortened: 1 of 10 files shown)


Toflar:
Can you explain why you use the `ampersand()` call there?

leofeyer:
Sure. The router returns paths like `/contao?do=undo&ref=ABC`, which will not validate as valid HTML markup. `&` should be `&amp;` and that's exactly what `ampersand()` does :)

Toflar:
Yeah I know but it's mixture of two things in the same method. You cannot call `route()` without ampersands :) Hmm...okay it's only in the Template class and it never makes sense to do it the other way around in that context...okay, then I can accept that :)

PR #60, opened 2015-03-08
[WIP] Use events to initialize the Contao framework

leofeyer:
As a proof of concept, I have moved the request token check to a separate event listener. IMHO we could optimize the `InitializeSystemListener` this way:

- Define the SwiftMailer defaults in the `config.yml` file.
- Use a separate event listener to set the relative path in the `Environment` class.
- Use a separate event listener to set the default language.
- Use a separate event listener to validate the installation.
- Use a separate event listener to check the request token (POC included in this PR).

@contao/developers Is this generally something we want?

### DIFF
--- a/src/EventListener/CheckRequestTokenListener.php
+++ b/src/EventListener/CheckRequestTokenListener.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\Environment;
+use Contao\Input;
+use Contao\RequestToken;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+
+/**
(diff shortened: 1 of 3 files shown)


Toflar:
> @contao/developers Is this generally something we want?

I like the idea of having separate listeners for the different purposes as we can then easily remove one legacy booting feature after the other.

aschempp:
I don't really like it. It's all legacy stuff, so we want to get rid of all of it somewhen. It will be more obvious if everything remains in the same place. The only thing I might like is calling separate methods (inside the same class) using separate events.

leofeyer:
That's not entirely true. There are things which we will keep and therefore should be refactored as event listeners:

 * The helpers (at least `functions.php` and `mbstring.php`)
 * Generating the symlinks (could be taken over by a Composer script handler)
 * Validating the installation (regarding `/web` and an incomplete setup)
 * Checking the request token (unless we decide to use the Symfony CSRF token system)

The rest (aka. the real legacy stuff) can remain in the class.

discordier:
The helpers should be moved into namespaced classes and the old functions should use these.

PR #61, opened 2015-03-09
Remove the ignoreInsecureRoot flag

leofeyer:
This PR removes the `ignoreInsecureRoot` flag as discussed in Ahrweiler. @contao/developers Please vote.

### DIFF
--- a/contao/config/default.php
+++ b/contao/config/default.php
@@ -160,4 +160,3 @@ $GLOBALS['TL_CONFIG']['proxyServerIps']       = '';
 $GLOBALS['TL_CONFIG']['sslProxyDomain']       = '';
 $GLOBALS['TL_CONFIG']['maintenanceMode']      = true;
 $GLOBALS['TL_CONFIG']['errorReporting']       = E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED;
-$GLOBALS['TL_CONFIG']['ignoreInsecureRoot']   = false; // FIXME: remove this option?
--- a/contao/languages/en/exception.xlf
+++ b/contao/languages/en/exception.xlf
@@ -126,7 +126,7 @@
         <source>Please set the document root to the &lt;code&gt;web/&lt;/code&gt; subfolder. This is usually done in the domain routing section of your server administration panel or in the vhosts file of your web server.</source>
       </trans-unit>
       <trans-unit id="XPT.insecureExplain">
-        <source>Contao 4 does not rely on &lt;code&gt;.htaccess&lt;/code&gt; files anymore and uses a public subfolder as document root instead. Anything above the document root must not be accessible via HTTP, otherwise anyone could download non-public resources including sensible data such as configuration files.</source>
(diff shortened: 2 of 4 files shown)


Toflar:
Oh, I need to fix my document root then :D Good to merge imho :+1:

leofeyer:
I'd suggest to make it an event listener eventually then it could be deactivated by removing the listener.

Toflar:
No, it's a security issue. I think we should force people to properly configure their setups :)

ausi:
Would it still be possible to run Contao in a subfolder on an apache server with a customized htaccess?

leofeyer:
@ausi That depends on what you mean by "customized htaccess".

aschempp:
Ich denke sowas wie Variant 1 von https://www.cyon.ch/support/a/210/kann-ich-eine-domain-aus-einem-unterordner-aufrufen ist gemeint

ausi:
Something like this to run contao in `example.com/my-subfolder`:

```
RewriteEngine on
RewriteBase /my-subfolder
RewriteRule (.*) /web/app.php%{REQUEST_URI} [L]
```

leofeyer:
Das müsste man testen.

aschempp:
As I commented previously, the saying of "not relying on .htaccess" is wrong. We still need .htaccess or a similar method (e.g. for Nginx) for URL rewriting, otherwise the `app.php` remains in the URL. There's no difference to Contao 3.

leofeyer:
The wording is correct. We have been relying on `.htaccess` files in Contao 3 and we are no longer in Contao 4. Of course your web server needs to support URL rewriting, but you do not necessarily need an `.htaccess` file for it - not even on Apache ;)

aschempp:
Where did we rely on `.htaccess` files in Contao 3 except for URL rewriting?

leofeyer:
Folder protection in `/files`?

aschempp:
Aah, forgot about that. I still think it's misleading as most people used `.htaccess` for URL rewriting. Maybe the text could say something like that:

> Contao 4 does no longer rely on `.htaccess` for file access control but uses a public subfolder as document root instead. …`

Toflar:
And I think it should say

`[...]a public subfolder (by default "web") as document [...]`

instead. :)

leofeyer:
@aschempp I have adjusted the wording in 51021d1333c9c8916b5129cb94fc6e45a035a476.

@Toflar If we'd say "by default" people would assume they could change the folder. However they cannot, because references to `/web` are currently hardcoded. We might change this of course but that would be a separate feature.

Can I merge the PR now?

aschempp:
I think it should be *but* instead of *and*, maybe a native speaker can review the sentence?

> Contao 4 does no longer rely on `.htaccess` files to protect folders and uses a public subfolder as document root instead. Anything above the document root must not be accessible via HTTP, otherwise anyone could download non-public resources including sensible data such as configuration files.

/cc @BlairWinans

leofeyer:
Changed and merged in 9da41d6608a52b53eb20b02eafdb17de27ccddc2.

blairwinans:
Just seeing this now—if you want it to make the most sense in EN, I would say:

>Contao 4 no longer relies on .htaccess files to protect folders, and instead uses a public subfolder as document root instead. Anything above the document root must not be accessible via HTTP, otherwise anyone could download non-public resources including sensitive data such as configuration files.

PR #63, opened 2015-03-10
The legacy code must only be initialized on master request

aschempp:
Otherwise subrequests will not be possible because of redeclared functions, `ini_set` after session-start etc.


Toflar:
That's right!

aschempp:
There's a problem on commands too thought, at least when using https://github.com/contao/contao/issues/42

PR #65, opened 2015-03-18
Make private what does not need to be overridden

Toflar:
Most of the code actually was already `private` :)


aschempp:
`getPublicFolders` will be dropped anyways in #55 but otherwise +1

leofeyer:
You are right, most of the protected stuff was in the autoloading classes, which we have dropped for the time being. Nevertheless, thanks for taking the time :)

PR #67, opened 2015-03-19
do not check installation in environments test and dev

dtomasi:
Why do you force the check also in test and dev environment. I think this is not needed because app_dev.php restricts 127.0.0.1.

### DIFF
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -42,14 +42,20 @@ class InitializeSystemListener
     private $rootDir;
 
     /**
+     * @var string
+     */
+    private $env;
+
+    /**
      * Constructor.
      *
-     * @param RouterInterface $router  The router object
-     * @param string          $rootDir The kernel root directory
+     * @param RouterInterface $router      The router object
+     * @param string          $rootDir     The kernel root directory
+     * @param string          $environment Current Environment
      */
-    public function __construct(RouterInterface $router, $rootDir)
+    public function __construct(RouterInterface $router, $rootDir, $environment)
     {
-        $this->router  = $router;
+        $this->router = $router;
         $this->rootDir = dirname($rootDir);
     }
 
@@ -300,7 +306,7 @@ class InitializeSystemListener
         }
 
         // Show the "insecure document root" message
-        if ('/web' === substr(Environment::get('path'), -4)) {
+        if ('/web' === substr(Environment::get('path'), -4) && $this->env == 'prod') {
             die_nicely('be_insecure', 'Your installation is not secure. Please set the document root to the <code>/web</code> subfolder.');
         }
 
--- a/src/Resources/config/services.yml
+++ b/src/Resources/config/services.yml
@@ -17,6 +17,7 @@ services:
         arguments:
             - "@router"
             - "%kernel.root_dir%"
+            - "%kernel.environment%"
         tags:
             # Priority must be higher than any other listener that uses the Contao framework and lower than the
             # priority of the route_listener, which defaults to 32.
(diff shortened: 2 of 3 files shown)


aschempp:
Nice idea! `app_dev` will potentially be secured by backend login or similar instead of localhost, but your point is still valid. I'm not sure if we should check for the environment or better for `kernel.debug`.

dtomasi:
I also thought about a config-parameter to skip the check. But this seems to be the easiest way. Why would you like to check `kernel.debug` instead of `kernel.environment` ? Is this for security reasons?

dtomasi:
Oh, looks like I have to check the tests....

leofeyer:
This is a **very bad idea**! The point of the warning is to make the user aware that they are using an insecure document root. This has nothing to do with `app.php` or `app_dev.php`, nor with `app_dev.php` being secured!

If your document root is not set to `/web`, you will be able to access **any** file of the installation!

aschempp:
Which for development, e.g. locally, is totally fine. That's what the development environment is for.

leofeyer:
That's not correct. Locally, I might as well access the `app.php` script without the document root set to `/web`. And on the server, I must never access **any** script without the correct document root.

dtomasi:
Thats the point! If I work locally in development I can use the `app_dev.php` for calling. On a live-system `app.php` is used.
I think its a bad idea to force every developer has to setup a virtual-host for every installation. There are a lot of people out there who use one document-root for all there projects.

leofeyer:
> I think its a bad idea to force every developer has to setup a virtual-host for every installation. There are a lot of people out there who use one document-root for all there projects. 

This is exactly what Symfony does, too, with the `/web` folder. It is a security measure and there should not be an (easy) opt-out, because otherwise people will assume they have a choice whereas they actually have not!

If you really want to bypass the feature, you **must take additional measures** to secure your installation. It is crucial to understand this.

dtomasi:
> This is exactly what Symfony does, too, with the /web folder.

Thats right! And also Typo3 has a public and secured area using symlinks. But thats not the fact. This also will be available on prod-environments. But there is no need to check this if the request comes form `localhost`. I really even don´t understand the reason, why this should be a bad idea.

```
// This check prevents access to debug front controllers that are deployed by accident to production servers.
// Feel free to remove this, extend it, or make something more sophisticated.
if (isset($_SERVER['HTTP_CLIENT_IP'])
    || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
    || !(in_array(@$_SERVER['REMOTE_ADDR'], array('127.0.0.1', 'fe80::1', '::1')) || php_sapi_name() === 'cli-server')
) {
    header('HTTP/1.0 403 Forbidden');
    exit('You are not allowed to access this file. Check ' . basename(__FILE__) . ' for more information.');
}
```

I think this shoud do the job!! And thats Symfony...

aschempp:
I have to agree with both. Currently, I have to remove the `die_nicely` in my local installation to use Contao 4 because I am not using virtual hosts. It would be nice to have a solution for that. However:

1. @dtomasi you must be aware that we discussed to remove the localhost-check in the `app_dev.php` and replace it with another means of authentication (see https://github.com/contao/symfony-todo/issues/3).
2. It would make sense to allow development access to an insecure root. If that is applied to a hosted environment, it would still configuration files are accessible for a development installation, which again is a no-go.

How about we secure the installation with a check on the current host? If host=localhost then ignore the insecure root? Independent of the entry point script?

leofeyer:
> If host=localhost then ignore the insecure root? Independent of the entry point script?

Maybe, but if yes then not independent of the entry point. If I develop locally, I might want to check the prod environment as well, don't I?

aschempp:
> Maybe, but if yes then not independent of the entry point. If I develop locally, I might want to check the prod environment as well, don't I?

That's what I mean. We should not check for environment but only for host.

leofeyer:
+1

discordier:
Fine for me.

PR #68, opened 2015-03-19
[RTM] Symfony Session Handling

Toflar:
Requires the following changes in `contao/standard-edition`'s `config.yml`:

```diff
-    session: false
+    csrf_protection: ~
+    session:
+        handler_id: ~
```

(I did not make a separate PR because of the effort :-)).

I know it's mixing session handling with CSRF token handling but they sort of belong together and I had to make sure I could log in again, so I hacked it all together :-)

## ToDo's
* [x] Should we add an `ArrayAccess` wrapper for `$_SESSION`? Should writing and fetching from `$_SESSION['BE_DATA']` and `$_SESSION['FE_DATA']` still be supported? Ideas?
* [x] Should we add `public function getToken()` to `Template` because one should not use `REQUEST_TOKEN` anymore?
* [x] Tests
* [x] CS
* [x] Remove all internal usages of any PHP `session_*` function calls (e.g. in `User::authenticate()` - maybe @aschempp needs to have a look?)
* [x] Fix `Session::setData()` mapping
* [x] Check Request Token only if needed

### DIFF
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -56,6 +56,11 @@ class InitializeSystemListener extends ScopeAwareListener
     private $rootDir;
 
     /**
+     * @var string
+     */
+    private $csrfTokenName;
+
+    /**
      * @var bool
      */
     private static $booted = false;
@@ -66,17 +71,20 @@ class InitializeSystemListener extends ScopeAwareListener
      * @param CsrfTokenManagerInterface $tokenManager The token manager service
      * @param SessionInterface          $session The session service
      * @param string                    $rootDir The kernel root directory
+     * @param string                    $csrfTokenName The name of the token to be used
      */
     public function __construct(
         RouterInterface $router,
         CsrfTokenManagerInterface $tokenManager,
         SessionInterface $session,
-        $rootDir
+        $rootDir,
+        $csrfTokenName = null
     ) {
-        $this->router       = $router;
-        $this->tokenManager = $tokenManager;
-        $this->session      = $session;
-        $this->rootDir      = dirname($rootDir);
+        $this->router        = $router;
+        $this->tokenManager  = $tokenManager;
+        $this->session       = $session;
+        $this->rootDir       = dirname($rootDir);
+        $this->csrfTokenName = $csrfTokenName;
     }
 
     /**
@@ -366,12 +374,12 @@ class InitializeSystemListener extends ScopeAwareListener
         // Backwards compatibility
         if (!defined('REQUEST_TOKEN')) {
             /** @var CsrfToken $token */
-            $token = $this->tokenManager->getToken('_csrf');
+            $token = $this->tokenManager->getToken($this->csrfTokenName);
             define('REQUEST_TOKEN', $token->getValue());
         }
 
         // Check the request token upon POST requests
-        $token = new CsrfToken('_csrf', Input::post('REQUEST_TOKEN'));
+        $token = new CsrfToken($this->csrfTokenName, Input::post('REQUEST_TOKEN'));
 
         // FIXME: This forces all routes handling POST data to pase a REQUEST_TOKEN
         if ($_POST && !$this->tokenManager->isTokenValid($token)) {
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -13,7 +13,7 @@ namespace Contao\CoreBundle\EventListener;
 use Contao\ClassLoader;
 use Contao\Config;
 use Contao\CoreBundle\Command\ContaoFrameworkDependentInterface;
-use Contao\CoreBundle\Session\Attribute\AttributeBagMirror;
+use Contao\CoreBundle\Session\Attribute\AttributeBagAdapter;
 use Contao\Environment;
 use Contao\Input;
 use Contao\System;
@@ -406,7 +406,7 @@ class InitializeSystemListener extends ScopeAwareListener
         /** @var AttributeBagInterface $beBag */
         $beBag = $this->session->getBag('contao_backend');
 
-        $_SESSION['FE_DATA'] = new AttributeBagMirror($feBag);
-        $_SESSION['BE_DATA'] = new AttributeBagMirror($beBag);
+        $_SESSION['FE_DATA'] = new AttributeBagAdapter($feBag);
+        $_SESSION['BE_DATA'] = new AttributeBagAdapter($beBag);
     }
 }
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -62,9 +62,10 @@ class InitializeSystemListener extends ScopeAwareListener
 
     /**
      * Constructor.
-     *
-     * @param RouterInterface $router  The router object
-     * @param string          $rootDir The kernel root directory
+     * @param RouterInterface           $router The router service
+     * @param CsrfTokenManagerInterface $tokenManager The token manager service
+     * @param SessionInterface          $session The session service
+     * @param string                    $rootDir The kernel root directory
      */
     public function __construct(
         RouterInterface $router,
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -15,12 +15,13 @@ use Contao\Config;
 use Contao\CoreBundle\Command\ContaoFrameworkDependentInterface;
 use Contao\Environment;
 use Contao\Input;
-use Contao\RequestToken;
 use Contao\System;
 use Symfony\Component\Console\Event\ConsoleCommandEvent;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 
 /**
  * Initializes the Contao framework.
@@ -35,6 +36,11 @@ class InitializeSystemListener extends ScopeAwareListener
      * @var RouterInterface
      */
     private $router;
+
+    /**
+     * @var CsrfTokenManagerInterface
+     */
+    private $tokenManager;
 
     /**
      * @var string
@@ -52,10 +58,14 @@ class InitializeSystemListener extends ScopeAwareListener
      * @param RouterInterface $router  The router object
      * @param string          $rootDir The kernel root directory
      */
-    public function __construct(RouterInterface $router, $rootDir)
-    {
-        $this->router  = $router;
-        $this->rootDir = dirname($rootDir);
+    public function __construct(
+        RouterInterface $router,
+        CsrfTokenManagerInterface $tokenManager,
+        $rootDir
+    ) {
+        $this->router       = $router;
+        $this->tokenManager = $tokenManager;
+        $this->rootDir      = dirname($rootDir);
     }
 
     /**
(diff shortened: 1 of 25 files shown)


Toflar:
I am now initializing the FE and BE attribute bags in the legacy Session class which is obviously wrong. The question is, should we provide those bags to the users (which means we add them via a generic listener by default)? Needs some discussion on Mumble. There's a lot of work to do here!

aschempp:
Looks good so far! I was not aware we can replace our request token with CSRF, but if that's possible I'd prefer to have that in a separate feature branch and merged as soon as it works (which appears to be fairly simple).

Toflar:
I won't be separating that. They belong together. If I implemented CSRF Handling first, I would need to configure and test native PHP session handling. If I do Session things first, I cannot log into the back end to see if session values are correctly stored. I'll keep it all together therefore.

Toflar:
Native session calls are removed or `FIXME`s are added where I'm not sure what to do. Also, the Contao frontend and backend session attribute bags are now registered via listener.

Toflar:
Our old Session class is now a simple mirror to the Symfony session :-) Now, I wanted to take on that `ArrayAccess` thing for `$_SESSION['BE_DATA']` and `$_SESSION['FE_DATA']` but I'm not sure if we should do that because if you did that, you did **not** use the API of Contao and you cannot expect this to work in a future version. Also I wonder if we should write that for `$_SESSION` because writing directly to it will still be supported and it will always be because it's a PHP superglobal. It's wrong but we cannot protect the users from doing it anyway so why should we add a component that "magically" maps them to the Symfony Session? I think it doesn't make sense. What do you think @contao/developers?

aschempp:
I think we should add the `ArrayAccess` mirror to `$SESSION['FE/BE_DATA']`. We know this is going to cause problems and there is no reason not to add it. It should obviously be deprecated in Contao 4 and removed in Contao 5 (maybe even add a deprecated notice).

Toflar:
> I think we should add the ArrayAccess mirror to $SESSION['FE/BE_DATA']. We know this is going to cause problems and there is no reason not to add it. It should obviously be deprecated in Contao 4 and removed in Contao 5 (maybe even add a deprecated notice).

@discordier and @leofeyer, what do you think?

a) Do we want the layer at all?
b) If yes, should it trigger an `E_USER_DEPRECATED`?

leofeyer:
What do we need the layer for?

Toflar:
Possible in Contao 3:
```php
$_SESSION['FE_DATA']['foo'] = 'bar';

echo \Session::getInstance()->get('foo');
```

This will not work in Contao 4 anymore. We could make it work by replacing `$_SESSION['FE_DATA']` and `$_SESSION['BE_DATA']` with a class implementing `ArrayAccess` and mapping it to the SF2 Session.

Toflar:
I have implemented everything (and more :-)) we've talked about:

* The `$_SESSION` keys `FE_DATA` and `BE_DATA` are now mirrored.
* I have added a FIXME for the token story
* User session is now invalidated
* I've replaced the `File::sendToBrowser()` by the SF `BinaryFileResponse` which is better suited than `StreamedResponse` and uses `stream_copy_to_stream`. This way I could get rid of the last native `session_*` call and I've added a `FIXME` there as the response must be thrown there using the upcoming `ResponseException`.

Removed `WIP` as this should be good to be merged now.

aschempp:
Maybe you should rebase to fix the merge conflicts?

Toflar:
That's because of 7b1d8995253edecdd71a11ee4d9eda1f861b009a, I can try :)

contao/classes/RebuildIndex.php
```diff
@@ -9,6 +9,8 @@
  */
 
 namespace Contao;
+use Symfony\Component\HttpKernel\KernelInterface;
```
aschempp: missing empty line

contao/controllers/BackendInstall.php
```diff
@@ -9,6 +9,8 @@
  */
 
 namespace Contao;
+use Symfony\Component\HttpKernel\KernelInterface;
```
aschempp: also missing empty line

contao/controllers/BackendMain.php
```diff
@@ -11,6 +11,8 @@
 namespace Contao;
 
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
```
aschempp: here too …

contao/library/Contao/File.php
```diff
@@ -9,6 +9,8 @@
  */
 
 namespace Contao;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
```
aschempp: another one

Toflar:
Stop commenting about CS, that's useless. Leo takes care of it anyway.

contao/library/Contao/File.php
```diff
@@ -681,32 +683,23 @@ public function resizeTo($width, $height, $mode='')
...
-		// Disable zlib.output_compression (see #6717)
-		ini_set('zlib.output_compression', 'Off');
-
-		// Open the "save as …" dialogue
-		header('Content-Type: ' . $this->mime);
-		header('Content-Transfer-Encoding: binary');
-		header('Content-Disposition: attachment; filename="' . ($filename ?: $this->basename) . '"');
-		header('Content-Length: ' . $this->filesize);
-		header('Cache-Control: must-revalidate, post-check=0, pre-check=0');
-		header('Pragma: public');
```
aschempp: some of these headers are gone? Did you test if it still does open the "save as …" dialog?

aschempp:
I wonder what's happening to CSRF if there is no security enabled? That's the case for the install tool for example.

src/Session/Attribute/AttributeBagMirror.php
```diff
@@ -0,0 +1,82 @@
...
+namespace Contao\CoreBundle\Session\Attribute;
+use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBag;
+
+/**
+ * Provides an ArrayAccess mirror for a session AttributeBag.
+ * Used for BC for $_SESSION['FE_DATA'] and $_SESSION['BE_DATA'].
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+class AttributeBagMirror implements \ArrayAccess
```
aschempp: I think this should be called `Adapter` instead of `Mirror` to follow the design pattern.

aschempp:
> $feBag of type object<Symfony\Compon...SessionBagInterface> is not a sub-type of object<Symfony\Compon...ribute\AttributeBag>. It seems like you assume a concrete implementation of the interface Symfony\Component\HttpFo...ion\SessionBagInterface to be always present.

What does that mean?
https://scrutinizer-ci.com/g/contao/core-bundle/inspections/af5ac3b4-8bc7-4b59-bb7e-4e610ae17673/issues/files/src/EventListener/InitializeSystemListener.php?status=new&orderField=path&order=asc

Toflar:
> What does that mean?
> https://scrutinizer-ci.com/g/contao/core-bundle/inspections/af5ac3b4-8bc7-4b59-bb7e-4e610ae17673/issues/files/src/EventListener/InitializeSystemListener.php?status=new&orderField=path&order=asc

This is actually correct. `Session::getBag()` returns a `SessionBagInterface` instead of an `AttributeBagInterface` (which extends `SessionBagInterface`). Do I have to add a check or just `/* @var $feBag AttributeBagInterface */`to get rid of this?

src/Session/Attribute/AttributeBagMirror.php
Toflar: And I think it should be `Mirror` instead of `Adapter`.

Toflar:
> I wonder what's happening to CSRF if there is no security enabled? That's the case for the install tool for example.

I don't get the question but it does not seem like that's something I need to worry about.

contao/library/Contao/File.php
Toflar: Of course I tested! And the headers are either set by default via the BinaryFileResponse class or useless. I was pretty careful about that! It works like a charm for me :)

aschempp:
> I wonder what's happening to CSRF if there is no security enabled? That's the case for the install tool for example.

I thought CSRF was related to security because the services are in that category. But according to your initial description `csrf_protection` is separately configured in the `config.yml` so that should be fine.

However, the unit tests coverage dropped significantly.

Toflar:
Yeah, no idea why. My code is 100% covered.

src/Resources/contao/library/Contao/File.php
```diff
@@ -681,32 +683,23 @@ public function resizeTo($width, $height, $mode='')
 	 */
 	public function sendToBrowser($filename=null)
 	{
-		// Make sure no output buffer is active
```
discordier: Why is this in your PR?
Mind if I steal it into mine?

Toflar: I can create a separate PR for it so @leofeyer can merge it first and we then both rebase?

discordier: Might be also a proper solution.
Another thing, this PR is already "[RTM]" therefore if @leofeyer is willing to merge, I can rebase and add the exception.

leofeyer: I'll rebase tonight or tomorrow.

src/EventListener/InitializeSystemListener.php
```diff
@@ -151,6 +167,9 @@ private function boot($routeName, $basePath)
         // Log PHP errors
         $this->iniSet('error_log', $this->rootDir . '/system/logs/error.log');
 
+        // Support legacy Session access
+        $this->initializeLegacySessionAccess();
```
leofeyer: Why is the method call here and not in line 163, when `$this->startSession()` used to be?

src/Resources/config/services.yml
```diff
@@ -29,6 +31,14 @@ services:
             # Priority must be lower than the one of the Contao initialize system listener.
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 26 }
 
+    contao.listener.session:
+        class: Contao\CoreBundle\EventListener\SessionListener
+        arguments:
+            - "@session"
+        tags:
+            - { name: kernel.event_listener, event: kernel.request, method: registerContaoAttributeBags, priority: 64 }
```
leofeyer: Why did you set the priority to 64? We should add an explaining comment.

src/EventListener/InitializeSystemListener.php
Toflar: Because I wanted this to be as early as possible. There's no specific reason - I would have added a comment if that was the case :)

leofeyer: Then may I ask you to move it down to where it used to be? We should not mess with the initialization order too much, because we cannot foresee the effects.

src/Resources/config/services.yml
Toflar: Added in 42686927457ea104cbd899bb6c660ffbad173732.

src/EventListener/InitializeSystemListener.php
Toflar: I can, I just saw that it is now before `Config::preload();` which means that if somebody did something like
```php
$_SESSION['FE_DATA']['foo'] = 'bar';
```
in his `localconfig.php` it would still work. If I move it down, it won't work anymore.

aschempp: It should probably be where the session was started previously?

aschempp: I mean accessing the session in `localconfig.php` did not work before, because it was not started?

Toflar: True, then I'll move it down.

Toflar: Done in 63587f9d911a6bfb68e57a506ac3d4aa032108fa.

Toflar:
Rebased.

leofeyer:
Merged in d30310394fd209481ed6301f85d07e702300e444.

PR #69, opened 2015-03-20
Fix the redirect urls in the controller.

discordier:
I was kept being redirected to `http://host//app.php/foo` which caused the router to not find any valid route anymore (note the double slash).

### DIFF
--- a/contao/library/Contao/Controller.php
+++ b/contao/library/Contao/Controller.php
@@ -1008,7 +1008,7 @@ abstract class Controller extends \System
 		// Make the location an absolute URL
 		if (!preg_match('@^https?://@i', $strLocation))
 		{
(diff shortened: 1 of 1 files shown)


aschempp:
Hacky but probably necessary so :+1:

leofeyer:
We should probably try to figure out what causes the issue ;)

aschempp:
The issue is the difference between an URL generated by the Symfony router and the legacy Contao code. Symfony router URLs are absolute by default.

leofeyer:
Fixed in 5f14064c822b25969af05c1342b869ecae66b885 (follow the link to see what the actual problem was).

discordier:
Yeah this fix is much better.

PR #70, opened 2015-03-20
[RTM] Feature/exception listener.

discordier:
This PR is currently a moving target.

Pretty error rendering is enabled by default keep the behaviour known from Contao 3.X.
To disable it, put the following into your `config_dev.yml`:
```diff
+contao:
+    pretty_error_screens: false
```

TODO
=====
- [X] Introduce Exception classes for all `die_nicely()` calls.
- [X] Add rendering of exceptions in exception handler.
- [X] Replace occurences of `exit` with proper exception throwing where possible.
- [X] Introduce proper error logging (probably a separate PR). Postponed to future error handling PR.
- [X] ~~Absorb 25dec35453f393debdf9d33c1a53d3a4551429bd from @Toflar's PR for streaming files.~~ Has been merged meanwhile and adjusted accordingly.
- [X] Decide how to handle the error templates.
  - Rewrite using twig! (see 9044208)
  - ~~Drop usage of class `Backend` at least in `be_error` to allow boot exceptions as `be_error.html5` breaks when an exception is encountered before `InitializeSystemListener::boot()` has been called due to `Backend::getTheme()`.~~
  - ~~Unify to a single or at least less templates (see https://github.com/contao/core-bundle/pull/70#issuecomment-87829443).~~

Caveats and stuff to be aware of:
==========================
1. the error logging is still being set in the `InitializeSystemListener::boot()`.
2. I had to alter the error handling to remove E_NOTICE etc. in the beginning of `InitializeSystemListener::boot()` as otherwise the whole legacy code will break in the console (automator command) as it is riddled with notices.
3. I had to remove the exception and error handlers from `functions.php` as they were rendering the default symfony error and exception handling useless.
4. The whole monolog handling is still missing, I wonder if we should configure it in the standard-distribution and inject the kernel and/or logger into the listener. I am as of now undecided which way to go.

How to use the new exceptions:
=========================

Ajax responses with "No Content".
----------------------------------------------
Old:
```php
    header('HTTP/1.1 204 No Content');
    header('X-Ajax-Location: ' . $strLocation);
    exit;
```
New:
```php
throw AjaxRedirectResponseException::create($location);
```

Redirect:
----------------------------------------------
Old:
```php
// Simple redirect.
Controller::redirect($strLocation);
// Simple redirect with status.
Controller::redirect($strLocation, 301);
```

New:
```php
// Simple redirect (HTTP 303).
throw RedirectResponseException::create($location);
// Simple redirect with status 301.
throw RedirectResponseException::create($location, 301);
// Provide additional header.
throw RedirectResponseException::create($location, 301, ['Auth-Token' => 'something']);
```

File::sendToBrowser()
----------------------------
For this one we will use a streaming response to lower load.
Fetch code from: 25dec35453f393debdf9d33c1a53d3a4551429bd


aschempp:
Similar to the session stuff, I think this is mixing things and therefore more complicated (to review) than necessary. The exceptions, especially replacing `die_nicely` is nowhere related to error logging. I suggest to separate the exception handling from error logging, so we are probably faster in finishing the first.

Things I would change:
 1. There should not be one exception for all `die_nicely` replacements. We must have a `NoRootFoundException` and `IncompleteInstallationException` etc.
 2. The parent class of all these should be a `TemplateResponseException` which in turn takes care itself to populate `getResponse()` (instead of the exception handler searching for the template).
 3. We must never check for an environment. There should be a parameter `contao.show_exceptions` which defaults to `%kernel.debug%` (and could be configured per environment).

discordier:
The error logging is related, as otherwise I will not end up in my Listener (or in the symfony handler at all) - we can split it out in a PR that has to be merged forehand but otherwise it can not be done.

On:

1. Yes there should be more than one exception. This will get changed.
2. I agree on having a common ResponseException but I strongly disagree that it shall populate itself. Exceptions should not encapsulate much business logic and especially none that interfaces with other classes. They are a "Notification of an error" which gets either caught or bubbled to the "global scope" to be handled by an handler class and maybe rendered/presented to the user.
3. And how shall I proceed there? As written, even if we have such a flag, the workflow is different.for the different environments. If symfony debug is loaded, the exception will be correctly rendered in there. If not, the exception get's bubbled to the twig handler as I am not handling the exception because I want the symfony debug bar to handle it. Which way to go then? I have to know either the environment or check the Symfony Debug bar if it is loaded. Of these the first one appears to be less harmful to me.

aschempp:
I think you're right with 2.

I don't understand what you mean with 3. Having a different configuration per environment is already supported with `config_dev.yml`?

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,143 @@
...
+            $event->setResponse($response);
+
+            return;
+        }
+
+        // FIXME: log the exception here if we do not find a better place to log via monolog from outside.
+
+        // FIXME: do we really want this?
+        // If not in prod, exit here.
+        if ($this->environment !== 'prod') {
```
Toflar: Does this mean the SF core exception handling would get triggered then? :)

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,143 @@
...
+            return;
+        }
+
+        $exception = $event->getException();
```
Toflar: This should go into a `checkDieNicelyException` just the same like `checkResponseException`.

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,305 @@
...
+     *                                   (only applicable when renderErrorScreens is true).
+     */
+    public function __construct($renderErrorScreens, $rootDir)
+    {
+        $this->renderErrorScreens = $renderErrorScreens;
+        $this->rootDir = dirname($rootDir);
+    }
+
+    /**
+     * Forwards the request to the Frontend class if there is a page object.
```
aschempp: That description sounds weird, sure it's not a copy-paste? ;-)

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,305 @@
...
+        'Contao\CoreBundle\Exception\InsecureInstallationHttpException'   => 'be_insecure',
+        'Contao\CoreBundle\Exception\MaintenanceModeActiveHttpException'  => 'be_unavailable',
+        'Contao\CoreBundle\Exception\NoLayoutHttpException'               => 'be_no_layout',
+        'Contao\CoreBundle\Exception\NoPagesFoundHttpException'           => 'be_no_active',
+        'Contao\CoreBundle\Exception\NotFoundHttpException'               => 'be_no_page',
+        'Contao\CoreBundle\Exception\RootNotFoundHttpException'           => 'be_no_root',
+    ];
+
+    /**
+     * Create a new instance.
```
aschempp: Symfony usually adds "Constructor." as a comment for this.

src/Exception/AbstractResponseException.php
```diff
@@ -0,0 +1,52 @@
...
+    private $response;
+
+    /**
+     * Construct the exception.
+     *
+     * @param Response $response   The Response to send
+     *
+     * @param \Exception $previous The previous exception used for the exception chaining
+     */
+    public function __construct($response, \Exception $previous = null)
```
aschempp: The interface is missing (first parameter)

src/Exception/BadRequestTokenHttpException.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ */
+class BadRequestTokenHttpException extends BadRequestHttpException
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(
+        $message = 'Invalid request token. Please <a href="javascript:window.location.href=window.location.href">go back</a> and try again.',
```
aschempp: This javascript looks wrong.

discordier: It is 1:1 copied from https://github.com/contao/core/blob/5d878de67eb218d5eb7ba967dbc669859b8489ec/system/initialize.php#L271
What do you suggest to change it to?

discordier:
IMO we should homogenize the error screens into a single error template which contains the appropriate values.
In this template we could then use something like `$this->errorType` which will contain the "name" of the error. That way the users can still display different error screens for different errors but we will have way less templates to manage.

If you think this is a good idea, I will update the PR accordingly (including changes of language files).

Toflar:
YES, please!

discordier:
> YES, please!

Please what?

Toflar:
Unifying error templates :)

discordier:
@leofeyer Are you ok with unifying them? If so, I'll add this to this PR.

leofeyer:
Don't know if this is easily possible. The error templates have quite different layouts.

discordier:
The following have the exact same layout (aside from languag key naming):
- `be_forbidden.html5`
- `be_no_active.html5`
- `be_no_forward.html5`
- `be_no_page.html5`
- `be_no_root.html5`

The following additionally have a `$lang->more` and `$lang->no*Explain`:
- `be_no_layout.html5` 
- `be_incomplete.html5`

The only one not fitting so far is `be_error` but that one has to be altered anyway as the message about adding `$GLOBALS['TL_CONFIG']['displayErrors'] = true;` is obsolete (does not work anymore).

aschempp:
I would rename the following:
`RedirectResponseException` => `RedirectException`
`RedirectResponseException::createAjax` => `AjaxRedirectException::create`

discordier:
I disagree as then the naming of the exceptions is mixed.
Currently we have `*ResponseException` and `*HttpException` which is like the Symfony way.
This way you know the type of the exception class without reading the code.
Following your suggestion, you will encounter problems when havin a real exception that is thrown when a redirect is wrong or something like that.

I'd rather do:
Keep `RedirectResponseException` for normal redirects.
`RedirectResponseException::createAjax` => `AjaxRedirectResponseException::create`
If it really hurts to have both in the same class. ;)

aschempp:
You might be right. 

Somehow I feel this could be simplified. In Symfony, the `Response` does not have an interface. Why do we need an interface and an abstract class for the `ResponseException`?

discordier:
Because they are exceptions and not responses.
They have different static create methods which are incompatible to each other and to reduce duplicate code we have the abstract class.
We have the interface for type checking.

aschempp:
> They have different static create methods which are incompatible to each other

Same for responses…

tests/EventListener/InitializeSystemListenerTest.php
```diff
@@ -157,6 +157,58 @@ public function testRequestWithoutScope()
...
 
     /**
+     * Tests a request without scope.
+     *
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     *
+     * @expectedException \Contao\CoreBundle\Exception\InsecureInstallationHttpException
+     */
+    public function testValidateInstallation()
```
leofeyer: Why did you add this test? Is it related to the exception handling?

src/Resources/contao/templates/backend/be_no_root.html5
```diff
@@ -16,8 +16,14 @@ $lang = (object) $GLOBALS['TL_LANG']['XPT'];
   <meta name="viewport" content="width=device-width,initial-scale=1.0">
 
   <style><?php
-    include TL_ROOT . '/system/themes/' . $theme . '/basic.css';
-    include TL_ROOT . '/system/themes/' . $theme . '/error.css';
+    $includeIfExists = function ($file) {
```
leofeyer: Why is this change required? In which scenario will the files not exist?

src/Resources/contao/modules/ModuleArticle.php
```diff
@@ -366,7 +366,7 @@ public function generatePdf()
 		// Close and output PDF document
 		$pdf->lastPage();
 		$pdf->Output(standardize(ampersand($this->title, false)) . '.pdf', 'D');
-
+		// FIXME: This has to be changed to a ResponseException somehow - rather write to temp file and stream it then?
```
leofeyer: I could live with using `$pdf->Output()` in this case.

src/Resources/contao/library/Contao/Template.php
```diff
@@ -265,7 +265,7 @@ public function output()
 
 		header('Vary: User-Agent', false);
 		header('Content-Type: ' . $this->strContentType . '; charset=' . \Config::get('characterSet'));
-
+		 // FIXME: would be cool to throw a response exception here but sadly the hook is in the way.
```
leofeyer: So what's the point of the FIXME then? Do you want to talk about removing the hook? Or do you just want to mention it; in this case, you should remove the `FIXME:` and leave the comment only.

src/Resources/contao/config/default.php
```diff
@@ -159,4 +159,3 @@
 $GLOBALS['TL_CONFIG']['proxyServerIps']       = '';
 $GLOBALS['TL_CONFIG']['sslProxyDomain']       = '';
 $GLOBALS['TL_CONFIG']['maintenanceMode']      = true;
-$GLOBALS['TL_CONFIG']['errorReporting']       = E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED;
```
leofeyer: Why is this gone? We need to be able to configure the error reporting level somewhere.

src/Resources/contao/classes/Ajax.php
```diff
@@ -393,20 +386,20 @@ public function executePostActions(\DataContainer $dc)
...
 					}
 				}
-				exit; break;
+				throw ResponseException::create('', 204);
 
 			// DropZone file upload
 			case 'fileupload':
 				$dc->move();
-				exit; break;
+				throw ResponseException::create('', 204);
```
leofeyer: Is there no such thing as a `NoContentResponseException`?

src/Resources/config/services.yml
```diff
@@ -12,6 +12,14 @@ services:
             - { name: kernel.event_listener, event: kernel.controller, method: onKernelController, priority: 254 }
             - { name: kernel.event_listener, event: kernel.finish_request, method: onKernelFinishRequest, priority: -254 }
 
+    contao.listener.handle_exception:
+        class: Contao\CoreBundle\EventListener\ExceptionListener
+        tags:
+            - { name: kernel.event_listener, event: kernel.exception, method: onKernelException, priority: 200 }
```
leofeyer: Why did you set the priority to 200? We should add an explaining comment.

src/Exception/RedirectResponseException.php
```diff
@@ -0,0 +1,58 @@
...
+     *
+     *     throw RedirectResponseException::createAjax('https://example.org/target.html');
+     *
+     * @param string $location The target URL
+     * @param int    $status   The response status code (defaults to 204)
+     * @param array  $headers  An array of response headers
+     *
+     * @return RedirectResponseException
+     */
+    public static function createAjax($location, $status = 204, $headers = array())
```
leofeyer: I'd prefer an `AjaxRedirectReponseException` class.

src/Exception/MaintenanceModeActiveHttpException.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ */
+class MaintenanceModeActiveHttpException extends ServiceUnavailableHttpException
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(
+        // FIXME: should we initialize this to a sane default?
```
leofeyer: Probably one hour?

src/Exception/BadRequestTokenHttpException.php
leofeyer: It is correct. We need to reload the page here.

src/EventListener/InitializeSystemListener.php
```diff
@@ -347,17 +365,11 @@ private function checkRequestToken()
...
-                header('HTTP/1.1 400 Bad Request');
-                die_nicely(
-                    'be_referer',
-                    'Invalid request token. Please <a href="javascript:window.location.href=window.location.href">go back</a> and try again.'
-                );
+                throw RedirectResponseException::createAjax(Environment::get('base') . 'contao/');
             }
 
-            exit; // FIXME: throw a ResponseException instead
+            // FIXME: obsolete when using symfony security?
```
leofeyer: @Toflar /cc

src/EventListener/InitializeSystemListener.php
```diff
@@ -285,14 +290,28 @@ private function validateInstallation(Config $config, Request $request = null)
...
+     * Ensure we do not have any error level active that could be harmful to the Contao 3 Framework.
+     */
+    private function sanitizeErrorHandling()
+    {
+        // Check if we have "simple errors" and make sure to kill them as the legacy code will otherwise break.
+        // This happens (at least) on console.
+        $level    = error_reporting();
+        $newLevel = $level & ~(E_NOTICE | E_STRICT | E_DEPRECATED);
+        if ($level & $newLevel) {
+            error_reporting($newLevel);
```
leofeyer: Not sure about this.

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,307 @@
...
+     * Try to read a template.
+     *
+     * @param string $template The file path to test.
+     *
+     * @return null|string
+     */
+    private function tryReadTemplate($template)
+    {
+        if (file_exists($template)) {
+            // Isolate the template parsing, the "unused" root dir variable will get used in the template.
```
leofeyer: I don't understand this :(

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,307 @@
...
+     * @param \Exception $exception The exception to check.
+     *
+     * @return null|Response
+     *
+     * @throws \Exception When anything went really haywire during rendering of the page.
+     */
+    private function tryToRenderContao404($exception)
+    {
+        // FIXME: I guess it is hopeless to render a 404 page from within a root page when no root page is published.
+        // We might want to skip RootNotFoundHttpException then and also NoPagesFoundHttpException is an candidate.
```
leofeyer: Do we really want to render the Contao 404 page upon **every** exception?

src/Resources/contao/classes/Ajax.php
discordier: Not yet but I already thought of implementing one.

src/Resources/contao/config/default.php
discordier: As long as you are ok with the fact that the users cause the system to not boot anymore I can re add it.

tests/EventListener/InitializeSystemListenerTest.php
discordier: As it expects the `InsecureInstallationHttpException` to be thrown which is part of my PR, yes it is related?

src/Resources/contao/templates/backend/be_no_root.html5
discordier: Every test scenario you can imagine, incomplete installation (no symlinks yet).

src/Resources/contao/modules/ModuleArticle.php
discordier: IMO we should never echo something as we might have sub requests etc. in the future.

src/Resources/contao/library/Contao/Template.php
discordier: Ok, I will demote it to a plain comment. I simply wanted to note why we did not use a response exception here.

src/Resources/config/services.yml
discordier: I have to check again which exact priority the symfony handler has. Will change accordingly and add comment.

src/Exception/MaintenanceModeActiveHttpException.php
discordier: Sounds reasonable to me.

src/EventListener/InitializeSystemListener.php
discordier: It is the result of heavy testing on my machine which is also the reason of killing the configurable error levels.
If we do not kill at least these, the core did not boot - will get even worse with extensions involved which cause recoverable errors due to improper type hints etc.
The whole installation will die then.
We could of course change our code to be safe against all these errors by putting heavy effort into fixing them but it is not worth the hassle.
I only dislike that we will loose the benefit of "enforcing" the users to write proper code for new extensions (code not emitting notices etc.).
My idea was to write an error handler for dev mode which will check where the notice originates from and accordingly either ignore it or not. Should not be a big problem as new code should never end up in there and therefore not slow down the system.
If you agree, I will provide an additional PR for this then.

src/EventListener/ExceptionListener.php
discordier: The error templates should not derive the `$this` context of the exception listener.

src/EventListener/ExceptionListener.php
discordier: Not every exception but only every descendant class of `NotFoundHttpException`, see the very next line.

src/Exception/MaintenanceModeActiveHttpException.php
aschempp: I don't think we should set a value. If we don't, the header will not be added. If we add a sane default, it should be set where the exception is thrown.

src/Resources/contao/config/default.php
leofeyer: What do you mean?

tests/EventListener/InitializeSystemListenerTest.php
leofeyer: Sorry, my bad. Didn't read the phpDoc comments :)

src/Resources/contao/config/default.php
aschempp: I think this will be part of an error-handler PR anyway?

src/Resources/contao/templates/backend/be_no_root.html5
leofeyer: Should we possibly be using Twig templates and the Symfony exception handler instead?

src/Resources/contao/library/Contao/Template.php
leofeyer: The `output()` method is deprecated anyway.

src/EventListener/InitializeSystemListener.php
leofeyer: We'll have to discuss this.

src/Resources/contao/modules/ModuleArticle.php
aschempp: We should use the `S` option as second parameter which returns the value. Probably should send it to a stream and use a stream response. I have never used streams though ;-)

src/EventListener/ExceptionListener.php
leofeyer: > The error templates should not derive the `$this` context of the exception listener.

This would probably be the better comment. Now I do understand :)

src/Resources/contao/templates/backend/be_no_root.html5
aschempp: We thought about this before. But that's not BC because I did change the templates for existing installations already (by copying to `/templates`).

src/EventListener/ExceptionListener.php
leofeyer: But there are places (even in Contao 3) where it is impossible to render the 404 page, because it would end in an endless loop.

src/Resources/contao/templates/backend/be_no_root.html5
leofeyer: You can still override them by copying them to `app/config/Resources`, can't you?

src/Exception/BadRequestTokenHttpException.php
aschempp: Then we should probably use `window.location.reload()`?

src/Resources/contao/templates/backend/be_no_root.html5
aschempp: Depending on the implementation, yes. But my existing adjustments will be gone. It would also require the backend template editor to support Twig etc., so I think this should be in a future version.

src/Exception/BadRequestTokenHttpException.php
leofeyer: Not sure but I recall that this did not work. The page has to be loaded again so the PHP session is initialized with a request token.

src/Resources/contao/templates/backend/be_no_root.html5
leofeyer: I don't agree. The Twig template loader might be a future feature but everything else is already there and could/should be used.

src/EventListener/ExceptionListener.php
aschempp: FYI this is what Symfony does in this case: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Routing/Loader/PhpFileLoader.php#L60-L71

src/Resources/contao/modules/ModuleArticle.php
Toflar: The `BinaryFileResponse` has to be used here as in 99.9% of all our cases. I don't have a valid use case for `StreamedResponse` in Contao, yet.

src/EventListener/ExceptionListener.php
leofeyer: I like the static method better. But we don't have to change it.

src/Resources/contao/templates/backend/be_no_root.html5
aschempp: It would mean error templates cannot be adjusted from the backend, which was currently possible.

src/Exception/BadRequestTokenHttpException.php
aschempp: maybe your code does a force-reload. And it would probably also reload the top frame instead of an iframe (which is better in theory). Guess we should live with that, though I don't like the button name. It should say `reload the page` ;-)

src/Resources/contao/templates/backend/be_no_root.html5
leofeyer: So what? We are only talking about error templates.

It was also possible to disable modules in the backend. And it was possible to define the URL suffix and some other settings. We did not care about this either, did we?

src/EventListener/InitializeSystemListener.php
Toflar: No idea either.

src/Exception/BadRequestTokenHttpException.php
leofeyer: > It should say `reload the page` ;-)

Agreed.

src/EventListener/ExceptionListener.php
discordier: Have a look at line 183 which covers exactly this case.

src/EventListener/ExceptionListener.php
discordier: @aschempp For symfony this works but we are in Contao and swallow errors and therefore I guess we will have a warning `assuming $this from incompatible context` and you will still have `$this` available.
Have not checked though.

aschempp: You will not have `$this` as it's a static method call. At least I assume so otherwise it would be a bug in Symfony ;-)

src/Exception/RedirectResponseException.php
discordier: Changed in 66b68e3c5cbfa6d5d4ee2888a238b83b25d76b4f.

src/EventListener/ExceptionListener.php
discordier: Changed in 62bd6d587f573625e34023d21a37ac42bef86e40.

src/Exception/BadRequestTokenHttpException.php
discordier: Changed in 45926cb3ea58f1abd7405983d9e315ea1c53810a

src/Resources/contao/config/default.php
discordier: If it needs to be configurable, yes that would be the right place then.

src/Resources/contao/library/Contao/Template.php
discordier: I remember that we agreed that it is but not marked as such in the code yet - changed in e408bf2.

src/Resources/contao/modules/ModuleArticle.php
discordier: The `Download` content element is a perfectly valid use case as it probably will stream a very large file.

discordier: This is btw. the reason why I wanted to steal the implementation from https://github.com/contao/core-bundle/pull/68#discussion_r27342929 into my PR.

Toflar: I said " I don't have a valid use case for StreamedResponse in Contao, yet.". I did not say I don't see a valid use case for streaming at all. I just said the `BinaryFileResponse` is the best fit for us as it does stream as well ;)

src/Resources/contao/config/default.php
leofeyer: Not necessarily. If we are using it in the new code, it should be defined in the `parameters.yml` file.

src/Resources/config/services.yml
discordier: I must be before `twig.exception_listener` and I guess before `security.exception_listener` as well but I am unsure about the latter one, maybe @Toflar or @aschempp know more what checks might be needed before.

src/Exception/MaintenanceModeActiveHttpException.php
discordier: So we have now 1:1, @Toflar up to you to vote.
I am still undecided but tend towards not specifying a default.

discordier:
I'd promote this to RTM if there are no objections left.
Only thing remaining to be resolved is https://github.com/contao/core-bundle/pull/70#discussion_r28164826

src/Exception/MaintenanceModeActiveHttpException.php
leofeyer: I have changed my vote to "don't set a default" value.

Toflar: I'm voting for setting a default of one hour. The RFC 2616 section "10.5.4 503 Service Unavailable" states

> If no Retry-After is given, the client SHOULD handle the response as it would for a 500 response. 

which again means 

> The server encountered an unexpected condition which prevented it from fulfilling the request.

This, however, is not true for our exception. It's not an unexpected condition that prevent the server from fulfilling the request. It's a scheduled and wanted maintenance mode and the client can retry the very same request somewhen in the future again.

leofeyer: So, if the crawler comes by every 5 minutes, we are sending it away for one hour? Even if the maintenance window was only 10 minutes?

Toflar: Yes, that's exactly what we do :-) Then it will come back in an hour and everything will be fine. If it comes back in an hour and the site is still in maintenance mode it will come back in yet another hour. However, if we do not send a `Retry-After` header, the crawler itself has to decide whether it wants to crawl again or it just declares this page as completely broken and comes back in 30 days...who knows :)

leofeyer: No, that's not right. A page which is crawled every five minutes should not have to wait for one hour. Instead, the crawler should come back at his regular cycle (every five minutes), which it will do.

Toflar: I'm pretty sure that if it comes back every 5 minutes and gets a `500 Internal Server Error` every time it will somewhen stop crawling that page :)

leofeyer: Maybe, but that's the point: it gets a 503 and not a 500! And I'm pretty sure that Google knows the difference :)

Toflar: Yeah, and that's why I quoted the RFC before which says every client that gets a 503 without a `Retry-After` header should treat this one just the same as a 500 :-)

leofeyer: Jeez, do you really think Google isn't any smarter than this?

Why are we discussing this? @aschempp voted -1, I voted -1 and @discordier tends to voting -1.

Toflar: Yeah, and I vote +1 because of the RFC. You never need to discuss if you don't feel like.

leofeyer: No hard feelings.

Toflar: It's simple: If you don't want to discuss, don't do it. It's 3 downvotes and one upvote. I'm fine with leaving it as it is now, go ahead an merge it so we can move on :)

aschempp:
Didn't we agree yesterday to remove the default messages for the exceptions?

discordier:
Which I did not change yesterday as you know I had to leave.
My comment about that there is nothing left to discuss was way before we discussed.

aschempp:
:+1:

leofeyer:
> My comment about that there is nothing left to discuss was way before we discussed.

LOL

leofeyer:
@discordier I have merged @Toflar's "stop using legacy code" PR and I have adjusted the test classes to not load the `Contao\Fixtures\Config` class by default. You might want to rebase and re-run your tests.

discordier:
Will do so.

leofeyer:
Here are things I stumbled across:

- The `LanguageHelper` is only used in the class `Fixtures\System`. It should therefore be a fixtures class, too.
- I think you can remove the `ExceptionListenerTest::setUp()` method.
- Can you please revert the `parent::setUp()` insertions and either create a separate PR or a ticket for it? The changes are not related to the exception handling and unnecessarily blow up the PR.
- Your deprecation comment for the `Template::output()` method suggests to throw a response exception, but the correct alternative is to use the `Template::getResponse()` method instead, isn't it?
- Since the class is called `ExceptionListener`, the service key should be `contao.listener.exception`.
- There are still a lot of exception classes with a default message (e.g. `$message = 'No active pages found'`). I don't remember what we have agreed on regarding this?
- I don't fully understand the purpose of the `InitializeSystemListener:: sanitizeErrorHandling()` method, but I'll be playing with it :)

Aside from these things, the PR looks good to me :)

leofeyer:
Wow, the new error screens look nice :)

But the error screens should not show up in debug mode (`app_dev.yml`), should they?

discordier:
The `setUp ()` calls have been extracted into #194.

The service key has been corrected in 6f85791.

The `InitializeSystemListener:: sanitizeErrorHandling()` has been simplified.

discordier:
The remaining issue https://scrutinizer-ci.com/g/contao/core-bundle/inspections/b5bcb4d4-b43e-4d3a-bc11-843787e4cdff/issues/files/src/EventListener/ExceptionListener.php?status=new&orderField=path&order=asc is a false positive.

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,416 @@
...
+     * Check if the Contao 404 should be rendered.
+     *
+     * @param \Exception $exception The exception to test.
+     *
+     * @return bool
+     */
+    private function isRenderingOfContao404PossibleFor($exception)
+    {
+        // If not an 404 exception or the frontend has not been booted, we can not render.
+        if (!($exception instanceof NotFoundHttpException && defined('BE_USER_LOGGED_IN'))) {
```
aschempp: Should we check the container scope instead of `BE_USER_LOGGED_IN`?

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,416 @@
...
+     *
+     * @throws \Exception When anything went really haywire during rendering of the page.
+     */
+    private function tryToRenderContao404($exception)
+    {
+        if (!$this->isRenderingOfContao404PossibleFor($exception)) {
+            return null;
+        }
+
+        static $processing;
```
aschempp: I think this should be a private static class variable (Symfony does that in `Symfony\Component\Config\Loader\FileLoader`).

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,416 @@
...
+            $processing = false;
+
+            // We can safely assume that when we get an not found exception when trying to render a not found
+            // page, that we have to render the internal page.
+            return null;
+        } catch (\Exception $pageException) {
+            $processing = false;
+            // Throwing the page exception here is intentional as we have some error handler problems!
+            // We should never end up here.
+            // Fixing error handling code paths is more important than the original exception.
```
aschempp: In this case we would not render a pretty error screen on a production system, right?

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,416 @@
...
+    /**
+     * Check if the exception is understood by us.
+     *
+     * @param \Exception $exception The exception to walk on.
+     *
+     * @param Request    $request   The HTTP request.
+     *
+     * @return null|Response
+     */
+    private function checkHttpException($exception, $request)
```
aschempp: The method arguments are missing type hints (e.g. `checkHttpException(\Exception $exception, Request $request)`. Same applies to other methods.

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,416 @@
...
+     *
+     * @return array|null
+     *
+     * @internal
+     *
+     * FIXME: replace this with a real - non legacy - language string loader.
+     */
+    protected function loadLanguageStrings()
+    {
+        System::loadLanguageFile('exception');
```
aschempp: This can only be called if the legacy framework is booted. Are we certain it is?

src/EventListener/InitializeSystemListener.php
```diff
@@ -388,17 +407,14 @@ private function handleRequestToken(Request $request = null)
...
-                header('HTTP/1.1 400 Bad Request');
-                die_nicely(
-                    'be_referer',
-                    'Invalid request token. Please <a href="javascript:window.location.href=window.location.href">go back</a> and try again.'
-                );
+                throw AjaxRedirectResponseException::create($this->router->generate('contao_backend'));
             }
 
-            exit; // FIXME: throw a ResponseException instead
+            // FIXME: obsolete when using symfony security?
```
aschempp: Can we not use the Symfony `InvalidCsrfTokenException` directly? Also interesting that their exception is not a "HTTP bad request".

src/Exception/AccessDeniedHttpException.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Contao\CoreBundle\Exception;
+
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException as BaseAccessDeniedHttpException;
+
+/**
+ * This exception is thrown when the access to a resource has been denied.
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ */
+class AccessDeniedHttpException extends BaseAccessDeniedHttpException
```
aschempp: It's somewhat strange we do handle 404 in our code to render the Contao page, but we don't do the same for a 403. What does a developer do if he wants to disallow access to a page? Throwing an `AccessDeniedHttpException` would be wrong I guess?

src/Exception/BadRequestHttpException.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Contao\CoreBundle\Exception;
+
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException as BaseBadRequestHttpException;
+
+/**
+ * This exception is thrown when there is an issue with the request.
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ */
+class BadRequestHttpException extends BaseBadRequestHttpException
```
aschempp: Why did we have to introduce a new exception for this? Are we intentionally only handling the Contao exceptions in our listener, but "ignoring" regular Symfony exceptions? That would mean if my extension uses a third-party bundle that throws such an exception, we would not pretty-render it?

src/Exception/NoContentResponseException.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Contao\CoreBundle\Exception;
+
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Exception to send an empty response and exit the execution in the Contao workflow.
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ */
+class NoContentResponseException extends AbstractResponseException
```
aschempp: I think that should be `EmptyResponseException` (even your comment says "send an empty response").

src/Resources/config/services.yml
```diff
@@ -36,6 +36,15 @@ services:
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 30 }
             - { name: kernel.event_listener, event: kernel.finish_request, method: onKernelFinishRequest, priority: -254 }
 
+    contao.listener.exception:
+        class: Contao\CoreBundle\EventListener\ExceptionListener
+        tags:
+            - { name: kernel.event_listener, event: kernel.exception, method: onKernelException }
+        arguments:
+            - "%contao.pretty_error_screens%"
+            - "@templating"
```
aschempp: As far as I know, you can use `@twig` directly in Symfony 2.7 (see http://symfony.com/blog/new-in-symfony-2-7-twig-as-a-first-class-citizen#using-twig-without-the-templating-sub-system)

src/Exception/NoContentResponseException.php
```diff
@@ -0,0 +1,38 @@
...
+     * Example:
+     *
+     *     throw NoContentResponseException::create();
+     *
+     * @param int   $status  The response status code
+     * @param array $headers An array of response headers
+     *
+     * @return ResponseException
+     */
+    public static function create($status = 204, $headers = array())
```
aschempp: This is somewhat strange. We cannot `throw new NoContentResponeException` without passing an empty response? Why is the factory necessary at all?

src/Resources/views/Error/error_layout.html.twig
```diff
@@ -0,0 +1,83 @@
+{% spaceless %}
```
aschempp: Isn't it common practice to call this file "base.html.twig" ?

src/Resources/views/Error/error_layout.html.twig
```diff
@@ -0,0 +1,83 @@
...
+<html>
+<head>
+    {% block head %}
+    <meta charset="utf-8">
+    <title>{% block title %}{{ statusCode }} - ERROR{% endblock %}</title>
+    <meta name="generator" content="Contao Open Source CMS">
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <link href='http://fonts.googleapis.com/css?family=Open+Sans:400,600,700,800,300' rel='stylesheet' type='text/css'>
+    <style>
+        {% include '@ContaoCore/error.css.twig' %}
```
aschempp: Why is the `error.css` not in the `Error` folder?

aschempp:
Looks very good to me! I have added a few comment and ideas. If we really need to merge now, I'm fine with merging first and then invalidate or resolve my questions :-)

src/EventListener/InitializeSystemListener.php
```diff
@@ -319,29 +324,43 @@ private function validateInstallation(Request $request = null)
...
     private function configureErrorHandling()
     {
         // Always show error messages if logged into the install tool (see #5001)
+        // FIXME: remove when the install tool is gone.
         if (Input::cookie('TL_INSTALL_AUTH') && !empty($_SESSION['TL_INSTALL_AUTH']) && Input::cookie('TL_INSTALL_AUTH') == $_SESSION['TL_INSTALL_AUTH'] && $_SESSION['TL_INSTALL_EXPIRE'] > time()) {
             $this->config->set('displayErrors', 1);
         }
 
         $this->iniSet('display_errors', ($this->config->get('displayErrors') ? 1 : 0));
-        error_reporting(($this->config->get('displayErrors') || $this->config->get('logErrors')) ? $this->config->get('errorReporting') : 0);
```
leofeyer: I actually had to re-add this line (now using `$this->container->getParameter('contao.error_level')` of course). Why did you remove it?

src/Resources/contao/classes/Ajax.php
```diff
@@ -155,11 +158,13 @@ public function executePreActions()
 					if ($e->getCode() == 0)
 					{
 						\System::loadLanguageFile('tl_maintenance');
-						echo '<p class="tl_error">' . $GLOBALS['TL_LANG']['tl_maintenance']['notWriteable'] . '</p>';
-						exit; break;
+						throw new ResponseException(
+							new Response('<p class="tl_error">' . $GLOBALS['TL_LANG']['tl_maintenance']['notWriteable'] . '</p>'),
+							$e
```
leofeyer: This seems wrong. You are passing the exception object but the `ResponseException` class actually expects an integer value (`$status`).

leofeyer:
I generally dislike the `create()` methods of the response exception classes.

```php
throw throw NoContentResponseException::create();
throw ResponseException::create($objWidget->generate());
```

Instead, why don't we modify the `AbstractResponseException` class as follows:

```php

```

leofeyer:
I generally dislike the `create()` methods of the response exception classes.

```php
throw NoContentResponseException::create();
throw ResponseException::create($objWidget->generate());
```

Instead, why don't we modify the `AbstractResponseException` class as follows:

```php
public function __construct($response = '', \Exception $previous = null)
{
    parent::__construct('Contao Response', 0, $previous);

    if (!is_object($response)) {
        $response = new Response($response);
    }

    $this->response = $response;
}
```

Then we can use the regular syntax to create exceptions:

```php
throw new NoContentResponseException();
throw new ResponseException($objWidget->generate());
```

Toflar:
I like the `create()` method because in contrast to what you (@leofeyer) propose, it's more explicit. If you have a string, you use  `create()`. If you have a `Response` instance, you don't :-) But I guess it's just a matter of preference but I definitely like it more the way it is now.

src/EventListener/InitializeSystemListener.php
discordier: Because it is useless.
The error level is set in the `sanitizeErrorLevel` method way before this line is called and setting it twice to the very same level is simply stupid.

src/EventListener/ExceptionListener.php
discordier: I can not. The main reason this exists is because of the lines the page model like:
https://github.com/contao/core-bundle/blob/52a32d2ee7cea17960e94a509e5f700948552de5/src/Resources/contao/models/PageModel.php#L275
When we end up here, the constant is not necessarily defined and will trigger a warning then.
Therefore it is impossible to use the page handler object (which uses the page model in return).

src/EventListener/ExceptionListener.php
discordier: And does it the way I do it in https://github.com/symfony/HttpKernel/blob/master/EventListener/ExceptionListener.php#L42 (which is actually where I got the idea from).

src/EventListener/ExceptionListener.php
discordier: Right, but as written in the comment this indicates an bug in the exception rendering itself and therefore we do not render the pretty screen.
We might want to change this into some external handler as written in the FIXME in the future though.

src/EventListener/ExceptionListener.php
discordier: Fixed in 6a24bf2eee9100e6b3b19c45ffa5afd123997efc.

src/EventListener/ExceptionListener.php
discordier: I added another check in f747359 but this also only a base test.
To be really bulled proof, I'd have to check for the class `System`, the constant `TL_ROOT` etc. as it is all wired together.

src/Resources/contao/classes/Ajax.php
aschempp: @leofeyer you're wrong, it does not (see `AbstractResponseException`).

aschempp:
I agree with both @leofeyer and @Toflar. For the general `ResponseException` I would accept a `Response` in `__construct`. But for a `NoContentResponseException` that does not make sense, it does not need a Response if there is no content. So the `NoContentResponseException` should have a different constructor, not a `create` method. Same with the `RedirectResponseException`.

src/EventListener/InitializeSystemListener.php
discordier: It is not even an `HttpException` and would therefore ultimately end up as being an `HTTP_INTERNAL_SERVER_ERROR`.

src/EventListener/ExceptionListener.php
aschempp: Ok then, fine with me :D

src/EventListener/ExceptionListener.php
aschempp: Well we could simply render our 500 error page, couldn't we? I'm not sure if it's an error in the exception handling, it's just an unknown error in rendering the Contao 404 page.

src/EventListener/InitializeSystemListener.php
aschempp: I know it's not, but I wonder if that does matter? I mean the goal of our handler is to show pretty errors & hide sensitive information instead of rendering exceptions. To me it would totally make sense to render our pretty screen for appropriate Symfony exceptions too.

src/EventListener/ExceptionListener.php
aschempp: I did mean `$this->container->isScopeActive('frontend')` instead of checking the constant. I assume you're checking the constant to see if we are in frontend mode?

src/Exception/AccessDeniedHttpException.php
discordier: Thought about that one too but have not yet implemented it as we might want to export the rendering into an sub request or the like.
Implementing it also in here will add the same amount to the PR.
Downside is also, that the 403 is throwing 404 exceptions internally and even redirect exceptions therefore implementing and testing would imply extracting of the rendering into an own sub handler.
I also thought about pushing parameters like the root page into the request attributes but doing so is beyond the basic functionality of this PR.

src/Exception/BadRequestHttpException.php
discordier: Exactly. If a third party bundle throws any exception you are responsible of catching and chaining/rethrowing it.
If any "unknown" exception is encountered in our "root exception handler", it is supposed to be a HTTP_INTERNAL_SERVER_ERROR as it is not intended for Contao and/or does not follow the normal workflow.

src/Exception/NoContentResponseException.php
discordier: The naming here reflects the HTTP status code which is `204 No Content`.
I tried to stay as close as possible to the HTTP status codes instead of making up arbitrary names from which you can not guess the status code.
Regarding the comment, I think "Exception to send an empty response" is easier to read than "Exception to send a response without content" or "Exception to send a response with no content".

@Toflar and @leofeyer what do you think?

src/Exception/BadRequestHttpException.php
aschempp: But that would mean our pretty error screen is not rendered for "regular" exceptions. I think that's wrong, we should render our error 500 pretty screen for everything that is not catched.

src/Exception/AccessDeniedHttpException.php
aschempp: I agree and created #197 to not forget about it :-)

src/EventListener/InitializeSystemListener.php
leofeyer: That's the point. It is **not** set twice to the same level. If neither `displayErrors` nor `logErrors` is enabled, it is set to `0`.

src/Exception/NoContentResponseException.php
leofeyer: +1 for the current name (`NoContentResponseException`).

aschempp: You're right, using the name of the HTTP code seems reasonable.

discordier:
I totally dislike the approach suggested by @leofeyer. 
Using the current method, you can use the factory method to create "common" use cases but have still the freedom of supplying the Response object on your own with code completion and parameter type hints.

Bear in mind that objects are perfectly valid content as well, as the `Response` invokes `__toString()` on them.
This will break then:
```php
throw new ResponseException(
    new \Exception('an Exception just as example class providing __toString()')
);
```
Second problem I see, how to specify the http status code when the constructor is changed instead of using a factory method?
Add it as another argument or do we force the users then to create an response if they want to use anything aside from 200?

discordier:
The `NoContentResponseException` could be changed though.
```php
/**
 * Construct the exception.
 *
 * @param int        $status   The response status code
 * @param array      $headers  An array of response headers
 * @param \Exception $previous The previous exception used for the exception chaining
 */
public function __construct($status = 204, $headers = array(), \Exception $previous = null)
{
    parent::__construct(new Response('', $status, $headers), $previous);
}
```

src/EventListener/InitializeSystemListener.php
discordier: It is set twice if any of the flags is specified. It get's reset to 0 otherwise which is of even more problematic nature as it is effectively killing the complete error handling of symfony, in all environments, I doubt that this is intended (it even corrupted the unit tests on my side which is actually the reason I dropped the setting of 0).
IMO we could drop the whole method as the logging should get redirected to the app/logs no matter what the displayErrors or logErrors flags of Contao say (logging is still an upcoming PR) and therefore honor what the application config says instead of arbitrarily modding the handlers.
I'd accept it if this would be restricted to the install tool - which is scheduled for removal anyway - but this modifies the behaviour of the whole application in a manner that can not be configured in the symfony way - the `contao.error_level` is already a hack I dislike but it is at least still in symfony environment.

src/Resources/views/Error/error_layout.html.twig
discordier: Umm... good question. Maybe because I added it later on and simply put it next to the reset CSS.

However we might want to add a `Css` folder etc. where we put the "flexible" stuff etc.

src/Exception/NoContentResponseException.php
Toflar: +1 for the current name (NoContentResponseException).

src/Resources/views/Error/error_layout.html.twig
discordier: Well, the `AcmeDemoBundle` notes `{% extends 'AcmeDemoBundle::layout.html.twig' %}` and as we are not the "main" framework template but only the frame for errors, I prefixed it with `error_`

discordier: The acme bundle is in fact even extending `TwigBundle::layout.html.twig`

src/Resources/config/services.yml
discordier: As both https://github.com/contao/standard-edition/blob/b0f82c6fb3210c40e47135b4d83d01ffad777dcb/composer.json#L19
and https://github.com/contao/core-bundle/blob/f7061486148e0c9b1ba9db72988912986133ee1e/composer.json#L21
are still depending on `~2.6` I'd keep it like it is.
Can be evaluated and changed later on - I simply currently lack the time to read this up.

src/Exception/BadRequestHttpException.php
discordier: Eh, it is for every exception.
The omitted exceptions are only unknown exception classes implementing `HttpExceptionInterface` for which I naturally do not have a template and therefore can not handle them properly.
But these exceptions will get swapped with the 500-error screen, when pretty screens are switched on, in lack of another exception template See (https://github.com/contao/core-bundle/pull/70/files#diff-cdf99235fbbeccfa12764a4be188ae32R114).

Affected (known) exceptions currently are:
```php
\Symfony\Component\HttpKernel\Exception\HttpException
\Symfony\Component\HttpKernel\Exception\BadRequestHttpException
\Symfony\Component\HttpKernel\Exception\UnsupportedMediaTypeHttpException
\Symfony\Component\HttpKernel\Exception\ConflictHttpException
\Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException
\Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException
\Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException
\Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException
\Symfony\Component\HttpKernel\Exception\UnprocessableEntityHttpException
\Symfony\Component\HttpKernel\Exception\PreconditionFailedHttpException
\Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException
\Symfony\Component\HttpKernel\Exception\PreconditionRequiredHttpException
\Symfony\Component\HttpKernel\Exception\GoneHttpException
\Symfony\Component\HttpKernel\Exception\NotFoundHttpException
\Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException
\Symfony\Component\HttpKernel\Exception\LengthRequiredHttpException
```

src/EventListener/InitializeSystemListener.php
discordier: See https://github.com/contao/core-bundle/pull/70#discussion_r28315881 for details, it is hidden but not with an error screen reflecting the intended result but the generic 500 (as it was in legacy code).

leofeyer:
Since we cannot wait any longer to release a beta version, I have adjusted and merged the branch in 9a29ae04fdec5c19053a707a041985369851ae9b. There are still some FIXMEs in the code, which we should take care of.

@discordier I have only read your comment regarding the removal of the `Exception::create()` method right now, but I had already changed it in my branch. We still can discuss it and change it again later on. Sorry about that.

Toflar:
BTW: I would like to thank @discordier for this PR, there's some excellent work that has been done here! :+1:

src/EventListener/InitializeSystemListener.php
leofeyer: We can discuss this. But if both `displayErrors` and `logErrors` is disabled, what do you need error reporting for at all?

src/Resources/views/Error/error_layout.html.twig
```diff
@@ -0,0 +1,83 @@
+{% spaceless %}
+<!DOCTYPE html>
+<html>
+<head>
+    {% block head %}
+    <meta charset="utf-8">
+    <title>{% block title %}{{ statusCode }} - ERROR{% endblock %}</title>
+    <meta name="generator" content="Contao Open Source CMS">
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <link href='http://fonts.googleapis.com/css?family=Open+Sans:400,600,700,800,300' rel='stylesheet' type='text/css'>
```
leofeyer: Do we really need all this font weights? It seems there are only two in use.

src/Resources/views/Error/error_layout.html.twig
leofeyer: I think it should be `ContaoCoreBundle:Error:layout.html.twig`.

leofeyer:
There is a problem in the `ExceptionListener` class on line 232:

```php
$response = $pageHandler->getResponse(false);
```

The `getResponse()` method expects the numeric page ID to be passed as argument and we are no longer handling the `$blnUnusedGet` case. This will lead to wrong log entries.

src/Resources/views/Error/error_layout.html.twig
discordier: This is up to the css finetuning which will be provided by @frontendschlampe.

discordier:
Regarding the value of passing `false` as `pageId` that is the reason why I left the original handlers in place so far and I stole the fallback value of false from: https://github.com/contao/core-bundle/blob/d0a75e4ba59143869c8d7995c4f377570f6c82e7/src/Resources/contao/controllers/FrontendIndex.php#L83 where it is indeed false. So the problem is already present in Contao 3.

discordier:
I will provide PRs as amendment regarding the stuff that has not been merged as soon as I find the time for it.

I see however another problem, as you have dropped the Contao exception classes and use the base classes from Symfony now, we will hijack all exceptions inheriting from that base class no matter if it is intended for us or not. Might be an undesired side effect. see also https://github.com/contao/core-bundle/pull/70#discussion_r28315881

The Problem described in https://github.com/contao/core-bundle/pull/70#discussion_r28313005 is also present again in current code.

leofeyer:
I don't fully understand, but we'll have time tomorrow to discuss it.

src/Resources/views/Error/error_layout.html.twig
aschempp: Why are they templates at all? They should simply be assets?

leofeyer: I guess they are templates because the CSS code shall be rendered inline.

src/Resources/views/Error/error_layout.html.twig
aschempp: I think so too, because we're already in the `Error` "namespace" the `error_` prefix is unnecessary.

leofeyer: Changed in 40a1a4cc1c73955791d066426615ab80500718d2.

PR #72, opened 2015-03-20
[RTM] Insecure document root is allowed for localhost

aschempp:
see #67 

/cc @dtomasi

### DIFF
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -301,6 +301,7 @@ class InitializeSystemListener
         }
 
         // Show the "insecure document root" message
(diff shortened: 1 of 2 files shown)


Toflar:
Tests are missing :)

aschempp:
I have updated the code to check the IP instead of host to allow for IPv6.

@Toflar it's currently using `die_nicely` so tests would be hard ;-)

dtomasi:
+1

Thanks! That´s also fine for me!

Toflar:
> @Toflar it's currently using die_nicely so tests would be hard ;-)

Then we need to add a  `FIXME` :)

aschempp:
Agreed, `FIXME` added.

PR #73, opened 2015-03-21
[RTM] Remove module checks

aschempp:
These checks should be removed. They are not compatible with the current ResourceProvider and they won't be compatible with the new solution I'm working on.

I just extracted that so we could *discuss* that separately.

### DIFF
--- a/src/Resources/contao/dca/tl_templates.php
+++ b/src/Resources/contao/dca/tl_templates.php
@@ -227,7 +227,7 @@ class tl_templates extends Backend
 		foreach ($kernel->getContainer()->get('contao.resource_provider')->getResourcesPaths() as $strModule => $strFolder)
 		{
 			// Continue if there is no templates folder
-			if ($strModule == 'repository' || !is_dir($strFolder))
+			if (!is_dir($strFolder))
 			{
 				continue;
 			}
--- a/src/Resources/contao/library/Contao/Database/Installer.php
+++ b/src/Resources/contao/library/Contao/Database/Installer.php
@@ -337,13 +337,8 @@ class Installer extends \Controller
 		$return = array();
 
 		// Only check the active modules (see #4541)
-		foreach ($kernel->getContainer()->get('contao.resource_provider')->getResourcesPaths() as $strModule => $strFolder)
+		foreach ($kernel->getContainer()->get('contao.resource_provider')->getResourcesPaths() as $strFolder)
 		{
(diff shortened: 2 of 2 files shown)


leofeyer:
We can remove neither of them. It will still be possible to install the ER2 as a legacy bundle, so we have to keep the first check. And as long as we support using `database.sql` files, we have to keep the second check, too.

aschempp:
I disagree. Installing the repository should not be supported. Obviously people could try to extract the extension from Contao 3, but that would result in templates showing up. Fine with me…

Regarding the second, that's a hack because the same problem would affect every extension but it's hardcoded to the core ones. Also, we agreed that it should be possible to run Contao 3 extensions in Contao 4, but we don't want to put efforts in an update process. This "fix" would only affect systems updated from Contao 2.

src/Resources/contao/dca/tl_templates.php
```diff
@@ -227,7 +227,7 @@ public function addNewTemplate()
 		foreach ($kernel->getContainer()->get('contao.resource_provider')->getResourcesPaths() as $strModule => $strFolder)
 		{
 			// Continue if there is no templates folder
-			if ($strModule == 'repository' || !is_dir($strFolder))
```
leofeyer: I think this is wrong. Isn't `$strFolder` the Contao resource path? Should it not be `$strFolder . '/templates`?

aschempp: You're right, but this is already fixed in my PR to `resource_locator`.

leofeyer:
> Also, we agreed that it should be possible to run Contao 3 extensions in Contao 4, but we don't want to put efforts in an update process.

That's not true. We don't want to put efforts in an update process right now, but certainly soon.

> This "fix" would only affect systems updated from Contao 2.

That's not true either. The `database.sql` files are supported in Contao 3, too. They have not been deprecated yet.

src/Resources/contao/dca/tl_templates.php
leofeyer: Can I still ask you to fix it here? It's kind of confusing to merge faulty code.

aschempp: That will result in a conflict in the other branch…

aschempp:
> That's not true either. The database.sql files are supported in Contao 3, too. They have not been deprecated yet.

Correct, but that's only for the core modules if they contain SQL files. And that can only be the case for a Contao 2 upgrade.

PR #74, opened 2015-03-22
[RTM] Set TL_SCRIPT to the relative base path

aschempp:
Unfortunately, the fix in #69 was wrong, because `$route` was an absolute path including the base path.


leofeyer:
Can you fix the code please, so the tests no longer fail?

aschempp:
Done

PR #75, opened 2015-03-22
[RTM] Replace ResourceProvider with resource locators

aschempp:
Now that resources are in `@Bundle/Resources/contao`, we have to rewrite the resource handling (see https://github.com/contao/symfony-todo/issues/6#issuecomment-84031211)

Be aware that the commits also contain #73 as it's an essential part of rewriting the resource finder.

## Locating files

The main component is our `contao.resource_locator` which can locate resources in a bundle. See http://symfony.com/doc/current/components/config/resources.html#locating-resources

Contrary to the previous approach, the resource locator does not find arbitrary locations (i.e. glob or regex support), but it can also find folders inside the `Resources/contao` directories.

There is also an additional `contao.cached_resource_locator` which checks for an internal cache file before finding resources in bundles. This greatly simplifies the logic to load resources either from cache or bundle directories.

### Adding custom locators

If someone really does have resources in other places then `Resources/contao`, they can add a service tagged with `contao.resource_locator`. A compiler pass will take care of registering a chain service to ask all resource locators (this is similar to how Twig does loaders in https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/TwigLoaderPass.php#L39)

## Dumping *(aka building internal cache)*

Working through our code and the Symfony code, I stumbled upon several version of `DumperInterface` in the Symfony world. This made me aware of how building the internal cache should work. So I also rewrote a dumper class that can combine several files into one.

### Using file loaders

Because we have PHP files but also XLF files for languages, I implemented different file loaders to take care of extracting content of the individual files. 
As an example, this would also make it fairly easy to add new means of adding DCA configuration (e.g. YML files).

## Open Questions
 - [x] Languages are currently cached in `system/cache/language/..`. Can we rename that folder to `languages/..` instead (adding the *s*)? This would allow to use the `contao.cached_resource_locator`. I doubt that would be a legacy problem?
 - [x] Public folders are only found in `system/modules`. Public files in Symfony bundles should be located in `Resources/public` (which is already the case for our bundles). Can we agree this is ok?
 - [x] Do we need a priority option for the tagged resource locators? Currently the order is unknown.
 - [x] Any reason we cannot use the kernel cache directory for our internal cache?

## TODOs
 - [x] ~~A compiler pass should alias `contao.resource_locator` to `contao.cached_resource_provider` if `bypassCache` is enabled in the system configuration (or kernel is in debug mode).~~ 
 - [x] Write unit tests


src/Config/Dumper/DumperInteface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+namespace Contao\CoreBundle\Config\Dumper;
+
+/**
+ * DumperInteface is the interface implemented by internal cache dumpers.
+ *
+ * @author Andreas Schempp <https://github.com/aschempp>
+ */
+interface DumperInteface
```
xchs: I suppose the file name, the interface name etc. should read "DumperInte**r**face", shouldn't it?

src/Config/Dumper/CombinedFileDumper.php
```diff
@@ -0,0 +1,97 @@
...
+
+use Symfony\Component\Config\Loader\LoaderInterface;
+
+/**
+ * CombinedFileDumper combines multiple files into one PHP file
+ *
+ * @author Andreas Schempp <https://github.com/aschempp>
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class CombinedFileDumper implements DumperInteface
```
xchs: Typo: DumperInte**r**face

src/DependencyInjection/Compiler/ResourceLocatorPass.php
```diff
@@ -0,0 +1,47 @@
...
+ * @author Andreas Schempp <https://github.com/aschempp>
+ */
+class ResourceLocatorPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        // register additional resource locators
```
xchs: **R**egister

src/Config/Dumper/DumperInteface.php
aschempp: You're absolutely right, fixed in eaedf77

xchs:
Regarding eaedf773b3fb4ebbafa449bc56ca391d537235c0: Did you already fix the typo in the file name (DumperInte**r**face.php)?

aschempp:
@xchs thanks again, now I did (in 6b8a4d5c7773aee7189e1464f359ee813f54db1d) ;-)

xchs:
Okay, well done. :+1:

leofeyer:
> Languages are currently cached in `system/cache/language/...`. Can we rename that folder to `languages/..` instead (adding the *s*)

Yes, we can.

> Public folders are only found in `system/modules`. Public files in Symfony bundles should be located in `Resources/public` (which is already the case for our bundles). Can we agree this is ok?

Yes.

> Do we need a priority option for the tagged resource locators? Currently the order is unknown. 

It should still be possible to override config settings, DCA settings, labels and templates, so I guess the loading order matters?

 Any reason we cannot use the kernel cache directory for our internal cache?

As long as it is possible to purge the cache via Automator (back end and command line), I have no objections.

aschempp:
I have changed the languages stuff and added priority sorting for locators. Any other opinions on the public folder part?

> As long as it is possible to purge the cache via Automator (back end and command line), I have no objections.

Purging the Symfony cache is currently done using `console cache:clear`, I guess we should build a maintenance task for that anyway.

src/Config/ChainFileLocator.php
```diff
@@ -0,0 +1,85 @@
...
+
+        if (!$first) {
+            return $files;
+        }
+
+        throw new \InvalidArgumentException(sprintf('No locator was able to find the file "%s".', $name));
+    }
+
+    /**
+     * If we're looking for all files, revers locator order so that higher priority overwrites lower priority locators.
```
Toflar: Spelling mistake

aschempp: fixed

aschempp:
I'm done here :-)

src/Resources/config/services.yml
```diff
@@ -35,10 +35,29 @@ services:
...
+            - "@kernel"
+        tags:
+            - { name: contao.resource_locator }
+
+    contao.resource_locator.chain:
+        class: Contao\CoreBundle\Config\ChainFileLocator
+        public: false
+
+    contao.resource_locator.cache:
+            class: Contao\CoreBundle\Config\CombinedFileLocator
```
Toflar: Should be `public: false`.

src/Resources/contao/library/Contao/Automator.php
```diff
@@ -492,31 +497,15 @@ public function generateConfigCache()
 		/** @var KernelInterface $kernel */
 		global $kernel;
 
-		$resources = $kernel->getContainer()->get('contao.resource_provider');
+		$objLocator = $kernel->getContainer()->get('contao.resource_locator');
+		$objDumper  = new CombinedFileDumper(
+            $kernel->getContainer()->get('filesystem'),
```
Toflar: CS

src/Resources/contao/library/Contao/Automator.php
```diff
@@ -563,23 +552,34 @@ public function generateDcaCache()
...
-			$cacheFiles[$fileName]->append(static::readPhpFileWithoutTags($file->getPathname()));
 		}
 
-		foreach ($cacheFiles as $cacheFile) {
-			$cacheFile->close();
+		$arrFiles   = array_values(array_unique($arrFiles));
+		$objLocator = $kernel->getContainer()->get('contao.resource_locator');
+		$objDumper  = new CombinedFileDumper(
+            $kernel->getContainer()->get('filesystem'),
+            new PhpFileLoader(),
```
Toflar: CS

src/Resources/contao/library/Contao/Automator.php
```diff
@@ -616,22 +616,43 @@ public function generateLanguageCache()
 		}
 
 		$arrLanguages = array_unique($arrLanguages);
-		$arrResourcesPaths = $kernel->getContainer()->get('contao.resource_provider')->getResourcesPaths();
+		$objLocator   = $kernel->getContainer()->get('contao.resource_locator');
+		$objDumper    = new CombinedFileDumper(
+            $kernel->getContainer()->get('filesystem'),
```
Toflar: CS

src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -331,26 +324,15 @@ public function getFromDca()
 	 */
 	public function getFromFile()
 	{
+        /** @var KernelInterface $kernel */
```
Toflar: CS

src/Command/SymlinksCommand.php
```diff
@@ -135,6 +151,18 @@ private function symlinkThemes($rootDir, $output)
     }
 
     /**
+     * Generates a symlink relative to the given path
+     *
+     * @param string $path    The path to create symlink of
+     * @param string $rootDir The root directory
+     * @param string $output  The output object
+     */
+    private function relativeSymlink($path, $rootDir, OutputInterface $output)
```
leofeyer: Can we rename the method to `createSymlink()`?

src/Config/CombinedFileLocator.php
```diff
@@ -0,0 +1,66 @@
...
+     * @param bool        $first       If true and no cache file exists, the exception will be thrown.
+     *
+     * @return string|array The full path to the file or an array of file paths
+     *
+     * @throws \InvalidArgumentException When file is not found
+     */
+    public function locate($name, $currentPath = null, $first = false)
+    {
+        // FIXME: we should inject this configuration
+        if (!\Config::get('bypassCache')) {
```
leofeyer: This is wrong. If we need this setting in the new code, it has to be moved to the `parameters.yml` file. However, it should just take the cache file if it exists just like anything else in Symfony.

src/Config/CombinedFileLocator.php
```diff
@@ -0,0 +1,66 @@
...
+            $cacheFile = $this->cacheDir . DIRECTORY_SEPARATOR . $name;
+
+            if (file_exists($cacheFile)) {
+                return $first ? $cacheFile : [$cacheFile];
+            }
+        }
+
+        // If $first is true, the cached file was wanted so don't ask the locator
+        if (!$first && null !== $this->locator) {
```
leofeyer: I don't really understand what `$first` is for?

src/Config/FileLocator.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param string|null $currentPath Not used
+     * @param bool        $first       Whether to return the first occurrence or an array of filenames
+     *
+     * @return string|array The full path to the file or an array of file paths
+     *
+     * @throws \InvalidArgumentException When file is not found and $first is true
+     */
+    public function locate($name, $currentPath = null, $first = false)
+    {
+        if ('' == $name) {
```
leofeyer: Should be `===`.

src/Config/Loader/XliffFileLoader.php
```diff
@@ -0,0 +1,189 @@
...
+     * Read the contents of a PHP file, stripping the opening and closing PHP tags
+     *
+     * @param string      $file A PHP file path
+     * @param string|null $type The resource type
+     *
+     * @return string The PHP code without the PHP tags
+     */
+    public function load($file, $type = null)
+    {
+        $language = $type ?: 'en';
```
leofeyer: Why is there a `$type` variable, which also works as language tag?

src/Config/Loader/XliffFileLoader.php
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * @return string
+     *
+     * @throws \OutOfBoundsException If less than 2 or more than 4 chunks are given.
+     */
+    private function getStringRepresentation(array $chunks, $value)
+    {
+        switch (count($chunks)) {
+            case 2:
+                return "\$GLOBALS['TL_LANG']['" . $chunks[0] . "'][" . $this->quoteKey($chunks[1]) . "] = " . $this->quoteValue($value) . ";\n";
```
leofeyer: Now the values are no longer actually set. It this correct?

src/ContaoCoreBundle.php
```diff
@@ -47,10 +46,8 @@ public function boot()
      */
     public function build(ContainerBuilder $container)
     {
-        $container->addCompilerPass(new AddContaoResourcesPass($this->getPath() . '/Resources/contao'));
+        parent::build($container);
```
leofeyer: There is no need to call `parent::build()` (check the method signature).

src/Resources/config/services.yml
```diff
@@ -35,10 +35,30 @@ services:
...
+        public: false
+
+    contao.resource_locator.cache:
+        class: Contao\CoreBundle\Config\CombinedFileLocator
+        public: false
+        arguments:
+            - '%kernel.cache_dir%/contao'
+            - "@contao.resource_locator"
+
+    contao.resource_locator:
```
leofeyer: What is this alias for?

src/Config/CombinedFileLocator.php
aschempp: The `$first` parameter is defined in the interface. It tells the locator to return the first file found (instead of all files found).

src/Config/Loader/XliffFileLoader.php
aschempp: Because the interface defines this variable, and it can be used for anything suitable for the loader.

src/Config/Loader/XliffFileLoader.php
aschempp: They are set in `addGlobal`

src/Resources/config/services.yml
aschempp: The alias is what should actually be used in the code. The others are pre-definitions for the container builder pass. Depending on if there are loaders or not, the alias will be changed. This is the same as Twig does with it's loaders.

src/ContaoCoreBundle.php
aschempp: We don't know if that will change. I followed the Symfony example here: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/FrameworkBundle.php#L66

src/Command/SymlinksCommand.php
aschempp: I don't think that would be clearer? There's already a method called `symlink`, and now one `createSymlink` :D

src/Config/CombinedFileLocator.php
aschempp: I agree, that's why there is a FIXME. However, I was unsure about the option anyway, like you said Symfony usually uses cache if there is any etc.
There would be a lot of other places to adjust if we remove the option, so I think it's fine to keep for this PR but let's open an issue to remove the option or move it to parameters.

leofeyer: I don't agree. Since this is new code, it should follow the Symfony standard instead of using legacy stuff.

aschempp: But refactoring this is not part of this PR, don't you think so? It would also include removing language files etc. And we would need to discuss how we handle such options in the future.

aschempp: Also, it should mean we don't want internal cache in dev environment etc.

leofeyer: If the cache exists, it is being used (Symfony default). To bypass the internal cache, we should rather check for `kernel.debug`, shouldn't we?

aschempp: > If the cache exists, it is being used (Symfony default).

Not necessarily. For example the container is re-built on every request in dev mode (or debug mode I guess).

> To bypass the internal cache, we should rather check for kernel.debug, shouldn't we?

We should actually create a new property that defaults to `kernel.debug`, but can be changed independently. Same as `%twig.debug%` (see https://github.com/contao/standard-edition/blob/develop/app/config/config.yml#L24)

src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -331,26 +324,15 @@ public function getFromDca()
...
 		global $kernel;
 
 		$table = '';
 		$return = array();
 
 		// Only check the active modules (see #4541)
-		foreach ($kernel->getContainer()->get('contao.resource_provider')->getResourcesPaths() as $strModule => $strFolder)
+		foreach ($kernel->getContainer()->get('contao.resource_locator')->locate('config/database.sql') as $strFile)
 		{
-			if (in_array($strModule, array('calendar', 'comments', 'faq', 'listing', 'news', 'newsletter')))
```
leofeyer: This check must not be removed until Contao 5 (see #162).

leofeyer: This check must not be removed until Contao 5 (see #162).

aschempp: I still disagree. Did you think about that? The modules have been moved from `system/modules` to `vendor/contao/*-bundle`. Do you really think the `database.sql` will still be there?

leofeyer: Not in the bundle, but in the old modules (e.g. `system/modules/calendar`). But this is a different problem, which has to be taken care of in the autoloader, so you are right eventually.

aschempp: You mean there could be a `system/modules/core` in a Contao 4 because of an update-leftover?
I think we should not handle that. Simply because the same could happen to Isotope, and there is no other correct solution than removing it manually.

src/Config/CombinedFileLocator.php
leofeyer: > We should actually create a new property that defaults to `kernel.debug`, but can be changed independently. Same as `%twig.debug%` (see https://github.com/contao/standard-edition/blob/develop/app/config/config.yml#L24)

Agreed. Go ahead :)

src/Config/Loader/XliffFileLoader.php
discordier: I somewhat dislike that we may only have 2-4 chunks.

Have a look at my implementation at https://github.com/cyberspectrum/contao-toolbox/blob/master/src/CyberSpectrum/Translation/Contao/File.php#L182 which allows arbitrary length in transcoding from xlf to contao php and vice versa.

src/ContaoCoreBundle.php
discordier: It is always a good idea to call the parent implementation as long as it is not abstract unless one intentionally wants to replace the functionality. I am with @aschempp on this one.

src/Config/Loader/XliffFileLoader.php
aschempp: I agree, but this is just the current implementation. We could also use the Symfony XliffLoader, but I guess we will move to the translator somewhen and this will be old stuff anyways.

discordier: Unfortunately we can not use the Symfony loader let alone the Symfony Translator framework. 
They are using source => target instead of the id of the translation in XLIFF and are translating in the gettext style using translation domains.

aschempp: That's only partially true (I did write an importer once), but I know what you mean. I did mean we might support the translator for new language files somewhen and should not need improve the old one.

leofeyer:
Again, some open questions remain:

1. The `ChainFileLocator::getLocators()` method did not work correctly, because you did not use the return value of `array_reverse()`. Now that I have fixed it, the `ChainFileLocatorTest::testLocateAll()` fails due of a wrong array order.

Why do we need to reverse the locator order at all?

2. The tests fail because of the `ClassLoader` class requiring a non-existing service:

`You have requested a non-existent service "contao.cached_resource_locator"`

leofeyer:
And this:

```yml
    contao.resource_locator.bundles:
        class: Contao\CoreBundle\Config\FileLocator
        public: false
        arguments:
            - "@kernel" # FIXME: kernel.bundles?
        tags:
            - { name: contao.resource_locator }
```

Why are we not passing `kernel.bundles` as argument?

aschempp:
Fixes available in #164 (still WIP). Should close this discussion.

PR #76, opened 2015-03-22
[RTM] Check if framework is available

### LINKED ISSUE #71: Contao framework not initialized upon invalid route
If you enter a URL that is not matched by a route (e.g. `http://contao.local/`), the Contao framework does not get initialized:

```
Fatal error: Class 'Contao\Frontend' not found in ~/contao/standard-edition/vendor/contao/core-bundle/src/EventListener/OutputFromCacheListener.php on line 31
```

The problem is that the Symfony exception listener is a sub request and we have recently added the following lines:

```php
// InitializeSystemListener::onKernelRequest()
if (!$event->isMasterRequest()) {
    return;
}
```

@aschempp How do we handle this case?

aschempp:
fixes #71


src/EventListener/AddToSearchIndexListener.php
```diff
@@ -27,6 +27,12 @@ class AddToSearchIndexListener
      */
     public function onKernelTerminate(PostResponseEvent $event)
     {
+        if (!defined('TL_ROOT')) {
```
Toflar: Add a `FIXME:` here please, because we must not use constants anymore.

src/EventListener/AddToSearchIndexListener.php
```diff
@@ -27,6 +27,12 @@ class AddToSearchIndexListener
...
+            return;
+        }
+
+        // TODO: we should also find a way to check for the master reqeust
```
Toflar: Switch this to a `FIXME` :)

aschempp: done

aschempp:
I have updated this PR to include the part about not booting the framework if there is no request scope. I also figured out we incorrectly checked for the request scope instead of the container scope in all places, but that's fixed here too.

src/DependencyInjection/Compiler/AddPackagesPass.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        foreach ($composerData as $package) {
+            $name = str_replace("'", "\\'", $package['name']);
+            $version = substr($package['version_normalized'], 0, strrpos($package['version_normalized'], '.'));
+
+            if (preg_match('/^[0-9]+\.[0-9]+\.[0-9]+$/', $version)) {
+                $packages[$name] = $version;
+            } elseif (isset($package['extra']['branch-alias'][$package['version_normalized']])) {
+                // FIXME: this is wrong, 4.0.x-dev would result in 4.0.999999
```
leofeyer: Why do you think this is wrong?

src/EventListener/InitializeSystemListener.php
```diff
@@ -301,7 +299,7 @@ private function validateInstallation(Config $config, $routeName)
 
         // Show the "insecure document root" message
         if ('/web' === substr(Environment::get('path'), -4)) {
-            die_nicely('be_insecure', 'Your installation is not secure. Please set the document root to the <code>/web</code> subfolder.');
+            //die_nicely('be_insecure', 'Your installation is not secure. Please set the document root to the <code>/web</code> subfolder.');
```
leofeyer: We should not commit this, should we?

leofeyer:
> I also figured out we incorrectly checked for the request scope instead of the container scope

What did we do wrong?

src/DependencyInjection/Compiler/AddPackagesPass.php
aschempp: Because 4.0.x-dev should be smaller than 4.0.0.

src/EventListener/InitializeSystemListener.php
aschempp: doh, that was from local testing. Fixed it

aschempp:
> > I also figured out we incorrectly checked for the request scope instead of the container scope

> What did we do wrong?

We must not check `$request->attributes->get('_scope')` instead we must check `$container->isScopeActive('…')`.

I also moved the packages information to the container. Apart from the advantage of caching and not being in the System class, it was necessary to have the Contao version available in the new code (in this case for the profiler) even if the legacy framework is not booted.

For all the changed I changed the title back to RFC for @contao/developers to review.

src/DependencyInjection/Compiler/AddPackagesPass.php
leofeyer: Nope: https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L208

aschempp: I see, so they assume that the `4.0.x-dev`-branch is a maintenance version of `4.0`. That would only be correct for hotfix branched, but not for `develop`. But they also keep the dev-flag, we should probably keep that too?

tests/bootstrap.php
```diff
@@ -47,5 +47,6 @@ class_alias("Contao\\$class", $class);
 
 /** @var Composer\Autoload\ClassLoader $loader */
 $loader->addPsr4('Contao\\CoreBundle\\Test\\', __DIR__);
+$loader->addPsr4('Contao\\Fixtures\\', __DIR__.'/Fixtures');
```
leofeyer: This cannot be right. Do you mean `'Contao\\CoreBundle\\Test\\Fixtures'`?

src/Command/InstallCommand.php
```diff
@@ -64,15 +64,14 @@ protected function configure()
      */
     protected function executeLocked(InputInterface $input, OutputInterface $output)
     {
-        $fs      = new Filesystem();
-        $rootDir = dirname($this->getContainer()->getParameter('kernel.root_dir'));
+        $fs = new Filesystem();
 
         foreach ($this->emptyDirs as $path) {
-            $this->addEmptyDir("$rootDir/$path", $fs, $output);
+            $this->addEmptyDir(getcwd() . "/$path", $fs, $output);
```
leofeyer: Why did you change this to `getcwd()`? Now the script can only be run from the installation root directory, which is wrong. It is ok to use `getcwd()` in a Composer script handler, because Composer **always** runs in the installation root directory. But `app/console` does not, so please revert this.

src/EventListener/AddToSearchIndexListener.php
```diff
@@ -27,6 +27,11 @@ class AddToSearchIndexListener
      */
     public function onKernelTerminate(PostResponseEvent $event)
     {
+        // FIXME: should be replaced with "Response implements Indexable" (see https://github.com/contao/symfony-todo/issues/3)
+        if (!defined('TL_ROOT')) {
```
leofeyer: Why did you add an additional check? `Frontend::indexPageIfApplicable()` will decide itself when indexing is applicable (hence the method name).

src/EventListener/ContainerScopeListener.php
```diff
@@ -40,9 +41,9 @@ public function __construct(ContainerInterface $container)
     /**
      * Enters the container scope when a route has been found.
      *
-     * @param FilterControllerEvent $event
+     * @param GetResponseEvent $event
      */
-    public function onKernelController(FilterControllerEvent $event)
+    public function onKernelRequest(GetResponseEvent $event)
```
leofeyer: Why are we now listening on the `kernel.request` event?

src/Resources/config/services.yml
```diff
@@ -9,7 +9,7 @@ services:
         arguments:
             - "@service_container"
         tags:
-            - { name: kernel.event_listener, event: kernel.controller, method: onKernelController, priority: 254 }
+            - { name: kernel.event_listener, event: kernel.controller, method: onKernelRequest, priority: 31 }
```
leofeyer: I assume you have set the priority to `31` on purpose. Can you elaborate please?

leofeyer:
Aside from my comments above and the failing tests, this PR looks good to me. Can you finish it and then change the title to `[RTM]` please?

tests/bootstrap.php
aschempp: This is part of #159, @dtomasi needs to answer that.

src/Command/InstallCommand.php
aschempp: That's merged from your commit (f02f9ac8faf14b4f4c1433fbe0fe4b813e5d6b54)

src/EventListener/AddToSearchIndexListener.php
aschempp: Class `Frontend` would not be available without the framework.

src/EventListener/ContainerScopeListener.php
aschempp: Because otherwise the container scope would not be set in the `InitializeSystemListener`

src/Resources/config/services.yml
aschempp: Priority 32 is the routing listener, priority 30 is the `InitializeSystemListener`. The container scope must be set between them.

src/Command/InstallCommand.php
leofeyer: Oh, good we noticed this :)

src/Resources/config/services.yml
leofeyer: Should we probably set the priority of the `InitializeSystemListener` to 28 and the one of the `ContainerScopeListener` to 30 then? In case anyone needs to add a custom listener in between any of them.

aschempp: I have changed all priorities accordingly in 36ba667a261dee51f830e0237af523aab1419220

aschempp:
Fixed and rebased everything.

tests/bootstrap.php
discordier: The namespace is used in https://github.com/contao/core-bundle/pull/76/files#diff-f95982ace6744c19f506f6fb5e0b26b6R11, therefore he had to add this here.

Problem here is, we defined our fixtures with the old autoloader compat to reside in `tests/Fixtures/library` and having the complete namespace of `Contao\\Fixtures`. I guess @dtomasi was confused by this unorthodox directory layout.

src/EventListener/InitializeSystemListener.php
```diff
@@ -86,27 +86,28 @@ public function onKernelRequest(GetResponseEvent $event)
...
      */
-    private function getScopeFromRequest(Request $request)
+    private function getScopeFromContainer()
     {
-        if ($request->attributes->has('_scope') && 'backend' === $request->attributes->get('_scope')) {
+        if ($this->container->isScopeActive('backend')) {
             return 'BE';
+        } elseif ($this->container->isScopeActive('frontend')) {
+            return 'FE';
+        } else {
```
discordier: This `else` is not needed and will increase cyclomatic complexity with some tools.
Please rewrite as:
```php
if ($this->container->isScopeActive('backend')) {
    return 'BE';
} elseif ($this->container->isScopeActive('frontend')) {
    return 'FE';
}
return null;
```

tests/bootstrap.php
dtomasi: @discordier is absolutely right. That's a bit unorthodox in SF2's world. So I set this namespace as usual.

src/EventListener/InitializeSystemListener.php
aschempp: Changed in b82467f88249a211b5efc778c42b444dd51712fa

aschempp:
I have again optimized the code to get only A's in Scrutinizer :-)

leofeyer:
Merged in b207989883927382a6bbda552b93f084425384e6. There are some open questions though:

1. https://github.com/contao/core-bundle/blob/develop/src/Resources/config/services.yml#L73

My IDE says "missing parameter" here.

2. https://github.com/contao/core-bundle/blob/develop/tests/TestCase.php#L119-L120

We are using `new Scope('…')` here, but we are using `new Scope('…', 'request')` in the `ContaoCoreBundle` class:

https://github.com/contao/core-bundle/blob/develop/src/ContaoCoreBundle.php#L42-L43

aschempp:
> https://github.com/contao/core-bundle/blob/develop/src/Resources/config/services.yml#L73
My IDE says "missing parameter" here.

That's because you did not run dev environment. The inspector uses data from the cache folder.

> We are using new Scope('…') here, but we are using new Scope('…', 'request') in the ContaoCoreBundle class: https://github.com/contao/core-bundle/blob/develop/src/ContaoCoreBundle.php#L42-L43
Which one is correct?

`request` would be the parent one. In the tests, there is no parent scope of type request (because the `HttpKernel` did not handle it. So the current is correct.

PR #64, opened 2015-03-12
[WIP] No more magic

leofeyer:
This is the PR to contao/symfony-todo#5 as discussed on Mumble yesterday.

PR #77, opened 2015-03-22
[RTM] Adds a catch-all route

aschempp:
fixes #64, so that all routes are now handled by the legacy code.

Be aware that the problem of https://github.com/contao/core/issues/7666 still exists.

### TODOs
 - [x] Rename the parameter from `{alias}` to something like `{_catch_all}`

### DIFF
--- a/src/Resources/contao/controllers/BackendIndex.php
+++ b/src/Resources/contao/controllers/BackendIndex.php
@@ -102,7 +102,6 @@ class BackendIndex extends \Backend
 		$objTemplate->username = $GLOBALS['TL_LANG']['tl_user']['username'][0];
 		$objTemplate->password = $GLOBALS['TL_LANG']['MSC']['password'][0];
 		$objTemplate->feLink = $GLOBALS['TL_LANG']['MSC']['feLink'];
-		$objTemplate->frontendFile = \Environment::get('base');
 		$objTemplate->disableCron = \Config::get('disableCron');
 		$objTemplate->default = $GLOBALS['TL_LANG']['MSC']['default'];
 
--- a/src/Resources/contao/controllers/BackendPreview.php
+++ b/src/Resources/contao/controllers/BackendPreview.php
@@ -72,7 +72,7 @@ class BackendPreview extends \Backend
 		}
 		else
 		{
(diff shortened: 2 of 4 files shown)


leofeyer:
This does not seem correct. We only need a route for `/` as it used to be in the old implementation of the loader. Why would we want `/foo.bar` to be handled by Contao?

aschempp:
Because we want `/foo.bar` to be handled by the legacy code and redirected to the 404 page.

leofeyer:
This is not the case in Contao 3. And I don't think it is wise to run **everything** through the PHP parser.

aschempp:
> And I don't think it is wise to run everything through the PHP parser.

But that was the case with Contao 3, right? If we don't do that, we need to write Symfony code to handle these cases (e.g. forward to the fallback language).

PS: it does not mean non `.html` URL are forwarded, the URL rewriting will still ignore them by default.

leofeyer:
> But that was the case with Contao 3, right?

No. Only HTML documents were rewritten.

aschempp:
> Only HTML documents were rewritten.

That's not true. The `.htaccess` takes care of forwarding requests to the entry point script, and by default it only forwarded `.html`. But if you removed that, everything was handled by Contao.
Btw, if that would be true, the requests would still not be handled with my proposed change, because the legacy code is still the same ;-)

leofeyer:
You could be right. Then we should at least name the route `contao_catchall` instead of `contao_root`.

aschempp:
I named it `contao_root` because we use it to redirect people to the root page.

leofeyer:
Yes, but that's misleading. Routes are not named after where they redirect. A `*_root` route usually relates to `/`.

aschempp:
@contao/developers other opinions? To me `contao_root` is just fine as that's what we use the name for.

Toflar:
I can't follow the discussion. Somebody needs to explain it to me tomorrow.

leofeyer:
As discussed on Mumble on March, 25th, we want to add two routes: `contao_root`, which matches only `/` and `contao_catch_all`, which comes as last route and catches everything which has not been caught before.

aschempp:
Changed to RFC

leofeyer:
Looks good to me. Can we change the status to `RTM`?

aschempp:
Changed, assuming noone else has something to say ;-)

PR #157, opened 2015-03-26
[RTM] Added a ConfigAdapter

Toflar:
This allows us to inject the `ConfigAdapter` instead of using `Config::get()` all the time. Will be useful also for @aschempp's PR for the resource locator.


aschempp:
:-1: for me, can't see the advantage.

src/Config/ConfigAdapter.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Return true if the installation is complete.
+     *
+     * @return boolean
+     */
+    public function isComplete()
+    {
+        $this->instantiate();
+
+        var_dump($this->config);
```
leofeyer: To be removed :)

discordier: Or wrap it with
```php
ob_start();
ob_get_clean();
```
So it won't cause side effects. ;)

leofeyer:
I've thought about this again and it would make things easier for me if the adapter behaved more than the original class. This e.g. means no magic instantiation.

```php
echo $adapter->get('urlSuffix'); // null

$adapter->preload();
echo $adapter->get('urlSuffix'); // .html (default value)

$adapter->initialize();
echo $adapter->get('urlSuffix'); // .xhtml (custom value)
```

If you don't call `preload()` and `initialize()`, the class is not fully loaded. This is the same with the original `Config` class.

aschempp:
What do you mean with "magic instantiation" ?

leofeyer:
Right now, the `initialize()` method is called upon the first usage of the `get()` method.

aschempp:
That does absolutely not matter for any class you want to test that does use the Config class.

leofeyer:
It might not matter, but it is still wrong. The adapter should not behave differently from the adapted class. It should just, well, adapt :)

Toflar:
This is RTM. Travis should give its OK in a few minutes :)

leofeyer:
Great, thanks!

aschempp:
I think `preload()` should not instantiate the class (see https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Config.php#L445)

aschempp:
Another point: Couldn't we simply change our code to `Config::getInstance()->get(…)` instead of `Config::get()` and then it would all be mockable? We need to change all the code anyway to a ConfigAdapter?

leofeyer:
I'd like to avoid changing the legacy code.

aschempp:
We dont need to change the legacy code. Just look at the PR where the ConfigAdapter is used.

src/EventListener/InitializeSystemListener.php
```diff
@@ -181,10 +190,7 @@ protected function boot(Request $request = null)
         $this->initializeLegacySessionAccess();
         $this->setDefaultLanguage();
 
-        // Fully load the configuration
-        $objConfig = Config::getInstance();
```
leofeyer: Here we should call `$this->config->instantiate()`. The adapter must not instantiate the class magically.

Toflar: No, the whole point about an adapter is that you can use its methods without relying on calling another method first in your tests.

Toflar:
My goodness, I'm not going to explain it again, @discordier and I already did it for like 20 minutes when we were calling. We **need** adapters to mock the old static method classes properly, period. Even if it would work differently we have a shitload of other classes like `System`, `Controller` etc. for which we will eventually write adapters so we can use and test them in new code. I'd even go that far and introduce a general `Adapter` namespace where we put all our adapters, so it'll look like this:

* Adapter
    * Config
    * Controller
    * System
    * etc.

Toflar:
> I think preload() should not instantiate the class (see https://github.com/contao/core/blob/master/system/modules/core/library/Contao/Config.php#L445)

That's correct. Fixed in 6b8927c03d79b0dc9b756c16797c45915af7ce03.

src/EventListener/InitializeSystemListener.php
leofeyer: In this case, it does make a huge difference when `preload()` and `instantiate()` are called. If the adapter changes the program flow, I cannot merge it, because we cannot foresee the ramifications.

Toflar: > because we cannot foresee the ramifications.

We can. It's a pretty simple piece of code and adapters will always be untestable, that's why we write them.

leofeyer: You don't (want to) understand.

Toflar:
Squashed all the commits so it's cleaned up now and contains your `instantiate()` part.

src/Config/ConfigAdapter.php
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Return a configuration value
+     *
+     * @param string $strKey The short key (e.g. "displayErrors")
+     *
+     * @return mixed|null The configuration value
+     */
+    public function get($strKey)
+    {
+        return \Config::get($strKey);
```
leofeyer: Why are you not using `$this->config` here? Static methods can be called non-statically.

Toflar:
I still think we should move it to a more general `Adapter` namespace where we can then also place `System`, `Controller` etc. What do you think?

leofeyer:
Yes, although we are neither using `System` nor `Controller` at the moment.

Toflar:
Yes, but we should still change it now already :)

leofeyer:
Why? Right now we only have one adapter and we don't know when and if others will follow?

Toflar:
But we know they will follow and changing the namespace of the ConfigAdapter later will be a BC break and only allowed in v5 :)

leofeyer:
> But we know they will follow

I don't know this for sure. How come you do?

Toflar:
Still the same answer. Static methods cannot be mocked and writing an adapter classes is the easiest and best way to use the legacy classes in new code while maintaining testability. I also thought about writing a small script that generates an adapter class from any other class:

* Same class name plus `Adapter`
* Same methods, static methods are turned into object methods
* Same method docblocks
* `getInstance()` is turned into `initialize()` if anybody wants to use it or did additional work in it (like we did in `Config`)

Doesn't need to be in Contao, it can be a library I manage or so. This should make it pretty easy to generate adapters ;-)

leofeyer:
No, you got me wrong.

I know why adapters are good. I just cannot foresee if we will need more of them in the future or not. Maybe we refactor the code in a way that we never need a `SystemAdapter`.

But anyway, I'll change it.

Toflar:
Ah, yeah right. We don't know but if we should happen to need one then it would be certainly wrong in the `Config` namespace so that's why I think it would be best to have a general `Adapter` one. Just to be prepared :+1:

leofeyer:
Changed in 77936df87477a053d2382f6b7699fd8759dd07f9.

PR #159, opened 2015-03-26
[RFC] Implementation of ContaoFrameworkDependentInterface

dtomasi:
Boot Contao´s Framework if Command implements this interface. see: https://github.com/contao/standard-edition/issues/2


src/Command/ContaoFrameworkDependentInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+namespace Contao\CoreBundle\Command;
+
+/**
+ * Interface FrameworkDependentCommandInterface
+ * Only boot Contao-Framework if Command implements this Interface
+ *
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ */
+interface ContaoFrameworkDependentInterface {
```
aschempp: there should be a line break here and a line break as last line of the file to follow PSR

aschempp:
Be aware that #76 would probably break or be affected by adding a `$booted` property, so I suggest wait with that until #76 is merged (and then rebase).

src/Command/ContaoFrameworkDependentInterface.php
dtomasi: isn´t it? see: dfcec2a9b86ce8e9986df90d9a705688454d1d8c

aschempp: I did also mean line breaks in the control structures (see https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#11-example for an example)

dtomasi: Ok, fixed! :)

dtomasi:
Shouldn´t the constants only be defined if `boot()` is called?
see: https://github.com/ContaoBlackforest/core-bundle/blob/develop/src/EventListener/InitializeSystemListener.php#L93

aschempp:
I agree, but thats also related to my PR. Do you want me to add the change? I don't know which PR @leofeyer can merge first, but they probably conflict with each other

dtomasi:
Ok, I think this need to be discussed and decided by the order of merging the PRs.
If I should do the change, simply tell me. :)

aschempp:
Actually, GIT is our friend! I have merged your pull-request into mine (see #76), so @leofeyer only has to merge one. We can still keep this one open and you can improve it, I will simply merge follow-up changes if anything changes.

dtomasi:
:+1:

dtomasi:
@aschempp @leofeyer 
Would you like to pull this too? I put it in a separated branch if you won´t... 
see: https://github.com/ContaoBlackforest/core-bundle/commit/25f664ba60f57be3e8df43f80a19e2ee4015332a

dtomasi:
Here is a working and tested solution for prevent booting the Framework twice.
@leofeyer @aschempp 
Would you like to pull this in #76 also? 

Greetz

PR #161, opened 2015-03-27
[RFC] Order the profiler templates

aschempp:
This will re-order the profiler templates so that Contao is the first icon in the web profiler toolbar. Should be merged after #76 as it does also affect the icon.


leofeyer:
How does it look if the debug bar is collapsed? Does the square logo work?

leofeyer:
Haha, I've just checked it and when collapsed, the web profiler still displays the Symfony logo :)

leofeyer:
Also, we should probably use the flat version of the Contao logo:

Contao-Logo

leofeyer:
Actually, I'd be just fine if the Symfony logo remained the fist. @contao/developers What do you think?

Toflar:
I don't care about the position of the logo, really :) But as @aschempp has done the work already, why not :)

leofeyer:
Because he has not done all of the work yet (see my comments above). And I don't know any other application which has replaced the Symfony logo in the profiler. Why would we want this?

Toflar:
> Why would we want this?

I don't see any reason either.

aschempp:
I think the icon looks perfect compared to the other icons. Only the Symfony logo is flat, but that's because their logo is always flat.
Regarding the order, it will only get interesting with my other PR where it will then show the version in the toolbar. So you will have  `C 4.0.0 | S 2.6.4 | …` in the toolbar.

leofeyer:
> So you will have  `C 4.0.0 | S 2.6.4 | …` in the toolbar.

Yeah, but why do you want to have this? Isn't it still a Symfony app? Or do you just want to show off :)

aschempp:
Because to 98% of the people its more important what Contao version they run and the Symfony version does not matter…

leofeyer:
Of course. But this information is also available if Contao is not the first item ;)

aschempp:
I think you don't understand, so here's a screenshot ;-)
bildschirmfoto 2015-04-01 um 15 38 21

PS: we should re-consider the packages versions.

aschempp:
And now I'm re-considering your icon ^^

leofeyer:
> And now I'm re-considering your icon ^^

Thanks.

> I think you don't understand

I do, however, understand. But as I said, I don't see any reason why we would want to put the Contao icon first.

aschempp:
bildschirmfoto 2015-04-01 um 17 04 01

aschempp:
I have updated the PR. The Contao icon was moved the the end again, but I have restored the application version. This is a feature intended in Symfony so I guess that should be good to use.

bildschirmfoto 2015-04-09 um 21 46 02

The Contao icon now shows additional information: The current container scope, if the legacy framework is booted, and also the legacy modules. Also, if the framework is not loaded, there will be no Contao panel in the profiler page (cause there is no information or DB queries to show). 

bildschirmfoto 2015-04-09 um 21 43 40

I have also removed the memory and runtime information (`TL_START`) cause Symfony already provides that information.

discordier:
What is a legacy module in this context? A bundle? Module in the theme? Module used in current page layout?
We usually have the term "module" for elements in the page and the "modules" under "system/modules" are normally referred to as "extensions". You might want to rephrase this here.

aschempp:
You're right, these are "Contao extensions", things in `system/modules`.

Now that you mention it, maybe it would be interesting to show the current page layout for example?

discordier:
We also have backend routes? :)
I'd not pollute this too much from the beginning, if someone needs something in there we can add it later.

aschempp:
Not sure what you mean, but we obviously can distinguish between backend and frontend. I agree showing the current layout can be another task ^^

discordier:
I was referring to the fact that it will be pointless to show the layout etc. when it will all be empty in the backend.
You will clutter the badge with information of which the one half will display "N/A" when being in the frontend and the other half will display "N/A" while being in the backend.
I repeat, display as much as needed but as least as possible.

leofeyer:
I think the tab info is confusing.

- Instead of `Framework: Contao & Symfony` I'd prefer `Contao 3 framework: loaded/not loaded`.
- Instead of `Contao Modules` I'd prefer `Legacy modules`. Or the removal of the entire entry.

aschempp:
@discordier I would obviously show the relevant info and not show N/A. The modules was just an idea what useful info we could show.

@leofeyer "Contao 3 framework" is too long, but is it really the Contao 3 framework, not just the Contao framework? I also tried not to name it "Legacy" ;-)

leofeyer:
For me, the container scope is the only valuable information of the three. Let's just remove the other two :)

aschempp:
I have removed the modules completely and do not display the framework anymore. On the other hand, I have added the page layout (for front end requests).
bildschirmfoto 2015-04-12 um 16 00 49

leofeyer:
I still don't like it. I'm fine with showing things like the page layout in the full tab, but for the popup, we should only show things from the Symfony context, such as the container scope or the request token.

leofeyer:
Oh, and on the full tab, we should rather dump the whole page object than just the page layout, shouldn't we?

PR #164, opened 2015-03-31
[RTM] Fix locator unit tests 

aschempp:


aschempp:
Added resource locator factory and fixed a major bug ^^
Should be RTM now.

src/Config/ChainFileLocator.php
```diff
@@ -63,18 +66,23 @@ public function locate($name, $currentPath = null, $first = false)
...
      */
-    private function getLocators($first)
+    private function mergeFiles(array $newFiles, array $allFiles)
     {
-        if (false === $first) {
-            return array_reverse($this->locators);
+        foreach ($newFiles as $k => $v) {
+            if (is_numeric($k)) {
+                $allFiles[] = $v;
+            } elseif (!isset($allFiles[$k])) {
```
leofeyer: This means the locator with the highest priority wins, right? Because it is the first one that gets to set non-numeric entries.

leofeyer:
Hm… https://travis-ci.org/contao/core-bundle/jobs/56562754

PR #166, opened 2015-03-31
[RFC] Prevent boot twice

aschempp:
Includes commits from @dtomasi 

The framework is now also booted on a subrequest if the master request did not boot it.


dtomasi:
+1 conventions are great :)

PR #169, opened 2015-04-02
[RFC] Reworking the resource locators

leofeyer:
This is the PR as discussed yesterday on Mumble.

**Locators**

There are now four locators involved:

- A chain locator (`ChainFileLocator`), which loads all services tagged with `contao.resource_locator` and processes them until one of them returns something.

- A bundle file locator (`BundleFileLocator`), which creates a `FileLocator` object from the bundle resources paths.

- A file locator aggregate (`FileLocatorAggregate`), which aggregates the results of multiple file loaders.

- A strict file locator (`StrictFileLocator`), which looks for files but not for directories.

**Chain**

The cache file locator and the file locator aggregate are both tagged with `contao.resource_locator` by default, whereby the cache file locator has the higher priority. Thus if a cache file exists, it will automatically be loaded, otherwise the file locator aggregate will look for all matching files.

**Prerequisites**

I have changed two prerequisites, so the code can be less blown up.

- No more support for `system/runonce.php`. Runonce files need to be stored in the `config/` directory of a bundle or module. Thus we can skip the additional check.

- Instead of showing the bundle names in the template editor, I now display the folder name, e.g. `contao/core-bundle` instead of `ContaoCoreBundle`. Thus we don't need the bundle names in the locators and can use the default Symfony `FileLocator` class.

**Extras**

This PR also fixes the `FileLocatorInterface` compatibility and moves the internal cache to `app/cache`.


src/Config/FileLocatorFactory.php
```diff
@@ -0,0 +1,61 @@
...
+use Symfony\Component\HttpKernel\Bundle\BundleInterface;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * Creates a FileLocator object with the bundle paths.
+ *
+ * @author Andreas Schempp <https://github.com/aschempp>
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class FileLocatorFactory
```
aschempp: This should be called `BundleFileLocatorFactory`. Factories should always have an interface.

leofeyer: Why do you want to prepend `Bundle`?

src/Config/CacheFileLocator.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Contao\CoreBundle\Config;
+
+use Symfony\Component\Config\FileLocatorInterface;
+
+/**
+ * Searches the cache paths for files.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class CacheFileLocator implements FileLocatorInterface
```
aschempp: This class name is wrong. The class does not know about cache, it simply knows about paths. It's simply a `FileLocator`.

src/Resources/config/services.yml
```diff
@@ -43,31 +43,34 @@ services:
...
+    contao.resource_locator:
+        alias: contao.resource_locator.chain
+
+    contao.resource_locator.aggregate:
+        class: Contao\CoreBundle\Config\FileLocatorAggregate
         public: false
-        factory: [Contao\CoreBundle\Config\FileLocator, createFromKernelBundles]
         arguments:
-            - "@kernel"
+            - ["@contao.resource_locator.factory"]
```
aschempp: How can this be correct? The factory is not a file locator?

src/Config/CacheFileLocator.php
leofeyer: Partially true, but it is not simply a file locator, because it only checks for files and not for directories.

src/Resources/config/services.yml
leofeyer: But it creates one?

leofeyer:
I have adjusted the implementation according to @aschempp's comments.

1. There is now an interface for the locator factory.

2. The locator factory can create file locators with the kernel bundle paths or the cache path.

3. I have removed the `CacheFileLocator` class and added a `StrictFileLocator` class instead. The strict file locator is a decorator for the regular one and returns only file and no directories.

discordier:
> No more support for system/runonce.php. Runonce files need to be stored in the config/ directory of a bundle or module. Thus we can skip the additional check.

This disables running the `runonce.php` files from the composer plugin as it put a compiled `runonce.php` in `system`. This issue will resolve itself with the new package manager but I  wanted to mention it.

leofeyer:
Is the plugin compatible with Contao 4? If so, you should probably store the file in `system/modules/!composer/config/runonce.php`. On the other hand, if the plugin lies in `system/modules`, it is a legacy module ;)

src/Config/FileLocatorFactoryInterface.php
```diff
@@ -0,0 +1,49 @@
...
+
+use Symfony\Component\Config\FileLocatorInterface;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * Interface for file locator factories.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+interface FileLocatorFactoryInterface
```
discordier: Interfaces are for ensuring compatibility between objects I doubt that it makes sense to have a static only interface.
IMO we should either convert the factory to a plain worker class or remove the interface as it will never get passed/used anywhere.

leofeyer: @aschempp noted that every factory class should have an interface and indeed, every factory in Symfony seems to follow this principle.

discordier: Yes, but they are not static classes, at least quick scanning did not find any.
Can you please show me the static (only!) factories in Symfony implementing an interface? :)

Until then I explain that it feels plain wrong to implement it like this and show counter examples.

First example: [```php class FormFactory implements FormFactoryInterface```](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/Forms.php#L24)
```php
use Symfony\Component\Form\Forms;
// The static call creates the factory instead of _BEING_ the factory.
$formFactory = Forms::createFormFactory(); 
$form = $formFactory->createBuilder()
    ->add('firstName', 'text')
    ->add('lastName', 'text')
    ->add('age', 'integer')
    ->add('gender', 'choice', array(
        'choices' => array('m' => 'Male', 'f' => 'Female'),
    ))
    ->getForm();
```
It is perfectly valid to have an interface here, as you get a factory instance which will perform the work.
You do NOT have any object here performing work but merely have a bunch of static functions stuck together into a "namespace" (class) which feels plain wrong.

Next example: [MetadataFactoryInterface](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Mapping/Factory/MetadataFactoryInterface.php) implemented by [BlackHoleMetadataFactory](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Mapping/Factory/BlackHoleMetadataFactory.php) and [LazyLoadingMetadataFactory](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Mapping/Factory/LazyLoadingMetadataFactory.php) is also not static (You might have to look at the deprecated interface until it get's moved to the new location in 3.0 over [here](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/MetadataFactoryInterface.php)).

You need the interface when implementing the "abstract factory pattern", not when the factory class is the "final" implementation.

leofeyer:
I have reworked the implementation again based on @discordier's comments. I have removed the factory and created an adapter class instead (`BundleFileLocator`).

aschempp:
1. I can't see a reason for the `StrictFileLocator`, can you explain the need for that?
2. What's the `FileLocatorAggregate` supposed to do in your implementation? It's not used anywhere.
3. Why did you remove the count-check on the `ResourceLocatorPass`? Did you understand the implementation? It follows the Symfony standard…

leofeyer:
By the way, it is wrong that the aggregate is never used. It is tagged as service!

https://github.com/contao/core-bundle/blob/feature/resource-locator/src/Resources/config/services.yml#L55

aschempp:
1. I don't get why we need a locator that is limited to files. If you specify a file name, I doubt there will be a directory named `config.php`.
2. I wanted to point out that this is wrong ;-)
3. Which is wrong too…

Here's how the concept previously, and how I think it should be (using the new classes and discussed concept):

 - `FileLocator` should be a basic class accepting a list of paths. We might be able to use the basic Symfony version.
 - Either we create a factory, or a child class of the Symfony `FileLocator` that retrieves the bundle paths. The current delegating behavior is pretty strange, as it's exactly what a super class should do. That's what I created in the first place but then replaced with a factory which is easier to test.
 - The `ChainFileLocator` must chain over multiple `FileLocator` instances and retrieve the first result. In the core, it will always have two locators: the cached and the regular one. This will make it hit the cache if found. This is necessary for the `contao.cached_resource_locator` service.
 - The `FileLocatorAggregate` should retrieve values from all tagged locators. That is the core one (searching in `Resources/contao`, and any custom locators if someone needs to place resources anywhere else. Contrary to the `ChainFileLocator`, it will return a merge of all results (if `$first = false`).
 - If there is only one tagged service (which is true in most cases), we don't need the `FileLocatorAggregate`.

leofeyer:
> - The `ChainFileLocator` must chain over multiple `FileLocator` instances and retrieve the first result. In the core, it will always have two locators: the cached and the regular one. This will make it hit the cache if found. This is necessary for the `contao.cached_resource_locator` service.

This is exactly how it works, except that my concept does not need a `cached_resource_locator`. There is only one resource locator in my code, which is the chain. It can handle as many locators as you want.

> - The `FileLocatorAggregate` should retrieve values from all tagged locators. That is the core one (searching in `Resources/contao`, and any custom locators if someone needs to place resources anywhere else. Contrary to the `ChainFileLocator`, it will return a merge of all results (if `$first = false`).

This is wrong. It would conflict with the cache.

> - If there is only one tagged service (which is true in most cases), we don't need the `FileLocatorAggregate`.

This is also wrong (see my comment above). First of all, there are at least **two** tagged locators and second of all, the aggregate ensures that the service always returns at least an empty array, which saves us from using `try {} catch() {}` all the time.

aschempp:
> This is exactly how it works, except that my concept does not need a cached_resource_locator. There is only one resource locator in my code, which is the chain. It can handle as many locators as you want.

I noticed now you delete the cache before building it, so the cache is never hit. Bit of a workaround, but that could work. For the languages though, we now miss the bundle names, which means a PHP and XLF file in the same directory would both be parsed.

>> The FileLocatorAggregate should retrieve values from all tagged locators. That is the core one (searching in Resources/contao, and any custom locators if someone needs to place resources anywhere else. Contrary to the ChainFileLocator, it will return a merge of all results (if $first = false).

> This is wrong. It would conflict with the cache.

Why would it? And how do you expect third-party locators to add paths to the system? If the first locator wins, no files will ever be found after the regular locator.

leofeyer:
While working on the cache warming, I have realized that using file locators is actually wrong and that the whole implementation is totally over engineered, whereby the latter is resulting from working around the fact that file locators are not the right tool for the task.

I will therefore revert the merge and prepare a pull request with a more reasonable approach, which we can discuss this afternoon.

leofeyer:
See #172.

PR #170, opened 2015-04-03
[RTM] Always use double backslashes in string literals

leofeyer:
To be merged after #169.


aschempp:
I've seen this before, but is there a reason to use two backslashes?

leofeyer:
Yes, it is wrong to use just one, because it actually escapes the character behind it. Even if it will still work in 95% of the cases, it is technically correct to use double backslashes.

PR #172, opened 2015-04-08
[RTM] Reworking the resource locators 2

leofeyer:
This is the new and much better implementation of the resources locator. It now uses the Symfony finder again, which gives us a lot more flexibility.


aschempp:
I can't see why this should be better than the resource locators. I could not find any advantages…

1. It is not **impossible** for third-party developers to add their custom resources path to the locator.
1. It does not follow the Symfony way, see [locating templates](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/Templating/Loader/TemplateLocator.php), [locating configuration files](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L51) or [loading anything from bundles](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpKernel/Kernel.php#L713-L725)
2. Loading the combined version is now delegated to the actual loading code, possibly resulting in duplicate code.
3. The finder is certainly much more expensive in locating files and folders (because it's more powerful, but we don't need that power)

leofeyer:
Resource locators are fine if you want to locate a certain template in several locations. However, we want to find multiple resources in a specific folder, which is what the finder is for. This is also a Symfony way by the way :)

aschempp:
The case of not knowing which DCA/language files there are is only valid for building the cache. Maybe we should add that information to the container. Appart from that, we are exactly doing what templates do: find `dca/default.php` in all bundles.

Can you point to an example for Symfony using the finder class to locate **resources**?

leofeyer:
Sure:

- https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bridge/Twig/Command/LintCommand.php#L140
- https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/Command/AssetsInstallCommand.php#L148
- https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/Command/CacheClearCommand.php#L144
- https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/Command/YamlLintCommand.php#L96
- https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L700

Shall I go on?

aschempp:
Perfect examples that the `Finder` should be used for unknown resources, but a `FileLoacator` is used for known resources. I also noticed a `FileLoader` is most of the time used in conjunction with a `LoaderInterface`, maybe we should elaborate on that.

leofeyer:
We'll discuss this this afternoon.

src/Resources/contao/dca/tl_user_group.php
```diff
@@ -308,29 +308,24 @@ public function getExcludedFields()
 		/** @var Symfony\Component\HttpKernel\KernelInterface $kernel */
 		global $kernel;
 
-		$included = array();
+		$processed = array();
 
-		foreach ($kernel->getContainer()->get('contao.resource_locator')->locate('dca') as $strDir)
+		/** @var SplFileInfo[] $files */
+		$files = $kernel->getContainer()->get('contao.resource_finder')->in('dca')->files()->name('*.php');
```
leofeyer: > Perfect examples that the `Finder` should be used for unknown resources

@aschempp Like here?

src/Resources/contao/library/Contao/Automator.php
```diff
@@ -560,41 +565,30 @@ public function generateDcaCache()
...
+		(
+			$kernel->getContainer()->get('filesystem'),
+			new PhpFileLoader(),
+			$kernel->getCacheDir() . '/contao'
+		);
+
+		$processed = array();
+
+		/** @var SplFileInfo[] $files */
+		$files = $kernel->getContainer()->get('contao.resource_finder')->in('dca')->files()->name('*.php');
```
leofeyer: > Perfect examples that the `Finder` should be used for unknown resources

@aschempp Or like here?

src/Resources/contao/library/Contao/Automator.php
```diff
@@ -711,42 +679,32 @@ public function generateDcaExtracts()
 		/** @var KernelInterface $kernel */
 		global $kernel;
 
-		$included = array();
-		$arrExtracts = array();
+		$processed = array();
+
+		/** @var SplFileInfo[] $files */
+		$files = $kernel->getContainer()->get('contao.resource_finder')->in('dca')->files()->name('*.php');
```
leofeyer: > Perfect examples that the `Finder` should be used for unknown resources

@aschempp Also here …

src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -272,37 +273,30 @@ public function getFromDca()
...
+		$processed = array();
 
 		// Ignore the internal cache
 		$blnBypassCache = \Config::get('bypassCache');
 		\Config::set('bypassCache', true);
 
-		// Only check the active modules (see #4541)
-		foreach ($kernel->getContainer()->get('contao.resource_locator')->locate('dca') as $strDir)
+		/** @var SplFileInfo[] $files */
+		$files = $kernel->getContainer()->get('contao.resource_finder')->in('dca')->files()->name('*.php');
```
leofeyer: > Perfect examples that the `Finder` should be used for unknown resources

@aschempp And here.

src/Resources/contao/library/Contao/Database/Updater.php
```diff
@@ -736,33 +737,20 @@ public function updateFileTreeFields()
 		/** @var KernelInterface $kernel */
 		global $kernel;
 
-		$arrFiles = array();
+		$processed = array();
+
+		/** @var SplFileInfo[] $files */
+		$files = $kernel->getContainer()->get('contao.resource_finder')->in('dca')->files()->name('*.php');
```
leofeyer: > Perfect examples that the `Finder` should be used for unknown resources

@aschempp You figured …

leofeyer:
As discussed on Mumble on April 8th, we will keep the finder in the cache warmer and use a file locator in places like this:

https://github.com/contao/core-bundle/pull/172/files#diff-e8c8926b79a5f155115d86ad49d4f574R218

The file locator is to be created by a [factory service](http://symfony.com/doc/current/components/dependency_injection/factories.html).

leofeyer:
I have updated the PR as discussed. I have also found a way to set the `kernel.resources_paths` parameters in a compiler pass.

leofeyer:
It is not entirely wrong.

> You are creating a second instance of the bundles already instantiated

This is true but not wrong. Symfony supports this, which is why I get the class names in `kernel.bundle`.

> Also, you do not know the constructor arguments.

AFAIK, bundles never have constructor arguments. Otherwise it would not make sense to pass the class names in `kernel.bundles`, would it?

> Or, some bundles might need arguments 

Well, if you know that bundles can actually have constructor arguments, the solution is very simple:

```php
// AppKernel.php
{
    $bundles = [
        new Contao\CoreBundle\ContaoCoreBundle($this),
    ];

    return $bundles;
}

// ContaoCoreBundle.php
class ContaoCoreBundle extends Bundle
{
    private $bundles;

    public function __construct(KernelInterface $kernel)
    {
        $this->bundles = $kernel->getBundles();
    }
```

But I am not sure if this is allowed.

src/Resources/contao/library/Contao/System.php
```diff
@@ -314,32 +315,29 @@ public static function loadLanguageFile($strName, $strLanguage=null, $blnNoCache
...
-				include $kernel->getContainer()->get('contao.cached_resource_locator')->locate('languages/' . $strCreateLang . '/' . $strName . '.php', null, true);
+				include $kernel->getCacheDir() . '/contao/languages/' . $strCreateLang . '/' . $strName . '.php';
 			}
-			catch (\InvalidArgumentException $e)
+			else
 			{
-				$objLocator = $kernel->getContainer()->get('contao.resource_locator');
-				$xlfLoader  = new XliffFileLoader($kernel->getRootDir(), true);
+				/** @var SplFileInfo[] $files */
+				$files = $kernel->getContainer()->get('contao.resource_finder')->in('languages/' . $strCreateLang)->files()->name($strName . '.php');
```
aschempp: Should use the `contao.resource_locator` here.

src/Resources/contao/library/Contao/System.php
```diff
@@ -314,32 +315,29 @@ public static function loadLanguageFile($strName, $strLanguage=null, $blnNoCache
...
+				$loader = new XliffFileLoader($kernel->getRootDir(), true);
 
-				// XLIFF files will overwrite PHP files if both exist in the same bundle
-				$arrFiles = array_merge
-				(
-					$objLocator->locate('languages/' . $strLanguage . '/' . $strName . '.php'),
-					$objLocator->locate('languages/' . $strLanguage . '/' . $strName . '.xlf')
-				);
+				/** @var SplFileInfo[] $files */
+				$files = $kernel->getContainer()->get('contao.resource_finder')->in('languages/' . $strCreateLang)->files()->name($strName . '.xlf');
```
aschempp: Should use the `contao.resource_locator` here.

src/DependencyInjection/Compiler/AddResourcesPathsPass.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $container->setParameter('kernel.resources_paths', $this->getResourcesPath($container));
+    }
+
+    /**
+     * Returns the Contao resources path of a bundle.
```
aschempp: This description seems incorrect.

src/Finder/ResourceFinder.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Returns a new Finder object with the resource paths set.
+     *
+     * @param string $path An optional path
+     *
+     * @return Finder The Finder object
+     */
+    public function in($path = null)
```
aschempp: The service, class and this method name makes the service look like a `Finder` (adapter), but it's not. In a Finder, the order of method calls does not matter (`->directories()->in(…)`). We should find better name for the method.

It's also strange to call `->get('contao.resource_finder')->in()->…` if you want to search in the resource folders directly. Maybe a separate method `findAll()` or similar would make sense?

src/Finder/ResourceFinder.php
```diff
@@ -0,0 +1,80 @@
...
+     * @return array The subpaths array
+     *
+     * @throws \InvalidArgumentException If the subpath does not exist
+     */
+    private function getExistingSubpaths($path)
+    {
+        $paths = [];
+
+        foreach ($this->paths as $key => $value) {
+            if (is_dir($dir = "$value/$path")) {
```
aschempp: Do we really need to check the folders? I think the finder will just do that and ignore what's not there?

src/Resources/config/services.yml
```diff
@@ -43,31 +43,15 @@ services:
...
-        class: Contao\CoreBundle\Config\ChainFileLocator
-        public: false
-
-    contao.resource_locator.cache:
-        class: Contao\CoreBundle\Config\CombinedFileLocator
-        public: false
-        arguments:
-            - '%kernel.cache_dir%/contao'
-            - "@contao.resource_locator"
+            - "%kernel.resources_paths%"
```
aschempp: Parameter is `%contao.resources_paths%`

src/Resources/config/services.yml
```diff
@@ -43,31 +43,15 @@ services:
...
 
     contao.resource_locator:
-        alias: contao.resource_locator.bundles
-
-    contao.cached_resource_locator:
-        alias: contao.resource_locator.cache
+        class: Symfony\Component\Config\FileLocator
+        arguments:
+            - "%kernel.resources_paths%"
```
aschempp: Parameter is `%contao.resources_paths%`

src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -336,10 +330,12 @@ public function getFromFile()
 		$table = '';
 		$return = array();
 
-		// Only check the active modules (see #4541)
-		foreach ($kernel->getContainer()->get('contao.resource_locator')->locate('config/database.sql') as $strFile)
+		/** @var SplFileInfo[] $files */
+		$files = $kernel->getContainer()->get('contao.resource_finder')->in('config')->files()->name('database.sql');
```
aschempp: Should use `contao.resource_locator` here.

src/Resources/contao/library/Contao/Database/Installer.php
```diff
@@ -336,10 +330,12 @@ public function getFromFile()
...
+
+		foreach ($files as $file)
 		{
-			$data = file($strFile);
+			$data = file($file);
```
aschempp: Should we create a loader for this similar to the `XliffFileLoader`?

src/Resources/contao/library/Contao/System.php
leofeyer: I have not used it here on purpose, because the file locator might throw an exception if there is no file (which is the case for `.php` files). Would have to wrap it in a try catch block. The finder, by contrast, simply returns an empty array.

src/DependencyInjection/Compiler/AddResourcesPathsPass.php
leofeyer: Fixed.

src/Finder/ResourceFinder.php
leofeyer: Not if the path you pass to `in()` does not exist.

src/Finder/ResourceFinder.php
leofeyer: You are right. I will rename the method to `find()` and add a `findIn()` method, which you can pass a subfolder as argument.

src/Resources/config/services.yml
leofeyer: Changed.

src/Resources/contao/library/Contao/Database/Installer.php
leofeyer: Or stick with the finder, because it will always return an array?

src/Resources/contao/library/Contao/Database/Installer.php
leofeyer: For a deprecated feature which is to be removed in the next major version?

aschempp: True, we agreed to remove support for `database.sql` in Contao 5.

src/DependencyInjection/Compiler/AddResourcesPathsPass.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    private function getResourcesPath(ContainerBuilder $container)
+    {
+        $paths   = [];
+        $rootDir = dirname($container->getParameter('kernel.root_dir'));
+
+        foreach ($container->getParameter('kernel.bundles') as $name => $class) {
+            $reflection = new \ReflectionClass($class);
+
+            if ('ContaoModuleBundle' === $reflection->getShortName()) {
```
aschempp: We don't really need a `ReflectionClass` for the `ContaoModuleBundle` because we do know the absolute namespace. Up to you if you wanna change that though…

aschempp: However it would have the advantage of not conflict with a same-named class in a different bundle (for whatever reason)

leofeyer: You know that any other solution to get the short class name is ugly, don't you?

```php
echo substr($classNameWithNamespace, strrpos($classNameWithNamespace, '\\') + 1);
```

src/Command/SymlinksCommand.php
```diff
@@ -94,7 +94,7 @@ public function generateSymlinks($rootDir, OutputInterface $output)
      */
     private function symlinkFiles($uploadPath, $rootDir, OutputInterface $output)
     {
-        $finder = $this->findIn('.public', "$rootDir/$uploadPath");
+        $finder = $this->findIn("$rootDir/$uploadPath")->files()->name('.public');
```
bytehead: What's the reason for filtering the `files` directory for `.public` files? Shouldn't the `files` directory be public anyway?

leofeyer: Nope: https://github.com/contao/core-bundle/blob/develop/UPGRADE.md#protected-folders

bytehead: Ah, thanks :+1:

PR #173, opened 2015-04-08
[WIP] Generate the internal cache upon cache:clear

leofeyer:
This PR moves the internal cache routines from the `Automator` class to a Symfony cache warmer/clearer event listener.

It also removes the `ContaoFrameworkDependentInterface`, because despite all efforts, we will always have to boot the Contao framework on the command line. The `cache:clear` command does not implement the `ContaoFrameworkDependentInterface` interface, still we need the Contao framework for our cache warmer.


leofeyer:
Sorry, wrong target branch :(

PR #174, opened 2015-04-08
[RFC] Generate the internal cache upon cache:clear

leofeyer:
This PR moves the internal cache routines from the `Automator` class to a Symfony cache warmer/clearer event listener.

It also removes the `ContaoFrameworkDependentInterface`, because despite all efforts, we will always have to boot the Contao framework on the command line. The `cache:clear` command does not implement the `ContaoFrameworkDependentInterface` interface, still we need the Contao framework for our cache warmer.


leofeyer:
As discussed on Mumble on April 8th, the PR can be merged after the FIXMEs have been taken care of. Regarding the Contao framework see# 175.

Issue #128, opened 2015-03-23
Trusted proxies


importer:
> <a href="https://github.com/aschempp"><img src="https://avatars.githubusercontent.com/u/1073273?v=3" align="left" width="42" height="42" hspace="10"></img></a> [Issue](https://github.com/contao/contao/issues/50) by @aschempp
Wednesday Feb 18, 2015 at 08:00 GMT

The trusted proxies from Contao configuration should be set in the Symfony container.

http://symfony.com/doc/current/reference/configuration/framework.html#trusted-proxies

trevorff:
Thank yo a lot for sharing. As for me, the most trusted proxies are the socks5 I buy from http://buy.fineproxy.org/eng/ website. They are really realiable and never gave up on me. Thanks again ;)

trevorff:
Thank yo a lot for sharing. As for me, the most trusted proxies are the socks5 I buy from http://buy.fineproxy.org/eng/ website. They are really realiable and never gave up on me. Thanks again ;)

Toflar:
As discussed on Mumble on April 8th, we need to do the following:

* Remove the legacy proxy settings as upgrading from Contao 3 to 4 is not possible anyway
* Adjust the `Environment` class to use the Symfony proxy settings for BC

leofeyer:
See #176.

PR #176, opened 2015-04-08
[RTM] Feature/trusted proxy support

Toflar:
This is a PR for https://github.com/contao/core-bundle/issues/128. I checked the code and imho our legacy code does exactly what Symfony's `Request::getClientIp()` does.

Please double check this though :)

### DIFF
--- a/src/Resources/contao/config/default.php
+++ b/src/Resources/contao/config/default.php
@@ -156,7 +156,6 @@ $GLOBALS['TL_CONFIG']['loadGoogleFonts']      = false;
 $GLOBALS['TL_CONFIG']['defaultFileChmod']     = 0644;
 $GLOBALS['TL_CONFIG']['defaultFolderChmod']   = 0755;
 $GLOBALS['TL_CONFIG']['maxPaginationLinks']   = 7;
-$GLOBALS['TL_CONFIG']['proxyServerIps']       = '';
 $GLOBALS['TL_CONFIG']['sslProxyDomain']       = '';
 $GLOBALS['TL_CONFIG']['maintenanceMode']      = true;
 $GLOBALS['TL_CONFIG']['errorReporting']       = E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED;
--- a/src/Resources/contao/dca/tl_settings.php
+++ b/src/Resources/contao/dca/tl_settings.php
@@ -27,7 +27,7 @@ $GLOBALS['TL_DCA']['tl_settings'] = array
 	(
 		'__selector__'                => array('useSMTP'),
 		// FIXME: adjust the palette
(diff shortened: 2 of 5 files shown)


aschempp:
I've noticed that too in https://github.com/contao/core-bundle/pull/72/files#diff-2f0eb5d18e0f5fb0dbaef526bb165d41R305

Should we add the `TL_CONFIG` variable for BC? We currently do that for all other variables, and I do have some customization where I use the variable.

Toflar:
> Should we add the TL_CONFIG variable for BC? We currently do that for all other variables, and I do have some customization where I use the variable.

BC of what? What would it help to add the variable?

aschempp:
I mean adding it here:
https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/library/Contao/Config.php#L484-L505

Toflar:
But what does it help? It sets a TL_CONFIG value that is **never** used.

aschempp:
We don't know if extension developers used it. Same with SMTP, it's unlikely someone used it because the `Email` class does automatically use it, but it would not harm to add it for the sake of completeness.

Toflar:
But it's not really possible because we did not support all the notations `IpUtils` supports. One can configure an IP range in the trusted_proxy settings and we would need to resolve this to a list of comma separated IP's to set that value. I mean, of course it's possible but is it worth the effort?

aschempp:
That's true, assuming people will start using IP ranges…

leofeyer:
Looks good to me. Let me know when it is RTM :)

Toflar:
Okay for you and @aschempp so it's RTM for me :)

leofeyer:
Merging this will invalidate #128, right?

Toflar:
Not invalidate but close :-)

aschempp:
We should also close my PR in contao/core for ip-range support and mention this

leofeyer:
Can you fix the tests please?

```
PHP Fatal error:  Call to a member function getContainer() on null in /var/www/contao/standard-edition/vendor/contao/core-bundle/src/Resources/contao/library/Contao/Environment.php on line 376
```

The method is probably missing in our kernel mock.

aschempp:
`global $kernel` is missing ;)

Toflar:
Fixed. But now I have a new problem that's probably caused by some component overriding the `$_SERVER` or whatever. I can't find the issue...

aschempp:
I guess here's the problem:
https://github.com/contao/core-bundle/blob/develop/tests/TestCase.php#L96

Toflar:
True, why is this?

aschempp:
I guess because a default value is alway necessary… 

Maybe we should change that to something like this:
```
if (empty($_SERVER['HTTP_ACCEPT_LANGUAGE'])) {
    $_SERVER['HTTP_ACCEPT_LANGUAGE'] = '';
}
```

leofeyer:
I have no idea. I think the `mockKernel()` method is from @discordier?

leofeyer:
Do the tests still work if you remove the line?

discordier:
Having similar issues in my pr. Stay tuned. Will search a solution tomorrow.

leofeyer:
I have fixed the tests and merged the PR in c375b7c3cb441565962e97493ba8bd7e594c9f28.

PR #177, opened 2015-04-09
[RTM] Fixed DEPRECATED.md description and related code changes

Toflar:


src/EventListener/InitializeSystemListener.php
```diff
@@ -138,10 +138,12 @@ public function onConsoleCommand()
     protected function setConstants($mode, $route)
     {
         // The constants are deprecated and will be removed in version 5.0.
+        $time = microtime(true);
```
leofeyer: Please revert this change. It has nothing to do with the `DEPRECATED.md` file.

DEPRECATED.md
```diff
@@ -15,7 +15,36 @@ You can use the static helper methods such as `System::loadLanguageFile()` or
 The constants `TL_ROOT`, `TL_MODE`, `TL_START` and `TL_SCRIPT` have been
 deprecated and will be removed in Contao 5.0.
 
-// FIXME: show the alternatives
+Instead of `TL_ROOT` use:
+
+```php
+global $kernel;
+
+$rootDir = $kernel->getContainer()->getParameter('kernel.root_dir');
```
aschempp: This will not give the same value, you need to add a `dirname()`

DEPRECATED.md
```diff
@@ -15,7 +15,36 @@ You can use the static helper methods such as `System::loadLanguageFile()` or
...
+Instead of `TL_MODE` to check for `BE` and `FE` use:
+
+```php
+global $kernel;
+
+$isFE = $kernel->getContainer()->isScopeActive('frontend');
+$isBE = $kernel->getContainer()->isScopeActive('backend');
+```
+
+Note: You can also use the `ScopeAwareListener` in your own event listeners.
```
aschempp: What do you mean by that?

DEPRECATED.md
```diff
@@ -15,7 +15,36 @@ You can use the static helper methods such as `System::loadLanguageFile()` or
...
+Instead of `TL_START` use:
+```php
+global $kernel;
+
+$startTime = $kernel->getStartTime();
+```
+
+Note: This will only work if Symfony is in debug mode (`kernel.debug`).
+
+Instead of `TL_SCRIPT` use the Symfony routing component to generate proper URLs.
```
leofeyer: Aren't people mainly using `TL_SCRIPT` to check for the current entry point?

```php
if (TL_SCRIPT == 'contao/main.php') {
}
```

We should probably describe both cases, shouldn't we?

DEPRECATED.md
Toflar: I mean what is written. You can use the `ScopeAwareListener` in your own event listeners if you like.

DEPRECATED.md
Toflar: No idea, I never ever used `TL_SCRIPT`.

src/EventListener/InitializeSystemListener.php
Toflar: Well, we should't use `TL_START` either if we're deprecating it ;-)

leofeyer: So let's just add a second example (assuming the following code is correct):

```php
if ('contao_backend_main' === $this->container->get('request')->get('_route')) {
}
```

DEPRECATED.md
leofeyer: So let's just add a second example (assuming that the following code is correct):

```php
if ('contao_backend_main' === $this->container->get('request')->get('_route')) {
}
```

src/EventListener/InitializeSystemListener.php
leofeyer: If `TL_REFERER_ID` is **not** deprecated, I agree.

Toflar: It's not (yet).

DEPRECATED.md
Toflar: Better?

DEPRECATED.md
Toflar: Fixed 136a52740cd1a0977bf46d4ce7c3c640d11bd56f.

src/EventListener/InitializeSystemListener.php
discordier: But should be.
IMO we should deprecate every constant to encourage writing properly testable code.
Constants not be used anymore not by legacy extensions and not even in the core-bundle but this will take a migration phase during the 4.x series.
To allow this, we should deprecate them all as of now.

Toflar: Of course it should be but we cannot deprecate anything as long as there's no alternative to be used :)

leofeyer: And there is no alternative to the referer ID yet.

Toflar: I'll try to find one.

leofeyer: Ok. In the meantime, please revert this change, since it is not related to the `DEPRECATED.md` file.

Toflar:
Reverted changes in c8fd4cffaaf0f02a67c65d74a93b074d36efb094.

PR #180, opened 2015-04-09
Fix wrong path in is_writeable

bytehead:
The `TL_ROOT` is wrong here, `$strFile` contains the full path already as you pass in `__FILE__` (see https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/controllers/BackendInstall.php#L66).

### DIFF
--- a/src/Resources/contao/library/Contao/Files.php
+++ b/src/Resources/contao/library/Contao/Files.php
@@ -303,7 +303,7 @@ class Files
 	{
 		$this->validate($strFile);
 
-		return is_writeable(TL_ROOT . '/' . $strFile);
+		return is_writeable($strFile);
 	}
 
 


aschempp:
The old classes always use relative paths. And there is a `str_replace` on TL_ROOT ?

bytehead:
Ok. But [here](https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/library/Contao/Files.php#L306) is another TL_ROOT - not replaced and probably not needed anymore?

aschempp:
That is added because the passed-in path always needs to be relative to TL_ROOT

bytehead:
Ah, I see. I tried to run the installer via browser on `contao/install` and that gave me absolute paths in `$strFile` because of the replacement of `TL_ROOT` with `__FILE__`.

aschempp:
Why would that give an absolute path?

bytehead:
`__FILE__` is absolute:
> __FILE__ always contains an absolute path with symlinks resolved whereas in older versions it contained relative path under some circumstances

http://php.net/manual/en/language.constants.predefined.php

aschempp:
Yeah, and the TL_ROOT part is str_replaced to have it relative ;-)

bytehead:
You're right. My fault. Symlinks are resolved, that was my problem ;-)

Issue #178, opened 2015-04-09
Replace the encryption key with the kernel secret


leofeyer:
As discussed, we want to replace the Contao encryption key with the Symfony kernel secret.

@contao/developers Pull request anyone?

Toflar:
Can we close this in favor of #181?

aschempp:
:+1:

PR #181, opened 2015-04-10
[RTM] Replace the encryption key with the kernel secret

bytehead:
Attempt to replace the encryption key with the kernel secret (see #178)

### DIFF
--- a/src/Resources/contao/library/Contao/Encryption.php
+++ b/src/Resources/contao/library/Contao/Encryption.php
@@ -10,6 +10,7 @@
 
 namespace Contao;
 
+use Symfony\Component\HttpKernel\KernelInterface;
 
 /**
  * Encrypts and decrypts data
@@ -73,7 +74,11 @@ class Encryption
 
 		if (!$strKey)
 		{
-			$strKey = \Config::get('encryptionKey');
+			/** @var KernelInterface $kernel */
+			global $kernel;
+
+			$container = $kernel->getContainer();
+			$strKey = $container->getParameter('framework.secret');
 		}
 
 		$iv = mcrypt_create_iv(mcrypt_enc_get_iv_size(static::$resTd), MCRYPT_RAND);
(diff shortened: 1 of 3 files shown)


aschempp:
You should also map the key in the legacy class (see https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/library/Contao/Config.php#L492)

bytehead:
Good point! Missed that.

src/Resources/contao/library/Contao/Encryption.php
```diff
@@ -157,7 +166,12 @@ protected static function initialize()
 			throw new \Exception('Error initializing encryption module');
 		}
 
-		if (\Config::get('encryptionKey') == '')
+		/** @var KernelInterface $kernel */
+		global $kernel;
+
+		$container = $kernel->getContainer();
+
+		if ($container->getParameter('framework.secret') == '')
```
bytehead: I'm not sure, if this check is still needed - Symfony will raise a `ParameterNotFoundException` when `secret` is empty...

Toflar: I think we can go without the check :)

aschempp: Empty is not equal to "not set", so the check makes sense to me?

Toflar: Yeah, the check would already throw an exception now. If you wanted to keep it you'd need to check for `->has()` :)

aschempp: That's not what I mean. It would throw an `ParameterNotFoundException` if the parameter is not set, which should never happen. However, I can set the secret to actually be an empty string, which this check would then catch.

Toflar: Ah, true. I thought they validate for not empty in the `Configuration` class of the `FrameworkBundle`, but they don't :(

leofeyer:
Thanks a lot for your help :)

src/Resources/contao/library/Contao/Encryption.php
discordier: Should be `===` instead of `==` though.

aschempp: No, because it's legacy code…

discordier: Arguable but ok... :)

PR #182, opened 2015-04-10
[RTM] Map the database classes to the Doctrine connection

leofeyer:
This is the reworked version of #31, which fixes the implementation and removes the child classes.


leofeyer:
I tested every method of the classes as well as the back end and front end. Everything seems to work, so this PR is RTM. Awaiting your comments though :)

src/Resources/contao/library/Contao/Database/Result.php
```diff
@@ -75,20 +83,16 @@
...
 	 * Validate the connection resource and store the query string
 	 *
-	 * @param resource $resResult The database result
-	 * @param string   $strQuery  The query string
-	 *
-	 * @throws \Exception If $resResult is not a valid resource
+	 * @param DoctrineStatement $statement The database statement
+	 * @param string            $strQuery  The query string
+     *
+     * @todo Try to find a solution that works without fetchAll().
```
Toflar: CS

Toflar:
I don't understand why you want to provide

```php
$db = Database::getInstance('doctrine.dbal.custom_connection');
```

In Contao 3 you can pass an array which is not possible anymore. This is a BC break. If you want to introduce that, I'm fine with it as long as it's documented. However, I don't understand why you want to provide a new solution to use a different connection with the **old** database class? You can simply take the default one and say that if you want to have another connection just follow http://symfony.com/doc/current/cookbook/doctrine/multiple_entity_managers.html?

leofeyer:
> If you want to introduce that, I'm fine with it as long as it's documented.

It is documented (see the `UPGRADE.md` file).

> You can simply take the default one and say that if you want to have another connection just follow

I know. I have adjusted the configuration file of the standard edition accordingly:

https://github.com/contao/standard-edition/commit/4882825e74ab200870b7c883854ab61d524d6b89

> I don't understand why you want to provide

I just think the correct way of having multiple DB connections is to configure them in Symfony instead of passing configuration arrays to the `Database` class. Don't you think so?

Toflar:
> I just think the correct way of having multiple DB connections is to configure them in Symfony instead of passing configuration arrays to the Database class. Don't you think so?

Yeah, but I think you miss my point. We are now using `Database` and we can now pass a configuration array for a second or third connection. If you drop this functionality now then no piece of code using the old configuration array will work anymore, so the developer will have to fix it in his code base. Now, why do you want to provide the possibility to pass a doctrine connection key if the developer should use the container and the DBAL directly anyway? :)

leofeyer:
You do have a point, this is a BC break. I just don't want to support the old feature anymore. If we make people adjust the Doctrine configuration, maybe we they will directly use Doctrine :)

It is just a matter of preference, so I guess we should discuss this and/or vote here.

aschempp:
> so this PR is RTM. Awaiting your comments though :)

It should be RFC then ;-)

Due to rewriting the classes, it's difficult to diff. Maybe you can quickly explain what has changed?

PS: are I correctly assuming custom drivers will no longer work?

src/Resources/contao/library/Contao/Database/Statement.php
```diff
@@ -375,43 +355,11 @@ protected function escapeParams($arrValues)
 	/**
 	 * Debug a query
 	 *
-	 * @param \Database\Result $objResult An optional result object
+	 * @deprecated Deprecated in Contao 4.0, to be removed in Contao 5.0.
```
discordier: Maybe better note: `@deprecated Deprecated in Contao 4.0, to be removed in Contao 5.0. - No op implementation until then.`
To tell developers that it is doing nothing anymore ans is only preserved to keep ABI compatibility.

discordier: On the other hand, IMO you can also remove it, as the compatibility has changed anyway as you removed the abstract protected methods below and the descendant classes implementing them removed.
It was protected and only called internally, I do not see any usecase where any developer would end up calling it from his code.

discordier:
I am with @Toflar on the custom configuration.
We should drop the possibility completely instead of providing an backwards incompatible workaround to use the old database classes.
If developers need a custom connection, they have to alter their code and then should be forced to alter it using the correct way and not using an workaround forcing them to alter the same code again when the old database classes will get removed.

PR #183, opened 2015-04-10
[RTM] Introduced a token generator for the referer id

Toflar:
This PR aims to replace the `TL_REFERER_ID`. First, I wanted to implement a service but then I realized that our referer ID is in fact nothing else than an attribute of the current request because it changes for every request :-)

### DIFF
--- a/src/RefererId/TokenGenerator.php
+++ b/src/RefererId/TokenGenerator.php
@@ -1,0 +1,29 @@
+<?php
+
+/**
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\RefererId;
+use Symfony\Component\Security\Csrf\TokenGenerator\UriSafeTokenGenerator;
+
+/**
+ * Generates a 0 characters short token for referer management in the URL.
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+
+class TokenGenerator extends UriSafeTokenGenerator
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function generateToken()
+    {
+        return substr(parent::generateToken(), 0, 8);
+    }
+}
--- a/src/RefererId/TokenGenerator.php
+++ b/src/RefererId/TokenGenerator.php
@@ -12,7 +12,7 @@ namespace Contao\CoreBundle\RefererId;
 use Symfony\Component\Security\Csrf\TokenGenerator\UriSafeTokenGenerator;
 
 /**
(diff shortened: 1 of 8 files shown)


src/RefererId/TokenGenerator.php
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\RefererId;
+use Symfony\Component\Security\Csrf\TokenGenerator\UriSafeTokenGenerator;
+
+/**
+ * Generates a 0 characters short token for referer management in the URL.
```
aschempp: I guess the `0` here is wrong, should be 8 ?

Toflar: Yes :D

aschempp:
Looks good to me, though I feel uncomfortable to add a lot of arbitrary information to the request attributes. Not sure if that's good practice in Symfony?

src/RefererId/TokenGenerator.php
```diff
@@ -0,0 +1,29 @@
...
+class TokenGenerator extends UriSafeTokenGenerator
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function generateToken()
+    {
+        return substr(parent::generateToken(), 0, 8);
+    }
+}
```
discordier: Could you please, please, please go to `File -> Settings` in there go to `Editor -> General` there check "Ensure line feed at file end on Save". :)

Toflar:
@discordier Done. This time I'll leave it to @leofeyer to fix CS a last time :D Also, I think we need to adjust the event listener priorities because I had to add it exactly in between of two which is why I took 29 for now. I did not change this because I didn't want to adjust all the priorities and then be out of sync of the `develop` again.

leofeyer:
> though I feel uncomfortable to add a lot of arbitrary information to the request attributes. Not sure if that's good practice in Symfony?

Not sure, either.

Toflar:
I am very sure. In fact, that's **exactly** what it's meant for. It's for "custom attributes" (https://github.com/symfony/HttpFoundation/blob/master/Request.php#L84-91) that relate to the request object which is exactly what our referer ID does.

aschempp:
@Toflar is probably right

http://symfony.com/doc/current/book/http_fundamentals.html#book-fundamentals-attributes

> The Request class also has a public attributes property, which holds special data related to how the application works internally. For the Symfony framework, the attributes holds the values returned by the matched route, like _controller, id (if you have an {id} wildcard), and even the name of the matched route (_route). The attributes property exists entirely to be a place where you can prepare and store context-specific information about the request.

leofeyer:
Great.

Issue #179, opened 2015-04-09
Problem with users and session


aschempp:
The user session storage currently does not work. Sessions are stored using a `register_shutdown_handler`, but at that point Symfony has already closed the session to send the response.

My quick idea was to use a `kernel.response` or `kernel.terminate` handler (same as Symfony uses for session). What do you think?

leofeyer:
+1

Toflar:
I remember, there was another ticket regarding "referer management broken" (I can't find it, wtf), that's 99% related to this one :)

leofeyer:
It is here: #143

leofeyer:
This issue has been fixed, but it seems that the fix introduced a new issue: https://github.com/contao/core-bundle/issues/143#issuecomment-94388947

PR #184, opened 2015-04-10
[WIP] Move the BackendUser session storage process to a separate listener

Toflar:
This restores the back end user session storage process so you can navigate again. We can make it even more beautiful as soon as my other PR #183 is merged :-)

Also, tests are still missing but that's a starting point!

### DIFF
--- a/src/EventListener/StoreSessionListener.php
+++ b/src/EventListener/StoreSessionListener.php
@@ -1,0 +1,164 @@
+<?php
+
+/**
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\BackendUser;
+use Contao\CoreBundle\Session\Attribute\AttributeBagAdapter;
+use Doctrine\DBAL\Driver\Connection;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\Event\PostResponseEvent;
+
+/**
+ * Stores the session data of a back end user so the state can be recreated
+ * on the next login.
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+class StoreSessionListener extends ScopeAwareListener
+{
+    /**
+     * @var SessionInterface
+     */
+    private $session;
+
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    /**
+     * Constructor.
+     *
+     * @param SessionInterface $session
+     */
+    public function __construct(
+        SessionInterface $session,
+        Connection $connection
+    ) {
+        $this->session = $session;
+        $this->connection = $connection;
+    }
+
+    /**
+     * Stores the referer data of a back end user so the state can be recreated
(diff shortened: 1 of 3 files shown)


src/EventListener/StoreSessionListener.php
```diff
@@ -0,0 +1,164 @@
...
+            $session[$key][TL_REFERER_ID]['last'] = '';
+        }
+
+        while (count($session[$key]) >= 25) {
+            array_shift($session[$key]);
+        }
+
+        $ref = \Input::get('ref');
+
+        if ($ref != '' && isset($session[$key][$ref])) {
```
aschempp: `!==`

Toflar: That's all legacy code here, see the FIXME

aschempp:
Looks good on a quick review. Two questions:
1. this only works for `BackendUser`, but we also need it for `FrontendUser`?
2. Shouldn't we use the same listener to restore the session on the next request?

Toflar:
> this only works for BackendUser, but we also need it for FrontendUser?

Ah, did not know we do that for the FE as well, but it should maybe be a different listener.

> 2. Shouldn't we use the same listener to restore the session on the next request?

If so, then the listener name would be slighty wrong :) But of course we could do that.

aschempp:
Probably fixed #179 ?

Toflar:
> Probably fixes #179 ?

Probably? Wtf, of course it does, I did it because of this :D

src/Resources/config/services.yml
```diff
@@ -92,6 +92,17 @@ services:
         class: Contao\CoreBundle\Security\User\ContaoUserProvider
         public: false
 
+    contao.store.session:
```
leofeyer: This should be `contao.listener.store_session`.

Toflar: True, but maybe we rename it to something more general anyway if it should also load the session on kernel.request again?

leofeyer: The namespace `contao.listener.` is already used for all the other listeners. And since the class is called `StoreSessionListener`, the key must be `store_session`.

Toflar: Yes, I know :-D But what I was saying is that as this is till WIP it will be renamed completely if we decide that it should also be responsible to load the values again on kernel request :)

leofeyer: Where are the values loaded now? In the `User` class?

Toflar: No idea, probably yes...all still legacy behavior

leofeyer: We don't need a listener to load the values. It is part of the `User::setFromDatabase()` logic.

Toflar: Yeah but I'm with @aschempp here. Someday we will have to relocate this logic into a listener and he's not completely wrong when he says that it might also go into that listener so it would be responsible for both, storing as well as reloading those values for the current user.

leofeyer: Then let's change it someday :)

aschempp: Changing it would greatly help to implement advanced security (see #122). I think we should change both if we change one of them…

leofeyer: Ok, then do it. I'm fine either way.

leofeyer:
We need this fixed ASAP, before the RC release. Can I merge the PR and we'll add the second feature later?

Toflar:
I am sorry but this is not even close to a state that can be merged.

leofeyer:
Ok, so do you want to work on it or do you want me to fix the issue?

Toflar:
I'll work on it asap, I know it's important.

Toflar:
Closed in favor of #207;

PR #185, opened 2015-04-10
[RTM] Do not use legacy code

Toflar:
We need to get rid of as much legacy code as possible in the `InitializeSystemListener`. @discordier and I are having issues with all tests failing due to (newly introducing) proper error handling and need to fix those tests which is impossible without those cleanups :)

### DIFF
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -291,6 +291,7 @@ class InitializeSystemListener extends ScopeAwareListener
 
             foreach ($langs as $lang) {
                 if (is_dir(__DIR__ . '/../../src/Resources/contao/languages/' . str_replace('-', '_', $lang))) {
+                    $_SESSION['TL_LANGUAGE'] = $lang; // backwards compatibility
                     $this->session->set('TL_LANGUAGE', $lang);
                     break;
                 }
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -13,7 +13,6 @@ namespace Contao\CoreBundle\EventListener;
 use Contao\ClassLoader;
 use Contao\Config;
 use Contao\CoreBundle\Session\Attribute\AttributeBagAdapter;
-use Contao\Environment;
 use Contao\Input;
 use Contao\System;
 use Symfony\Component\HttpFoundation\Request;
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -179,7 +179,7 @@ class InitializeSystemListener extends ScopeAwareListener
 
         $this->setRelativePath($request ? $request->getBasePath() : '');
         $this->initializeLegacySessionAccess();
-        $this->setDefaultLanguage();
+        $this->setDefaultLanguage($request);
 
         // Fully load the configuration
         $objConfig = Config::getInstance();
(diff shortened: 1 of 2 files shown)


discordier:
+1 as I had to fix this in my PR as well in f252a8e18aa498892117dc00a18c282142e65ac6

src/EventListener/InitializeSystemListener.php
```diff
@@ -275,29 +274,32 @@ class_alias('Contao\\ModuleLoader', 'ModuleLoader');
...
      * Sets the default language.
+     *
+     * @param Request $request
      */
-    private function setDefaultLanguage()
+    private function setDefaultLanguage(Request $request = null)
     {
-        if (!isset($_SESSION['TL_LANGUAGE'])) {
-            $langs = Environment::get('httpAcceptLanguage');
+        if (!$this->session->has('TL_LANGUAGE')) {
```
aschempp: I guess this will not work as `$this->session` will be the parameter bag and not the "root".

Toflar: It will work because we later do `$this->session->set('TL_LANGUAGE')`.

discordier: I do not know what you mean by `root` but it works.

src/EventListener/InitializeSystemListener.php
```diff
@@ -275,29 +274,32 @@ class_alias('Contao\\ModuleLoader', 'ModuleLoader');
      */
     private function setRelativePath($basePath)
     {
-        Environment::set('path', $basePath);
```
leofeyer: This must not be removed! It will break the `Environment` class, which relies on `path` being set during system initialization.

leofeyer: Ah, I see. You have changed the `Environment` class :)

src/EventListener/InitializeSystemListener.php
```diff
@@ -275,29 +274,32 @@ class_alias('Contao\\ModuleLoader', 'ModuleLoader');
...
-                    $_SESSION['TL_LANGUAGE'] = $lang;
+                    $_SESSION['TL_LANGUAGE'] = $lang; // backwards compatibility
+                    $this->session->set('TL_LANGUAGE', $lang);
                     break;
                 }
             }
         }
 
-        $GLOBALS['TL_LANGUAGE'] = $_SESSION['TL_LANGUAGE'];
+        if ($this->session->has('TL_LANGUAGE')) {
```
leofeyer: The if block is unnecessary, because `$langs` always includes `en` and `src/Resources/contao/languages/en` always exists. Therefore `$this->session->has('TL_LANGUAGE')` should always be true.

leofeyer:
Aside from my one comment, this looks good to me. Please add the `RTM` prefix if it is ready to merge.

src/EventListener/InitializeSystemListener.php
Toflar: By `root` @aschempp meant the root of `$_SESSION`. Because the SF2 Session handling stores everything in `$_SESSION['_sf2_attributes']` by default. But again, it will work because we also set it on the session service later :)

src/EventListener/InitializeSystemListener.php
Toflar: You're right. Fixed in 74b1193904dc922b181702ec8b62051eb13dcec3.

Toflar:
Updated and RTM.

aschempp:
I wonder if we should add `TL_LANGUAGE` to the new session parameter bag? The language should be part of the request, which it already is in `_locale` if language-URLs are enabled. Otherwise we should maybe set it?

PR #186, opened 2015-04-11
[RFC] Make the trailing slash for the backend route (/contao) optional

bytehead:
I'm (and probably others are) used to just type `http://example.com/contao` to get to the backend login.

Without this change, you have to type in the trailing slash (`http://example.com/contao/`) to get to the backend without any exception.

What do you think about?

### DIFF
--- a/src/Controller/BackendController.php
+++ b/src/Controller/BackendController.php
@@ -40,7 +40,7 @@ class BackendController extends Controller
      *
      * @return Response
      *
-     * @Route("/", name="contao_backend")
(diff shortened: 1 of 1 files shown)


aschempp:
I would prefer a redirect "route" for that.

bytehead:
Like in #187?

aschempp:
http://symfony.com/doc/current/cookbook/routing/redirect_in_config.html

leofeyer:
I'm closing this PR in favor of #187.

PR #187, opened 2015-04-11
[RFC] Add redirect route for the contao backend

bytehead:
See #186

### DIFF
--- a/src/Controller/BackendController.php
+++ b/src/Controller/BackendController.php
@@ -24,6 +24,7 @@ use Contao\BackendSwitch;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 
 /**
  * Handles the Contao backend routes.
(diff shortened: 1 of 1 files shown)


leofeyer:
I think we should add the redirect route in the `app/config/routing.yml` file instead of creating a controller method, which does nothing. This is also recommended here:

http://symfony.com/doc/current/cookbook/routing/redirect_in_config.html

bytehead:
Makes sense to me. But that would be in contao/standard-edition? I already have some working code.

leofeyer:
>  But that would be in contao/standard-edition?

Exactly. You don't have to create another PR though. I could just port your suggestion.

bytehead:
I added this before the `ContaoCoreBundle:` in `app/config/routing.yml`:

```yml
contao_backend_alias:
    path: /contao
    defaults:
        _controller: FrameworkBundle:Redirect:redirect
        route: contao_backend
        permanent: true
        _scope: backend

```

leofeyer:
I committed it on your behalf in contao/standard-edition@ff9df8671eb8c026611bba1b4f33b4e56e20dd37 and 3a4319326e80f86eb92131f4eb99880108d6d8d9.

bytehead:
Thanks! And cleaner that way (from `/contao/` to `/contao`). :+1:

PR #191, opened 2015-04-12
[WIP/RFC] Regular and normalized package versions

aschempp:
People usually want the "real" installed version and not a normalized version of a package. This is especially true for our profiler, which currently shows *4.0.9999999* is installed.

This PR adds the Composer version parser and a new parameter to the kernel. We don't actually need the normalized version (because we only need packages for jQuery and MooTools), so we might think about not adding that information to the kernel but let Devs use the VersionParser directly.

### DIFF
--- a/src/Composer/VersionParser.php
+++ b/src/Composer/VersionParser.php
@@ -1,0 +1,211 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Contao\CoreBundle\Composer;
+
+/**
+ * Version parser
+ *
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ */
+class VersionParser
+{
+    private static $modifierRegex = '[._-]?(?:(stable|beta|b|RC|alpha|a|patch|pl|p)(?:[.-]?(\d+))?)?([.-]?dev)?';
+
+    /**
+     * Returns the stability of a version
+     *
+     * @param  string $version
+     * @return string
+     */
+    public static function parseStability($version)
+    {
+        $version = preg_replace('{#.+$}i', '', $version);
+
+        if ('dev-' === substr($version, 0, 4) || '-dev' === substr($version, -4)) {
+            return 'dev';
+        }
+
+        preg_match('{'.self::$modifierRegex.'$}i', strtolower($version), $match);
+        if (!empty($match[3])) {
+            return 'dev';
+        }
+
+        if (!empty($match[1])) {
+            if ('beta' === $match[1] || 'b' === $match[1]) {
+                return 'beta';
+            }
+            if ('alpha' === $match[1] || 'a' === $match[1]) {
+                return 'alpha';
+            }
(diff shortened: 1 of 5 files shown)


leofeyer:
-1 from me. Even though 4.0.9999999 might be confusing at the first glance, it is the correct version. Using the Composer version parser does not change a thing, because `normalizeBranch()` would also give you 4.0.9999999.

But let's see what the others, especially @discordier, say :)

leofeyer:
The only other option I could think of was not to normalize the version number and output `4.0.x-dev` instead of `4.0.9999999`.

aschempp:
Thats exactly what it does ;-)

discordier:
Add the MIT license to the repository under which the code has been released and it is correct. :)

leofeyer:
We don't have to copy the class and add it to our bundle. It will be enough if we just remove the [extra normalization step](https://github.com/contao/core-bundle/blob/develop/src/DependencyInjection/Compiler/AddPackagesPass.php#L125). If someone wants to use the Composer class, they should require it in their `composer.json` and use the original class.

aschempp:
I know that we did that previously, but I disagree. The Composer code does way more, for example it would also detect branch aliases (`dev-develop as 1.0.0`). It does support mapping `dev-master` to a comparable version (`9999999-dev`) and probably more.

There are two versions we need: One that we can show to the user (and use for the GoogleAPI, because it will be the same), and one that is useable for `version_compare`.

leofeyer:
Ok, then add the necessary requirements and use the original class.

aschempp:
> Ok, then add the necessary requirements and use the original class.

I thought about this too, but are we willing to include the WHOLE composer package? https://github.com/composer/composer

Remember that people might need to manually upload that stuff, at least for 4.0 if they don't have Composer on the server.

leofeyer:
I don't care. It is wrong to copy a class from another package instead of adding the package as dependency.

> but are we willing to include the WHOLE composer package?

No. You are the one who wants to have the version parser. I'm totally fine with the current solution, which fulfills all our needs!

aschempp:
> It is wrong to copy a class from another package instead of adding the package as dependency.

Just FYI it's not a complete copy. And your example would be copying code too, just less because you copied less of the functionality ;-)

Other opinions @discordier @Toflar ?

Toflar:
1. Copying is wrong.
2. Including the whole package is not the best way but the only way currently.

Also see https://github.com/composer/composer/issues/3545

leofeyer:
> Including the whole package is not the best way but the only way currently.

No, there is another way: https://github.com/contao/core-bundle/pull/191#issuecomment-92241605

leofeyer:
Until we have agreed on a final solution, I have removed the `x-dev` normalization in 83bf0cc3c555eb1a19ed3056bef45cd73d3d8e07, because this seems to be the only thing we all want :)

Toflar:
So we definitely need this, otherwise I cannot install my own dev version. My `composer.json`:
```json
 "require": {
        "contao/core-bundle": "dev-feature/user-session-listener as 4.0.0"
    },
    "repositories": [
      {
        "type": "git",
        "url": "/local/path/to/contao/core-bundle"
      }
    ]
```

This will result in an exception:

> ContextErrorException in SetApplicationPass.php line 36: Notice: Undefined index: contao/core-bundle

leofeyer:
Simplified as discussed on Mumble on April 15th in 44f67aa887acd1e16555809acd0a137f05dba6aa.

Issue #97, opened 2015-03-23
Use the Symfony event dispatcher for hooks and callbacks


importer:
> <a href="https://github.com/leofeyer"><img src="https://avatars.githubusercontent.com/u/1192057?v=3" align="left" width="42" height="42" hspace="10"></img></a> [Issue](https://github.com/contao/contao/issues/19) by @leofeyer
Thursday Aug 14, 2014 at 19:00 GMT

We want to use the Symfony event dispatcher for hooks and callbacks in a backwards compatible manner.

leofeyer:
Closed in favor of #204.

PR #204, opened 2015-04-15
[WIP] Use the Symfony event dispatcher for hooks and callbacks

leofeyer:
This is the PR for #97. Can you please confirm that this is the way to go, so I can adjust the other hooks as well?


Toflar:
Looks good to me!! My input: I would place all hook listeners in an extra subnamespace so it becomes `Contao\CoreBundle\EventListener\Hooks\InitializeSystemListener`. Otherwise the `EventListener` namespace containing new code already becomes really cluttered with listeners that only provide support for old hooks and will get removed in v5 anyway :)

leofeyer:
Good point, will change it accordingly.

aschempp:
I would suggest to use one class for all hook events. That will make the old stuff much cleaner. Also we only need one service with just a lot of tags.

leofeyer:
I don't think Insights and Scrutinizer would like that. They already don't like that we have only two controllers with too many methods.

Toflar:
Well, Scruitinizer should not be the issue for code that only provides BC layers. https://scrutinizer-ci.com/docs/reviews/excluding_files_from_analysis

leofeyer:
I know. Still we should always follow the best practices, because people will look at how we've done things.

Toflar:
Of course and I prefer your variant.

leofeyer:
This will also solve our arguments order problem:

```php
// Call
$strTemp = $this->$callback[0]->$callback[1]($row, $blnWriteToFile, $vars, $parent);

// Usage
public function compileDefinition($row, $blnWriteToFile, $vars, $parent)
{
    // do something
}
```

But if we are using events, it will all be in the event object:

```php
// Call
$event = new CompileDefinitionEvent();
$event->setRow($row);
$event->setWriteToFile($blnWriteToFile);
$event->setVars($vars);
$event->setParent($parent);

$this->container->get('event_dispatcher')->dispatch('contao.initialize_system', $event);

// Usage
public function compileDefinition(CompileDefinitionEvent $event)
{
    print_r($event->getRow());
}
```

If we ever want to add or remove parameters, we don't have to stick to an argument order anymore :)

Toflar:
Did you only realize this now? :D

leofeyer:
No, I just wanted to mention it :)

aschempp:
I'm still for one class, also using event subscribers (see http://symfony.com/doc/current/components/event_dispatcher/introduction.html#using-event-subscribers). Let's discuss that later today.

leofeyer:
Yes. I have still added another example to my POC to see how it works with parameters.

discordier:
You might want to have a look how we did it in dc-general.

src/Event/InitializeSystemEvent.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->rootDir = $rootDir;
+    }
+
+    /**
+     * Returns the root directory.
+     *
+     * @return string The root directory
+     */
+    public function getRootDir()
```
aschempp: Looks like a setter method is missing here.

aschempp:
In addition to setting the event variables from hook references, you also need to set the variables where the event is fired. Otherwise the variables could still not be changed by reference. Example:

```php
// Trigger the getPageLayout hook (see #4736)
$event = new PageEvent($objPage, $objLayout, $this);
$kernel->getContainer()->get('event_dispatcher')->dispatch(ContaoEvents::GENERATE_PAGE, $event);
$objPage = $event->getPage();
$objLayout = $event->getLayout();
```

src/Event/InitializeSystemEvent.php
leofeyer: There is no setter method. The variable is only injected to include the `initialize.php` file.

leofeyer:
No, I found a better solution. Check my latest commit :)

aschempp:
Not sure if that's not a problem. Now we are deciding wether the variables are passed by reference, instead of the hook developer…

leofeyer:
The developer will still have to put an ampersand before the argument, otherwise they will not receive the reference. Here's how you can test:

```php
// vendor/contao/calendar-bundle/src/Resources/config.php
$GLOBALS['TL_HOOKS']['modifyFrontendPage'][] = function ($page, $template, $handler) {
    $template = 'fe_foo';

    return $page;
};
```

The template name will not be changed, therefore `fe_page` is rendered. However, if you add the ampersand and make `$template` a reference:

```php
// vendor/contao/calendar-bundle/src/Resources/config.php
$GLOBALS['TL_HOOKS']['modifyFrontendPage'][] = function ($page, &$template, $handler) {
    $template = 'fe_foo';

    return $page;
};
```

Then you get a `Could not find template "fe_foo"` exception.

leofeyer:
I'm going to write the tests tonight.

dtomasi:
As shortly spoken in #234, here are my suggestions this.

1. There should be a own `EventDispatcher` for Contao, which handles only its events.
2. The naming convention inner `ContaoEventDispatcher` should be based on domains. This can look like this `contao.page.generate`, `contao.page.get_id_from_url`,`contao.user.check_credentials` ... This is a common practice used for example in Doctrine2 and helps developers to debug.
3. IMHO there should be a Proxy for `EventDispatcher`, which can handle `preDispatch` and `postDispatch` actions to convert or transform Events to Hooks. So you have the ability to check if there is a according hook to trigger in every dispatched event inner `ContaoEventDispatcher`. This harms to have a forest of listeners for the hooks and makes it much more easier to eliminate the BC for hooks.
4. As pre and post action handler there should be something like `ContaoHookHandler` which can check if there is a hook to trigger according the events name or class. 

I am looking forward to your comments. :)

leofeyer:
> There should be a own `EventDispatcher` for Contao, which handles only its events.

What would be the advantages?

>  The naming convention inner `ContaoEventDispatcher` should be based on domains.

We are following another scheme here: convert the old camel-case notation into an underscore notation and prefix the result with `contao.`.

> IMHO there should be a Proxy for `EventDispatcher`, which can handle `preDispatch` and `postDispatch` actions to convert or transform Events to Hooks.

Can you elaborate on this one? No idea what you have in mind.

dtomasi:
@leofeyer I just has spoken to @aschempp at mumble. This changed my point of view in some cases, as he told me CoreBundle should not be the base to a standalone contao-application. We also discussed the Proxy I suggested and figured out that it's about as broad as it's long. The mass of code that is needed is exactly the same. So this is just a matter of taste how this is implemented. In this Case I retract my suggestions from above because in the end every way of implementation has the same result.

Thank you.

leofeyer:
As discussed on Mumble on April 29th, it will save us a lot of work if we leave the existing hooks in the legacy code and simply insert the event trigger before them. Then we won't have to create all the additional listeners.

bytehead:
:+1: Is it still realistic for 4.1? :)

leofeyer:
Here is the list of the currently available hooks in the core distribution:

 * activateAccount
 * activateRecipient
 * addComment
 * addCustomRegexp
 * addFileMetaInformationToRequest
 * addLogEntry
 * checkCredentials
 * closeAccount
 * colorizeLogEntries
 * compareThemeFiles
 * compileArticle
 * compileDefinition
 * compileFormFields
 * createDefinition
 * createNewUser
 * customizeSearch
 * executePostActions
 * executePreActions
 * executeResize
 * exportTheme
 * extractThemeFiles
 * generateBreadcrumb
 * generateFrontendUrl
 * generatePage
 * generateXmlFiles
 * getAllEvents
 * getArticle
 * getArticles
 * getAttributesFromDca
 * getCacheKey
 * getCombinedFile
 * getContentElement
 * getCountries
 * getForm
 * getFrontendModule
 * getImage
 * getLanguages
 * getPageIdFromUrl
 * getPageLayout
 * getPageStatusIcon
 * getRootPageFromUrl
 * getSearchablePages
 * getSystemMessages
 * getUserNavigation
 * importUser
 * indexPage
 * initializeSystem
 * insertTagFlags
 * isAllowedToEditComment
 * isVisibleElement
 * listComments
 * loadDataContainer
 * loadFormField
 * loadLanguageFile
 * modifyFrontendPage
 * newsListCountItems
 * newsListFetchItems
 * outputBackendTemplate
 * outputFrontendTemplate
 * parseBackendTemplate
 * parseDate
 * parseFrontendTemplate
 * parseTemplate
 * parseWidget
 * postAuthenticate
 * postDownload
 * postLogin
 * postLogout
 * postUpload
 * prepareFormData
 * printArticleAsPdf
 * processFormData
 * removeOldFeeds
 * removeRecipient
 * replaceDynamicScriptTags
 * replaceInsertTags
 * reviseTable
 * sendNewsletter
 * setCookie
 * setNewPassword
 * sqlCompileCommands
 * sqlGetFromDB
 * sqlGetFromDca
 * sqlGetFromFile
 * storeFormData
 * updatePersonalData
 * validateFormField

leofeyer:
As discussed on Mumble, @aschempp wants to try to set up a general service which allows to map events to hooks. If he succeeds, we do not need this PR at all.

leofeyer:
Closed in favor of contao/core-bundle#376.

PR #205, opened 2015-04-15
[RFC] Split the services.yml file into separate files

leofeyer:
As discussed a few weeks ago, this PR splits the `services.yml` file into separate files to make it less confusing.


Toflar:
Yeah!

bytehead:
:+1:

src/Resources/config/listener.yml
```diff
@@ -0,0 +1,89 @@
...
+        tags:
+            - { name: kernel.event_listener, event: kernel.terminate, method: onKernelTerminate }
+
+    contao.listener.container_scope:
+        class: Contao\CoreBundle\EventListener\ContainerScopeListener
+        arguments:
+            - "@service_container"
+        tags:
+            - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 30 }
+            - { name: kernel.event_listener, event: kernel.finish_request, method: onKernelFinishRequest, priority: -254 }
```
leofeyer: @aschempp Why is the priority -254 and not -30?

aschempp: To really be the last one… but I guess -30 is fine too.

aschempp:
Looks good to me (assuming you did not change the actual content, just move stuff around).

leofeyer:
I didn't change anything.

src/Resources/config/listener.yml
leofeyer: Can we be sure that it is the last one when the priority is -254? Symfony defines things with -1000 if I remember correctly.

aschempp: It will not be guaranteed to be the last one, no matter what priority we choose (except [`-INF`](http://php.net/manual/en/math.constants.php#constant.inf) ^^. The [Symfony documentation](http://symfony.com/doc/current/cookbook/service_container/event_listener.html) says to choose a priority between +255 and -255

> There is an additional tag option priority that is optional and defaults to 0. This value can be from -255 to 255, and the listeners will be executed in the order of their priority (highest to lowest). This is useful when you need to guarantee that one listener is executed before another.

Though that's not really correct, as it can be any integer value.

leofeyer: Then why did you not choose -255?

Issue #206, opened 2015-04-15
Destructors don’t get executed


ausi:
Because the registered shutdown function `BackendUser::storeSession()` results in a fatal error *“Failed to start the session because headers have already been sent”* some `__destruct`ors don’t get executed, e.g. `Config::__destruct()`.

Because of that, things like changing settings or disabling maintenance mode via the backend doesn’t work and fails silently without an error message.

Toflar:
I tackled the User issue with: https://github.com/contao/core-bundle/pull/207

leofeyer:
This should be resolved after we have merged #207.

aschempp:
?
This can be closed if @Toflar is fixing that in #207

leofeyer:
I left it open so I don't forget to check if the PR really fixes the issue.

leofeyer:
It did :)

Issue #143, opened 2015-03-23
Referrer management broken


importer:
> <a href="https://github.com/leofeyer"><img src="https://avatars.githubusercontent.com/u/1192057?v=3" align="left" width="42" height="42" hspace="10"></img></a> [Issue](https://github.com/contao/contao/issues/65) by @leofeyer
Monday Mar 09, 2015 at 18:49 GMT

Somewhere along the way we broke the referrer management in the back end. Steps to reproduce:

- Log into the back end.
- Open the article module.
- Open any article for editing.
- Edit any content element.
- Click on the "back" button.

It keeps sending me to the welcome screen.

Toflar:
This one can be closed in favor of #207.

leofeyer:
There is still a little problem with the referer URLs:

1. Log into the back end.
2. Open the articles module.
3. Open an article for editing.
4. Open a content element for editing.
5. Click on the "back" button (correctly takes you back to the article).
6. Click on the "back" button of the article (takes you to the home page instead of the article listing).

@Toflar Can you please check if this is a bug in the new session listeners?

aschempp:
@leofeyer please also try this in the Demo, I'm pretty sure that problem always existed…

leofeyer:
I did try it in the demo and it works in Contao 3. Otherwise I would not have reopened the ticket.

leofeyer:
Fixed in b1794950629cd8e2a953d12c3fadf88732008f01.

PR #207, opened 2015-04-15
[RTM] Move the user session handling to a separate listener

Toflar:
Follow-Up PR to https://github.com/contao/core-bundle/pull/184.

Now also contains the FrontendUser handling and does the loading process as well. In addition, `TL_REFERER_ID` has been replaced as well.

Tests still to be added.

### DIFF
--- a/src/EventListener/UserSessionListener.php
+++ b/src/EventListener/UserSessionListener.php
@@ -218,14 +218,13 @@ class UserSessionListener extends ScopeAwareListener
         $bag = $this->getSessionBag();
 
         $refererOld = $bag->get('referer');
-
         if (!$this->canModifyFrontendSession($request, $refererOld)) {
             $this->storeSession();
             return;
         }
 
         $refererNew = [
-            'last'      => $refererOld['current'],
+            'last'      => (string) $refererOld['current'],
             'current'   => $this->getRelativeRequestUri($request)
         ];
 
--- a/src/EventListener/UserSessionListener.php
+++ b/src/EventListener/UserSessionListener.php
@@ -147,20 +147,13 @@ class UserSessionListener extends ScopeAwareListener
 
         $ref = $request->query->get('ref', '');
 
+        // If the referer param is in the URL and in the session,
+        // current is moved to last
         if ('' !== $ref && isset($refererOld[$ref])) {
-            if (!isset($refererOld[$refererId])) {
-                $refererOld[$refererId] = [];
-            }
-
-            $refererNew[$refererId] = array_merge(
-                $refererOld[$refererId],
-                $refererOld[$ref]
-            );
             $refererNew[$refererId]['last'] = $refererOld[$ref]['current'];
-        } elseif (count($refererOld) > 1) {
-            $refererNew[$refererId] = end($refererOld);
         }
 
+        // Set new current referer
         $refererNew[$refererId]['current'] = $this->getRelativeRequestUri($request);
 
         $bag->set($key, $refererNew);
--- a/src/EventListener/UserSessionListener.php
+++ b/src/EventListener/UserSessionListener.php
@@ -228,7 +228,7 @@ class UserSessionListener extends ScopeAwareListener
             'current'   => $this->getRelativeRequestUri($request)
         ];
 
-        $this->session->set('referer', $refererNew);
+        $bag->set('referer', $refererNew);
 
         $this->storeSession();
     }
@@ -252,7 +252,9 @@ class UserSessionListener extends ScopeAwareListener
         if (!$request->query->has('pdf')
             && !$request->query->has('file')
             && !$request->query->has('id')
-            && $refererOld['current'] !==  $this->getRelativeRequestUri($request)
+            && (isset($refererOld['current'])
+                && $refererOld['current'] !== $this->getRelativeRequestUri($request)
+            )
         ) {
             return true;
         }
--- a/src/EventListener/UserSessionListener.php
+++ b/src/EventListener/UserSessionListener.php
@@ -142,21 +142,20 @@ class UserSessionListener extends ScopeAwareListener
         $key        = $request->query->has('popup') ? 'popupReferer' : 'referer';
         $refererId  = $request->attributes->get('_contao_referer_id');
         $bag        = $this->getSessionBag();
-        $refererOld = $this->prepareBackendReferer($bag->get($key), $refererId);
-        $refererNew = [];
+        $referer    = $this->prepareBackendReferer($bag->get($key), $refererId);
 
         $ref = $request->query->get('ref', '');
 
         // If the referer param is in the URL and in the session,
         // current is moved to last
-        if ('' !== $ref && isset($refererOld[$ref])) {
-            $refererNew[$refererId]['last'] = $refererOld[$ref]['current'];
+        if ('' !== $ref && isset($referer[$ref])) {
+            $referer[$refererId]['last'] = $referer[$ref]['current'];
         }
 
         // Set new current referer
-        $refererNew[$refererId]['current'] = $this->getRelativeRequestUri($request);
+        $referer[$refererId]['current'] = $this->getRelativeRequestUri($request);
 
-        $bag->set($key, $refererNew);
+        $bag->set($key, $referer);
 
         $this->storeSession();
     }
@@ -283,7 +282,7 @@ class UserSessionListener extends ScopeAwareListener
      */
     private function storeSession()
     {
-        $user   = $this->getUserObject();
+        $user = $this->getUserObject();
 
         $this->connection->prepare('UPDATE ' . $user->getTable() . ' SET session=? WHERE id=?')
             ->execute([
@@ -312,7 +311,7 @@ class UserSessionListener extends ScopeAwareListener
      */
     private function getRelativeRequestUri(Request $request)
     {
-        return substr(
+        return (string) substr(
             $request->getRequestUri(),
             strlen($request->getBasePath()) + 1
         );
--- a/tests/EventListener/UserSessionListenerTest.php
+++ b/tests/EventListener/UserSessionListenerTest.php
@@ -257,6 +257,11 @@ class UserSessionListenerTest extends TestCase
         $request->attributes->set('_contao_referer_id', 'dummyTestRefererId');
         $request->server->set('REQUEST_URI', '/path/of/contao?having&query&string=1');
 
+        $requestWithRefInUrl = new Request();
+        $requestWithRefInUrl->attributes->set('_contao_referer_id', 'dummyTestRefererId');
+        $requestWithRefInUrl->server->set('REQUEST_URI', '/path/of/contao?having&query&string=1');
+        $requestWithRefInUrl->query->set('ref', 'dummyTestRefererId');
+
         return [
             'Test current referer null returns correct new referer for back end scope' => [
                 ContaoCoreBundle::SCOPE_BACKEND,
@@ -269,6 +274,27 @@ class UserSessionListenerTest extends TestCase
                 [
                     'dummyTestRefererId' => [
                         'last'      => '',
+                        // Make sure this one never contains a / at the beginning
+                        'current'   => 'path/of/contao?having&query&string=1'
+                    ]
+                ]
+            ],
+            'Test referer returns correct new referer for back end scope' => [
+                ContaoCoreBundle::SCOPE_BACKEND,
+                'contao_backend',
+                'Contao\\BackendUser',
+                'tl_user',
+                $requestWithRefInUrl,
+                'referer',
+                [
+                    'dummyTestRefererId' => [
+                        'last'      => '',
+                        'current'   => 'hi/I/am/your_current_referer.html'
+                    ]
+                ],
+                [
+                    'dummyTestRefererId' => [
+                        'last'      => 'hi/I/am/your_current_referer.html',
                         // Make sure this one never contains a / at the beginning
                         'current'   => 'path/of/contao?having&query&string=1'
                     ]
--- a/tests/EventListener/UserSessionListenerTest.php
+++ b/tests/EventListener/UserSessionListenerTest.php
@@ -19,6 +19,7 @@ use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Security\Core\Authentication\Token\AnonymousToken;
 
 /**
  * Tests the UserSessionListener class.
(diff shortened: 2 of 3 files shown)


ausi:
This PR fixes #206. But the session doesn’t work for the theme import in [Theme.php:78](https://github.com/Toflar/core-bundle/blob/a741505a7a0602e464dc98bfe64b82586e6b1afa/src/Resources/contao/classes/Theme.php#L78) and [Theme.php:107](https://github.com/Toflar/core-bundle/blob/a741505a7a0602e464dc98bfe64b82586e6b1afa/src/Resources/contao/classes/Theme.php#L107). Is this issue related to this PR or should I create a new ticket for it?

Toflar:
The legacy `Session` class is transparently mapped. Why does it not work? See https://github.com/contao/core-bundle/pull/68/files#diff-2f0eb5d18e0f5fb0dbaef526bb165d41R415

ausi:
It does work if I remove the line [UserSessionListener.php:163](https://github.com/Toflar/core-bundle/blob/a741505a7a0602e464dc98bfe64b82586e6b1afa/src/EventListener/UserSessionListener.php#L163) `!$request->query->has('key')`. The URL of the theme import is `/contao?do=themes&key=importTheme&rt=...&ref=...`

Toflar:
And why does it work in 3.5? https://github.com/contao/core/blob/master/system/modules/core/classes/BackendUser.php#L108

ausi:
Because `$this->Database->prepare("UPDATE...` is outside of the `if` in Contao 3.5: https://github.com/contao/core/blob/8f678aaf08102432f943241a90d13c1c308741bd/system/modules/core/classes/BackendUser.php#L150-L155

Toflar:
Okay, so storing the session should always be done.

Toflar:
Can you try c2cd469?

ausi:
Theme import works with c2cd469797386f10b02f29fb4648ec18071f3d32, thanks!

But it seems there are problems with the referrer management. If you edit a content element and hit “save & close” it takes you back to the backend home page.

bytehead:
See #143

Toflar:
Should work in c04d6fe.

ausi:
> Should work in c04d6fe96a7099485728fa0b33f3d34c2869896d.

Referrer management works now, thanks again!

src/EventListener/UserSessionListener.php
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * Checks whether the user has session data stored and passes it to the
+     * session service if so.
+     *
+     * @param GetResponseEvent $event The event object
+     */
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        if (!$this->isFrontendMasterRequest($event)
```
aschempp: Will that mean we do not have a session on subrequests?

src/EventListener/UserSessionListener.php
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    /**
+     * Stores the session for the front end.
+     *
+     * @param Request $request
+     */
+    private function storeFrontendSession(Request $request)
+    {
+        /** @var AttributeBagInterface $bag */
```
aschempp: Why are these type hints necessary if the method has a correct `@return` block?

src/EventListener/UserSessionListener.php
```diff
@@ -0,0 +1,274 @@
...
+        return $this->session->getBag($bag);
+    }
+
+    /**
+     * Stores the session data in the database.
+     */
+    private function storeSession()
+    {
+        $user   = $this->getUserObject();
+        $table  = $this->isFrontendScope() ? 'tl_member' : 'tl_user';
```
aschempp: It would be better to add a `getTable` method to the user object.

Toflar: I don't want to change the legacy code too much.

src/EventListener/UserSessionListener.php
Toflar: Ah, that was needed before I put this in its own method. Removed in 4bb2e5c952f443529c76332796389cb12ccaf011.

src/EventListener/UserSessionListener.php
Toflar: Yes, which is correct in my opinion. If you want to have session data you should take the session of the master request.

src/Resources/config/listener.yml
```diff
@@ -87,3 +87,17 @@ services:
...
+        calls:
+            - [setContainer, ["@service_container"]]
+        arguments:
+            - '@session'
+            - '@doctrine.dbal.default_connection'
+        tags:
+            # Priority must be lower than contao.listener.session (64)
+            - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
+            # Priority must be higher than the framework bundle save session listener (-1000)
+            - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse }
```
leofeyer: I assume that Symfony uses the `kernel.response` event as well and you have copied it from there? Otherwise, why don't we use the `kernel.terminate` event?

src/EventListener/UserSessionListener.php
leofeyer: Since `User::storeSession()` is a public function, why don't you just do `$this->getUserObject()->storeSession()`? We could leave most of the legacy code in the `*User` classes then.

src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1003,6 +1003,7 @@ public static function redirect($strLocation, $intStatus=303)
 			exit;
 		}
 
+		$strLocation = ltrim($strLocation, '/');
```
leofeyer: Why is this necessary?

src/Resources/config/listener.yml
Toflar: I did not copy, no. Actually, this is the whole thing about the problem. We have to read and store away the session into the database before the framework bundle listener closes the session.

src/EventListener/UserSessionListener.php
Toflar: Because I think we want to keep this functionality even if we remove our user classes once. That way, the work is already done.

src/Resources/config/listener.yml
leofeyer: And the framework bundle listener listens to the `kernel.response` event?

Toflar: https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpKernel/EventListener/SaveSessionListener.php#L63

src/Resources/contao/library/Contao/Controller.php
Toflar: Because `$request->getRequestUri()` returns the URI including the leading `/` which is correct. Our `Controller::redirect()` method would turn things into something like that: `http://domain.de//path/to/whatever`. I thought it's better to ensure there's never a leading `/` than doing stupid `str_replace()` calls **everytime** we pass something into that legacy method.

src/EventListener/UserSessionListener.php
leofeyer: Not sure about that. Look at this issue in our [Ahrweiler-TODO](https://github.com/contao/symfony-todo/issues/3):

> Nur State-Informationen in die DB-Session, Referer und Token in `$_SESSION`

This means that we have to rework the session handling, therefore I don't think we will keep the current code.

We should either keep this change as simple as possible or incorporate the new feature, shouldn't we?

src/Resources/config/listener.yml
leofeyer: Ok, then this is fine :)

src/Resources/contao/library/Contao/Controller.php
leofeyer: But this also means that we won't be able to detect wrong input anymore. Also, I think the `ltrim()` call should be on line 1013:

```php
// Make the location an absolute URL
if (!preg_match('@^https?://@i', $strLocation))
{
	$strLocation = \Environment::get('base') . ltrim($strLocation, '/');
}
```

src/EventListener/UserSessionListener.php
Toflar: Well, basically this is already done then :-) If you look at the code carefully you notice that this works well together with my first PR that integrated the session using the session bags `contao_frontend` and `contao_backend`. The listener here **only** considers those two!! This results in the easy fact that if you want to have Contao store away your session values you add it to either the `contao_frontend` or the `contao_backend` AttributeBag. If you want to have it as stateless you just write to the default SF2 attribute bags or your own :-)

src/Resources/contao/library/Contao/Controller.php
Toflar: Why would you want to detect wrong input. The method just doesn't work if you pass a leading `/` so why do you think it's wrong if it ensures that there's none automatically? And yes, we could move it down there if you prefer.

src/EventListener/UserSessionListener.php
leofeyer: Ok, that's great. But I'd still prefer to leave the legacy code in the `*User` classes. It would also make it easier to synchronize it with the Contao 3 repository.

src/Resources/contao/library/Contao/Controller.php
leofeyer: > Why would you want to detect wrong input

Obviously, because I would want to fix it :)

Toflar: That only makes sense if the automated fixing could generate false positives. But in this case it cannot. 
Your proposal should look like this

```php
if ($strLocation[0] == '/')
{
    throw new InvalidArgumentException('The target location string contains a leading slash which it must not!');
}
```

**THAT** would be detecting wrong input. Right now it just does nothing. It passes it on and generates wrong URL's. Either you fix it automatically (which you can do in this case) or you have to throw an Exception :-)

leofeyer: You are right. Then keep it as is and only move it down to line 1013.

Toflar: Okay. Done in ea0299a156f67d5e79da6698a1c91c2ce7168f57.

src/EventListener/UserSessionListener.php
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * Stores the session for the back end.
+     *
+     * @param Request $request
+     */
+    private function storeBackendSession(Request $request)
+    {
+        if (!$this->canModifyBackendSession($request)) {
+            $this->storeSession();
```
leofeyer: Why are you calling `$this->storeSession()` if the session cannot be modified anyway?

src/EventListener/UserSessionListener.php
aschempp: I'm for keeping @Toflar's code, because we will need the new code at some point (e.g. when introducing advanced security).

Toflar: The old code uses old classes such as `Database`, `Session` and `Input` and the constant  `TL_REFERER_ID`. I already did all the refactoring work and it's a lot better now and we can even write tests for it. Especially as synchronizing with the Contao 3 repository is not an argument at all because if it was one it would mean that we can never move on to new things because if we do syncing with Contao 3 is complicated :-)

src/EventListener/UserSessionListener.php
Toflar: Because you always need to store the data of the session? We only modify the referer part but there are a lot of other things going on. This is also done in Contao 3 (storing is outside the `if` clause). Also see @ausi's comment here: https://github.com/contao/core-bundle/pull/207#issuecomment-93442689

src/EventListener/UserSessionListener.php
leofeyer: Ok, but then you might as well add the `getTable()` method as @aschempp mentioned above :)

leofeyer:
Looks good to me. @Toflar Do you happen to have time to move the referer handling to a separate session bag, which is not stored in the database? The same applies to the request token and the referer ID.

Or would you prefer to create a separate PR for this?

src/EventListener/UserSessionListener.php
Toflar: That's true, shall I?

Toflar:
> Looks good to me. @Toflar Do you happen to have time to move the referer handling to a separate session bag, which is not stored in the database? The same applies to the request token and the referer ID.

I wanted to do some more work on the weekend but would have spent that on writing tests for this PR because I think it's more important to finish this first. What's your opinion?

Toflar:
> Or would you prefer to create a separate PR for this?

Oh, and yes!

src/EventListener/UserSessionListener.php
leofeyer: +1

leofeyer:
> I wanted to do some more work on the weekend but would have spent that on writing tests for this PR because I think it's more important to finish this first. What's your opinion?

+1

src/EventListener/UserSessionListener.php
Toflar: Done in ec4b90a3c68c3e16463c5d049ec9562913304c0e.

src/Resources/contao/library/Contao/Controller.php
aschempp: Sorry, but this must be wrong. If the input contains a leading slash, it's an absolute URL (probably generated by the Symfony router). Removing the leading slash does not (always) correctly convert an absolute to a relative URL! We had exactly the same problems with TL_SCRIPT.

Toflar: Fixed in ed1a2b37d8f5cb7fa84b1c58d9764e401b7cfc43.

leofeyer:
Is this RTM then?

Toflar:
Like 3 hours ago you were +1 on writing tests for it? How can it possibly be RTM now??

leofeyer:
I though you wanted to add the tests on the weekend, rather than implementing the new feature. But I was not aware that I have to wait for the tests to be completed to merge this PR.

But it probably makes sense this way :smiley:

Toflar:
@discordier, @leofeyer: I tried to mock `__get()` for 2.5 hours now and I did it in any possible way but I just can't figure out why it doesn't work :-( Giving up for today.

discordier:
@Toflar see https://github.com/Toflar/core-bundle/pull/1 ;)

When mocking special methods like `__get()` which have a coded arguments, you have to provide a base class implementing it. This was missing in our fixture class.

I also fixed a second problem which happened to occur in `Contao\CoreBundle\Test\Security\Authentication\ContaoAuthenticatorTest::testAuthenticateToken()` and tried to use `Contao\User` as `User` while the name was already been taken.
This is most likely still due our massive usage of `class_alias()`.

The error was: `PHP Fatal error:  Cannot use Contao\User as User because the name is already in use in tests/Fixtures/library/FrontendUser.php on line 5`

Toflar:
Ah again because of those stupid fixtures. Can't we get rid of them finally? @aschempp: Any reason why you are using fixtures in your tests instead of mocking the classes?

leofeyer:
They are actually dummy classes and not fixtures. And I don't think we can get rid of them easily. Couldn't write tests for the event listener without using dummy classes :(

Toflar:
PHPUnit can mock classes that don't exist using `getMockBuilder()`. We **must** get rid of those dummy classes using the mock builder and/or adapter classes. Green PHPUnit outputs won't be satisfying me as long as we have unpredictable dummy classes in our test suite. But that's not something you need to care about. I've added an issue https://github.com/contao/core-bundle/issues/226 so I don't forget about it.

Toflar:
\*phew* - this is as good as I can get it for now. It's 94.39% coverage on lines. We can still add tests but I guess it's good enough to merge now. Changing to RTM.

PR #214, opened 2015-04-16
[RTM] Rework the exception listeners

leofeyer:
This is the PR regarding our exception listener discussion yesterday. There are now three exception listeners:

 - `ResponseExceptionListener`: takes care of all response exceptions.
 - `ExceptionConverterListener`: converts runtime exceptions into HTTP exceptions.
 - `PrettyErrorScreenListener`: converts HTTP exceptions into pretty error screens.


src/EventListener/ExceptionConverterListener.php
```diff
@@ -0,0 +1,92 @@
...
+        'Contao\\CoreBundle\\Exception\\InvalidRequestTokenException'    => 'BadRequestHttpException',
+        'Contao\\CoreBundle\\Exception\\NoActivePageFoundException'      => 'InternalServerErrorHttpException',
+        'Contao\\CoreBundle\\Exception\\NoLayoutSpecifiedException'      => 'InternalServerErrorHttpException',
+        'Contao\\CoreBundle\\Exception\\NoRootPageFoundException'        => 'InternalServerErrorHttpException',
+        'Contao\\CoreBundle\\Exception\\PageNotFoundException'           => 'NotFoundHttpException',
+        'Contao\\CoreBundle\\Exception\\ServiceUnavailableException'     => 'ServiceUnavailableHttpException',
+    ];
+
+    /**
+     * Adds the referer ID to the request.
```
aschempp: ?

src/EventListener/ExceptionConverterListener.php
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @param GetResponseForExceptionEvent $event The event object
+     */
+    public function onKernelException(GetResponseForExceptionEvent $event)
+    {
+        $exception = $event->getException();
+        $class     = get_class($exception);
```
aschempp: We should check for inheritance too.

src/EventListener/ExceptionConverterListener.php
```diff
@@ -0,0 +1,92 @@
...
+     * Converts an exception to an HTTP exception.
+     *
+     * @param \Exception $exception The exception object
+     * @param string     $target    The target exception
+     *
+     * @return HttpException|null The HTTP exception or null
+     */
+    private function convertToHttpException(\Exception $exception, $target)
+    {
+        switch ($target) {
```
aschempp: This can be simplified like so:

```php
if ('ServiceUnavailableHttpException' === $target)
    return new ServiceUnavailableHttpException(null, $exception->getMessage(), $exception);
}

return new $target($exception->getMessage(), $exception);
```

src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -0,0 +1,312 @@
...
+     *
+     * @param string $type The error type
+     *
+     * @return Response|null The response object or null
+     */
+    private function renderPageHandler($type)
+    {
+        $type = "error_$type";
+
+        if (!isset($GLOBALS['TL_PTY'][$type]) || !class_exists($GLOBALS['TL_PTY'][$type])) {
```
aschempp: This can't work as type would be `403` or `404`, but should be `error_403` or `error_404` (if I remember correctly).

src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -0,0 +1,312 @@
...
+    {
+        $event->setResponse($this->renderTemplate('service_unavailable', 503));
+    }
+
+    /**
+     * Checks the exception chain for a known exception.
+     *
+     * @param GetResponseForExceptionEvent $event The event object
+     */
+    private function checkExceptionChain(GetResponseForExceptionEvent $event)
```
aschempp: That's a bit confusing. I would rather call the method something like `setResponseFromExceptionChain`.

src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -0,0 +1,312 @@
...
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException;
+
+/**
+ * Renders pretty error screens for exceptions.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
```
aschempp: I think you should add @discordier here too

src/EventListener/ResponseExceptionListener.php
```diff
@@ -0,0 +1,45 @@
...
+            return;
+        }
+
+        $response = $exception->getResponse();
+
+        if (!$response instanceof Response) {
+            return;
+        }
+
+        $event->setResponse($response);
```
aschempp: You are no longer setting the `X-Status-Code` flag.

src/EventListener/ExceptionConverterListener.php
leofeyer: Maps known exceptions to HTTP exceptions?

aschempp: Not sure if my "?" was not clear enough: the method description is wrong ;-)

src/EventListener/ExceptionConverterListener.php
leofeyer: No, it cannot (meaning it can be simplified but not like this). You have removed the "return `null` if the exception is unknown" case.

src/EventListener/PrettyErrorScreenListener.php
leofeyer: If we change the name, it should begin with `render` like the other methods, shouldn't it?

src/Resources/config/listener.yml
```diff
@@ -33,14 +41,10 @@ services:
...
-        class: Contao\CoreBundle\EventListener\ExceptionListener
-        arguments:
-            - "%contao.pretty_error_screens%"
-            - "@templating"
-            - "@contao.adapter.config"
+    contao.listener.exception_converter:
+        class: Contao\CoreBundle\EventListener\ExceptionConverterListener
         tags:
-            - { name: kernel.event_listener, event: kernel.exception, method: onKernelException }
+            - { name: kernel.event_listener, event: kernel.exception, method: onKernelException, priority: 64 }
```
aschempp: Why do we need to set priorities at all?

src/EventListener/PrettyErrorScreenListener.php
leofeyer: I'll fix this with the next commit. Had been removed accidentally.

src/EventListener/ResponseExceptionListener.php
leofeyer: No. Don't have to.

src/EventListener/ExceptionConverterListener.php
aschempp: We are passing in the value of our mapper array, so we always know the cases? There can never be `null`?

src/Resources/config/listener.yml
leofeyer: Because it is vital that the exception converter listener comes before the pretty error screen listener?

src/EventListener/ExceptionConverterListener.php
leofeyer: I know and I suggested to use "Maps known exceptions to HTTP exceptions" instead. You like?

src/Resources/config/listener.yml
aschempp: I don't think so, as we are running the same chain again (and not pass our converted exception to the next listener).

src/EventListener/ExceptionConverterListener.php
leofeyer: You are right.

src/EventListener/ResponseExceptionListener.php
aschempp: I had a long discussion with @discordier why that is necessary. Can you explain why you think its not?

src/EventListener/PrettyErrorScreenListener.php
aschempp: That would work I guess.

src/EventListener/ExceptionConverterListener.php
aschempp: Ah, sure that would work.

src/EventListener/ExceptionConverterListener.php
leofeyer: It still does not work, because `$target` is the short class name only.

src/Resources/config/listener.yml
leofeyer: Uh, nope.

src/EventListener/ResponseExceptionListener.php
leofeyer: It does not make a difference in the current implementation.

src/EventListener/ExceptionConverterListener.php
aschempp: doh. Maybe we can pass in the mapper array key?

leofeyer: It would work like this:

```php
        if ('ServiceUnavailableHttpException' === $target) {
            return new ServiceUnavailableHttpException(null, $exception->getMessage(), $exception);
        }
        
        $class = "Symfony\\Component\\HttpKernel\\Exception\\$target";

        return new $target($exception->getMessage(), $exception);
```

But I think the current way is cleaner.

src/EventListener/ResponseExceptionListener.php
aschempp: I disagree. The status code of your `Response` would not be sent to the browser.

src/EventListener/InitializeSystemListener.php
```diff
@@ -325,12 +325,12 @@ private function validateInstallation(Request $request = null)
 
         // Show the "insecure document root" message
         if (!in_array($request->getClientIp(), ['127.0.0.1', 'fe80::1', '::1']) && '/web' === substr($request->getBasePath(), -4)) {
-            throw new InsecureInstallationHttpException(null, 'Your installation is not secure. Please set the document root to the <code>/web</code> subfolder.');
+            throw new InsecureInstallationException('Your installation is not secure. Please set the document root to the <code>/web</code> subfolder.');
```
discordier: We now also should remove the `<code>` tag from the message as it is not an http exception anymore.

src/EventListener/InitializeSystemListener.php
```diff
@@ -382,7 +382,7 @@ private function handleRequestToken(Request $request = null)
                 throw new AjaxRedirectResponseException($this->router->generate('contao_backend'));
             }
 
-            throw new BadRequestTokenHttpException(
+            throw new InvalidRequestTokenException(
                 'Invalid request token. Please <a href="javascript:window.location.href=window.location.href">reload the page</a> and try again.'
```
discordier: HTML code should be stripped from the exception message here as well.

src/EventListener/PrettyErrorScreenListener.php
discordier: Correct, but rendering them like this in the legacy way is still ugly.
I still would prefer providing route names and sub requests.

src/Exception/ResponseException.php
```diff
@@ -13,25 +13,40 @@
...
-     * @param int   $status   The response status code (defaults to 204)
-     * @param array $headers  An array of response headers
+     * @param Response   $response The Response object
+     * @param \Exception $previous The previous exception
+     */
+    public function __construct(Response $response, \Exception $previous = null)
+    {
+        $this->response = $response;
+
+        parent::__construct($response->getContent(), 0, $previous);
```
discordier: This is bad as the content of the response may be changed but the message not.

leofeyer: Fixed in b920073ef6bab8fdf0645cc51991e334d5dd4b16.

src/EventListener/ResponseExceptionListener.php
discordier: Has been restored in 6e9d445e837f429efc6d8a2cabe6705689c66ddc and improved in 01812bf87610ddcddf1acc0a9d60103082c806dd

leofeyer:
I have updated the status to RTM.

PR #215, opened 2015-04-16
[RTM] Fix symlinks command

bytehead:
This commit fixes an issue with the symlink command (`php app/console contao:symlinks`) to symlink the `files`-directory (upload directory).

**Before (e.g. `$uploadPath = 'files'`):**
`files` gets a `/` added and the result is `../../files` and ends up with

```bash
[Symfony\Component\Filesystem\Exception\IOException]                                                                                              
  Failed to create symbolic link from "../../files/" to "/var/www/dev/contao-4/standard-edition-develop/web/files/".
```

**After (same `$uploadPath`):**
```bash
Added web/files as symlink to ../files.
Added system/themes/flexible as symlink to ../../vendor/contao/core-bundle/src/Resources/contao/themes/flexible.
Added web/assets as symlink to ../assets.
Added web/system/themes as symlink to ../../system/themes.
Added system/logs as symlink to ../app/logs.
```
Secondary, I changed some code style within the same file.

### DIFF
--- a/src/Command/SymlinksCommand.php
+++ b/src/Command/SymlinksCommand.php
@@ -212,11 +212,11 @@ class SymlinksCommand extends LockedCommand implements ContainerAwareInterface
      */
     private function validateSymlink($source, $target, $rootDir)
     {
-        if ($source == '') {
+        if ($source === '') {
             throw new \InvalidArgumentException('The symlink source must not be empty.');
         }
 
-        if ($target == '') {
+        if ($target === '') {
             throw new \InvalidArgumentException('The symlink target must not be empty.');
         }
 
--- a/src/Command/SymlinksCommand.php
+++ b/src/Command/SymlinksCommand.php
@@ -159,7 +159,13 @@ class SymlinksCommand extends LockedCommand implements ContainerAwareInterface
     {
         /** @var SplFileInfo $file */
         foreach ($finder as $file) {
-            $path = $prepend . '/' . $file->getRelativePath();
+            $path = $prepend;
+
+            // Don't add a trailing slash, when relativePath is ''
+            if (strlen($file->getRelativePath())) {
+                $path .= '/' . $file->getRelativePath();
+            }
+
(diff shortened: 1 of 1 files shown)


bytehead:
I could also extend the tests on this command/class if you want.

src/Command/SymlinksCommand.php
```diff
@@ -206,11 +212,11 @@ private function symlink($source, $target, $rootDir, OutputInterface $output)
      */
     private function validateSymlink($source, $target, $rootDir)
     {
-        if ($source == '') {
+        if ($source === '') {
```
leofeyer: Should be `('' === $source)`, although of course this is not related to this PR :)

leofeyer:
How do I reproduce the initial problem?

src/Command/SymlinksCommand.php
bytehead: I was not sure about that, but that's fine. I will change that.

bytehead:
I can just run `php app/console contao:symlinks` in the root directory. And be sure to have a `.public` file in the `/files` folder (the reason for my question yesterday).

leofeyer:
Ah, I see. I had not tested putting a `.public` file into the upload folder root yet :)

bytehead:
That was my thought. ;-) I stumbled while installing the system from scratch. There is some other stuff going wrong (also stuff, that probably needs to be discussed about), but that will come for discussion in other tickets.

leofeyer:
Fixed in c81b36c288509153a8e1ce7ffe0b151f5ab50234. I used `rtrim($path, '/')` to make sure there is never a trailing slash.

bytehead:
Even better :)

discordier:
Could you guys please write unit tests reproducing the original issues along with the fixes?
Otherwise we could have [regression bugs](http://en.wikipedia.org/wiki/Regression_testing).

For essential stuff like the symlinking it is definately worth the hassle.

PR #221, opened 2015-04-17
Database schema: Specified key was too long

bytehead:
During a fresh install on `contao/install` while creating the tables:

```
An exception occurred while executing 'CREATE TABLE `tl_files` (
`id` int(10) unsigned NOT NULL auto_increment,
`pid` binary(16) NULL,
`tstamp` int(10) unsigned NOT NULL default '0',
`uuid` binary(16) NULL,
`type` varchar(16) NOT NULL default '',
`path` varchar(1022) NOT NULL default '',
`extension` varchar(16) NOT NULL default '',
`hash` varchar(32) NOT NULL default '',
`found` char(1) NOT NULL default '1',
`name` varchar(255) NOT NULL default '',
`importantPartX` int(10) NOT NULL default '0',
`importantPartY` int(10) NOT NULL default '0',
`importantPartWidth` int(10) NOT NULL default '0',
`importantPartHeight` int(10) NOT NULL default '0',
`meta` blob NULL,
PRIMARY KEY (`id`),
KEY `pid` (`pid`),
UNIQUE KEY `uuid` (`uuid`),
KEY `path` (`path`),
KEY `extension` (`extension`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE utf8_general_ci;':

SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 1000 bytes
```

It would acutally work with `varchar(333)` - max key length = 1000 bytes, UTF-8 character uses 3 bytes, so max length of the `varchar` is 333 (see 5fbe4a9).

What's the reason to use `varchar(1022)` for the field `path` ?

### DIFF
--- a/src/Resources/contao/dca/tl_files.php
+++ b/src/Resources/contao/dca/tl_files.php
@@ -148,7 +148,7 @@ $GLOBALS['TL_DCA']['tl_files'] = array
 		'path' => array
 		(
 			'eval'                    => array('unique'=>true),
-			'sql'                     => "varchar(1022) NOT NULL default ''",
+			'sql'                     => "varchar(333) NOT NULL default ''",
 		),
 		'extension' => array
 		(


leofeyer:
We need the field to be longer than 255 characters. And we still want to set a btree index, because in most of the cases, the paths will be shorter than 333 characters.

The discussion is here: https://github.com/contao/core/issues/5556

leofeyer:
I just tried to reproduce the issue, but the tables were all created without problems. Which MySQL version are you running?

bytehead:
5.6.12

bytehead:
I will give another shot as well (with newest develop).

leofeyer:
Are you using InnoDB? It seems you can fix it by changing the [innodb_large_prefix](http://dev.mysql.com/doc/refman/5.6/en/innodb-parameters.html#sysvar_innodb_large_prefix) setting.

bytehead:
Using `MyISAM`, charset `utf8`, collation `utf8_general_ci`

bytehead:
And still getting the exception on a fresh created database.

leofeyer:
No idea. Are you running MySQL in strict mode?

bytehead:
That's a good question! It actually looks like that:

```
mysql> SELECT @@GLOBAL.sql_mode;;
+--------------------------------------------+
| @@GLOBAL.sql_mode                          |
+--------------------------------------------+
| STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION |
+--------------------------------------------+
1 row in set (0,00 sec)
```

bytehead:
Disabling `strict` mode works. But isn't `strict` mode a default setting?

leofeyer:
Yes, we are trying to maintain strict mode compatibility.

leofeyer:
Fixed in 84f70298454b1311d7b0f6e503b47bfef82cc76e. Will have to back port this to Contao 3.5.

bytehead:
Nice, thanks!

Issue #218, opened 2015-04-17
Template::getTemplate() is no longer used by TemplateInheritance


ausi:
In Contao 3 it was possible to extend from `FrontendTemplate` and change how the template paths get resolved by overwriting the `getTemplate` method. E.g.:

```php
class MyTemplate extends FrontendTemplate {
	public static function getTemplate($template, $format = 'html5') {
		// Custom logic...
		return $path;
	}
```

This is no longer possible because `TemplateInheritance` calls `getTemplate()` on `Controller` in [TemplateInheritance.php:81](https://github.com/contao/core-bundle/blob/808bc63862816b7143b1f6abb8182cf8967a95e6/src/Resources/contao/library/Contao/TemplateInheritance.php#L81).

I’m using that in an extension, but I don’t know how common that is and if other extensions are doing the same. If you agree that this is an issue I can make a pull request.

leofeyer:
How would you resolve it?

ausi:
I’m not sure. Adding a `getTemplate` method to `TemplateInheritance` and calling `Controller::getTemplate()` from there?

leofeyer:
But the `Controller` class already has a `getTemplate()` method and since the `Template` class extends `Controller`, I don't know if that would work. But we could simply give the method another name :)

ausi:
> But we could simply give the method another name :)

It wouldn’t be backwards compatible then but much cleaner than overwriting a method of a parent class with a trait I guess :)

Should I create a pull request with a `TemplateInheritance::getTemplatePath($strTemplate, $strFormat='html5', $default = false)` method?

leofeyer:
Yes. And the method should be protected.

PR #225, opened 2015-04-18
[RTM] Add the TemplateInheritance::getTemplatePath() method (see #218).

ausi:
As discussed in #218

### DIFF
--- a/src/Resources/contao/library/Contao/TemplateInheritance.php
+++ b/src/Resources/contao/library/Contao/TemplateInheritance.php
@@ -78,7 +78,7 @@ trait TemplateInheritance
 		while ($this->strParent !== null)
 		{
 			$strCurrent = $this->strParent;
-			$strParent = $this->strDefault ?: \Controller::getTemplate($this->strParent, $this->strFormat);
+			$strParent = $this->strDefault ?: $this->getTemplatePath($this->strParent, $this->strFormat);
 
 			// Reset the flags
 			$this->strParent = null;
@@ -94,7 +94,7 @@ trait TemplateInheritance
 			}
 			elseif ($this->strParent == $strCurrent)
 			{
-				$this->strDefault = \TemplateLoader::getDefaultPath($this->strParent, $this->strFormat);
+				$this->strDefault = $this->getTemplatePath($this->strParent, $this->strFormat, true);
 			}
 
 			ob_end_clean();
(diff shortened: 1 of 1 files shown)


src/Resources/contao/library/Contao/TemplateInheritance.php
```diff
@@ -78,7 +78,7 @@ public function inherit()
 		while ($this->strParent !== null)
 		{
 			$strCurrent = $this->strParent;
-			$strParent = $this->strDefault ?: \Controller::getTemplate($this->strParent, $this->strFormat);
+			$strParent = $this->strDefault ?: static::getTemplatePath($this->strParent, $this->strFormat);
```
leofeyer: Why did you make the method static? Does `$this->getTemplatePath()` not work?

ausi: `$this->getTemplatePath()` would work. Should I change only the call to `$this->getTemplatePath()` or should I also make the method non-static?

leofeyer: Please also make the method non-static.

ausi: Done in 7cdf8d7069487b6d7ed637321eec9ab327cc8e27

aschempp:
The method should be called `getTemplate` to be BC. That works totally fine, see http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.precedence

leofeyer:
Then the method would override `Controller::getTemplate()`, wouldn't it?

aschempp:
Yes it would. But that's totally fine, as it calls `\Controller::getTemplate` internally.

leofeyer:
Not sure. @ausi What do you think?

ausi:
As I commented in #218

> > But we could simply give the method another name :)

> It wouldn’t be backwards compatible then but much cleaner than overwriting a method of a parent class with a trait I guess :)

I think overwriting a method of the parent class and calling that parent class from there could be confusing. IMO the question is if we want it clean or backwards compatible.

leofeyer:
+1 for clean.

aschempp:
It's legacy code, it does not need to be "clean", it needs to be BC ;-)
According to the PHP docs, it would be totally valid to do…

Issue #190, opened 2015-04-12
Set _locale for backend user


aschempp:
see http://symfony.com/doc/current/cookbook/session/locale_sticky_session.html

This should replace `TL_LANGUAGE` both in `$GLOBALS` and `$_SESSION`.

Toflar:
:+1:

leofeyer:
Will you create a PR?

aschempp:
I'm on it.

aschempp:
See #237

PR #227, opened 2015-04-19
[RTM] The URL suffix is not the format

aschempp:
While working on the locale listener (see #190), I noticed that it's possible to have container parameter in routes. After playing around with it, I realized that our current approach is wrong. The URL suffix is only the format in the default configuration. But allowing people to set it to *empty* or */* would make the `_format` property useless.

Therefore I propose to make it static.

Be aware that this means URLs like `/en/` (language root) would be routed by the `catch_all` route, but I think that's fine.

### DIFF
--- a/src/Resources/config/routing.yml
+++ b/src/Resources/config/routing.yml
@@ -15,3 +15,17 @@ contao_backend_redirect:
 contao_frontend:
     resource: .
     type: contao_frontend
+
+contao_root:
+    path: /
+    defaults:
+        _scope: frontend
+        _controller: ContaoCoreBundle:Frontend:index
+
+contao_catch_all:
+    path: /{_url_fragment}
+    defaults:
+        _scope: frontend
+        _controller: ContaoCoreBundle:Frontend:index
+    requirements:
+        _url_fragment: .*
--- a/src/Resources/contao/classes/Frontend.php
+++ b/src/Resources/contao/classes/Frontend.php
@@ -338,7 +338,7 @@ abstract class Frontend extends \Controller
 
 				$arrParams = array('_locale' => $objRootPage->language);
 
-				$strUrl = $kernel->getContainer()->get('router')->generate('contao_root', $arrParams);
+				$strUrl = $kernel->getContainer()->get('router')->generate('contao_index', $arrParams);
 				$strUrl = substr($strUrl, strlen(\Environment::get('path')) + 1);
 
 				static::redirect($strUrl, 301);
--- a/src/Resources/contao/classes/Frontend.php
+++ b/src/Resources/contao/classes/Frontend.php
@@ -336,11 +336,9 @@ abstract class Frontend extends \Controller
 				/** @var KernelInterface $kernel */
 				global $kernel;
 
-				$arrParams = array();
-				$arrParams['alias'] = '';
-				$arrParams['_locale'] = $objRootPage->language;
+				$arrParams = array('_locale' => $objRootPage->language);
 
-				$strUrl = $kernel->getContainer()->get('router')->generate('contao_frontend', $arrParams);
+				$strUrl = $kernel->getContainer()->get('router')->generate('contao_root', $arrParams);
 				$strUrl = substr($strUrl, strlen(\Environment::get('path')) + 1);
 
 				static::redirect($strUrl, 301);
@@ -426,7 +424,7 @@ abstract class Frontend extends \Controller
 		global $kernel;
 
 		$arrParams = array();
-		$arrParams['alias'] = $pageId . $strParams . \Config::get('urlSuffix');
+		$arrParams['alias'] = $pageId . $strParams;
 		$arrParams['_locale'] = $objPage->rootLanguage;
 
 		$strUrl = $kernel->getContainer()->get('router')->generate('contao_frontend', $arrParams);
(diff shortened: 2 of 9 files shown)


leofeyer:
+1

aschempp:
Just found a bug, wait with merging ;-)

aschempp:
I have updated and rebased the PR and set it to RFC.

Due to the (correct) changes on the `contao_frontend` route, the URL suffix is always added by the router. That's actually correct, previously the format was not added but we added the suffix in the alias.
To redirect to the language root, I had to change the `contao_root` route to include the locale (if appropriate).

src/Resources/contao/classes/Frontend.php
```diff
@@ -341,11 +341,9 @@ public static function getRootPageFromUrl()
 				/** @var KernelInterface $kernel */
 				global $kernel;
 
-				$arrParams = array();
-				$arrParams['alias'] = '';
-				$arrParams['_locale'] = $objRootPage->language;
+				$arrParams = array('_locale' => $objRootPage->language);
 
-				$strUrl = $kernel->getContainer()->get('router')->generate('contao_frontend', $arrParams);
+				$strUrl = $kernel->getContainer()->get('router')->generate('contao_root', $arrParams);
```
leofeyer: That's not correct. The `contao_root` route only refers to `/`, doesn't it?

src/Resources/contao/templates/backend/be_login.html5
```diff
@@ -81,7 +81,7 @@
         <p>Contao Open Source CMS :: Copyright ©2005-<?= date('Y') ?> by Leo Feyer :: Extensions are copyright of their respective owners :: Visit <a href="https://contao.org" target="_blank">contao.org</a> for more information :: Obstructing the appearance of this notice is prohibited by law!</p>
         <p>Contao is distributed in the hope that it will be useful but WITHOUT ANY WARRANTY. Without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details. Contao is free software. You can redistribute it and/or modify it under the terms of the GNU/LGPL as published by the Free Software Foundation.</p>
       </div>
-      <p id="go_to_frontend"><a href="<?= $this->route('contao_root') ?>" class="footer_preview" title="<?= $this->feLink ?>"><?= $this->feLink ?></a></p>
```
leofeyer: You change of the `contao_root` route has many ramifications. Wouldn't it be better to add a `contao_language_root` route?

src/Routing/FrontendLoader.php
```diff
@@ -85,16 +73,8 @@ public function supports($resource, $type = null)
...
-
-        // URL suffix
-        if ($this->format) {
-            $pattern .= '.{_format}';
-
-            $defaults['_format'] = $this->format;
-            $require['_format']  = $this->format;
-        }
+        $pattern = '/{alias}%contao.url_suffix%';
+        $require = ['alias' => '.+'];
```
leofeyer: Now the alias cannot be empty anymore?

aschempp: That's correct, otherwise the URL would be `de/.html` if you provide no alias with a german locale.

src/Resources/contao/classes/Frontend.php
aschempp: That's no longer the case (I guess you already noticed).

src/Resources/contao/templates/backend/be_login.html5
aschempp: It does not really have many ramifications. The existing ones are fixed.
It's a bit strange, but it's the best solution I could come up with. And I think it's actually correct, because `/` is not a valid URL if you have `prefix_locale` enabled. In cases like the backend login link, we're not actually linking to a known route, we're linking to the installation base path.

src/Resources/contao/classes/Frontend.php
leofeyer: No, you got me wrong. The route should only refer to `/`.

aschempp: I know what you mean, but that's no longer the case with my PR. Because if `prefix_locale` is enabled, the root always contains a language. There is no `/` route in that case.

src/Resources/contao/templates/backend/be_login.html5
leofeyer: I already consider it a ramification that you cannot use `$this->route('contao_root')` anymore. And `/` is a valid URL, even if `prepend_locale` is enabled. There is a "do not redirect the empty URL" option in the back end settings!

aschempp: I know, but we never generate that route. We only allow the page to be rendered (which honestly is kinda stupid, because it generates duplicate content). What language is generated in that case?

leofeyer: That does not matter for this discussion. Why don't you add a `contao_language_root` route instead?

aschempp: Because it would mean we need to `if-else` based on wether `prefix_locale` is enabled or not and use two different routes. What would you use `contao_root` route for if `prefix_locale` is enabled?

leofeyer: To redirect to the empty domain (`/`).

aschempp:
Route renamed and the unit tests now really test URL generation.

Unfortunately, the unit tests seem to fail, but due to the PrettyErrorScreen ?

leofeyer:
The pretty error screens should not be active in test mode, but I have to check. What is the error message?

aschempp:
Interestingly, there's no problem on Travis. Must be related to my localhost. It was not about the listener, but about the listener tests failing. Something about TwigEngine missing TwigEnvironment. But if Travis is ok, I guess we're fine and it's just a setup problem on my host.

aschempp:
I've added the missing stuff as discussed

src/Resources/config/routing.yml
```diff
@@ -15,3 +15,17 @@ contao_backend_redirect:
 contao_frontend:
     resource: .
     type: contao_frontend
+
+contao_root:
```
leofeyer: Isn't `contao_frontend` actually the catch all route? At least there is a comment saying "The fallback route must be the last one!" and now the route is no longer the last one :)

aschempp: That comment is outdated…

leofeyer: Then why don't you remove it?

leofeyer: I can do it, too, when merging the PR.

aschempp: Feel free to do so :-)

leofeyer:
Merged in 239e44e1e32a3ab9335e832e40de8e3f53db7944.

PR #229, opened 2015-04-21
[RTM] The Contao users must only be loaded if the correct scope is active

aschempp:
fixes https://github.com/contao/standard-edition/issues/7

### DIFF
--- a/src/Security/ContaoAuthenticator.php
+++ b/src/Security/ContaoAuthenticator.php
@@ -11,7 +11,6 @@
 namespace Contao\CoreBundle\Security;
 
 use Contao\CoreBundle\Security\Authentication\ContaoToken;
-use Contao\CoreBundle\Security\User\ContaoUserProvider;
 use Symfony\Component\Security\Core\Authentication\SimplePreAuthenticatorInterface;
 use Symfony\Component\Security\Core\Authentication\Token\AnonymousToken;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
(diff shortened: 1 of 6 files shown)


leofeyer:
Don't forget to add `[RTM]` if you are finished.

aschempp:
Should we add unit tests for that case?

leofeyer:
Yes.

leofeyer:
And please don't forget to fix the tests :)

aschempp:
I have rebased the PR and fixed/improved the unit tests.

src/Security/User/ContaoUserProvider.php
```diff
@@ -57,4 +59,20 @@ public function supportsClass($class)
     {
         return is_subclass_of($class, 'Contao\\User');
     }
+
+
+    private function isFrontend($username)
```
leofeyer: I'd prefer `isFrontendUser()` and `isBackendUser()`.

leofeyer:
Except for my one comment, the PR looks RTM to me :)

aschempp:
Name updated and Scrutinizer issue fixed.

leofeyer:
I could actually merge this PR if the tests weren't failing :)

aschempp:
Checking…

src/Security/ContaoAuthenticator.php
```diff
@@ -78,12 +63,16 @@ public function authenticateToken(TokenInterface $token, UserProviderInterface $
 
         try {
             $providerKey = $token->getKey();
-            $user        = $this->userProvider->loadUserByUsername($providerKey);
+            $user        = $userProvider->loadUserByUsername($providerKey);
```
leofeyer: IMHO, the code should be as follows:

```php
try {
    $user = $userProvider->loadUserByUsername($token->getKey());
} catch (UsernameNotFoundException $e) {
    return $token;
}

if ($user instanceof User) {
    return new ContaoToken($user);
}

return $token;
```

leofeyer: @aschempp Can I change the code as mentioned above?

aschempp: good idea, I just changed it.

leofeyer:
Merged in eb6ecce875875aec1d1d584d7b67c09bdf3796d5.

leofeyer:
@aschempp After merging this, I can open neither the back end nor the front end:

```
A Token was not found in the SecurityContext.
500 Internal Server Error - AuthenticationCredentialsNotFoundException
```

Issue #192, opened 2015-04-13
[RFC] Split up the InitializeSystemListener


Toflar:
We currently have the problem that our `InitializeSystemListener ` is pretty complex and does way too much to be tested thoroughly. For instance, it extends our `ScopeAwareListener` which means that whatever you test, you need to set a container to it even if that specific part of the listener you want to test is not related to any container at all.

Also related to https://github.com/contao/core-bundle/issues/175 I think we should really go about this again. I feel like the `InitializeSystemListener` is correctly placed but the components are wrong.

My approach (simplified):

```php
class InitializeSystemListener
{
    private $framework;

    public function __construct(ContaoFramework $framework)
    {
        $this->framework = $framework;
   }

    public function onKernelRequest(GetResponseEvent $event)
    {
        $this->framework->boot();
    }
}
```

```php
class ContaoFramework
{
    public function boot()
    {
        $bootEvent = new BootEvent();
        $this->eventDispatcher->dispatch(ContaoEvents::BOOT_FRAMEWORK, $bootEvent);
    }
}
```

This provides the possibility to have all our other responsibilities such as

* Setting constants
* Setting ini values
* Setting default language
* Error handling
* Validating the installation
* etc.

in their very own listeners listening to `ContaoEvents::BOOT_FRAMEWORK` and they only get the things injected they need :-)

* CoreBundle
    * Listener
        * BootFramework
            * SetConstantsListener.php
            * ValidateInstallationListener.php
            * etc.

RFC ;-)

leofeyer:
Reminds me a lot of #60, in which you guys have voted against changing the listener :)

Toflar:
“you guys” = @aschempp. If you read the comments...

leofeyer:
:D

leofeyer:
IMHO, we should at least create listeners for the non-legacy stuff such as checking for the `/web` subfolder.

aschempp:
I agree we could rewrite it, if we implement something like #175. But we should be aware that doing that will open third-party developers new possibilities for the legacy code. If we do that, we need to think about replacements for stuff like deprecating `iniconfig.php` or fire a real event for `initializeSystem` hook.

Same as @leofeyer wrote in #175, it does not have a very high priority to me (because the current solutions works, and we can easily change everything in any minor release because nothing is API).

Toflar:
> But we should be aware that doing that will open third-party developers new possibilities for the legacy code.

Why do you think it would give them new possibilities?

[...]

PR #234, opened 2015-04-24
[RFM] Split InitializeSystemListener

dtomasi:
see: #192 

Things that are done:

- splitted `IntitializeSystemListener` in 8 decoupled listeners
- Listeners in `Contao\CoreBundle\EventListener\Framework` are registered by a `CompilerPass` see: https://github.com/ContaoBlackForest/core-bundle/blob/feature/split-initialize-system-listener/src/ContaoCoreBundle.php#L64
- `InitializeSystemListenerTest` is rewritten to test `contao.framework.boot`-Event is dispatched and how often this happens.
- Some tests for listeners in `Contao\CoreBundle\EventListener\Framework` are already coded

Things that need to be done:

- Write tests for all listeners in `Contao\CoreBundle\EventListener\Framework`
- Write a test-setup that loads all required services and parameters to load the framework. This is needed for other Tests like `AddToSearchIndexListenerTest`.

Things that need to be discussed:

- basic structure
- naming of listeners
- should the listeners bei separated from listeners in `listeners.yml`? maybe `listeners_framework.yml` ? 
- should we also create a `contao.framework.terminate` and `contao.framework.exception`- Event ?

What are your opinions ?


src/Event/ContaoFrameworkBootEvent.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Set the Config
+     *
+     * @param ConfigAdapter $config
+     *
+     * @return $this
+     */
+    public function setConfig(ConfigAdapter $config)
```
Toflar: I don't really like this because it's not needed, right? Any listener can just get it injected :-)

Toflar:
Cool, thanks for the work so far! Looks promising :-)

> should the listeners bei separated from listeners in listeners.yml? maybe listeners_framework.yml ?

Yes, please!

> Should we also create a contao.framework.terminate and contao.framework.exception- Event ?

I think that's not within the scope of this PR but keep that in mind (or open a separate issue for it) :-)

src/Event/ContaoFrameworkBootEvent.php
dtomasi: @Toflar You are right. Injecting it makes more sense... I will change this.

src/EventListener/InitializeSystemListener.php
```diff
@@ -152,257 +85,10 @@ public function onConsoleCommand()
...
-    private function initializeLegacySessionAccess()
-    {
-        /** @var AttributeBagInterface $feBag */
-        $feBag = $this->session->getBag('contao_frontend');
-
-        /** @var AttributeBagInterface $beBag */
-        $beBag = $this->session->getBag('contao_backend');
+        $bootEvent = new ContaoFrameworkBootEvent();
+        $bootEvent->setConfig($this->config);
+        $bootEvent->setParameter('contao.mode','FE');
```
Toflar: Coding style: Add a ` ` here.

Toflar:
Saw that you generally don't put any ` ` after a `,` - this is wrong CS :)

dtomasi:
@Toflar I will do that as a short refactoring before RTM-Status. Thats missing on the list of todo´s

dtomasi:
@Toflar which standard do you use for phpcs? Is it a implemented or do you have your own?

Toflar:
PSR :) But we don't have any automated CS checker yet, afaik :)

aschempp:
Looks very promising! I think the parameter stuff looks strange, and the TL_MODE should be part of the event. See https://github.com/ContaoBlackForest/core-bundle/pull/1

Also, I think your priorities are wrong. The higher the number, the earlier the event is fired. Looks like the current sorting should be inverted.

dtomasi:
> Looks very promising!

Thanks!

> I think the parameter stuff looks strange, and the TL_MODE should be part of the event.

I introduced the `ParameterBag` for a ability to provide parameters to the listeners. In foresight to eliminate the legacy stuff piece by piece this can be used to provide thing like `TL_MODE` in a way that is more like symfony does it. I prefer to use a Parameter in `ContaoFrameworkBootEvent` like `contao.mode` instead of using `TL_MODE` like you did in ContaoBlackForest#1.

> Also, I think your priorities are wrong. The higher the number, the earlier the event is fired. Looks like the current sorting should be inverted.

You are totally right. I will change this asap.

Toflar:
Is there any reason why we're passing around the "mode" instead of the "scope" as it's called everywhere else now? :)

aschempp:
- Scope = Container Scope (`frontend` or `backend`)
- Mode = TL_MODE (`FE` or `BE`)

@dtomasi regarding the mode, I think I did understand your idea but I don't like it. There will be no additional parameters except if we add them by adding code to our listener. It's not that any third party developer could add parameters and use them in listeners. Therefore we can also simply adjust the event methods if necessary.

Toflar:
> Scope = Container Scope (frontend or backend)
> Mode = TL_MODE (FE or BE)

I know but why do we even need the mode anymore? If the scope is "backend" then for BC reasons we set `TL_MODE` to "BE" and the same for the front end. That can be hard coded because there are no other modes :)

aschempp:
> That can be hard coded because there are no other modes

Valid point I guess. I think we don't need to allow for any other `TL_MODE` constant right?

Toflar:
Exactly. They're here for BC anyway and will be dropped in 5 or so :)

dtomasi:
In this case we would neither use `TL_MODE` as argument for `ContaoFrameworkBootEvent` or in is ParameterBag. I will going to inject the container to `SetConstantsListener` and set `TL_MODE` by container´s scope. Do you agree?

Toflar:
Yes, that's what I wanted to point out. You can extend the `ScopeAwareListener` there and reuse the methods it provides :)

aschempp:
Exactly, we wouldn't need any parameters in the event. The container is already available in your `SetConstantsListener` as you did not apply my patch ;-)

dtomasi:
@aschempp I cherry-picked it, but thanks to PHPStorms local history :)

dtomasi:
I have some open questions:

- Is validating the installation really the right place here? This is CTO4 i thing and does not really belong to the initialization process of the legacy-framework? 
- Should a add a test for the Chain of all Listeners or do you think this is not needed? 
- Someone a simple idea to fix the broken test in `AddToSearchIndexListenerTest`? I do not really want to rewrite the `TestCase` in this PR.

aschempp:
> Is validating the installation really the right place here? This is CTO4 i thing and does not really belong to the initialization process of the legacy-framework?

That should be a separate issue and not threaded in this PR.

dtomasi:
@aschempp Ok, I will first open a new issue for that...

dtomasi:
RFC ??  :smile:

Toflar:
Yes, you can change the title - excellent work! @discordier and @leofeyer: your turn :)

dtomasi:
@Toflar @aschempp Thanks a lot for your help.

aschempp:
I have added yet another pull request at https://github.com/ContaoBlackForest/core-bundle/pull/3


Btw, do we still need to fire `contao.framework.initialize` event? I think that's unnecessary because there is now an event chain?

dtomasi:
I´m firing `contao.framework.initialize` because developers can then also use EventDispatcher instead of the legacy-hook. But if there is a other solution, we can drop this.

aschempp:
> But if there is a other solution, we can drop this.

1: We want to replace all hooks with events, so there will be another solution anyway
2: Use a `contao.framework.boot` listener with the correct priority ;-)

dtomasi:
Yeah, thats right ... ;-) So I will drop that!!

aschempp:
I'm also not sure if we need the `RegisterListenersPass`? We're not using that functionality at all?

dtomasi:
I think we improved the last issues and optimize everything according to this. @leofeyer and @discordier do you have some thing additional to this PR? Otherwise I think we can set it to RTM-Status, don´t you?

src/EventListener/InitializeSystemListener.php
```diff
@@ -114,30 +51,14 @@ public function __construct(
...
-
-        $this->setConstants(
-            $this->getModeFromContainerScope(),
-            substr($route, strlen($basePath) + 1),
-            $basePath,
-            $request->attributes->get('_contao_referer_id')
-        );
-
-        $this->boot($request);
+        $this->eventDispatcher->dispatch('contao.framework.boot', new ContaoFrameworkBootEvent($event->getRequest()));
```
leofeyer: I'd prefer `contao.boot_framework`.

leofeyer:
Hm, no hard feelings, but this looks a little over-engineered to me. I have commented somewhere that I would prefer a single listener to initialize the Contao 3 framework and just separate the non-legacy stuff (e.g. the insecure installation check).

Now we are adding own events and a lot of stuff which people might think they are supposed to use. However, this is exactly what we **don't** want. We are aiming to remove the legacy framework, so it would be helpful to "force" people to use the Symfony stuff only.

Toflar:
@leofeyer: This is not over-engineered at all. It's hardly needed! It makes a lot of things easier, especially testing and we get rid of a lot of useless dependency injections. I'll definitely vote +1 to merge this.

leofeyer:
I have to vote -1.

We don't need a separate event system just to initialize the Contao 3 framework. Especially because most of the stuff is deprecated and will be removed in the future.

dtomasi:
And this makes it more easier to remove the stuff ...

src/EventListener/InitializeSystemListener.php
dtomasi: IMO the second `.` is used to define a domain. So this is the domain `framework`. This can also by used for domains like `page`, `article` or `file`. While I included `RegisterListenerPass` as a `CompilerPass` this could be a base to structure and tag contao`s events in future.

leofeyer: I am working on adding events instead of hooks (see #204) and I'm using `contao.event_name` all the time. We should try to be consistent here.

dtomasi: Ok, I agree to have a consistence here. I would change this according to #204. 

BTW: Did you had a look at `RegisterListenerPass`. This maybe helpful for #204 
https://github.com/symfony/EventDispatcher/blob/master/DependencyInjection/RegisterListenersPass.php

leofeyer: How would this be helpful for #204?

dtomasi: Ok, I now saw that this is just a translation between hooks and events ... not the plan to handle events in future, right? Then `RegisterListenerPass` does not matter ...

leofeyer: We want to handle events in the future. Every hook is to be replaced by an event and as long as we support hooks, there is a listener that subscribes to an event and triggers the corresponding hook.

dtomasi: Ok, I understand what you are doing there, but I don´t like it.

leofeyer: What do you not like? I'm open for suggestions.

dtomasi: But thats off-topic here. I would comment to the PR.

leofeyer: Then please do this. The PR is to be merged next week.

aschempp:
As just discussed with @dtomasi, the `RegisterListenerPass` is only necessary if we would (eventually) have a custom event dispatcher for the Contao "namespace". In my opinion, we should then add the pass, and not now for things maybe coming in the future. @dtomasi correct me if I'm wrong :-)

dtomasi:
@aschempp I agree in removing it for now. 

But some additional information about this should be provided: 

Main functionality of `RegisterListenerPass` is to add tagged services to the given `EventDispatcher`. Not in every case this has to be another as the standard `EventDispatcher`. It just provides a way to use your own tags like this: 

```php
        // Add the Pass
        $container->addCompilerPass(
            new RegisterListenersPass(
                'event_dispatcher', // The EventDispatcher to add the listeners and subscribers
                'contao.event_listener', // The tag of listener-services which should be added
                'contao.event_subscriber' // The tag of subscriber-services which should be added
            ),
            PassConfig::TYPE_BEFORE_REMOVING
        );

```

```yaml

## Now you can register with `contao.event_listener`
    contao.listener.my_listener_to_add:
        class: Contao\CoreBundle\EventListener\MyListenerToAdd
        tags:
            - { name: contao.event_listener, event: contao.some_event, method: calledMethod , priority: 10 }

## instead of `kernel.event_listener`
    contao.listener.my_listener_to_add:
        class: Contao\CoreBundle\EventListener\MyListenerToAdd
        tags:
            - { name: kernel.event_listener, event: kernel.some_event, method: calledMethod , priority: 10 }

```

So thats like @aschempp said its not necessary for now. It just provides nice tagging of listeners and subscribers.

In this case I will remove this for now.

aschempp:
> It just provides nice tagging of listeners and subscribers.

I don't get why it would be "nice"? There's no advantage or anything?

dtomasi:
As discussed with @Toflar and @discordier last week on mumbel, this should be simplified and also extended :)

Things that were improved:

- `\Config` and `\Input` are now initialized in one Listener. -> `ContaoStaticClassesListener`
- `SetConstantsListener` now has only one method to set all constants
- `ContaoFrameworkBootEvent` is now named as  `ContaoFrameworkInitializeEvent`
- All Listeners that are used in Initialization-Process are wrapped in a own, freezed EventDispatcher.
- Initialization-Process can now be triggered by calling the `Initializer`-Service

aschempp:
What's the reason to use an `ImmutableEventDispatcher`? I can still add any listener by tagging my own service?

dtomasi:
Yeah... thats true ... I will fix this by adding the services hard coded in Compilerpass and remove the tags.

dtomasi:
IMHO RTM :smirk:

src/ContaoCoreBundle.php
```diff
@@ -59,5 +61,10 @@ public function build(ContainerBuilder $container)
 
         $container->addCompilerPass(new AddResourcesPathsPass());
         $container->addCompilerPass(new SetApplicationPass());
+
+        $container->addCompilerPass(
+            new CreateInitializerEventDispatcherPass(),
+            PassConfig::TYPE_BEFORE_REMOVING
```
Toflar: Any reason why you use `TYPE_BEFORE_REMOVING` and if so, can you add a one-liner as a comment in the code? :)

src/EventListener/Framework/ValidateInstallationListener.php
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Validates the Installation
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ * @author Yanick Witschi <https://github.com/toflar>
+ * @author Leo Feyer <https://github.com/leofeyer>
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ */
+class ValidateInstallationListener
```
Toflar: Why is this still part of this PR? I thought you wanted to split it up in https://github.com/contao/core-bundle/pull/240 ?

src/Legacy/Framework/Initializer.php
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+
+namespace Contao\CoreBundle\Legacy\Framework;
```
Toflar: I don't like the naming. Especially we want to omit `Legacy` because it's not. It is the "old way" but it's not legacy because there is no alternative to it yet.

I think `Contao\CoreBundle\Framework` is correct because we can then put other "stand-alone" services there as well. And I think the class should be named `FrameworkInitalizer` so it's better readable. I talked to @aschempp about the naming already and this makes sense to us both.

dtomasi: Thats ok to me. Is this definitely?

Toflar:
I have added my remarks. Apart from those naming issues this PR makes all initializing of our "old things" way more beautiful, easier to test and is RTM to me. A huge :+1: from me on this one.

src/EventListener/Framework/ValidateInstallationListener.php
dtomasi: Its depending on each other ... This is the actual state of `InitializeSystemListener` which is splitted in this PR.

Toflar: Ah yeah, missed that. Sorry.

src/ContaoCoreBundle.php
dtomasi: not really ... can be removed ... :smile:

src/Legacy/Framework/Initializer.php
Toflar: To me and @aschempp, yes :) But we'll see at the call tomorrow, thx for the changes! :)

dtomasi:
Includes now IncompleteInstallationListener like discussed in #239.

Toflar:
So can you guys @aschempp, @leofeyer and @discordier  provide a list of things you don't like so we can move on?

discordier:
Aside from the already mentioned naming issues I also think this is RTM.
I however have not put it into production and can not say if everything still works but for this we have the unit tests, don't we? :D

dtomasi:
I tested this in a production. Beyond the unit tests I had to fix some things. See f204aee and 0e312ab

leofeyer:
> So can you guys @aschempp, @leofeyer and @discordier provide a list of things you don't like so we can move on?

First of all, I recall we have agreed upon having one service instead of a bunch of event listeners to boot the Contao framework. The service shall have a `boot()` method, which boots the framework if it has not yet been booted, and it can be used in controllers or commands that rely on the Contao framework.

This PR, however, follow a completely different approach, which I have already voted -1 on on April 28th. I still think it is over-engineered and it does way too much given the fact that all of it is legacy code and will be removed in the near future.

@dtomasi I know you have put a lot of time and effort into this PR and I am really sorry for voting -1 therefore. But the PR just goes into a completely wrong direction IMHO (see my comments above).

dtomasi:
@leofeyer Yes, I have spent a lot of time for this PR and I created it without a knowledge of your roadmap. Maybe that was the real problem. If this is really the completely wrong direction you should close this.

leofeyer:
> and I created it without a knowledge of your roadmap

Well, that's not entirely true. I have mentioned it already [here](https://github.com/contao/core-bundle/issues/239#issuecomment-96611846) 10 days ago and [here](https://github.com/contao/core-bundle/pull/234#issuecomment-96984423) 9 days ago. From there on you knew and it was your own decision to continue to work on the PR.

dtomasi:
Yeah, maybe. I had a glimmer of hope this could change if I have done. However, just close it.

leofeyer:
Thanks for your honesty. Do we really have to close the PR or can it be adjusted to provide the boot service? Or would it make more sense to create a separate PR?

dtomasi:
That´s your choice ... But I don´t think this is really sensible because of the commit-history. There are just 7 commits but 28c7e79 is squashed ...

aschempp:
I do agree with @leofeyer on the "over-engineering", because there is no advantage on using an event dispatcher to run hard-coded events.

However, I also think we should further continue on this PR. Keeping history is always good in my opinion to see what we did, what we did "wrong" and why we did things.

Toflar:
So to sum up:

* Beautiful but too over-engineered for booting something we want to get rid of anyway
* The service itself is correct
* The `boot()` or `initialize()` call of the service should not be in a general listener but instead in the the respective classes that need the old classes.

To get rid of the over-engineered listeners we could simply inject the service container into the `FrameworkInitializer` service and do everything hardcoded there. It's not going to change anyway and there's no need for proper Dependency Injection there.

dtomasi:
So would this be the way you talked about ?? If this is so, I will write the tests.

aschempp:
@dtomasi I don'ft understand your question :(

dtomasi:
Sorry @aschempp 

I moved all the functionality of the listeners into a singel `FrameworkInitializer`-Service, removed the Listeners and the CompilerPass.
Like @Toflar has summarized above I think this should be the way, this PR should look like. 
I´m now asking if you all agree, before I will write the tests, because I would not do this vainly?

dtomasi:
ping

src/ContaoCoreBundle.php
```diff
@@ -12,6 +12,7 @@
 
 use Contao\CoreBundle\DependencyInjection\Compiler\AddPackagesPass;
 use Contao\CoreBundle\DependencyInjection\Compiler\AddResourcesPathsPass;
+use Contao\CoreBundle\DependencyInjection\Compiler\CreateFrameworkInitializerEventDispatcherPass;
```
leofeyer: This looks like it could be removed?

src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -30,6 +30,7 @@ class ContaoCoreExtension extends ConfigurableExtension
         'adapter.yml',
         'cache.yml',
         'listener.yml',
+        'listener_framework.yml',
```
leofeyer: This also looks like it could be removed?

src/Resources/config/listener.yml
```diff
@@ -137,4 +132,3 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
             - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse }
-
```
leofeyer: We always have an empty line at the end of PHP files.

src/Resources/config/services.yml
```diff
@@ -48,3 +48,7 @@ services:
         calls:
             - ['setContainer', ['@service_container']]
 
+    contao.framework.initializer:
```
leofeyer: This should be `contao.framework_initializer`.

tests/Bundle/ContaoCoreBundleTest.php
```diff
@@ -81,6 +81,7 @@ public function testBuild()
                 'Contao\\CoreBundle\\DependencyInjection\\Compiler\\AddPackagesPass',
                 'Contao\\CoreBundle\\DependencyInjection\\Compiler\\AddResourcesPathsPass',
                 'Contao\\CoreBundle\\DependencyInjection\\Compiler\\SetApplicationPass',
+                'Contao\\CoreBundle\\DependencyInjection\\Compiler\\CreateFrameworkInitializerEventDispatcherPass'
```
leofeyer: This looks like it could be removed together with the `CreateFrameworkInitializerEventDispatcherPassTest` class?

tests/EventListener/Framework/ContaoFrameworkHookListenerTest.php
```diff
@@ -0,0 +1,48 @@
...
+use Contao\CoreBundle\Test\TestCase;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+
+/**
+ * Tests the ContaoFrameworkHookListener class.
+ *
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ * @author Andreas Schempp <https://github.com/aschempp>
+ */
+class ContaoFrameworkHookListenerTest extends TestCase
```
leofeyer: Do you still use this?

tests/EventListener/Framework/ContaoRequestTokenListenerTest.php
```diff
@@ -0,0 +1,182 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Security\Csrf\CsrfTokenManager;
+
+/**
+ * Tests the ContaoRequestTokenListener class.
+ *
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ */
+class ContaoRequestTokenListenerTest extends TestCase
```
leofeyer: Are you still using this class?

tests/EventListener/Framework/ContaoSessionListenerTest.php
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+namespace Contao\CoreBundle\Test\EventListener\Framework;
+
+use Contao\CoreBundle\Event\ContaoFrameworkInitializeEvent;
+use Contao\CoreBundle\EventListener\Framework\ContaoSessionListener;
+use Contao\CoreBundle\Test\TestCase;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpFoundation\Session\Storage\MockArraySessionStorage;
```
leofeyer: Are you still using this class?

tests/EventListener/Framework/ContaoStaticClassesListenerTest.php
```diff
@@ -0,0 +1,70 @@
...
+use Contao\CoreBundle\Adapter\ConfigAdapter;
+use Contao\CoreBundle\EventListener\Framework\ContaoStaticClassesListener;
+use Contao\CoreBundle\Test\TestCase;
+
+/**
+ * Tests the ContaoFrameworkHookListener class.
+ *
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ */
+class ContaoConfigListenerTest extends TestCase
```
leofeyer: Are you still using this class?

tests/EventListener/Framework/EnvironmentConfigurationListenerTest.php
```diff
@@ -0,0 +1,125 @@
...
+use Contao\CoreBundle\EventListener\Framework\EnvironmentConfigurationListener;
+use Contao\CoreBundle\Test\TestCase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Tests the SetConstantsListener class.
+ *
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ */
+class EnvironmentConfigurationListenerTest extends TestCase
```
leofeyer: Are you still using this class?

tests/EventListener/Framework/IncompleteInstallationListenerTest.php
```diff
@@ -0,0 +1,59 @@
...
+use Contao\CoreBundle\Test\TestCase;
+use PHPUnit_Framework_MockObject_MockObject;
+
+/**
+ * Tests the ValidateInstallationListener class.
+ *
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ * @author Andreas Schempp <https://github.com/aschempp>
+ */
+class IncompleteInstallationListenerTest extends TestCase
```
leofeyer: Are you still using this class?

tests/EventListener/Framework/SetConstantsListenerTest.php
```diff
@@ -0,0 +1,168 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Tests the SetConstantsListener class.
+ *
+ * @author Dominik Tomasi <https://github.com/dtomasi>
+ * @author Andreas Schempp <https://github.com/aschempp>
+ */
+class SetConstantsListenerTest extends TestCase
```
leofeyer: Are you still using this class?

dtomasi:
@leofeyer First of all, can you tell me if this is the way you did imagine? If it so, I will work on and write the tests. Should be possible to have this done till 16:00.

leofeyer:
You would do me a great favor if you could remove all the left-over stuff in the first step. Then I could tell more easily :)

dtomasi:
This should be done with the last 5 commits ...

leofeyer:
Ah, ok. There are still classes such as `CreateFrameworkInitializerEventDispatcherPass`, which I figured were not used anymore. But maybe I didn't understand correctly :)

The approach is much better now, except we want to entirely drop the `InitializeSystemListener` class. Each controller or command that required the Contao framework has to use the initialization service.

dtomasi:
I thought I got them all ... but you were right ... I missed the `CreateFrameworkInitializerEventDispatcherPassTest` ... now this should be clean :)

leofeyer:
+1

May I ask you to rebase the PR once? I have made some CS changes to the `InitializeSystemListener` class today.

And let's discuss the rest this afternoon. Has @Toflar asked you to join us?

dtomasi:
Of course ... I will do the rebase! 

> Has @Toflar asked you to join us?

Yes, he did ... I will be at Mos Eisley then ...

leofeyer:
Great :)

leofeyer:
How is the status of this PR? Can I merge it and start fixing the CS (and tests if necessary)?

src/ContaoFramework.php
```diff
@@ -81,118 +90,62 @@ class InitializeSystemListener extends AbstractScopeAwareListener
...
      * @param CsrfTokenManagerInterface $tokenManager  The token manager service
      * @param string                    $csrfTokenName The name of the token
      * @param ConfigAdapter             $config        The config adapter object
+     * @param int                       $errorLevel    The PHP error level
      */
     public function __construct(
+        ContainerInterface $container,
         RouterInterface $router,
         SessionInterface $session,
         $rootDir,
```
leofeyer: We don't need to inject `$rootDir`. We can get it from the container, can't we?

src/ContaoFramework.php
```diff
@@ -81,118 +90,62 @@ class InitializeSystemListener extends AbstractScopeAwareListener
...
         CsrfTokenManagerInterface $tokenManager,
         $csrfTokenName,
-        ConfigAdapter $config
+        ConfigAdapter $config,
+        $errorLevel
```
leofeyer: We don't need to inject `$errorLevel`. We can get it from the container, can't we?

src/ContaoFramework.php
```diff
@@ -81,118 +90,62 @@ class InitializeSystemListener extends AbstractScopeAwareListener
...
      * @param ConfigAdapter             $config        The config adapter object
+     * @param int                       $errorLevel    The PHP error level
      */
     public function __construct(
+        ContainerInterface $container,
         RouterInterface $router,
         SessionInterface $session,
         $rootDir,
         CsrfTokenManagerInterface $tokenManager,
         $csrfTokenName,
```
leofeyer: We don't need to inject `$csrfTokenName `. We can get it from the container, can't we?

src/ContaoFramework.php
```diff
@@ -81,118 +90,62 @@ class InitializeSystemListener extends AbstractScopeAwareListener
...
      * @param RouterInterface           $router        The router service
      * @param SessionInterface          $session       The session service
      * @param string                    $rootDir       The kernel root directory
      * @param CsrfTokenManagerInterface $tokenManager  The token manager service
      * @param string                    $csrfTokenName The name of the token
      * @param ConfigAdapter             $config        The config adapter object
+     * @param int                       $errorLevel    The PHP error level
      */
     public function __construct(
+        ContainerInterface $container,
```
leofeyer: We really should not inject the container and get the scope from the request object instead (`$request->attributes->get('_scope')`).

dtomasi:
@leofeyer In done with the test in about 1hour ... is this fine for you? I will also have a look at your line-note above.

leofeyer:
+1

dtomasi:
There is a test for `initialize`-method and the 'do not boot twice'-thing pending. But time is running out ... I think this should be enough at this moment. What do you think ?

leofeyer:
It is. Just add a `// FIXME` comment wherever something is missing.

dtomasi:
Ok... There are also some errors in the tests I detected ... I´m now fixing this and give you notice

src/EventListener/AddToSearchIndexListener.php
```diff
@@ -27,7 +41,7 @@ class AddToSearchIndexListener
      */
     public function onKernelTerminate(PostResponseEvent $event)
     {
-        if (!defined('TL_ROOT')) {
+        if ($this->framework->isInitialized()) {
```
dtomasi: Could this be right? Shouldn´t is be `if (!$this->framework->isInitialized()) {` ???

dtomasi:
@leofeyer I´m done.

leofeyer:
Great, thank you very much for your help. I'll merge the PR this afternoon.

src/EventListener/AddToSearchIndexListener.php
aschempp: You're probably right

leofeyer:
Merged in a70f46d1d0b82a4edcb8e45bef19b5ab26ac4582.

Issue #233, opened 2015-04-23
[RFC] Locale formats


aschempp:
# [RFC] Locale formats

For **several days** I now read on the topic of *Locales* in regards to our current handling and *the Symfony way*. I'm writing down here what I found so we're all up to date and to solve confusion once and for all. Loosely related to #190 and #171.

## Background

> In computing, a locale is a set of parameters that defines the user's language, country and any special variant preferences that the user wants to see in their user interface.

https://en.wikipedia.org/wiki/Locale

This is probably familiar to everyone of us. 

### Locale formats

These two formats are relevant for our case.

1. IETF [**Language Tag**][IETF], specified in [BCP 47], (aka RFC 5646, RFC 4646, RFC 3066, RFC 1766).  
 Used in:
 - HTTP Accept-Language header
 - XML & HTML documents *(e.g. &lt;html lang=""&gt;)*

2. **Locale ID**, according to the [International Components for Unicode (ICU)][ICU].  
 Used in:
 - [PHP Intl] extension
 - [Symfony Intl] component, a replacement layer for the PHP intl extension
 - Recommended format for the [Symfony Translation] component 
 - Returned by Symfony Request::getLanguages() (after parsing the *HTTP Accept-Language* header)

[Transifex] uses a *Locale ID* to represent regional languages (e.g. `zh_TW`) but a *Language Tag* to represent language scripts (e.g. `zh-Hant`) :-(

### Differences between *Language Tag* and *Locale ID*

As far as I understand, there is no major difference **for our use case**.

- A *Language Tag* uses a dash (`-`) as delimiter between *language*, *script* and *region*.
- A *Locale ID* uses an underscore (`_`) as delimiter between *language*, *script* and *country*.
- In a *Locale ID*, the third subtag is always a country (according to [ISO 3166]) whereas in a *Language Tag* it can also be an [UN M.49] region code.

A *Locale ID* also does allow to specify more details on locales like the currency, calendar or collation. However, they are (currently) not relevant for our case.

### Structure of *Language Tag* and *Locale ID*

Apart from the differences noted above, both *Language Tag* and *Locale ID* are very similar in their format:

1. The *language* subtag specified using a two- or three-letter lowercase code (using [ISO 639]-1 or [ISO 639]-2). 
2. An optional *script* subtag (specified in [ISO 15924])  
   (Examples: Latn = Latin, Cyrl = Cyrillic, Hans = Chinese Simplified, Hant = Chinese Traditional)
[...]

PR #237, opened 2015-04-26
[RTM] Locale handling

aschempp:
Because there will be no quick solution to #233, this pull request implements the most simple solution:

1. Use `kernel.request` event to convert our route locale (e.g. `de-CH`) to a locale ID (`de_CH`).
2. Set request and translator locale when page or user is loaded
3. Use new locale in `InitializeSystemListener`

#### TODO
 - [ ] Should we replace `$GLOBALS['TL_LANGUAGE']` everywhere (in legacy code)?

### DIFF
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -293,20 +293,14 @@ class InitializeSystemListener extends ScopeAwareListener
      */
     private function setDefaultLanguage(Request $request = null)
     {
-        if (!$this->session->has('TL_LANGUAGE')) {
-            $langs = $request ? $request->getLanguages() : [];
-            array_push($langs, 'en'); // see #6533
+        $language = 'en';
 
-            foreach ($langs as $lang) {
-                if (is_dir(__DIR__ . '/../../src/Resources/contao/languages/' . str_replace('-', '_', $lang))) {
-                    $this->session->set('TL_LANGUAGE', $lang);
-                    break;
-                }
-            }
+        if (null !== $request) {
+            $language = str_replace('_', '-', $request->getLocale());
         }
 
-        $GLOBALS['TL_LANGUAGE']  = $this->session->get('TL_LANGUAGE');
-        $_SESSION['TL_LANGUAGE'] = $this->session->get('TL_LANGUAGE'); // backwards compatibility
+        $GLOBALS['TL_LANGUAGE']  = $language;
+        $_SESSION['TL_LANGUAGE'] = $language; // backwards compatibility
     }
 
     /**
--- a/src/EventListener/InitializeSystemListener.php
+++ b/src/EventListener/InitializeSystemListener.php
@@ -299,7 +299,7 @@ class InitializeSystemListener extends ScopeAwareListener
             $language = str_replace('_', '-', $request->getLocale());
         }
 
-        $GLOBALS['TL_LANGUAGE']  = $language;
+        $GLOBALS['TL_LANGUAGE']  = $language; // backwards compatibility
         $_SESSION['TL_LANGUAGE'] = $language; // backwards compatibility
     }
 
--- a/src/EventListener/LocaleListener.php
+++ b/src/EventListener/LocaleListener.php
@@ -59,18 +59,15 @@ class LocaleListener extends ScopeAwareListener
         $request = $event->getRequest();
         $session = $request->getSession();
 
-        if ($locale = $request->attributes->get('_locale')) {
-            $locale = $this->formatLocaleID($locale);
-            $request->attributes->set('_locale', $locale);
-            $session->set('_locale', $locale);
-        } elseif ($locale = $session->get('_locale')) {
-            $request->attributes->set('_locale', $locale);
+        if ($request->attributes->has('_locale')) {
+            $locale = $this->formatLocaleID($request->attributes->get('_locale'));
+        } elseif (null !== $session && $session->has('_locale')) {
+            $locale = $session->get('_locale');
         } else {
             $locale = $this->getPreferredLocale($request);
-
-            $request->attributes->set('_locale', $locale);
-            $session->set('_locale', $locale);
         }
+
+        $this->saveLocale($request, $locale);
     }
 
     /**
@@ -127,5 +124,15 @@ class LocaleListener extends ScopeAwareListener
         }
 
         return $locale;
+    }
+
+
+    private function saveLocale(Request $request, $locale)
+    {
+        $request->attributes->set('_locale', $locale);
+
+        if ($request->hasSession()) {
+            $request->getSession()->set('_locale', $locale);
+        }
     }
 }
--- a/src/EventListener/LocaleListener.php
+++ b/src/EventListener/LocaleListener.php
@@ -60,7 +60,9 @@ class LocaleListener extends ScopeAwareListener
         $session = $request->getSession();
 
         if ($locale = $request->attributes->get('_locale')) {
-            $session->set('_locale', str_replace('-', '_', strtolower($locale)));
+            $locale = str_replace('-', '_', strtolower($locale));
+            $request->attributes->set('_locale', $locale);
+            $session->set('_locale', $locale);
         } elseif ($locale = $session->get('_locale')) {
             $request->attributes->set('_locale', $locale);
         } else {
--- a/src/EventListener/LocaleListener.php
+++ b/src/EventListener/LocaleListener.php
@@ -60,7 +60,7 @@ class LocaleListener extends ScopeAwareListener
         $session = $request->getSession();
 
         if ($locale = $request->attributes->get('_locale')) {
-            $locale = str_replace('-', '_', strtolower($locale));
+            $locale = $this->formatLocaleID($locale);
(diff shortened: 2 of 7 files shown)


leofeyer:
What do we need this for?

src/Resources/config/listener.yml
```diff
@@ -5,10 +5,15 @@
...
-# - 24: ToggleViewListener
-# - 22: OutputFromCacheListener
+# - 20: RefererIdListener
+# - 20: LocaleListener
+#
+# The priorites of the following listeners must be lower than the one of the
+# Symfony locale listener (defaults to 16).
+#
+# - 10: InitializeSystemListener
+# - 5:  ToggleViewListener
```
leofeyer: This is wrong. The `ToggleViewListener` needs to come before the `OutputFromCacheListener`.

src/Resources/contao/classes/BackendUser.php
```diff
@@ -357,7 +360,10 @@ protected function setUserFromDb()
 		}
 
 		$GLOBALS['TL_USERNAME'] = $this->username;
-		$GLOBALS['TL_LANGUAGE'] = str_replace('_', '-', $this->language);
+		$GLOBALS['TL_LANGUAGE'] = str_replace('_', '-', $this->language); // backwards compatibility
+
+		$kernel->getContainer()->get('request_stack')->getCurrentRequest()->setLocale($this->language);
+		$kernel->getContainer()->get('translator')->setLocale($this->language);
```
leofeyer: Are we using the translator service at all?

aschempp: Not yet, but we definitely want to, same as with Twig.

src/Resources/config/listener.yml
aschempp: Changed it

aschempp:
`$request->getLocale()` is the Symfony replacement for `$GLOBALS['TL_LANGUAGE']`. Additionally, the current `_locale` attribute in routes would be useless without this.

Regarding the listener priorities, I have changed the OutputFromCache and ToggleView listeners to negative priorities. I think they should come after the default `kernel.request` listeners, because they actually terminate the event.
I'm not sure if that makes sense. If the OutputFromCache would not use legacy code, it should probably render as soon as possible…

src/Resources/contao/classes/BackendUser.php
leofeyer: No, because we **are** using Twig but not the translator.

aschempp: But only since we moved the exception templates. But Twig was pre-configured before, and it makes sense to make the translator work at all if we provide a full-stack app.

leofeyer:
We should not change any listener priorities not related to this PR. There should be a separate PR instead.

src/Resources/contao/classes/BackendUser.php
leofeyer: I agree, but first we'll have to sort out the language code and locales issue.

aschempp: This PR already does for the relevant stuff…

aschempp:
I would agree, but we have to as the framework cannot boot without the request language…

dtomasi:
This would also touch #192. Therefor this should be merged before or I should cherry pick the changes.

src/EventListener/InitializeSystemListener.php
```diff
@@ -293,20 +293,14 @@ class_alias("Contao\\$class", $class);
...
-            }
+        $language = 'en';
+
+        if (null !== $request) {
+            $language = str_replace('_', '-', $request->getLocale());
         }
 
-        $GLOBALS['TL_LANGUAGE']  = $this->session->get('TL_LANGUAGE');
-        $_SESSION['TL_LANGUAGE'] = $this->session->get('TL_LANGUAGE'); // backwards compatibility
+        $GLOBALS['TL_LANGUAGE']  = $language;
```
Toflar: This should have a `// backwards compatibility` as well now right? :)

aschempp: Yeah, a minor detail ^^

Toflar: Of course, just add it :)

aschempp: done

src/EventListener/InitializeSystemListener.php
```diff
@@ -293,20 +293,14 @@ class_alias("Contao\\$class", $class);
...
-            foreach ($langs as $lang) {
-                if (is_dir(__DIR__ . '/../../src/Resources/contao/languages/' . str_replace('-', '_', $lang))) {
-                    $this->session->set('TL_LANGUAGE', $lang);
-                    break;
-                }
-            }
+        $language = 'en';
+
+        if (null !== $request) {
+            $language = str_replace('_', '-', $request->getLocale());
```
leofeyer: You have skipped the check if the language is installed.

src/EventListener/LocaleListener.php
```diff
@@ -0,0 +1,99 @@
...
+     * @return string The preferred locale ID
+     */
+    private function getPreferredLocale(Request $request)
+    {
+        $languages = array_values(
+            array_map(
+                function(SplFileInfo $file) {
+                    return $file->getFilename();
+                },
+                iterator_to_array($this->resourceFinder->findIn('languages')->depth(0)->directories())
```
leofeyer: As discussed on Mumble on April 29th, we should also check `app/Resources` here.

Toflar:
You have to check in `app/Resources` for language bundles as well.

src/Resources/config/listener.yml
```diff
@@ -5,10 +5,15 @@
 # Symfony route listener (defaults to 32).
 #
 # - 30: ContainerScopeListener
-# - 28: RefererIdListener
-# - 26: InitializeSystemListener
-# - 24: ToggleViewListener
-# - 22: OutputFromCacheListener
+# - 20: RefererIdListener
```
leofeyer: As discussed on Mumble on April 29th, the order should be adjusted as follows:

```
[Symfony route on 32]
- 30: ContainerScopeListener
- 20: RefererIdListener
- 20: LocaleListener
[Symfony locale on 16]
- 14: InitiaizeSystemListener
- 0: ToggleViewListener
- 0: OutputFromCacheListener
```

aschempp:
I have updated the PR to what we discussed today on Mumble. However, I just noticed we're missing unit tests, and I would not merge it without that. Will add that later today or tomorrow.

src/EventListener/LocaleListener.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        if (!$this->isContaoScope()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+        $session = $request->getSession();
+
+        if ($locale = $request->attributes->get('_locale')) {
+            $session->set('_locale', str_replace('-', '_', strtolower($locale)));
```
leofeyer: Shouldn't you set the locale in the request, too?

src/EventListener/LocaleListener.php
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @return string The preferred locale ID
+     */
+    private function getPreferredLocale(Request $request)
+    {
+        $finder = Finder::create()
+            ->directories()
+            ->depth(0)
+            ->in([
+                __DIR__ . '../Resources/contao/languages',
```
leofeyer: This cannot work. It should be `__DIR__ . '/../...`.

src/EventListener/LocaleListener.php
aschempp: No, because the Symfony listener will do that

aschempp: Ah, confusion here. You mean in the request attributes, and you're right. Fixed in dade40d9eec3a6bef27458af26f2e879cb11e8cb

aschempp:
I have added unit tests. However, there is one incomplete where we should validate the installed languages. Any suggestions how to do that? The bundle path is obviously not injected?

Toflar:
Like that:

```php
/**
 * Constructor.
 *
 * @param string  $defaultLocale The default locale
 * @param string  $rootDir       The kernel root directory
 * @param Finder  $dirsFinder    An optional Finder instance containing the language directories
 */
public function __construct($defaultLocale, $rootDir, $dirsFinder = null)
{
    $this->defaultLocale = $defaultLocale;
    $this->rootDir       = $rootDir;
    $this->dirsFinder    = $dirsFinder;

    if (null === $this->dirsFinder) {
        $this->dirsFinder = $this->getDefaultFinder();
    }
}

aschempp:
I dislike to add functionality only for unit tests. What do you think about injecting the `kernel.bundles` and use the path for `ContaoCoreBundle` from there?

Toflar:
You don't just inject it for unit tests. You inject it for reusability in other contexts and providing a fallback way is a valid option :) But if your way works I'm fine with that as well :)

aschempp:
I have fixed the listener and tests using a factory to generate the available locales.

leofeyer:
Merged in 125608ad5247de40a9f17dd87cecb116096a65dc.

Issue #239, opened 2015-04-27
Separate validating CTO4-Installation from initialize process of legacy framework


dtomasi:
Validating the installation of CTO4 does not belong to the initialize process of legacy framework. In my opinion this should be separated.

leofeyer:
+1

dtomasi:
@leofeyer Would you like to have a PR for this??

aschempp:
It depends… we're currently verifying if the `localconfig.php` etc. is set, which is only necessary for the legacy framework…

dtomasi:
That´s right. What´s the plan to provide Config/Parameters for BC in future? I´m asking because I thought about implementing an Interface to decouple this.

aschempp:
Well the config should be in the container. We already did that for a few properties (e.g. `prefix_locale`), but it only makes sense for stuff we migrate to the new framework (e.g. it does not make sense for `installPassword`.

For most things, the Container will verify there is a value, so we might actually not need a generic listener…

dtomasi:
I agree. Then it should be kept in mind for the future. :)

leofeyer:
As I have said before, I would like the "your installation is insecure" check to be an event listener. The `/web` subfolder is not related to the Contao 3 framework.

aschempp:
Valid point.

dtomasi:
What´s about split the up in `InsecureInstallationListener` and `IncompleteInstallationListener`?

`InsecureInstallationListener` should then be placed in `Contao\CoreBundle\EventListener` and `IncompleteInstallationListener` in `Contao\CoreBundle\EventListener\Framework`, where the legacy stuff is placed.

aschempp:
+1

leofeyer:
IMHO, we don't need separate event listeners for the Contao 3 framework; one `InitializeSystemListener` is enough. But +1 for the `InsecureInstallationListener`.

dtomasi:
But in #234 we need the `IncompleteInstallationListener. This PR splits up the whole `InitializeSystemListener`.

leofeyer:
I'm closing this ticket in favor of #234.

PR #240, opened 2015-04-27
[RFC] Split Validate-Installation-Check

dtomasi:
See: #239 

NOTE: This PR should be rebased if #234 is merged first.

### DIFF
--- a/src/EventListener/InsecureInstallationListener.php
+++ b/src/EventListener/InsecureInstallationListener.php
@@ -25,6 +25,11 @@ use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 class InsecureInstallationListener
 {
     /**
+     * @var array
+     */
+    protected $secureClientIps = array('127.0.0.1', 'fe80::1', '::1');
+
+    /**
      * Validates the installation.
      *
      * @param GetResponseEvent $event
@@ -34,7 +39,6 @@ class InsecureInstallationListener
      */
     public function onKernelRequest(GetResponseEvent $event)
     {
-
         $request = $event->getRequest();
 
         if (null === $request || 'contao_backend_install' === $request->attributes->get('_route')) {
@@ -42,11 +46,7 @@ class InsecureInstallationListener
         }
 
         // Show the "insecure document root" message
-        if (!in_array($request->getClientIp(), [
-                '127.0.0.1',
-                'fe80::1',
-                '::1'
-            ]) && '/web' === substr($request->getBasePath(), -4)
+        if (!in_array($request->getClientIp(), $this->secureClientIps) && '/web' === substr($request->getBasePath(), -4)
         ) {
             throw new InsecureInstallationException(
                 'Your installation is not secure. Please set the document root to the /web subfolder.'
--- a/src/EventListener/InsecureInstallationListener.php
+++ b/src/EventListener/InsecureInstallationListener.php
@@ -1,0 +1,56 @@
+<?php
+
+/**
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\Exception\IncompleteInstallationException;
+use Contao\CoreBundle\Exception\InsecureInstallationException;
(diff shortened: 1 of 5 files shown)


src/EventListener/InsecureInstallationListener.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $request = $event->getRequest();
+
+        if (null === $request || 'contao_backend_install' === $request->attributes->get('_route')) {
+            return;
+        }
+
+        // Show the "insecure document root" message
+        if (!in_array($request->getClientIp(), [
+                '127.0.0.1',
```
leofeyer: I'd prefer to make this a protected class property. Then we would not have to add the line breaks, either.

src/Resources/config/listener.yml
```diff
@@ -6,6 +6,7 @@
 #
 # - 30: ContainerScopeListener
 # - 28: RefererIdListener
+# - 27: InsecureInstallationListener
```
leofeyer: As long as the listener priorities are not reworked, we should keep one number in between each listener priority. This means it should be:

```
- 30: ContainerScopeListener
- 28: RefererIdListener
- 26: InsecureInstallationListener
- 24: InitializeSystemListener
- 22: ToggleViewListener
- 20: OutputFromCacheListener
```

leofeyer:
Looks good to me.

dtomasi:
@leofeyer looks bad to travis ... :)

leofeyer:
Sure, because now the `InsecureInstallationException` is no longer thrown in the `InitializeSystemListener` class :)

src/Resources/config/listener.yml
dtomasi: I do not really understand this. Why should there be a number between if it is change in every PR that needs to. I think it would be better to do something like that: 

```
- 300: ContainerScopeListener
- 280: RefererIdListener
- 260: InsecureInstallationListener
- 240: InitializeSystemListener
- 220: ToggleViewListener
- 200: OutputFromCacheListener
```

In this case you could rework the priority as a refactoring and not in PR itself ... Won´t you think this is better? In most reason changing the priority of all other listeners is not really in scope of a PR.

leofeyer: > The priorities of the following listeners must be lower than the one of the Symfony route listener (defaults to 32).

And we want to give users and developers a chance to register own listeners in between one of our listeners, therefore we keep one free number in between.

dtomasi: Ok, this makes sense ... didn´t had symfony´s priority in mind ...

src/EventListener/InsecureInstallationListener.php
```diff
@@ -0,0 +1,56 @@
...
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\Exception\IncompleteInstallationException;
+use Contao\CoreBundle\Exception\InsecureInstallationException;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+
+/**
+ * Validates the Installation
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
```
aschempp: It does not make much sense to me to add all authors of the old listener. I'd say you should only add yourself :)

aschempp:
We should wait for #234 to be merged. But the test class is currently bad, e.g. the name is incorrect and itt does not test all cases (see https://github.com/aschempp/contao-core-bundle/commit/8cbd0e809a316f4ebda2eb1d6d8e7ce374ee3d2e)


Regarding the priorities, I think we should stop trying to define a complete priority chain for everything. The `InsecureInstallationListener` is only depending on the route, it could as well have a priority of 30 as it does not need the `ContainerScopeListener`.

dtomasi:
Yes, I agree. This should be refactored after #234 is merged.

Toflar:
So we discussed this on Mumble today and there are some things we don't like about it yet :) Can we all Mumble together maybe tomorrow? :)

dtomasi:
Sure! What time?

Toflar:
Oh, I accidentally commented the wrong PR. But this is RTM for me imho :)

dtomasi:
+1

leofeyer:
Merged in 1aed66918cc0c40f5f023e0a82ba3c55edda6095.

PR #241, opened 2015-04-28
[WIP] Using adapter classes instead of untestable legacy code

Toflar:
This is a hell lot of work but the longer I'm using it the more convinced I am that this is the right way. It allows us to fully test our new code without relying on legacy code. I've introduced `AdapterInterface`'s for every single adapter. This will give us the ability to provide intermediate layers if we e.g. want to remove a concrete adapter class.

This is NOT RFC. I just add it as PR so you know I'm working on it and I have a place where I can ask questions related to it.


aschempp:
Why would we need adapters for everything like backend controllers? I thought we only need them for classes with static methods?

Toflar:
I explicitly wrote this is not RFC. Is it that hard? I will explain everything tomorrow.

discordier:
Just a remark about the naming.
Having the same interface name  in multiple namespaces might lead to problems.
Therefore it would be better to name them like the concrete implementation.
`\Contao\Backend\AdapterInterface` => `\Contao\Backend\BackendAdapterInterface` or `\Contao\BackendAdapterInterface` if you do not like the doubling of the Backend part.

I still suffer in MetaModels from having a "Simple" interface in various sub namespaces forcing me to alias them around in the use section and therefore know that there WILL be problems. :)

Toflar:
> Having the same interface name in multiple namespaces might lead to problems.

That's not the case here? :)

leofeyer:
As discussed on Mumble on May 6th, @Toflar will create a separate `ContaoAdapterBundle` with the adapters and the necessary services, which we can then use.

Toflar:
I was able to get a first usable version of the adapter bundle done. See https://github.com/terminal42/contao-adapter-bundle. It currently contains already 45 adapters and they are all automatically registered to the service container which itself is 100% unit tested. So this bundle should really be useable from now on! The real problem is not that bundle or generating the adapters, the problem is using them everywhere we use legacy classes now.
I managed to fix a hell lot of tests but I still have 56 of them failing when running `phpunit --testsuite latest`. Also I had to skip a few tests:

a) I skipped the whole `InitializeSystemListener` for now because that's about to be changed anyway because we need a service.
b) The `ContaoCacheWarmerTest` is actually not testing if `ContaoCacheWarmer` is working fine but rather if the `DcaExtractor` generates the correct extract which means it's actually not testing the new code but the legacy code and should thus be moved to the correct test suite.

Toflar:
I could well need some help here. I did a regex search for `use Contao\\(?!CoreBundle)` on the `src` directory to see where old Contao classes are used and it seems that I'm not far from it!
Those are the remaining usages:

screen shot 2015-05-08 at 23 30 28

For almost all of it the adapter classes should already be available! Anybody willing to lend me a helping hand?

Toflar:
Down to 51 errors now.

aschempp:
Great work so far!

Two things:
 1. I would use `contao.adapter.*` as prefix for the container. There is no need nor is it "best practice" (afaik) to prefix everything with a namespace in the container.
 2. The adapters **must only** contain methods that are static in the original class. It is wrong to have `ModelAdapter->row()`. Was my explanation not clear enough? ;-)

Toflar:
1. I think vendor prefixing is better, don't want to change that actually.
2. You're explanation was clear but it's ugly :-) What if I need a static and a regular method of a class? Do you want to inject two dependencies? This is a bad approach :-)

leofeyer:
1. Shouldn't the vendor prefix be `terminal42.contao_adapter.`?
2. An adapter needs to implement all public methods of the adapted class, otherwise it is not usable.

aschempp:
Sorry, but you're simply wrong. Examples:

**Old Code**
```php
$page = \PageModel::findByAlias('…');
$data = $page->row();
```

**New Code**
```php
$adapter = new PageModelAdapter();
$page    = $adapter->findByAlias('…');
$data    = $page->row();
```

The call to `row()` to get the data is **NOT** on the adapter, so it's pointless to add that method to the adapter. Example for unit tests:

```php
$adapter = $this->getMock('PageModelAdapter');
$model   = $this->getMock('PageModel');

$adapter
    ->expects($this->any())
    ->method('findByAlias')
    ->willReturn($model)
;

$model
    ->expected($this->any())
    ->method('row')
    ->willReturn([ … ])
;
```

leofeyer:
I think you've got it wrong:

```php
$adapter = new PageModelAdapter();
$page    = $adapter->findByAlias('…'); // this will return the PageModelAdapter object!
$data    = $page->row(); // the call is executed on the PageModelAdapter object
```

aschempp:
You're wrong, look at https://github.com/terminal42/contao-adapter-bundle/blob/develop/src/Adapter/PageModelAdapter.php#L107

Changing it would also not be API compatible because the adapter does not extend the old class.

leofeyer:
You may be right. Let's discuss this on Wednesday.

dmolineus:
I don't know what you've discussed and what's the strategy where Contao is heading to. I'm kind of unhappy with the adapter solution as proposed right now.

As far as I see and described [here](https://github.com/terminal42/contao-adapter-bundle) the main goal is making *legacy* code testable. Are the adapters gonna being used in the core bundle or is it just for 3rd party extensions?

If it's planned to integrate it in the official Contao core, wouldn't it be better to redefine the public API which Contao provides instead of adding dependencies of the adapter classes? The redefined API would be defined as interfaces, provided as services without static method calls. 

As long as the *legacy api* still exists, adapters are the right pattern to implement public interface. There would be a `ConfigInterface` and the `ConfigAdapter` implementing the interface. In a future release the adapter could be droppped when dropping legacy code. This approach would also allow to introduce new services by restructuring some methods.

As I mentioned above I don't know what's the roadmap you're heading in the next releases. So maybe my proposal does not fit anyway. Feel free to include it in your ongoing discussions.

Toflar:
Closed as discussed on Mumble in favor of https://github.com/contao/core-bundle/pull/346

PR #245, opened 2015-05-04
Prevent recursion when creating symlinks

aschempp:
Otherwise a .htaccess file in a subfolder of an already published folder would generate an exception. This is for example the case with Isotope eCommerce (see https://github.com/isotope/core/tree/master/system/modules/isotope)


leofeyer:
How do I reproduce this? I understand the problem but I don't like the solution :)

Do you really mean `.htaccess`? Or do you mean `.public`?

aschempp:
In my case it happend due to nested .htacess, but I'm pretty sure nested .public would have the same result. The Symlink command quit with an exception telling the parent folder it wanted to add toncreate the symlink already existed.

leofeyer:
Fixed in 67c2ba250f92e637f20abf0379aedc002d328051.

PR #247, opened 2015-05-08
[RTM] Unify customTpl handling (see #235)

leofeyer:
This is the PR regarding #235.

 - [x] `ContentHyperlink`: the template is set dynamically depending on the `useImage` field.
 - [x] `ModuleEventMenu`: the template is set dynamically depending on the `cal_format` field.
 - [x] `ModuleLogin`: the template is set dynamically depending on the `cols` field and on the login status.
 - [x] `ModuleSearch`: the template is set dynamically depending on the `searchType` field.
 - [x] `ModuleNewsMenu`: the template is set dynamically depending on the `news_format` field.

PR #249, opened 2015-05-11
[RFC] Consolidate all front end forms

leofeyer:
This is the PR for https://github.com/contao/core/issues/7292 and https://github.com/contao/core-bundle/pull/247. The goal is to consolidate all front end forms to use the same HTML markup and to remove the "tableless" option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Contao change log
 Version 4.0.0-RC1 (2014-05-XX)
 ------------------------------
 
+### Changed
+Consolidated the markup of all front end forms (see #249).
+
 ### Fixed
 Consolidate the templates and module keys (see #247).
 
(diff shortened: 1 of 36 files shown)


Toflar:
Very cool! I just don't see any `form.css` yet :-) But I'm :+1: on this!

leofeyer:
I have changed the status to RFC.

aschempp:
+1 in general. But why are we not using real widgets e.g. In form_login?

leofeyer:
Using real widgets would be the next step. Right now, I don't have the time, so I only changed the backwards incompatible stuff. We can always replace the current solution with real widgets now that the markup is the same.

leofeyer:
Merged in d10c826ae5dd7a4ec984dbe65325cd28ef50abc4.

frontendschlampe:
for first look it feels great - just the submit container in each widget are still in use ;-)

leofeyer:
?

leofeyer:
As quickly discussed on Mumble on May 12th, we also want to skip the "add submit button" option.

leofeyer:
Removed in 5b9d0918341862fb5283719883f29dc2ce809339.

frontendschlampe:
:+1:

PR #250, opened 2015-05-11
[RFC] Merge the CSS classes and remove the "space" setting

leofeyer:
This is a PR for https://github.com/contao/core/issues/6011 (and probably a dozen more). It merges the CSS classes if an element is included in another element and also removes the "space before/after" field.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,6 +4,20 @@ API changes
 Version 3.* to 4.0
 ------------------
 
+### Space before/after
+
+The field "space before/after" has been removed. Use a CSS class instead and
+define the spacing in your style sheet.
+
+
+### CSS classes of incuded elements
+
+If an element is included in another element, the CSS classes are now merged
+instead of overwritten, e.g. if content element A has the CSS class `elemA` and
+there is an alias content element with the CSS class `elemB` referring to
+element A, both CSS classes will be applied (`class="elemA elemB"`).
+
+
 ### Form field "headline"
 
 The form field "headline" has been removed in favor of the "explanation" field.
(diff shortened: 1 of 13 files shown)


Toflar:
I think you should call a separate `getCssClassesAndId()` (or so) which can be overwritten in the child classes.

leofeyer:
You can overwrite the class properties in the `generate()` method of the child class.

Toflar:
I know. It's just a bit tedious because everybody has to do the `trim()` stuff and I thought now that we're changing it, why not? :)

leofeyer:
If you override the parent method, don't you have to apply `trim()` as well?

Toflar:
If you do

```php
/**
 * @return string
 */
public function getCssId();

/**
 * @return array
 */
public function getCssClasses();
```

not :-) Just an idea :)

leofeyer:
I don't understand why a method saves us from using `trim()`. If I wanted to add a CSS class, here's what I'd do:

```php
public function getCssClasses()
{
    return trim(parent::getCssClasses() . ' custom-class');
}
```

Since I don't know if there is a parent CSS class, I **always** have to use `trim()`, don't I?

UPGRADE.md
```diff
@@ -4,6 +4,20 @@ API changes
 Version 3.* to 4.0
 ------------------
 
+### Space before/after
+
+The field "space before/after" has been removed. Use a CSS class instead and
+define the spacing in your style sheet.
+
+
+### CSS classes of incuded elements
```
xchs: Typo: "incuded" -> "included"

leofeyer: Thanks @xchs, fixed in 26128ea71e8337493dec02dc45ac811638c2926b.

leofeyer:
Merged in d2b1fa8663099c0060678c53d6ae8718175a566d.

PR #254, opened 2015-05-13
[RTM] Use the Symfony session instead of $this->Session

leofeyer:
This PR replaces `$this->Session` with the Symfony session. It also attempts to store non-state data in the default session bag, which will not be persisted in the database.

### DIFF
--- a/src/Resources/contao/widgets/PageSelector.php
+++ b/src/Resources/contao/widgets/PageSelector.php
@@ -149,17 +149,17 @@ class PageSelector extends \Widget
 		}
 		else
 		{
-            /** @var AttributeBagInterface $objBag */
-			$objBag = $objSession->getBag('contao_backend');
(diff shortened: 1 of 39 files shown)


src/Resources/contao/widgets/PageSelector.php
```diff
@@ -139,14 +149,17 @@ public function generate()
 		}
 		else
 		{
-			$strNode = $this->Session->get('tl_page_picker');
+            /** @var AttributeBagInterface $objSessionBag */
```
Toflar: CS

leofeyer:
I have changed the status to RFC.

Toflar:
:+1:

leofeyer:
I have moved the referer logic from the `UserSessionListener` into a `StoreRefererListener` class. Please review again :)

leofeyer:
I have changed the status to RTM.

PR #255, opened 2015-05-13
Inject the framework boot service for the commands that need it

Toflar:


dtomasi:
The tests needs also some refactoring, because of the new argument in constructor.

leofeyer:
Implemented in 176f273f35f0ae9e4255b025db2a3695fcf7e022.

Issue #257, opened 2015-05-14
Re-add System::setContainer() and System::getContainer()


leofeyer:
Once #234 and #254 have been merged, I would like to re-add the `System::setContainer()` and `System::getContainer()` methods. We are currently doing

```php
/** @var KernelInterface $kernel */
global $kernel;

$kernel->getContainer()->get('...');
```

over 300 times. The new framework initializer could set the container when initializing the framework.

What do you think?

leofeyer:
I'm closing this ticket in favor of #259.

PR #259, opened 2015-05-14
[RFC] Add System::getContainer() and System::setContainer()

leofeyer:
This is the PR for #257. Given the huge improvement of the unit tests (no more `globals $kernel` required) and the vast amount of saved lines, I assume you are fine with me merging it :)

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -49,34 +49,26 @@ have been deprecated and will be removed in Contao 5.0.
 Use the `kernel.root_dir` instead of `TL_ROOT`:
 
 ```php
-global $kernel;
-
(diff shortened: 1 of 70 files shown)


aschempp:
Well we have discussed this many times and said not to do it.

I do, however, see the nice side of this. It simplified the code. But what about we would add `System::getService()` and `System::getParameter()` instead? That would make it even better if we really want to add "new features" to the legacy framework.

leofeyer:
Then you would not have autocompletion in PhpStorm anymore.

aschempp:
> Then you would not have autocompletion in PhpStorm anymore.

That's a valid argument.

Issue #151, opened 2015-03-23
Kick all deprecated things for 4.0


importer:
> <a href="https://github.com/Toflar"><img src="https://avatars.githubusercontent.com/u/481937?v=3" align="left" width="42" height="42" hspace="10"></img></a> [Issue](https://github.com/contao/contao/issues/73) by @Toflar
Thursday Mar 19, 2015 at 14:03 GMT

We have tons of `@deprecated` statements that have been added in version 2 or 3 already, those methods/classes should be dropped in 4.0 :)

importer:
> <a href="https://github.com/leofeyer"><img src="https://avatars.githubusercontent.com/u/1192057?v=3" align="left" width="42" height="42" hspace="10"></img></a> Comment by @leofeyer
Thursday Mar 19, 2015 at 15:22 GMT

Fine with me, I guess. Would have to take a detailed look though.

importer:
> <a href="https://github.com/Toflar"><img src="https://avatars.githubusercontent.com/u/481937?v=3" align="left" width="42" height="42" hspace="10"></img></a> Comment by @Toflar
Thursday Mar 19, 2015 at 15:24 GMT

Would be pretty helpful because then all `@deprecated` hints are related to "from version 3 to version 4".

importer:
> <a href="https://github.com/discordier"><img src="https://avatars.githubusercontent.com/u/940331?v=3" align="left" width="42" height="42" hspace="10"></img></a> Comment by @discordier
Sunday Mar 22, 2015 at 02:04 GMT

I'd love to see a lot to disappear.

leofeyer:
I have searched for `backwards compatibility` and got about 160 results. Not all of them are things to be removed though. I therefore suggest to do the following:

 - Review each result to find the stuff to be remove.
 - Replace the `backwards compatibility` comment with `deprecated, to be removed in Contao 5.0`.
 - Then in Contao 5, we can eventually remove the stuff.

What do you think?

aschempp:
Sounds reasonable. Should we add this for our discussion to go through the stuff?

leofeyer:
If there are no objections, we don't need to discuss it IMHO.

discordier:
I think all BC layers introduced in C3 to be compatible with C2 shall be removed in C4, C2 is simply outdated.
Everything that got introduced in C3.X shall be marked deprecated to be removed in C5.

[...]

PR #260, opened 2015-05-15
[RTM] Add deprecation notices (see #151)

leofeyer:
This is the PR for #151. It adds deprecation notices to everything tagged with "backwards compatibility". It also remove the tag where it does not longer seem to be appropriate and adds `TODO` tags where I'm not sure why things are as they are :)

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -1,6 +1,69 @@
 Deprecated features
 ===================
 
+### "channel" token
+
+Using the simple token "channel" in newsletter subscription mails has been
+deprecated in Contao 4.0 and will no longer work in Contao 5.0. Use the
+"channels" token instead.
+
+
+### $this->arrCache
+
+Using `$this->arrCache`, which is defined in the `System` class, has been
+deprecated in Contao 4.0 and will no longer work in Contao 5.0. If you are
+using it in your class, make sure to define it as property.
+
+
+### $this->items in pagination templates
+
+Using `$this->items` in pagination templates has been deprecated in Contao 4.0
+and will no longer work in Contao 5.0. Use `$this->pages` instead.
+
+
+### TL_SCRIPT_URL and TL_PLUGINS_URL
+
+The constants `TL_SCRIPT_URL` and `TL_PLUGINS_URL` have been deprecated in
+Contao 4.0 and will be removed in Contao 5.0. Use `TL_ASSETS_URL` instead.
+
+
+### UnresolvableDependenciesException
+
+The `UnresolvableDependenciesException` class has been deprecated in Contao 4.0
+and will be removed in Contao 5.0.
+
+
+### $this->language in TinyMCE config files
+
+Using `$this->language` in TinyMCE configuration files has been deprecated in
+Contao 4.0 and will no longer work in Contao 5.0. Use the static method
+`Backend::getTinyMceLanguage()` instead.
+
+
+### $GLOBALS['TL_LANGUAGE'] and $_SESSION['TL_LANGUAGE']
+
+Using the globals `$GLOBALS['TL_LANGUAGE']` and `$_SESSION['TL_LANGUAGE']` has
+been deprecated in Contao 4.0 and will no longer work in Contao 5.0. Use the
+locale from the request object instead:
+
+```php
+$locale = System::getContainer()->get('request')->getLocale();
+```
+
+
+### Request.Mixed
+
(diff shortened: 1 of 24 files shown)


leofeyer:
Just in case you wonder why I have removed the "backwards compatibility" comments at the empty ptable occurrences: there is already an entry in the `DEPRECATED.md` file about that.

aschempp:
We should only deprecate things there are alternatives for, and document them. That's not the case for everything, e.g. `TL_SCRIPT_URL`.

leofeyer:
The constants do have documented alterantives: https://github.com/contao/core-bundle/blob/develop/DEPRECATED.md#constants

aschempp:
Not the one I mentioned…

leofeyer:
You are right. But there is an alternative: `TL_ASSETS_URL`.

So what you actually wanted to say: Leo, please make sure to document every deprecation notice in the `DEPRECATED.md` file and don't forget to describe the alternatives, right? :)

aschempp:
Ah my mistake, I did not realize `TL_SCRIPT_URL` was the old constant from Contao 2. I was referring to `TL_ASSETS_URL` where there is no alternative (yet).

leofeyer:
I have documented all deprecations which do not trigger a notice in the `DEPRECATED.md` file. The PR is now RTM to me.

DEPRECATED.md
```diff
@@ -1,6 +1,69 @@
...
+
+
+### $GLOBALS['TL_LANGUAGE'] and $_SESSION['TL_LANGUAGE']
+
+Using the globals `$GLOBALS['TL_LANGUAGE']` and `$_SESSION['TL_LANGUAGE']` has
+been deprecated in Contao 4.0 and will no longer work in Contao 5.0. Use the
+locale from the request object instead:
+
+```php
+$locale = System::getContainer()->get('request')->getLocale();
```
Toflar: You should always retrieve the `request_stack`, shouldn't you?

DEPRECATED.md
```diff
@@ -1,6 +1,69 @@
...
+```
+
+
+### Request.Mixed
+
+Using the old Request.Mixed class instead of Request.Contao has been deprecated
```
Toflar: Hint for `JavaScript` is missing. Everything else is referring to PHP things so this might be confusing.

DEPRECATED.md
leofeyer: Isn't `System::getContainer()->get('request')` always the current request?

aschempp: As far as I know, the `request` service is only there for legacy support and should never be used.

leofeyer: Good to know :)

leofeyer:
I have fixed all remaining issues and changed the status to RTM.

src/Resources/contao/forms/Form.php
```diff
@@ -510,6 +510,7 @@ protected function processFormData($arrSubmitted, $arrLabels, $arrFields)
 			}
 
 			// Do not use Models here (backwards compatibility)
+			// TODO: shall we keep it this way or force everyone to use models for their tables?
```
Toflar: Keep it as is, but change in v5

src/Resources/contao/modules/ModuleRegistration.php
```diff
@@ -312,6 +312,7 @@ protected function compile()
 		// Add the groups
 		foreach ($arrFields as $k=>$v)
 		{
+			// TODO: why is this necessary?
 			$this->Template->$k = $v; // backwards compatibility
```
Toflar: Because of this commit: https://github.com/contao/core-bundle/commit/0fb309d4439b21d96cdadb9535923ad7ff7b1e41

src/Resources/contao/modules/ModuleRegistration.php
```diff
@@ -322,6 +323,8 @@ protected function compile()
 		$this->Template->formId = $strFormId;
 		$this->Template->slabel = specialchars($GLOBALS['TL_LANG']['MSC']['register']);
 		$this->Template->action = \Environment::get('indexFreeRequest');
+
+		// TODO: why is this necessary?
```
Toflar: Same here: https://github.com/contao/core-bundle/commit/0fb309d4439b21d96cdadb9535923ad7ff7b1e41

Toflar:
Added all my comments. Other than that, it's RTM :)

Issue #262, opened 2015-05-15
[4.0.0-RC1] Frontend error: "" is not a supported locale


ausi:
I set up a fresh installation of Contao 4.0.0-RC1, after the setup I get the exception `"" is not a supported locale.` for all front-end requests e.g. `/app_dev.php/` or `/app_dev.php/home.html`. The backend works.

Changing [EventListener/LocaleListener.php:55](https://github.com/contao/core-bundle/blob/0b2194e8fa042e4da5972f7ad06c28415f37677f/src/EventListener/LocaleListener.php#L55) to:

```php
if ($request->attributes->has('_locale') && null !== $request->attributes->get('_locale')) {
```

fixes the issue.

AFAIK `_locale` is set to `null` in [Symfony/Component/HttpKernel/EventListener/RouterListener.php:149](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpKernel/EventListener/RouterListener.php#L149).

aschempp:
1. Did you use the `contao/standard-edition`?
2. Did you enable `prepend_locale`?
3. What version of Symfony did it install for you (visible in the dev console)?

ausi:
1. Yes, I installed it via `php composer.phar create-project contao/standard-edition <zielordner> 4.0.0-RC1`
2. `prepend_locale` is set to `false`
3. Symfony 2.6.7

ausi:
Just tested it also with `prepend_locale: true`, didn’t make a difference.

leofeyer:
@aschempp I was able to reproduce it by setting `prepend_locale` to `false` and the executing `composer run-script post-update-cmd` on the command line.

leofeyer:
I think the problem is here:

https://github.com/contao/core-bundle/blob/develop/src/Routing/FrontendLoader.php#L108

Shouldn't we set a default locale here?

```php
$route->addDefaults(['_locale' => 'en']);
```

This also fixes the issue on my system.

leofeyer:
It also helps to remove the entire line. Why set a `_locale` attribute at all if it is `null`?

leofeyer:
I think @ausi's fix is correct :)

leofeyer:
I have created a PR: https://github.com/contao/core-bundle/pull/264

Please review and vote +1 to merge it.

PR #264, opened 2015-05-15
[RTM] Correctly handle an empty `_locale` attribute

leofeyer:
This is the PR for #262.

### DIFF
--- a/src/EventListener/LocaleListener.php
+++ b/src/EventListener/LocaleListener.php
@@ -69,10 +69,8 @@ class LocaleListener
      */
     private function getLocale(Request $request)
     {
-        $attributes = $request->attributes;
-
-        if ($attributes->has('_locale') && null !== $attributes->get('_locale')) {
-            return $this->formatLocaleId($attributes->get('_locale'));
+        if (null !== $request->attributes->get('_locale')) {
+            return $this->formatLocaleId($request->attributes->get('_locale'));
         }
 
         if (null !== ($session = $request->getSession()) && $session->has('_locale')) {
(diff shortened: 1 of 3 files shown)


xchs:
+1

I got the same error as already mentioned in #262 and this PR fixes it (at least for me).

src/EventListener/LocaleListener.php
```diff
@@ -51,16 +51,35 @@ public function onKernelRequest(GetResponseEvent $event)
...
+     *
+     * @param Request $request The request object
+     *
+     * @return string The locale
+     */
+    private function getLocale(Request $request)
+    {
+        $attributes = $request->attributes;
+
+        if ($attributes->has('_locale') && null !== $attributes->get('_locale')) {
```
ausi: I think this can be reduced to

```php
if (null !== $attributes->get('_locale')) {
```

because `ParameterBag::get()` retuns `null` if the key doesn’t exist.

ausi:
Besides https://github.com/contao/core-bundle/pull/264#discussion_r30442244 +1

ausi:
Should we add `[null, 'en'],` to [LocaleListenerTest.php:251](https://github.com/contao/core-bundle/blob/0b2194e8fa042e4da5972f7ad06c28415f37677f/tests/EventListener/LocaleListenerTest.php#L251) and `[null, 'de', ['de', 'en']],` to [LocaleListenerTest.php:267](https://github.com/contao/core-bundle/blob/0b2194e8fa042e4da5972f7ad06c28415f37677f/tests/EventListener/LocaleListenerTest.php#L267)?

src/EventListener/LocaleListener.php
aschempp: I'd agree with that.

leofeyer:
I have changed the class accordingly and added the tests as proposed by @ausi. I have also changed the status of the PR to RTM.

PR #275, opened 2015-05-16
Use master request to initialize the framework

aschempp:
We should quickly discuss if that's the best option.

### DIFF
--- a/src/ContaoFramework.php
+++ b/src/ContaoFramework.php
@@ -123,7 +123,7 @@ class ContaoFramework
         $this->csrfTokenName = $csrfTokenName;
         $this->config        = $config;
         $this->errorLevel    = $errorLevel;
-        $this->request       = $requestStack->getCurrentRequest();
+        $this->request       = $requestStack->getMasterRequest();
     }
 
     /**


leofeyer:
Let's assume the master request is not one of the Contao routes (so no framework) but triggers a sub-request for one of the Contao routes. Would we want the framework to be initialized with the data of the master request or the current one? How would we want the generated URLs to look like?

aschempp:
Good question. Considering that our Input class does only support the master request, and it was not possible to have subtequests previously, it's probably correct to use master only. 

I'm not really experienced with subrequests though.

leofeyer:
I guess the safest solution would be to initialize the system with the current request. The first request that needs the framework should also have the correct context.

Toflar:
I think the current request is correct. But we can discuss that :)

leofeyer:
The original approach was correct: we have to inject the request stack and work with the current request. @aschempp Do you want to adjust your PR or do you want me to change it?

aschempp:
We haven't discussed the topic…

leofeyer:
Fine, let's wait until Wednesday. Although it is clear what to do :)

discordier:
As I already stated numerous times, always use the current request on the stack.

leofeyer:
So basically @Toflar, @discordier and I all share the same opinion. Shall we still waste time discussing this issue? :)

aschempp:
Yes please…

discordier:
Somewhat related to this.
I just realized that the User class is setting the locale etc. on the current request... this might bear problems when the request is a sub request and others expect it on the master request or vice versa.
The coders would have to reset the user from the DB for every sub request.
I wonder if we can just ignore that until we have a replacement for the legacy authentication layers.

leofeyer:
Ok, we **do** need to discuss this :)

leofeyer:
As discussed on Mumble on May 20th, we want to use the request stack in conjunction with `getCurrentRequest()`.

aschempp:
Is now part of #282

leofeyer:
Fixed in dcc2cf56cc08067ef8dbc854344949d92dcade08.

PR #277, opened 2015-05-18
The referer should only be stored for the backend route

aschempp:
see https://github.com/contao/core/blob/master/system/modules/core/classes/BackendUser.php#L112 for the Contao 3 version.

### DIFF
--- a/src/EventListener/StoreRefererListener.php
+++ b/src/EventListener/StoreRefererListener.php
@@ -67,7 +67,7 @@ class StoreRefererListener
      */
     private function storeBackendReferer(Request $request)
     {
-        if (!$this->canModifyBackendSession($request)) {
+        if ('contao_backend' !== $request->attributes->get('_route') || !$this->canModifyBackendSession($request)) {
             return;
         }
 


src/EventListener/StoreRefererListener.php
```diff
@@ -67,7 +67,7 @@ public function onKernelResponse(FilterResponseEvent $event)
      */
     private function storeBackendReferer(Request $request)
     {
-        if (!$this->canModifyBackendSession($request)) {
+        if ('contao_backend' !== $request->attributes->get('_route') || !$this->canModifyBackendSession($request)) {
```
leofeyer: Why didn't you add the condition to the `canModifyBackendSession()` method?

Toflar: Yeah, I wondered the same myself :D

aschempp: Because the method is called `canModifyBackendSession` and it previously was not the same `if` too. But that would obviously work either way…

leofeyer:
Fixed in 8ed603af96738cab3b492c74436dba8719a7a95c.

PR #279, opened 2015-05-18
Do not show maintenance message in debug mode

aschempp:
I think it does make sense to show the real website if the visitor is successfully using the `app_dev.php` entry point (either on localhost or using authentication).


leofeyer:
Fixed in 3997cb73e371757aca5a924aab1b94008d01a1ce.

PR #282, opened 2015-05-22
[RTM] Bring back the custom entry point

aschempp:
This will add support for legacy entry points. It worked fine with my local tests and the two entry points of Isotope eCommerce. Be aware that the legacy entry point does not support request-response, but I guess we have to live with that.


aschempp:
Fixed tests and Scrutinizer. Let me know your thoughts before I finish stuff like unit tests.

src/Security/ContaoAuthenticator.php
```diff
@@ -28,6 +30,21 @@
...
+     * @var ContainerInterface
+     */
+    private $container;
+
+    /**
+     * Constructor.
+     *
+     * @param ContainerInterface $container The service container
+     */
+    public function __construct(ContainerInterface $container)
```
leofeyer: This is an API change and cannot be done at the current point in time. We will have to extend `ContainerAware` instead.

leofeyer:
Phew, these are a lot more changes than expected. Why did you have to adjust the framework and the authenticator class?

Given the amount of work it takes to make custom entry points work, I'd almost vote for not supporting them at all. Would this be a problem e.g. for Isotope?

src/Security/ContaoAuthenticator.php
aschempp: I often hear @Toflar that constructors are not API. Not sure where that info is from though…

leofeyer: That's wrong. It was possible to instantiate the class without passing a container object before and it is no longer now. Definitely an API change.

aschempp:
> Why did you have to adjust the framework and the authenticator class?

The authenticator incorrectly booted the framework regardless of the container scope. That means our "lazy" booting was not working at all (because Security is a `kernel.request` event).

The framework changes are minimal. The constants should not be set because they are already initialized in every entry point script.

src/Security/ContaoAuthenticator.php
aschempp: True. Though we could also say this is a bug (see my other comment) in the RC ;-)

leofeyer: We should simply extend `ContainerAware` and handle the case `null === $this->container`. Problem solved without API change :)

aschempp: Not really. If the container is not there, we can't determine the scope, so not load the framework. That would change the whole class behavior.

Toflar: I certainly didn't say that. This is a BC break for sure.

leofeyer: > If the container is not there, we can't determine the scope, so not load the framework. That would change the whole class behavior.

I don't get it. Right now, there is no container, either, so how does the class behavior change?

aschempp: Currently it does load the user no matter the container scope. With the change, it would not, if the container is not injected (because that would be wrong).

leofeyer:
@aschempp You did not answer my question if it would be a problem for Isotope if we would not support custom entry points in Contao 4 anymore.

aschempp:
It certainly would. The currently implementation has two custom entry points. And I'm pretty sure there are quite a few other extensions. Just search the ER for files containing `define('TL_MODE'`

aschempp:
Btw, I don't really see this thing as so complicated? If you remove the `ContaoFramework` and `ContaoAuthenticator` changes, it barely consists of a frontend controller and an entry point script.

leofeyer:
I have fixed the `ContaoFramework` class in dcc2cf56cc08067ef8dbc854344949d92dcade08. I will fix the authenticator in a separate commit too, so we can see the actual changes of this PR.

leofeyer:
I have extracted the authentication related changes in 9ce022922cffd7dca19fd71779c6bb2ec34017e3. Can you rebase the PR now please?

aschempp:
I have rebased the PR

src/Controller/InitializeController.php
```diff
@@ -0,0 +1,64 @@
...
+use Contao\CoreBundle\ContaoCoreBundle;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Handles the Contao initialize.php route.
+ *
+ * @author Andreas Schempp <https://github.com/aschempp>
```
Toflar: Shouldn't this whole class be marked as `@deprecated`?

src/Controller/InitializeController.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $masterRequest = $this->get('request_stack')->getMasterRequest();
+        $realRequest   = Request::createFromGlobals();
+        $scope         = ContaoCoreBundle::SCOPE_FRONTEND;
+
+        if (defined('TL_MODE') && TL_MODE === 'BE') {
+            $scope = ContaoCoreBundle::SCOPE_BACKEND;
+        }
+
+        // Necessary to make the base path correct
```
leofeyer: Why is this necessary?

src/Controller/InitializeController.php
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @return Response
+     *
+     * @Route("/_initialize", name="contao_initialize")
+     */
+    public function indexAction()
+    {
+        trigger_error('Custom entry points are deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
+
+        $masterRequest = $this->get('request_stack')->getMasterRequest();
```
leofeyer: Why is the master request not the real request?

src/Controller/InitializeController.php
aschempp: The method `Request::getBasePath()` would return the path to the entry point script as relative URL all assets.

src/Controller/InitializeController.php
aschempp: Because the master request is the one being generated in our `initialize.php` with the custom route.

src/Controller/InitializeController.php
leofeyer: Shouldn't we replace the whole server bag with the one of the master request?

aschempp: Maybe that would work too…

aschempp: well no…

aschempp: The server bag of the master request is "incorrect" as well. The master request will point to the actual entry point script (which I think is good).

leofeyer:
Ok, so this is RTM then, isn't it? What about the `if (defined('TL_SCRIPT'))` check in the framework initializer class? No longer required?

aschempp:
Good point, they sure are still required! Must have got lost during the rebase.

aschempp:
Fixed the `defined` check

aschempp:
Found another problem and fixed it. Now the entry point works fine with the banner extension from @BugBuster1701

leofeyer:
Don't forget to change the status to `RTM` once it is ready to merge.

aschempp:
I just added another improvement. I had a very hard time yesterday to find the problem in af02d90843a6d0d9b352a1a4a5b7c141aea918d7 because the exception rendered by Symfony was not shown on the screen. This is RTM to me now.

leofeyer:
Two more questions:

1. Why are you sending a response with the status code `100`? This seems completely wrong, because a valid response has a status code greater or equal to 200 and less than 300.

2. If the response does not have a status code of `100`, you execute `$kernel->terminate()` and `exit;`. But if the status code is `100`, you execute neither of them. Is this correct?

aschempp:
1. From [Wikipedia](http://en.wikipedia.org/wiki/List_of_HTTP_status_codes)

    > **100 Continue**
This means that the server has received the request headers, and that the client should proceed to send the request body.

    I found this to be the most appropriate status code to tell our entry point script to continue the operation.

 2. Yes that's correct. If there is a Symfony response, we must send it and stop the script. Otherwise we leave the initialize.php and continue with the entry point script.

leofeyer:
It is not appropriate at all (and not even implemented in HTTP/1.0). The status 100 is meant for checking in advance if a server is willing to accept a rather big POST request, instead of directly sending the POST request. A status 100 response does not contain a body, either!

We definitely need to find another way of identifying the responses generated by the `InitializeController` class. And before we call `$response->send()`, we **must** set the response code to `200`.

aschempp:
You must have misread that. If the response code **is 100, we're not sending the response**. Therefore we don't need to set anything to *200*, and the actually used code does not matter at all for the client/server response. It's only a mean to communicate with the entry point script, because a controller **must** return a response.

leofeyer:
Ok, I might have gotten that wrong. Still we should not use a wrong status code. Why don't you add an X-Status-Header or similar instead?

aschempp:
The status code is not wrong… there’s just no status code for server-internal communication :D

We should use a status code that will certainly not be used by any other Symfony response, and I think that will be the case with a code 100.

leofeyer:
It's still wrong. @discordier What do you think?

leofeyer:
Merged in ffb9df4a4b5797531e55f9f77976d5dcca1ca5f3.

PR #288, opened 2015-06-01
[RFC] Fix the Symfony version display in the debug bar

leofeyer:
This PR fixes the Symfony version display in the debug bar. It will colorize the output depending on whether the Symfony version has known issues and this does not work if we inject the Contao version.

Also, it is wrong to inject the Contao version, because the end product is a Symfony app (as stated in the keynote). Contao is "just" a bundle and therefore should have the version number attached to its icon (see the screenshot above).


aschempp:
I agree with removing it from Symfony app. But I would prefer to show the container state as previously. The Contao version is available from the popup.

leofeyer:
So is the container scope :)

aschempp:
True. We should only add the most relevant information… Not sure what that would be though.

leofeyer:
I'd say the version number. People usually know whether they are in the back end or in the front end :)

aschempp:
Yeah, they usually also know the Contao version (which is visible in the backend) :-P

leofeyer:
True. Then just add no `sf-toolbar-info-piece-additional` at all?

aschempp:
Except if we can mimic the "is supported" information from Symfony…

leofeyer:
We'd have to find out how they check the version; they are probably using the [Security Advisories checker API](https://security.sensiolabs.org/api).

bytehead:
@leofeyer we did this already for contao < 4: [contao security advisories](https://github.com/1up-lab/contao-security-advisories). It also checks the installed extensions (by the composer-client) against the security lists. I would like to have a security check shipped with the core bundle. At least usable with a cli command.

aschempp:
Great idea, but probably for 4.1 (as 4.0 is RC)?

leofeyer:
I'll change the milestone after I have merged the fixes.

PR #291, opened 2015-06-04
[RFC] Added unit tests for Widget::getPost

aschempp:
Related to https://github.com/contao/core/issues/7855

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -18,6 +18,7 @@
             <exclude>./tests/Contao/GdImageTest.php</exclude>
             <exclude>./tests/Contao/ImageTest.php</exclude>
             <exclude>./tests/Contao/PictureTest.php</exclude>
+            <exclude>./tests/Contao/WidgetTest.php</exclude>
             <exclude>./tests/Fixtures</exclude>
         </testsuite>
         <testsuite name="legacy">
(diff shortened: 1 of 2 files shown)


aschempp:
We should discuss wether the data result 5 is correct.

leofeyer:
You mean `['foo[k1][0]', 'foo', ['k1'=>'bar'], 'bar'],`? It is not correct IMHO. It should be `['foo[k1][0]', 'foo', ['k1'=>['bar']], 'bar'],` instead.

aschempp:
No, the test is correct. It tests what happens if the requested subkey is not an array. The result is questionable, that's what is currently returned.

leofeyer:
If PHP behaves this way, our class should, too, don't you think?

aschempp:
Well PHP would probably return `b` and not `bar`? Also, if the index would not be numeric it would not return anything (I assume).

leofeyer:
Frankly, I neither see the point of the PR nor do I really understand your problem :)

aschempp:
well, there are two things:

 1. The primary point of the PR was to validate https://github.com/contao/core/issues/7855, to confirm that we do not need the `empty()` check. The code coverage tells us that the array is never empty.

 2. The question is something I noticed while adding test input and adding the actual output. I think it's wrong to return `bar` if the input is `foo => ['k1'=>'bar']` and we're looking for `'foo[k1][0]`

leofeyer:
Ok, then we should fix this when working on contao/core#7855, correct?

aschempp:
I would say so, yes. But I would change that both in 4.0 (and not 3.5)

aschempp:
Please merge this in the develop branch then…

aschempp:
Please don't forget about this

leofeyer:
I won't. I have kept the e-mail in my todo folder.

leofeyer:
Added in b4783ddecbab222c88d53c31658dad5ebe43cc10.

PR #319, opened 2015-07-28
[RTM] Feature/image size permission

qzminski:


src/Resources/contao/library/Contao/System.php
```diff
@@ -562,32 +562,74 @@ public static function getTimeZones()
 	/**
 	 * Return all image sizes as array
 	 *
+	 * @param bool $all Returns all sizes omitting the back end user permission check
+	 *
 	 * @return array The available image sizes
 	 */
-	public static function getImageSizes()
+	public static function getImageSizes($all = false)
```
aschempp: I would prefer to call this argument something like `$checkPermissions`.

src/Resources/contao/library/Contao/System.php
```diff
@@ -562,32 +562,74 @@ public static function getTimeZones()
...
 	{
-		if (empty(static::$arrImageSizes))
-		{
-			try
-			{
+		if (empty(static::$arrImageSizes)) {
```
aschempp: Please use the legacy code style, as this is a legacy class.

src/Resources/contao/library/Contao/System.php
```diff
@@ -562,32 +562,74 @@ public static function getTimeZones()
...
+
+		// Permission check
+		if (TL_MODE == 'BE' && $all === false) {
+			$user = \BackendUser::getInstance();
+
+			// Limit only if the user is not an admin
+			if (!$user->isAdmin) {
+				$allowedSizes = deserialize($user->imageSizes, true);
+
+				if (!empty($allowedSizes)) {
```
aschempp: This means if no option is checked then all image sizes are allowed? I think that's not how all the other permissions work?

src/Resources/contao/library/Contao/System.php
```diff
@@ -562,32 +562,74 @@ public static function getTimeZones()
...
+
+					$imageSizes = $filteredSizes;
+				}
+			}
+		}
+
+		// HOOK: allow to add custom logic
+		if (isset($GLOBALS['TL_HOOKS']['getImageSizes']) && is_array($GLOBALS['TL_HOOKS']['getImageSizes'])) {
+			foreach ($GLOBALS['TL_HOOKS']['getImageSizes'] as $callback) {
+				$imageSizes = static::importStatic($callback[0])->$callback[1]($imageSizes);
```
aschempp: You should pass the method argument here, too.

aschempp: Also, as this is Contao 4, we should use events and not add new hooks.

qzminski: > Also, as this is Contao 4, we should use events and not add new hooks.

Isn't it still the work in progress? https://github.com/contao/core-bundle/pull/204

src/Resources/contao/library/Contao/System.php
qzminski: Yes, if no option is checked then all image sizes are allowed. It is not how all other permissions work (not sure but probable) but is there ever a case you would not want to display any option to the user? If you are creating a gallery content element then you would either want to display all or limited options, but never none.

aschempp: Somewhat true, but if you enable the file manager and do not allow any folder, thats the same. I agree it might be useful to allow everything, but it is against the security rules, because things should always be only-on-allow…

src/Resources/contao/library/Contao/System.php
aschempp: It is, for converting the existing hooks. But I would not add new ones. I'm open to other opinions though :-)

src/Resources/contao/library/Contao/System.php
qzminski: Maybe we should have the "Limit image sizes" checkbox with a subpalette then?

aschempp: Unfortunately that does not work with the current permission-merging. And there is no such option anywhere in the system yet either.

qzminski: So I would have to check image size checkboxes every time I create a new user group or a new image size? That sounds like a tedious work. @Toflar ?

aschempp: That's the problem. But it's the same if you add a folder to the file manager. Or if you install a new backend module. Or if you add a new page to the page tree…

discordier:
It is totally correct to add an event for this but having the code for filtering inline in the system class is IMO totally wrong.
The whole stuff you have added in `System::getImageSizes()`, aside from triggering the event should be moved into an listener for the new event.
This allows developers to manipulate the array before the filtering by the core occurs if they want or do anything else the like by using a higher priority in their listeners.
It also helps keeping the business logic modular.

src/Resources/contao/library/Contao/System.php
```diff
@@ -587,7 +592,71 @@ public static function getImageSizes()
...
+							{
+								$filteredSizes[$group][] = $size;
+							}
+						}
+					}
+
+					$imageSizes = $filteredSizes;
+				}
+			}
+		}
```
discordier: As stated in https://github.com/contao/core-bundle/pull/319#issuecomment-126050836, all above here should be moved into a listener for the event being triggered below.

aschempp:
We should actually use security voters, right?

discordier:
Nope, as then each of the image sizes would have to become an attribute.
This will very easily get out of hand.
Filtering is correct here.

However, for the edit, delete etc operation checkers in the various dca files it would be correct. Yet I'd prefer using ACLs in ther instead of voters.

aschempp:
http://symfony.com/doc/current/cookbook/security/acl.html

> Using ACL's isn't trivial, and for simpler use cases, it may be overkill. If your permission logic could be described by just writing some code (e.g. to check if a Blog is owned by the current User), then consider using voters.

I think using a voter would be perfectly valid. The voter would receive the image size model. No idea why that's not used in the first place here ^^

discordier:
If you want to stick to voters, ok, I'm fine with that too (btw. I have read the documentation also, no need to cite it here). However I think we are writing a library (Contao core-bundle) where ACL are better in contrast to writing a concrete application where voters could be used for anything not yet covered by the ACLs defined via base libraries.

On using the voters again, what permissions do you want to pass to the voter? IMO in this use case only `USE` would make sense.
However, we could extend it then to check the permission if an user is allowed to `EDIT`, `CREATE`, `DELETE` the model and hereby centralize the check.

leofeyer:
Why do we need this in the first place?

aschempp:
In my first Contao 4 installation, I have about 20 different image size configurations. 15 of them are for custom developed frontend modules. Only 3-5 should be used by the user in content elements. I also don't want the user to set a custom size.

By limiting the available options, a user can only choose what I predefined.

aschempp:
We have updated the PR to provide a service instead of adjusting the legacy code. Please review and give feedback.

Toflar:
Apart from the missing unit tests this looks great!

leofeyer:
Right now we are not yet using the event system for hooks, so I can only merge this PR after #204 has been merged, correct?

aschempp:
#204 is about converting existing hooks. There's no reason to provide a hook for new events in my opinion.

src/Event/ContaoCoreEvents.php
```diff
@@ -0,0 +1,38 @@
...
+final class ContaoCoreEvents
+{
+    /**
+     * The contao.image_sizes.all event is triggered when all image sizes are generated.
+     *
+     * @var string
+     *
+     * @see Contao\CoreBundle\Event\GetImageSizesEvent
+     */
+    const IMAGE_SIZES_ALL = 'contao.image_sizes.all';
```
leofeyer: Should be `contao.image_sizes_all`.

src/Event/ContaoCoreEvents.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * The contao.image_sizes.user event is triggered when image sizes are generated for a user.
+     *
+     * @var string
+     *
+     * @see Contao\CoreBundle\Event\GetImageSizesEvent
+     */
+    const IMAGE_SIZES_USER = 'contao.image_sizes.user';
```
leofeyer: Should be `contao.image_sizes_user`.

src/Image/ImageSizes.php
```diff
@@ -0,0 +1,128 @@
...
+     * @param Connection               $db
+     * @param EventDispatcherInterface $eventDispatcher
+     */
+    public function __construct(Connection $db, EventDispatcherInterface $eventDispatcher)
+    {
+        $this->db              = $db;
+        $this->eventDispatcher = $eventDispatcher;
+    }
+
+    public function getAllOptions()
```
leofeyer: Almost all methods are missing a phpDoc comment.

src/Resources/contao/dca/tl_user.php
```diff
@@ -346,6 +346,18 @@
 			'eval'                    => array('multiple'=>true),
 			'sql'                     => "blob NULL"
 		),
+		'imageSizes' => array
+		(
+			'label'                   => &$GLOBALS['TL_LANG']['tl_user']['imageSizes'],
+			'exclude'                 => true,
+			'inputType'               => 'checkbox',
+			'options_callback'        => function () {
+				return System::getContainer()->get('contao.image.image_sizes')->getAllOptions();
```
leofeyer: The method should probably be called just `all()`.

src/Resources/contao/languages/en/tl_user.xlf
```diff
@@ -197,6 +203,9 @@
       <trans-unit id="tl_user.filemounts_legend">
         <source>Filemounts</source>
       </trans-unit>
+      <trans-unit id="tl_user.imagesizes_legend">
```
leofeyer: The key should be camel-cased (`imageSizes`) like the two above.

leofeyer:
Aside from my comments, this PR looks good to me.

src/Resources/contao/languages/en/tl_user.xlf
aschempp: Agreed, @qzminski can you change that?

src/Resources/contao/dca/tl_user.php
aschempp: No it should not. Because the service is not supposed to only return permissions. @ausi might use it for his PR on the image services, that's why the name is pretty specific.

src/Image/ImageSizes.php
aschempp: Yeah… it was a first draft :P

src/Event/ContaoCoreEvents.php
aschempp: True, because that's the constant's name. I guess we can live with the missing "options" in the name?

aschempp:
Be aware that we need the changes to the install tool (already in develop) for this to work, as we're now accessing the container in the DCA.

aschempp:
The necessary changes should be done

Toflar:
phpdoc and unit tests are still missing though

aschempp:
Docs and unit tests are added.

leofeyer:
Ist this one RTM? The title still says `RFC`.

aschempp:
I think so, unless someone has more feedback. @qzminski please change the title.

leofeyer:
Merged in ef296e37144daa96d043bedbb7cbe6184ce44cfa.

Issue #329, opened 2015-08-13
Resource paths not being registered for custom extenstions


jamesdevine:
There appears to be a problem in `DependencyInjection/Compiler/AddResourcesPathsPass.php` where the paths for extensions placed in 'system/modules' aren't being added to the resources paths variable.

I have a bundle registered in AppKernel.php for the extension:

````
$bundles = [
     ...
     new Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle('boxes', $this->getRootDir())
];
````

But the check on line (43) in `AddResourcesPathsPass.php` never returns true when looping through the bundles:

````
if ('Contao\\CoreBundle\\HttpKernel\\Bundle\\ContaoModuleBundle' === $class)
````

This seems to be preventing the install tool from finding the DCA files and syncing the database for custom extensions in `system/modules`.

Changing this line as follows cures the issue of detecting the class:
````
if (is_a($class,'Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle')) {
````

However there is a second issue where the wrong path is being used:

````
$paths[] = sprintf('%s/system/modules/%s', $rootDir, $name);
````

This should be:

````
$paths[] = sprintf('%s/system/modules/%s', $rootDir, $class->getName());
````

So in summary, this is how the whole block should be:

````
foreach ($container->getParameter('kernel.bundles') as $name => $class) {
            if (is_a($class,'Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle')) {
                $paths[] = sprintf('%s/system/modules/%s', $rootDir, $class->getName());
            } elseif (null !== ($path = $this->getResourcesPathFromClassName($class))) {
                $paths[] = $path;
            }
        }
````

After implementing that change the database tool on the install tool works properly for modules in `system/modules`.

aschempp:
`$class` should not be an object but a string. But I can agree that it does not work starting with Contao 4.0.2, I just didn't notice until just now (I didn't investigate why my backend module is gone).

jamesdevine:
A bit more info on this - it seems `$class` is only an object when running the install tool (via `install.php`).

When loading the CMS backend `$class` is a string as it should be and the current code works fine.

aschempp:
Then the install tool is wrong.

leofeyer:
Fixed in contao/installation-bundle@ff6f3db1d1389d916631f6267084cb21b5c9f194.

PR #330, opened 2015-08-13
Fix resource path discovery for custom modules

jamesdevine:


jamesdevine:
See issue #329

leofeyer:
See #329.

PR #331, opened 2015-08-13
[RTM] Backwards compatibility for old page types

Toflar:
As discussed on mumble.

### DIFF
--- a/src/Resources/contao/controllers/FrontendIndex.php
+++ b/src/Resources/contao/controllers/FrontendIndex.php
@@ -267,17 +267,11 @@ class FrontendIndex extends \Frontend
 					// Backwards Compatibility
 					if (!method_exists($objHandler, 'getResponse'))
 					{
-						$objResponse = new Response();
-
 						// Generate content
 						ob_start();
 						$objHandler->generate($objPage, true);
-						$strBuffer = ob_get_clean();
 
-						$objResponse->setContent($strBuffer);
-						$objResponse->setStatusCode(http_response_code());
-
-						return $objResponse;
+						return new Response(ob_get_clean(), http_response_code());
 					}
 
 					return $objHandler->getResponse($objPage, true);
--- a/src/Resources/contao/controllers/FrontendIndex.php
+++ b/src/Resources/contao/controllers/FrontendIndex.php
@@ -262,11 +262,25 @@ class FrontendIndex extends \Frontend
 					break;
 
 				default:
-					/** @var \PageRegular $objHandler */
 					$objHandler = new $GLOBALS['TL_PTY'][$objPage->type]();
 
+					// Backwards Compatibility
+					if (!method_exists($objHandler, 'getResponse'))
+					{
+						$objResponse = new Response();
+
+						// Generate content
+						ob_start();
+						$objHandler->generate($objPage, true);
+						$strBuffer = ob_get_clean();
+
+						$objResponse->setContent($strBuffer);
+						$objResponse->setStatusCode(http_response_code());
+
+						return $objResponse;
+					}
+
 					return $objHandler->getResponse($objPage, true);
-					break;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


Toflar:
ToDo: What should `PageRoot::getResponse()` do?

leofeyer:
It should return a `RedirectResponse` I guess.

leofeyer:
This is far from `RTM` to me, which probably results from the fact that there was no `RFC` phase :)

IMHO it is wrong to generate a response here if the page handler has no `getResponse()` method. And merging all the headers is over-engineering, because if someone calls `header()`, the headers will be send anyway.

Furthermore, you are mixing two different issues in your PR, namely adding the missing `getResponse()` method to the page handlers and adding a BC routine to handle legacy page handlers which do not have a `getResponse()` method.

leofeyer:
As discussed at the core workshop in August 2015, the merging should be done in an event listener if a route requests it.

aschempp:
Would it be possible to only register the listener if the legacy framework is booted?

leofeyer:
> Would it be possible to only register the listener if the legacy framework is booted?

Don't know if that makes sense, because you might want to apply it to your custom route, too.

aschempp:
No, my custom routes should always use the response. Only legacy code might use headers, and then the legacy framework is always required.

Toflar:
In theory it could make sense for any route. Everybody can send headers using `header` it's just not the right way in a Symfony environment. Anyway: We can also agree that we don't want to have this header merging. It's a @contao/developers decision.

aschempp:
I agree it would make sense, but it's wrong. We should only "support" it for legacy code, because there was no correct solution previously. There still is no correct solution to add headers in a frontend module for example…

leofeyer:
I'm in favor of an event listener, which only applies if it is a Contao request.

aschempp:
The event listener should simply check for ContaoFramework::isInitialized() :-)

Toflar:
Okay, I'll update this PR and will create another one for the listener :)

leofeyer:
Why don't we use the [ScopeAwareTrait](https://github.com/contao/core-bundle/blob/master/src/EventListener/ScopeAwareTrait.php#L67)?

aschempp:
Well that’s somewhat deprecated now with the Symfony container scope changes? I think that would be mixed functionality.

leofeyer:
I recall that we have agreed to keep using the container scopes, haven't we?

aschempp:
We have to keep using them, but we should not extend their functionality. Also, the scope is not related to headers at all.

leofeyer:
We are not extending their functionality. The `isContaoScope()` method exists. And I figured this would be the preferred way to determine if a request is a Contao request or not. Am I wrong?

aschempp:
Sure that’s the correct location. I thought you wanted to add the header catching routines to the scope trait.

leofeyer:
No, I just want the event listener to do `if (!$this->isContaoScope()) return;`.

aschempp:
Ah. Well I still prefer the framework and not the scope. Only legacy code should be handled, but it should also work for custom routes.

leofeyer:
@Toflar I'll leave the decision up to you :)

Toflar:
I'll see what I can do :)

Toflar:
This PR is adjusted to provide BC for page types without a `getResponse()` method.

aschempp:
So if a frontend module will set the headers that won't be handled, right?

Toflar:
Yes, because this is not part of this PR:

src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -262,11 +262,25 @@ public function run()
...
+
+						// Generate content
+						ob_start();
+						$objHandler->generate($objPage, true);
+						$strBuffer = ob_get_clean();
+
+						$objResponse->setContent($strBuffer);
+						$objResponse->setStatusCode(http_response_code());
+
+						return $objResponse;
```
leofeyer: This could be greatly simplified now:

```php
return new Response(ob_get_clean(), http_response_code());
```

Toflar: Done.

leofeyer:
Merged in 5b66e9f7f3baebe79b04266a47e559f283630d92.

PR #334, opened 2015-08-25
[RTM] Render pretty error screens for the master request and HTML format only

### LINKED ISSUE #298: PrettyErrorScreenListener renders error_404 page but not page_not_found template
There's some strange behavior in the `PrettyErrorScreenListener` class.

If I throw a `PageNotFoundException`, it is converted to a `NotFoundHttpException` by the `ExceptionConverterListener`. The `PrettyErrorScreenListener` will then check that exception.

 - if it is a `NotFoundHttpException`, it will search for a *page_404* type in your page structure and render that

- If there is no *page_404* exception, it will only render the `page_not_found` template if any of the parent exceptions is a `PageNotFoundException`.

That does not really make sense to me. I guess we need to discuss that in an upcoming Mumble call ^^

leofeyer:
Fixes #298.


aschempp:
The event listener should be stateless.

leofeyer:
I have updated the PR accordingly. Is it RTM now?

aschempp:
Maybe you should deprecate the exceptions we're not using anymore?

leofeyer:
We are still using all of the exceptions. We just do not render a pretty error screen for two of them anymore.

aschempp:
Ah, true. Then I think it’s RTM.

I thought about if anyone was using the current Twig templates, but I think we don’t need to keep BC about that.

leofeyer:
Templates != SemVer :)

aschempp:
I know, but it would still break every installation if the templates are gone ;-)

leofeyer:
True. But in this case, I highly doubt that anyone out there already used the templates :)

aschempp:
Agreed

Toflar:
:+1:

leofeyer:
Merged in 552193164633130ca27f6e75f21fcaff9ffd55ad.

Issue #228, opened 2015-04-20
Improve utf8_romanize


Toflar:
Ich habe bei meinen blackfire.io-Tests auch gesehen, dass diese Funktion einer der grössten Bottlenecks ist, die Contao noch hat. Ich denke also es wäre sehr sinnvoll, hier ein bisschen Research reinzustecken. Insbesondere, weil wir uns für die 3.5 für PHP 5.4 entschieden haben und nun den Transliterator nutzen könnten: http://php.net/manual/en/class.transliterator.php

Ausserdem hab ich noch ein Package mit 3.5 Millionen Downloads gefunden, welches versucht, die ganzen iconv, Intl und mbstring Sachen zu abstrahieren und je nach System das schnellste zu wählen: https://github.com/tchwork/utf8
Ein return \Patchwork\Utf8::toAscii($str); dürfte womöglich das gleiche Resultat wie bei uns bieten und einiges schneller sein. Die prüft übrigens auch darauf ob überhaupt UTF-8-Ranges betroffen sind (https://github.com/tchwork/utf8/blob/master/class/Patchwork/Utf8.php#L42).

leofeyer:
Siehe auch: https://github.com/contao/core/issues/7748

Toflar:
@aschempp are you trying to use the PHP core `Locale` for #190? Because if so, I guess we can get rid of our own language and country lists and use Symfony's `Intl` component which will provide a fallback for English if the php `intl` extension is not installed :) :) :) Obviously this will improve performance quite a bit!

```php
// Displays an "alphabetically" ordered list of country names in Japanese
use Symfony\Component\Intl\Intl;
\Locale::setDefault('ja');
$countries = Intl::getRegionBundle()->getCountryNames();
print_r($countries);
```

```php
// Displays an alphabetically ordered list of country names in French
use Symfony\Component\Intl\Intl;
\Locale::setDefault('fr');
$countries = Intl::getRegionBundle()->getCountryNames();
print_r($countries);
```

```php
Displays an alphabetically ordered list of languages in German (even "Schweizerdeutsch" is available :-D)
use Symfony\Component\Intl\Intl;
\Locale::setDefault('de');
$languages = Intl::getLanguageBundle()->getLanguageNames();
print_r($languages);
```

aschempp:
[...]

PR #336, opened 2015-08-25
[RTM] Use unit tested utf-8 library patchwork/utf-8

Toflar:
I've replaced all our `utf8_*` functions by the equivalents provided by `patchwork/utf-8`. Also, I removed all our `mb_*` compatibility implementations as the library does that as well.

For reasons of BC, I have moved all `utf8_*` functions to `functions.php` and marked them as `deprecated`.
However, there is one function which I don't quite know what to to with:

`utf8_decode_entities` is doing this now:

```php
	$str = preg_replace_callback('~&#x([0-9a-f]+);~i', function($matches) {
		return \Patchwork\Utf8::chr(hexdec($matches[1]));
	}, $str);
	$str = preg_replace_callback('~&#([0-9]+);~', function($matches) {
		return \Patchwork\Utf8::chr($matches[1]);
	}, $str);

	return $str;
```

I think this should be replaced by
```php
html_entity_decode($str, ENT_COMPAT|ENT_HTML401, 'utf-8');
```

However, I see that it is currently not replacing things like `&nbsp;`. `utf8_decode_entities` is only used in the `Input` class which obviously is a sensitive part of our application so I don't know if we can just change it? @ausi Do you have an opinion on that?

Moreover, should this PR already include the changes necessary to replace the lines of code using the `utf8_*` functions or will we remove the usages when working on Contao 5?


ausi:
I don’t see a security problem in changing `utf8_decode_entities` to `html_entity_decode` but it changes the current behavior by double decoding some entities, e.g.:

```php
$_GET['foo'] = '&#x26;amp;';
// Before:
Input::get('foo'); // returns &amp;
// After:
Input::get('foo'); // returns &
```

The DocComment of the current version also says “HTML entities like `&nbsp;` or `&quot;` will not be decoded.”. So IMO we should keep the current version.

Am I right that you didn’t deprecate `utf8_decode_entities` because it is still used in the `Input` class? If so, I would suggest to replace it with the two `preg_replace_callback` calls in the `Input` class and deprecate the `utf8_decode_entities` function.

Toflar:
> Am I right that you didn’t deprecate utf8_decode_entities because it is still used in the Input class? If so, I would suggest to replace it with the two preg_replace_callback calls in the Input class and deprecate the utf8_decode_entities function.

Yeah, that's exactly what I would've done but I guess we should try to understand and evaluate whether the current behavior of the `Input` class is correct or not.

leofeyer:
Can I close #228 in favor of this?

Toflar:
Yes, you can.

ausi:
> I guess we should try to understand and evaluate whether the current behavior of the `Input` class is correct or not.

I tried to understand what `Input::xssClean()` does, or should do. From a security point of view the method does the following:

* Not modify simple values: `null`, empty string, booleans, numbers and numeric strings.
* Remove null bytes
* Remove all occurrences of the substring `\0`

The rest of the filters are not relevant for security IMO because they can all be bypassed. But they are relevant for working with the `Input` class because they do weird things:

```php
\Input::xssClean("&amp \n\t;")  // returns '&amp;'
\Input::xssClean('Java Script') // returns 'JavaScript'
\Input::xssClean('Apple T-V') // returns 'AppleT-V'
\Input::xssClean('<a href="http://example.com/document.cookiereceipt.html">')
// returns empty string
\Input::xssClean('<a href="https://en.wikipedia.org/wiki/Expression (mathematics)">')
// returns empty string
\Input::xssClean('<a href="https://en.wikipedia.org/wiki/Expression">')
// returns '<a href="https://en.wikipedia.org/wiki/Expression">'
```

The two lines [Input.php:590-591](https://github.com/contao/core-bundle/blob/a1b9832f6c4a5ee96c13979b6144346e0790c2da/src/Resources/contao/library/Contao/Input.php#L590-L591) are a mystery for me, especially the second one. The git history of those lines points to “Initial commit” from August 2008.

Toflar:
> But they are relevant for working with the Input class because they do weird things:

:D 

> The two lines Input.php:590-591 are a mystery for me, especially the second one.

Yeah, it replaces itself by itself. Actually this line just does nothing except consuming time :D

Well, we should probably just leave it as is and mark it deprecated linking some resources about why input filtering is wrong and should not be done.

ausi:
> Well, we should probably just leave it as is and mark it deprecated linking some resources about why input filtering is wrong and should not be done.

+1

Toflar:
Any links/resources you could recommend?

ausi:
No, sorry.

discordier:
Input filtering is not bad at all but it is bad to try to "repair" broken input.
Some rules apply in general:
1. "All input is evil" (Michael Howard in "Writing Secure Code")
2. Invalid input must be discarded, not processed!
3. "All input is still evil"
4. Do not trust user input at all, therefore make sure that you properly escape database query parameters, data passed as html output, etc.

On everything above and definately worth reading:
http://phpsecurity.readthedocs.org/en/latest/Input-Validation.html

On the differerences between sanitation and validation:
https://vip.wordpress.com/2011/10/13/key-differences-between-validation-and-sanitization/

Some heads up on what is evil:
http://anuary.com/55/why-input-sanitization-is-evil

How to correctly apply filtering (by putting it where it belongs, the "controller", not the environment/framework)
https://www.owasp.org/index.php/Testing_for_Input_Validation
http://www.smashingmagazine.com/2011/01/keeping-web-users-safe-by-sanitizing-input-data/

leofeyer:
> The two lines [Input.php:590-591](https://github.com/contao/core-bundle/blob/a1b9832f6c4a5ee96c13979b6144346e0790c2da/src/Resources/contao/library/Contao/Input.php#L590-L591) are a mystery for me, especially the second one.

You can easily solve this mystery by searching for `/(&#*\w+)[\x00-\x20]+;/` on e.g. Google.

ausi:
>> The two lines [Input.php:590-591](https://github.com/contao/core-bundle/blob/a1b9832f6c4a5ee96c13979b6144346e0790c2da/src/Resources/contao/library/Contao/Input.php#L590-L591) are a mystery for me, especially the second one.

> You can easily solve this mystery by searching for `/(&#*\w+)[\x00-\x20]+;/` on e.g. Google.

I didn’t find an explanation of what this regular expression does on Google. Can you post the link please?

leofeyer:
E.g. here: https://gist.github.com/mbijon/1098477

leofeyer:
Here is an even better explanation: http://psoug.org/snippet/XSS_Sanitizer_Function_17.htm

ausi:
I still don’t get it. The comments above the regular expressions don’t make any sense to me.

Toflar:
I have added the hint to the Input class and deprecated the `utf8_decode_entities` function. It's RTM to me now.

leofeyer:
> The comments above the regular expressions don’t make any sense to me.

It will fix standard character entities and add a semi-colon at the end in case it is missing. There are browser that render these incomplete entities (with the missing semi-colon) but `html_entity_decode()` will not recognize them and thus not remove them.

I don't know how to exploit this, but obviously others do :)

ausi:
The two lines [Input.php:590-591](https://github.com/contao/core-bundle/blob/a1b9832f6c4a5ee96c13979b6144346e0790c2da/src/Resources/contao/library/Contao/Input.php#L590-L591) do not add a semicolon if it is missing. What does *“It will fix standard character entities”* mean? Which problem is fixed by replacing `"&amp \n\t;"` with `"&amp;"`?

leofeyer:
I am currently merging this PR and there are some issues I want to address:

**utf8_convert_encoding()**

You have removed the `iconv()` call, so now we rely on `mb_convert_encoding()` being present. Is this correct?

Also, can we remove the `@` operator from the `mb_convert_encoding()` call?

**utf8_detect_encoding()**

Again, we are now relying on `mb_detect_encoding()` to be present. Have we agreed on making the PHP `mbstr` extension mandatory?

**utf8_strlen()**

Same as above.

leofeyer:
I just looked it up: So far, the PHP `mbstring` extension is *optional*. After this change, it would be *mandatory*. @contao/developers /cc

leofeyer:
And obviously, if we make the `mbstring` extension required, we don't need the patchwork library anymore :)

Toflar:
I guess you should read the description of the library I used.

leofeyer:
Well I have, and the description suggest a completely different use case!

Instead of using the methods of the `Utf8` class, you are supposed to initialize the environment early in your bootstrap sequence and then use the regular `mb_`, `iconv_` and `grapheme_` functions, which the library will define if the PHP extensions are not available.

Toflar:
Ah I see what you mean now. Well imho only the `\Patchwork\Utf8\Bootup::initAll();` is missing to make sure shims are loaded if the extensions are not present. However, this is not related to not using the library at all because a lot of things are not possible with the functions. For example `Utf8::toAscii()`.

leofeyer:
In fact, `Utf8::toAscii()` is the only method we need from the `Utf8` class, because none of the PHP extensions provides it.

So, if we decide to use the portability layer (by adding `Bootup::initAll()`), we would also have to replace all `Patchwork\Utf8` usages with their `mb_` equivalents, right?

Toflar:
* `Utf8::chr()` ?
* `Utf8::ord()` ?
* etc.

To me, the `Utf8` class should be used because to me it's the general interface to all the handling you need. In fact it does transparent mapping to the native functions in most cases but it also makes sure bugs are handled properly (such as in `Utf8::stripos()`) or prefers the extensions over a native implementation if available (such as in `Utf8::str_split()`). That's how I understood that library.

leofeyer:
> because to me it's the general interface to all the handling you need

That's not right, because it only mirrors *some* of the functions, but not all of them. I've stumbled across it when looking for an `mb_convert_encoding()` method; there is none in the `Utf8` class, because you are supposed to use the native `mb_` method (which will be provided by the library if not available as PHP extension).

Toflar:
I know, I did? For example `utf8_strlen()` now uses `mb_strlen()`.

leofeyer:
That was **the only** function using an `mb_` function. All other functions were using the `Utf8` class!

Toflar:
Because they're available there and I think they're there for good reason because they check for additional things :-) Do you have an example for a method that is a) defined in `Utf8` and b) not doing anything else than passing it to an `mb_*` function?

leofeyer:
Yes, of course:

* `Utf8::strlen()`
* `Utf8::strpos()`
* `Utf8::strrpos()`
* `Utf8::strstr()`
* `Utf8::strrchr()`
* `Utf8::strrichr()`
* `Utf8::strtolower()`
* `Utf8::strtoupper()`

Enough?

leofeyer:
You can have a look at the current merge state here:

https://github.com/contao/core-bundle/blob/feature/use-patchwork-utf8/src/Resources/contao/helper/functions.php#L533-L942

I agree that using the library is inconsistent. But it is a fact that not all required functions are mirrored in the `Utf8` class (such as `mb_convert_encoding()`). What do you think?

Toflar:
Oh, then `utf8_strlen()` should use `Utf8::strlen()`. Must have missed this one.
What the library says: 

> The Patchwork\Utf8 class implements the quasi-complete set of native string functions that need UTF-8 grapheme clusters awareness. Function names, arguments and behavior carefully replicates native PHP string functions.

> Mirrored string functions are: strlen, substr, strpos, stripos, strrpos, strripos, strstr, stristr, strrchr, strrichr, strtolower, strtoupper, wordwrap, chr, count_chars, ltrim, ord, rtrim, trim, str_ireplace, str_pad, str_shuffle, str_split, str_word_count, strcmp, strnatcmp, strcasecmp, strnatcasecmp, strncasecmp, strncmp, strcspn, strpbrk, strrev, strspn, strtr, substr_compare, substr_count, substr_replace, ucfirst, lcfirst, ucwords, number_format, utf8_encode, utf8_decode, json_decode, filter_input, filter_input_array.

It just provides a general way to work with Utf8 so I don't quite understand why you would not want to use it. To me, `Utf8::strlen($string)` reads a lot clearer that `grapheme_strlen($string)`. That's all I'm saying. Do you prefer using the native Intl functions then? I mean, I can change that I just don't think it's great :-D I think the idea of Nicolas Grekas was that we use a general `Utf8` class when working with UTF-8 instead of using `pcre` here, `intl` there and `mstring` elsewhere.

Toflar:
> such as mb_convert_encoding()

Well, this one is probably not there because it converts encodings from one into another and does not belong to a `Utf8` class. `Utf8` obviously only handles `utf-8` :-D

leofeyer:
We also have to talk about removing the `USE_MBSTRING` constant, which technically is a BC break.

leofeyer:
@nicolas-grekas Can you please shed some light on this? What is the best practice to use the Patchwork UTF-8 library, using the native `mb_`, `iconv_` and `grapheme_` functions or using the `Utf8` class as much as possible?

nicolas-grekas:
If grapheme_strlen() applies to your use case, then using Utf8::strlen would be more consistent as I agree with you: **not** using pcre here, intl there and mbstring elsewhere is the purpose if the Utf8 class. (add workaround for bugs to the mix).

nicolas-grekas:
But note that you have to choose one measuring system and stick to it: either graphemes or unicode characters. That means either Utf8::* or mb_*. I can't choose for you on this topic.

leofeyer:
@nicolas-grekas Thanks a lot.

leofeyer:
Here is the current change set: https://github.com/contao/core-bundle/compare/66d380d37b9d9c848f5e33b3e2a6937d6e44d7fd...3a1c487418040ea6e7a47cf8b1a64fbeb6b67051

leofeyer:
Merged in cbb13debd16ee26817df63d2a9c39afe3eb1b4b4.

PR #338, opened 2015-08-27
[RTM] Use Monolog to log PHP errors and exceptions

### LINKED ISSUE #335: The file "app/logs/error.log" does not exist after an error occurred.
I tried to test the module newsletter list with Contao 4.0.2 but I received an error and the file "error.log" does not exist in "app/logs". I have only `newsletter_1.log` and `dev.log`.

leofeyer:
This PR includes three fixes:

* Use Monolog to log PHP errors.
* Add log entries in the pretty error screen listener.
* Use `app/logs` instead of `system/logs`.

The PR also fixes #335.

### DIFF
--- a/src/EventListener/PrettyErrorScreenListener.php
+++ b/src/EventListener/PrettyErrorScreenListener.php
@@ -274,13 +274,13 @@ class PrettyErrorScreenListener
                 'errorOccurred' => 'An error occurred while executing this script. Something does not work properly. '
                     . 'Additionally an error occurred while trying to display the error message.',
                 'howToFix' => 'How can I fix the issue?',
-                'errorFixOne' => 'Open the <code>app/logs/prod.log</code> file and find the associated error '
-                    . 'message (usually the last one).',
+                'errorFixOne' => 'Search the <code>app/logs</code> folder for the current log file and find the '
+                    . 'associated error message (usually the last one).',
                 'more' => 'Tell me more, please',
                 'errorExplain' => 'The script execution stopped, because something does not work properly. The '
                     . 'actual error message is hidden by this notice for security reasons and can be '
-                    . 'found in the <code>app/logs/prod.log</code> file (see above). If you do not '
-                    . 'understand the error message or do not know how to fix the problem, search the '
+                    . 'found in the current log file (see above). If you do not understand the error message or do '
+                    . 'not know how to fix the problem, search the '
                     . '<a href="https://contao.org/faq.html">Contao FAQs</a> or visit the '
                     . '<a href="https://contao.org/support.html">Contao support page</a>.',
             ],
--- a/src/EventListener/PrettyErrorScreenListener.php
+++ b/src/EventListener/PrettyErrorScreenListener.php
@@ -71,10 +71,10 @@ class PrettyErrorScreenListener
     /**
      * Constructor.
      *
-     * @param bool              $prettyErrorScreens True to render the error screens
-     * @param \Twig_Environment $twig               The twig environment
-     * @param ConfigAdapter     $config             The config adapter
-     * @param LoggerInterface   $logger             An optional logger service
+     * @param bool                 $prettyErrorScreens True to render the error screens
+     * @param \Twig_Environment    $twig               The twig environment
+     * @param ConfigAdapter        $config             The config adapter
+     * @param LoggerInterface|null $logger             An optional logger service
      */
     public function __construct(
         $prettyErrorScreens,
--- a/src/EventListener/PrettyErrorScreenListener.php
+++ b/src/EventListener/PrettyErrorScreenListener.php
@@ -15,6 +15,7 @@ use Contao\CoreBundle\Exception\InternalServerErrorHttpException;
 use Contao\CoreBundle\Exception\RedirectResponseException;
 use Contao\StringUtil;
 use Contao\System;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
@@ -49,6 +50,11 @@ class PrettyErrorScreenListener
     private $config;
 
     /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @var array
      */
     private $mapper = [
@@ -68,12 +74,18 @@ class PrettyErrorScreenListener
      * @param bool              $prettyErrorScreens True to render the error screens
      * @param \Twig_Environment $twig               The twig environment
      * @param ConfigAdapter     $config             The config adapter
+     * @param LoggerInterface   $logger             The logger service
      */
-    public function __construct($prettyErrorScreens, \Twig_Environment $twig, ConfigAdapter $config)
-    {
+    public function __construct(
+        $prettyErrorScreens,
+        \Twig_Environment $twig,
+        ConfigAdapter $config,
+        LoggerInterface $logger
+    ) {
         $this->prettyErrorScreens = $prettyErrorScreens;
         $this->twig = $twig;
         $this->config = $config;
+        $this->logger = $logger;
     }
 
     /**
@@ -118,6 +130,7 @@ class PrettyErrorScreenListener
                 break;
 
             default:
+                $this->logException($exception);
                 $this->renderTemplate('error', 500, $event);
                 break;
         }
@@ -197,6 +210,7 @@ class PrettyErrorScreenListener
             return;
         }
 
+        $this->logException($exception);
         $this->renderTemplate($template, $statusCode, $event);
     }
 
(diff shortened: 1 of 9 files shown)


src/Resources/contao/helper/functions.php
```diff
@@ -15,9 +15,9 @@
  * @param string $strMessage
  * @param string $strLog
  */
-function log_message($strMessage, $strLog='error.log')
```
Toflar: Imho this is wrong. It should be marked `@deprecated` and also use the logger service internally, no?

src/Resources/contao/library/Contao/Automator.php
```diff
@@ -462,11 +462,11 @@ public function generateInternalCache()
 	 */
 	public function rotateLogs()
```
Toflar: Monolog has a `RotatingFileHandler` for this, can't we reuse this?

tests/EventListener/PrettyErrorScreenListenerTest.php
```diff
@@ -262,7 +266,10 @@ public function testErrorScreen()
             })
         ;
 
-        $listener = new PrettyErrorScreenListener(true, $twig, new ConfigAdapter());
+        /** @var LoggerInterface $logger */
+        $logger = $this->getMock('Psr\\Log\\LoggerInterface');
+
+        $listener = new PrettyErrorScreenListener(true, $twig, new ConfigAdapter(), $logger);
```
Toflar: You should not only fix the tests but also include tests for the new functionality which is checking if the logger logs on exception :)

src/Resources/contao/helper/functions.php
leofeyer: The problem is that logging might occur before the container is set. I'm ok with deprecating the method though.

src/Resources/contao/library/Contao/Automator.php
leofeyer: Sure, didn't know about it.

tests/EventListener/PrettyErrorScreenListenerTest.php
leofeyer: You are right.

src/Resources/contao/helper/functions.php
Toflar: Ah, I see. I've been asking myself this for a long time already. What does SF2 itself do if there occur errors before the container is ready?

leofeyer: I suppose they fall back to the standard error handler. But our method is not (only) related to errors.

Toflar: And we cannot really use the container here anyway as the channels (target files) are already set and cannot be changed dynamically. So the only case where we could use the container here is if the container is built and ready to use and `$strLog` was not set explicitly.

discordier:
Die miserably with default error settings. By definition a container is always created in deployment as it is built from cli and therefore the developer sees if anything goes wrong.

src/Resources/contao/helper/functions.php
discordier: We already talked about this during our first Ahrweiler meeting and AFAIR we came to the conclusion that we can not support `$strLog` anymore and therefore will either have to prepend the message with the value of this or ignore the variable entirely.

For the fact of logging before the container is built. This would AFAIR only affect logging from `config.php` during container compilation, wouldn't it? IMO we should silently ignore the logging request then or use the old functionality as noted in this PR.

I think this pseudocode outlines the best approach:
```php
function log_message($strMessage)
 {
	// $container should be fetched here, if any available.
	if ((func_num_args() === 1) && $container) {
		$container->get('logger')->info($strMessage);
	} else {
		error_log(sprintf("[%s] %s\n", date('d-M-Y H:i:s'), $strMessage), 3, TL_ROOT . '/app/logs/' . $strLog);
		trigger_error(E_USER_DEPRECATED, 'Logging without container and/or to arbitrary destinations is not supported anymore.');
	}
 }
```

leofeyer:
So, any reason why I should not merge this PR?

leofeyer:
Sorry, I didn't realize that there are still some adjustments to be made :)

leofeyer:
Now this PR is really `RTM`. Can you please check?

Toflar:
Excellent :+1:

leofeyer:
The rotating file handler configuration is here: https://github.com/contao/standard-edition/commit/b6d2e0c837cad40076f2342e145a3871b058d25b

src/Resources/contao/helper/functions.php
```diff
@@ -14,10 +14,20 @@
...
 {
-	error_log(sprintf("[%s] %s\n", date('d-M-Y H:i:s'), $strMessage), 3, TL_ROOT . '/system/logs/' . $strLog);
+	trigger_error('Using log_message() has been deprecated and will no longer work in Contao 5.0. Use the logger service instead.', E_USER_DEPRECATED);
+
+	if ($strLog === null)
+	{
+		$strLog = 'prod-' . date('Y-m-d') . '.log';
+	}
+
+	error_log(sprintf("[%s] %s\n", date('d-M-Y H:i:s'), $strMessage), 3, TL_ROOT . '/app/logs/' . $strLog);
```
aschempp: Why not use the logger service from the container here?

src/Resources/contao/languages/en/exception.xlf
```diff
@@ -21,10 +21,10 @@
         <source>An error occurred while executing this script. Something does not work properly.</source>
       </trans-unit>
       <trans-unit id="XPT.errorFixOne">
-        <source>Open the &lt;code&gt;app/logs/error.log&lt;/code&gt; file and find the associated error message (usually the last one).</source>
+        <source>Serach the &lt;code&gt;app/logs&lt;/code&gt; folder for the current log file and find the associated error message (usually the last one).</source>
```
aschempp: Typing mistake

leofeyer: Fixed in adf2426c23e27cba2d0ed7d9c7554c235f59f2a6.

src/Resources/contao/helper/functions.php
leofeyer: Read through the comments :)

src/Resources/config/listener.yml
```diff
@@ -83,6 +83,7 @@ services:
             - "%contao.pretty_error_screens%"
             - "@twig"
             - "@contao.adapter.config"
+            - "@monolog.logger"
```
aschempp: You should use the generic `@logger` service (which usually is an alias to `monolog.logger`.

leofeyer: Changed in c93da7791ddddf994bc6eb68937343d3633df5a1.

src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -68,12 +74,18 @@ class PrettyErrorScreenListener
...
      * @param ConfigAdapter     $config             The config adapter
+     * @param LoggerInterface   $logger             The logger service
      */
-    public function __construct($prettyErrorScreens, \Twig_Environment $twig, ConfigAdapter $config)
-    {
+    public function __construct(
+        $prettyErrorScreens,
+        \Twig_Environment $twig,
+        ConfigAdapter $config,
+        LoggerInterface $logger
```
aschempp: As far as I could see in the Symfony code, the Logger is always an optional dependency (https://github.com/symfony/symfony/search?utf8=✓&q=%22%24logger+%3D+null%22&type=Code). This would also prevent a (arguably irrelevant) BC break.
I guess the logger will be NULL if logging should be disabled (we did have that option in Contao 3.5).

leofeyer: Does this really make sense? We are only using the logger if we render a pretty error screen and the pretty error screen says "check the log file for the actual error message". Would anyone want to render the pretty error screen without logging the actual error message?

aschempp: Well that's a matter of configuration, right? I could also change the templates… Just saying it looks like a convention to me, but I couldn't find any actually description/blog about it. Maybe someone else knows?

leofeyer: Changed in b179886c3daa7fff44b5361b9218aa73c9409db7.

leofeyer:
Merged in c2f92152eda891328995f123e86a00c9b04c81d8.

Toflar:
I'm sorry, I saw this too late but an optional dependency should be injected using setter injections, not optional constructor injection.

leofeyer:
However, there are [quite a few locations where Symfony is doing it this way](https://github.com/symfony/symfony/search?utf8=%E2%9C%93&q=%22%24logger+%3D+null%22&type=Code), too.

Toflar:
Imho they're doing it wrong too then :)

aschempp:
I think it's correct, because the dependency is not optional. The logger is always injected, but the service value can be null.

Toflar:
Lol, so if it can be null it is optional. There is nothing in between null and not null :D

leofeyer:
It is not only the logger. [Here](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Http/Firewall/RemoteUserAuthenticationListener.php#L31) is an example where the event dispatcher can be omitted.

Toflar:
Yeah, there might be even more of those. Sure. But again, they're all wrong imho.

leofeyer:
No idea really. There is a heated discussion on "optional dependencies" in general and I really don't know what is the current best practice.

Toflar:
I mean, I can live with it, obviously. I just like it when things are obvious. If we handle all optional dependencies within the core bundle with setter injection then I can rely on the fact that all constructor arguments are always needed. Now I have to check what happens if I don't add the logger service. And I have to do this for every service with optional constructor services. However, if those were setter injections it's the other way around: Only if I wanted it do do additional things, I have to check what additional calls I have to execute. You see the difference? But after all, it's just a decision we all have to make and then stick to :)

Toflar:
Plus, it is easier to extend!

leofeyer:
Maybe we should ask a Symfony developer for their current best practice? Just don't mention them here, otherwise they have to read through the whole thread just to answer a simple question :)

aschempp:
We're not following this in the core already, as there are setter injections that are sort of required (due to traits)…

Again, I fully agree with what Symfony does. The requirement is not option in the sense that you should always pass the respective container service. You're never supposed to not add the argument in the configuration. It's just possible for the service to be `null`.

Toflar:
And I stick to what I said too: There is nothing between null and not null. Whatever you are supposed to do or not as a developer, from the application point of view the logger service is either required or not. If it's allowed to be `null` it's not required (= optional) for the service to work. End of the story :)

src/Resources/contao/helper/functions.php
discordier: The comments are gone. I guess due to force pushing on your side?

PR #339, opened 2015-08-31
[RTM] Added a fully unit tested MergeHttpHeadersListener

Toflar:
This is a follow up to https://github.com/contao/core-bundle/pull/331. We still have a lot of "legacy" code that cannot properly send HTTP headers. The way to go currently is still using `header()`. However, those ones are not respected in a lot of Symfony Bundles that listen on `kernel.response` and read the `Response` object for headers and do more stuff accordingly. This PR provides a listener that merges the sent HTTP headers into the SF response object unless they were already present (so it's not possible to overwrite the ones in the response).

RFC: 

* [x] `listener.yml` and config is obviously still missing.
* [x] Is it correct not to replace already existing headers in the response object?
* [x] What about the status code? I cannot really use `http_response_code()` because how do I know whether the one in the `Response` object is correct or the one set using `header()`?

### DIFF
--- a/src/EventListener/MergeHttpHeadersListener.php
+++ b/src/EventListener/MergeHttpHeadersListener.php
@@ -30,11 +30,6 @@ use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 class MergeHttpHeadersListener
 {
     /**
-     * @var array
-     */
-    private $routeNames;
-
-    /**
      * @var ContaoFrameworkInterface
      */
     private $contaoFramework;
@@ -54,16 +49,11 @@ class MergeHttpHeadersListener
     /**
      * Constructor.
      *
-     * @param array                    $routeNames
      * @param ContaoFrameworkInterface $contaoFramework
      */
-    public function __construct(
-        array $routeNames,
-        ContaoFrameworkInterface $contaoFramework
-    ) {
-        $this->routeNames = $routeNames;
+    public function __construct(ContaoFrameworkInterface $contaoFramework)
+    {
         $this->contaoFramework = $contaoFramework;
-
         $this->setHeaders(headers_list());
     }
 
@@ -123,15 +113,6 @@ class MergeHttpHeadersListener
     public function onKernelResponse(FilterResponseEvent $event)
     {
         if (!$this->contaoFramework->isInitialized()) {
-
-            return;
-        }
-
-        $request = $event->getRequest();
-
-        if (!$request->attributes->has('_route')
-            || !in_array($request->attributes->get('_route'), $this->routeNames)
-        ) {
 
             return;
         }
--- a/src/EventListener/MergeHttpHeadersListener.php
+++ b/src/EventListener/MergeHttpHeadersListener.php
@@ -40,6 +40,11 @@ class MergeHttpHeadersListener
     private $contaoFramework;
 
     /**
+     * Remove old headers or not
+     * @var boolean
+     */
+    private $removeOldHeaders = true;
+
     /**
      * Headers
      * @var array
@@ -81,6 +86,27 @@ class MergeHttpHeadersListener
     {
         $this->headers = $headers;
     }
+
+    /**
+     * Gets whether old headers should be removed using header_remove()
+     * or not.
+     *
+     * @return boolean
+     */
+    public function getRemoveOldHeaders()
+    {
+        return $this->removeOldHeaders;
+    }
+
+    /**
+     * Sets whether old headers should be removed using header_remove()
+     * or not.
+     *
+     * @param boolean $removeOldHeaders
+     */
+    public function setRemoveOldHeaders($removeOldHeaders)
+    {
+        $this->removeOldHeaders = (bool) $removeOldHeaders;
     }
 
     /**
@@ -120,6 +146,11 @@ class MergeHttpHeadersListener
     {
         foreach ($this->getHeaders() as $header) {
             list($name, $content) = explode(':', $header, 2);
+
+            if ($this->removeOldHeaders) {
+                header_remove($name);
+            }
+
             $content = trim($content);
 
             if ($response->headers->has($name)) {
--- a/src/EventListener/MergeHttpHeadersListener.php
+++ b/src/EventListener/MergeHttpHeadersListener.php
@@ -103,9 +103,15 @@ class MergeHttpHeadersListener
      * or not.
      *
      * @param boolean $removeOldHeaders
+     *
+     * @throws \InvalidArgumentException
      */
     public function setRemoveOldHeaders($removeOldHeaders)
     {
+        if (false === $removeOldHeaders && 'cli' !== PHP_SAPI) {
+            throw new \InvalidArgumentException('You cannot disable removing old
+            headers when not on CLI SAPI.');
+        }
         $this->removeOldHeaders = (bool) $removeOldHeaders;
     }
 
--- a/src/EventListener/MergeHttpHeadersListener.php
+++ b/src/EventListener/MergeHttpHeadersListener.php
@@ -33,21 +33,54 @@ class MergeHttpHeadersListener
      * @var array
      */
     private $routeNames;
+
     /**
      * @var ContaoFrameworkInterface
      */
     private $contaoFramework;
 
     /**
+    /**
+     * Headers
+     * @var array
+     */
+    private $headers = [];
+
+    /**
      * Constructor.
      *
-     * @param array $routeNames
+     * @param array                    $routeNames
      * @param ContaoFrameworkInterface $contaoFramework
      */
-    public function __construct(array $routeNames, ContaoFrameworkInterface $contaoFramework)
-    {
+    public function __construct(
+        array $routeNames,
+        ContaoFrameworkInterface $contaoFramework
+    ) {
         $this->routeNames = $routeNames;
         $this->contaoFramework = $contaoFramework;
+
+        $this->setHeaders(headers_list());
+    }
+
+    /**
+     * Gets the headers.
+     *
+     * @return array
+     */
+    public function getHeaders()
+    {
+        return $this->headers;
+    }
+
+    /**
+     * Sets the headers.
+     *
+     * @param array $headers
+     */
+    public function setHeaders(array $headers)
+    {
+        $this->headers = $headers;
+    }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


aschempp:
As previously stated, I still think we should not hard-code the route names but simply rely on the framework being initialized. That will always be the case if legacy code (e.g. current frontend modules) is generated.

Regarding replacing the headers, what will happen in the browser? We're not canceling the headers, right? So they will be sent anyway. Also, afaik a lot of headers can be set multiple times or have multiple values, or maybe I'm wrong about this?

src/EventListener/MergeHttpHeadersListener.php
```diff
@@ -0,0 +1,107 @@
...
+        if ('cli' === PHP_SAPI && function_exists('xdebug_get_headers')) {
+            $headers = xdebug_get_headers();
+        } else {
+            $headers = headers_list();
+        }
+
+        foreach ($headers as $header) {
+            list($name, $content) = explode(':', $header, 2);
+            $content = trim($content);
+
```
discordier: IMO here `header_remove($name);` should be added.

tests/EventListener/MergeHttpHeadersListenerTest.php
```diff
@@ -0,0 +1,205 @@
...
+     * Test for xdebug_get_headers() function. If this doesn't exist,
+     * the tests are nonsense as on CLI the headers_list() function does not
+     * return any headers sent using header().
+     *
+     */
+    protected function setUp()
+    {
+        if (!function_exists('xdebug_get_headers')) {
+            $this->markTestSkipped('To test header() merging you need to have
+            xdebug enabled');
```
discordier: Why this line break?

src/EventListener/MergeHttpHeadersListener.php
```diff
@@ -0,0 +1,107 @@
...
+            if ($response->headers->has($name)) {
+                continue;
+            }
+
+            $response->headers->set($name, $content);
```
discordier: Headers may be present more than once, therefore I guess we should skip the check and merge the value by using 
```diff
-            if ($response->headers->has($name)) {
-                continue;
-            }
-
-            $response->headers->set($name, $content);
+            $response->headers->set($name, $content, false);
```

Toflar:
Not sure if this is the way to go actually. Maybe we should just provide a service that allows adding headers from legacy context and just ignore the fact that legacy code is not handled?

discordier:
I am not sure I understand what you mean by your last comment.
Do you mean the one "sending" a response (creating it) should query the service and let it add the missing headers to the response?
If so, the service will be most likely never called as when one is creating new routes and a proper response he will not use the native header methods anyway.

Toflar:
So tell me, how do you properly send headers in a Contao front end module as of Contao 4? :)

aschempp:
Not possible. But it begs the question if a frontend module should ever do that (except by throwing a `ResponseException`)…

discordier:
> So tell me, how do you properly send headers in a Contao front end module as of Contao 4? :)

See your upcoming PR regarding ESI :smiling_imp:

> Not possible. But it begs the question if a frontend module should ever do that (except by throwing a `ResponseException`)…

This is of course another part of the story, in general it is not needed at all.

leofeyer:
I also think that we should not tie this to a predefined set of routes but simply rely on the Contao 3 framework being initialized. Also, I think we can omit the CLI routines.

Toflar:
> See your upcoming PR regarding ESI

This will have absolutely no affect on this PR. Not in the least.

discordier:
Of course it will, as you will have to perform the merging in there then and unset all set headers afterwards.

leofeyer:
@Toflar Will you be working on this PR for Contao 4.1?

Toflar:
I will try, yes.

Toflar:
So, I looked into this again and I'm not quite sure what to do really.

1. There's no way we can perfectly merge legacy headers because I cannot determine if a header that is present should be merged or not. If somebody wanted to override a header sent by a module on purpose, this listener is going to kill it again.
2. This listener sort of encourages you to send headers using `header()` in non-symfony code. However, I guess it should be done using your own listener if you need this.
3. I think it would be better to work towards making our 4 ways of how you can integrate content in Contao a normal controller action as soon as possible.

leofeyer:
Hm, I figured we only need to implement the two changes suggested by @discordier and then we'd be done?

Toflar:
Yeah those changes are the ones in question. My version does not replace the header if already present in the response. Chris' version will add/merge it. Both can be correct or not.

discordier:
As stated before, there is no 100% bullet proof solution but I guess my approach covers the most common use cases.

leofeyer:
1. I don't think we should override headers which are already present in the response object. Someone must have put them there, so we should not mess with them.

2. I don't think we should handle the CLI case. The Contao 3 framework was not supposed to be run on the command line, therefore it does not make sense to cover it now, don't you think?

leofeyer:
As discussed on Mumble on October 14th, the listener should have a high priority so the legacy headers are available in the following listeners.

The CLI support is required to run the PHPUnit tests.

Toflar:
* Got rid of the xdebug dependency in https://github.com/Toflar/core-bundle/commit/e84a4df96ce0a68a155f9b3878dedf38d5268827
* `header_remove()` added in https://github.com/Toflar/core-bundle/commit/8932026fcc3e1ed0d00e7d80b52ea02e3a4e4e48 /cc @xtra
* Merge http headers if already present instead of not adding them at all (adjusted unit tests too) in https://github.com/Toflar/core-bundle/commit/3b6accbc13b97b099ca0f5417e10c0fbf1f8182f /cc @xtra
* Listener config added in https://github.com/Toflar/core-bundle/commit/ceef2139d5fc3e2476b0490639c2f2a93a59f871

@contao/developers Changing to RTM.
@leofeyer When you merge, note that you have to define all Contao routes where you want to merge the headers in the `config.yml` of the standard distribution. Here's my proposal:

```diff
# Contao configuration
contao:
    prepend_locale: "%prepend_locale%"
    encryption_key: "%kernel.secret%"
    merge_http_headers_routes: 
+        - contao_root
+        - contao_catch_all
+        - contao_frontend
+        - contao_frontend_cron # not sure about this 
+        - contao_frontend_share # not sure about this 
```

aschempp:
If we really need to specify the routes (and I don't really like that), then there should be in the bundle's default configuration!

leofeyer:
> As previously stated, I still think we should not hard-code the route names but simply rely on the framework being initialized.

> I also think that we should not tie this to a predefined set of routes but simply rely on the Contao 3 framework being initialized.

You did not really address this, did you?

src/EventListener/MergeHttpHeadersListener.php
```diff
@@ -0,0 +1,161 @@
...
+     * @param Response $response
+     *
+     * @return Response
+     */
+    private function mergeHttpHeaders(Response $response)
+    {
+        foreach ($this->getHeaders() as $header) {
+            list($name, $content) = explode(':', $header, 2);
+
+            if ($this->removeOldHeaders) {
```
leofeyer: Why do we need this to be configurable? We always have to remove the headers once they are merged.

Toflar: It's needed for the unit tests. Thats why they all set this to `false`. You cannot operate on headers on the CLI.

Toflar:
> If we really need to specify the routes (and I don't really like that), then there should be in the bundle's default configuration!

True. Added in cd33d0d49770a38fbc1a07bd3a8ff978cbe2d795.



> You did not really address this, did you?

Me? Or Andy? I obviously did not because nobody told me to do so. I'm convinced we need this because if I write an `/api` endpoint and I need the `ContaoFramework` there too, I certainly don't want it to interfere with my headers. This way I can easily disable the listener by not adding the route for `/api` to the config. Even better, if I want it to merge my headers I have to explicitly enable it.

aschempp:
I dont think that case will ever happen. If you create your /api entry point, you will most likely not have someone use header(). And if you do because you have booted the old framework and use an old module, you probabky want them merged anyway. 

How about a method in the event listener to disable it instead of defining the routes? Then your custom controller could simply get the service and make sure the headers are "ignored" if it really ever is a problem.

Toflar:
> How about a method in the event listener to disable it instead of defining the routes? Then your custom controller could simply get the service and make sure the headers are "ignored" if it really ever is a problem.

I don't like this approach. I think that's wrong.

aschempp:
What feels wrong about it? The advantage is that it could be set by any third-party bundle instead of requiring to change the distribution.

Toflar:
Any third party bundle can also just set the config parameters...

src/EventListener/MergeHttpHeadersListener.php
leofeyer: Then why should check `('cli' !== PHP_SAPI)`, otherwise people (like me) think the option is there to be used :)

Toflar: Not really. If they want them not to be removed, then why not. Who cares :-)

leofeyer:
Last time we have discussed the issue, we have agreed on applying the listener whenever the Contao 3 framework is booted. I am also open for the new solution, however, we obviously need to discuss it.

aschempp:
> Any third party bundle can also just set the config parameters...

Not as far as I know on an array parameters, they're not merged from all bundles.

src/EventListener/MergeHttpHeadersListener.php
leofeyer: I do :)

And using the CLI check is something Symfony does, too: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Response.php#L373

Toflar: Okay, implemented in afdb2d43995722b1a2f8ff96662b2a833c18c510.

Toflar:
> Last time we have discussed the issue, we have agreed on applying the listener whenever the Contao 3 framework is booted.

I guess I was not part of this discussion then. I think it provides more flexibility when it can be disabled per route but I can also just kick it and we'll re-implement it when we really need it. I want to finish this one finally.

aschempp:
> I can also just kick it and we'll re-implement it when we really need it

I would prefer that. Let's not over-engineer for potential use-cases.

Toflar:
Done in 1df016d011c5d1a61309c5cc7389a8d0ee6736fb. Let me know if that's all okay now, then I'll squash the commits so we don't pollute the history of the main repository too much.

leofeyer:
Merged in d29eb181ad3748179de11602962654089748a197.

PR #342, opened 2015-09-08
[RTM] Image and Picture as services

ausi:
This is a very early draft of the image handling as a service. Here are my ideas:

The new classes `Contao\CoreBundle\Image\Image` and `Contao\CoreBundle\Image\Picture` work similar to the current classes, but with dependency injection and without the `create` and legacy methods. The new `Image` class uses `Imagine` to manipulate image data. The Image and Picture objects are created via a factory which is registered in the service container.

The old classes `Contao\Image` and `Contao\Picture` are wrappers  around the new classes and services to keep backwards compatibility.

Does that make sense?

## Questions

* [x] How do we handle the the hooks `executeResize` and `getImage`? To be 100% backwards compatible we would have to pass an instance of the `Contao\Image` class.
* [x] Should we still use the `Contao\File` class?
* [x] `ImageFactory` and `PictureFactory` or `Factory` with `createImage` and `createPicture` methods?
* [x] Should we split the functionality of how the replaced images are saved/cached into a separate `contao.image.cache` service?
* [x] Is the `imagine_class` configuration node the right way to make the used Imagine class configurable?
* [x] We would need something like a `contao.adapter.model_repository` service to get access to the files and image size models.


aschempp:
To tell you my opinion on the basic questions:
 - we should not use `Contao\File`
 - Two factories is better, each should only have one `create` method 
 - The `contao.image.imagine` service looks pretty valid to me

Regarding the `model_repository`, I guess you misunderstood me on that. Each class which currently has a static method should get it's own adapter. So in your case you would probably need `contao.adapter.files_model` and `contao.adapter.image_sizes` services.

Also, we should use a compiler pass to automatically detect the available Imagine options and choose the best (= only fallback to GD2 if Imagick is not supported).

Toflar:
> Regarding the model_repository, I guess you misunderstood me on that. Each class which currently has a static method should get it's own adapter. So in your case you would probably need contao.adapter.files_model and contao.adapter.image_sizes services.

Thinking about this, I think a general `model_repository` would be more intelligent. It would also cover non-core models.

ausi:
> Thinking about this, I think a general `model_repository` would be more intelligent. It would also cover non-core models.

For the image service we would have to inject three models: `ImageSizeModel`, `ImageSizeItemModel` and `FilesModel`. Therefore I think a general `model_repository` would make it cleaner.

aschempp:
Just think about Doctrine. Each entity has it's repository…

Toflar:
but only one entity manager ;)

ausi:
But I can inject `doctrine` and call `->getRepository('...')` on it. IMO it would be great to have a single service to get any Contao model from.

Toflar:
> But I can inject doctrine and call ->getRepository('...') on it. IMO it would be great to have a single service to get any Contao model from.

:+1:

aschempp:
> But I can inject doctrine and call ->getRepository('...') on it. IMO it would be great to have a single service to get any Contao model from.

True! It just means we will improve our model stuff while we actually wanted to prefer Doctrine ^^

ausi:
> Also, we should use a compiler pass to automatically detect the available Imagine options and choose the best (= only fallback to GD2 if Imagick is not supported).

Like so aaf12ec7ba631b0daaa3d4810afc7c19bf73d063?

ausi:
> we should not use `Contao\File`

Where should we put the code for `Contao\File::__get('imageSize')` then? Into the new `Image` class itself?

ausi:
Can I start working on it with the class structure described above?
Or do we need to discuss something before I start?

aschempp:
I would love to discuss the topic on today's Contao call. The compiler pass looks pretty good.

ausi:
I will be there.

ausi:
I added some classes :)

The current proposal would work like this:

`$imageFactory->create('foo.jpg')` creates an `Image` and returns it.

`$imageFactory->create('foo.jpg', 2)` creates an `Image`, fetches the image size `2` from the database and passes the data to the `Resizer`. The `Resizer` passes the configuration to the `ResizeCalculator` and resizes the image according to the result and returns it.

`$pictureFactory->create('foo.jpg', 2)` creates an `Image`, fetches the image size `2` from the database and passes the data to the `PictureGenerator`. The `PictureGenerator` uses the `Resizer` to resize the images and returns a `Picture` object with the results.

@contao/developers what do you think?

src/DependencyInjection/Compiler/DetermineImagineImplementation.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    private function getImagineImplementation()
+    {
+        foreach (['Imagick', 'Gmagick', 'Gd'] as $name) {
+
+            $class = 'Imagine\\' . $name . '\\Imagine';
+
+            try {
+                new $class();
+                break;
```
aschempp: I just realized this can be simplified a lot. Simply return `$class` here, then you also don't need the if-condition below (just throw the exception).

src/Image/ImageFactory.php
```diff
@@ -0,0 +1,83 @@
...
+namespace Contao\CoreBundle\Image;
+
+use Imagine\Image\ImagineInterface;
+use Contao\CoreBundle\Adapter\AdapterFactoryInterface;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * Creates Image objects
+ *
+ * @author Martin Auswöger <martin@auswoeger.com>
```
aschempp: We usually use a link to our GitHub account link instead of a real email address.

src/DependencyInjection/Compiler/DetermineImagineImplementation.php
ausi: Done.

src/Image/Image.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Gets the dimensions
+     *
+     * @return ImageDimensions
+     */
+    public function getDimensions()
+    {
+        return $this->dimensions;
```
Toflar: Where does this come from?

src/Image/ImageFactory.php
ausi: Is this a requirement? I don’t like to link to GitHub here because I don’t control the github.com domain :)

src/Image/Image.php
ausi: The Image class will have to compute the dimensions using the injected Imagine service or read them from the SVG source.

Toflar: Ah I see

src/Image/ResizeConfiguration.php
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Sets the height.
+     *
+     * @param int $height the height
+     *
+     * @return self
+     */
+    public function setHeight($height)
+    {
+        $height = (int) $height;
```
aschempp: I would performe the casting after the value check simply in the assignment statement below (applies to all methods).

Toflar:
I think I like the basic approach :-) The only thing I would change to bring even more interoperability would be to use things like `ResizerInterface` and type hint this instead of a fixed dependency on `Resizer` so we can easily exchange those :)

ausi:
@Toflar thanks. I agree we should add some interfaces, but I would like to get a working implementation and then adding the interfaces, to better understand which interfaces could be useful.

src/Image/ResizeConfiguration.php
ausi: I don’t like that, because e.g. `-0.1` would than throw an error but `(int)-0.1` wouldn’t. In more complex methods, casting after the validation could be used to bypass the validation.

src/Image/ImageDimensions.php
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param BoxInterface $size     The size
+     * @param bool         $relative The relative flag
+     */
+    public function __construct(
+        BoxInterface $size,
+        $relative = false
+    ) {
+        $this->size = $size;
+        $this->relative = (bool) $relative;
```
aschempp: What is a relative image dimension?

src/Image/ResizeConfiguration.php
aschempp: But `-0.1` would be an invalid input?

src/DependencyInjection/Compiler/DetermineImagineImplementation.php
```diff
@@ -0,0 +1,53 @@
...
+    private function getImagineImplementation()
+    {
+        foreach (['Imagick', 'Gmagick', 'Gd'] as $name) {
+
+            $class = 'Imagine\\' . $name . '\\Imagine';
+
+            try {
+                new $class();
+                return $class;
+            } catch (RuntimeException $exception) {}
```
aschempp: An empty catch statement should always explain why it's empty :-)

ausi: You wanted the code to be that way ;)
Is `// Continue the foreach loop if $class is not available` OK as an explanation?

aschempp: I would rather describe something like `Tests the Imagine class which throws an exception if the parent PHP implementation is not available.`

src/Image/ResizeConfiguration.php
ausi: IMO not. `-0.1` would be casted to `0` and is therefore valid.

aschempp:
I also like the approach in general. I just wonder if the `Image` class should be mutable at all? Also, I think the factory should not accept an ID of an image configuration (from DB) but directly it's configuration.

src/DependencyInjection/Compiler/DetermineImagineImplementation.php
ausi: done

ausi:
> I just wonder if the `Image` class should be mutable at all?

You suggest to add `$importantPart` as a constructor argument?
I tried to follow the approach to add required attributes to the constructors and optional attributes as setters.

> I think the factory should not accept an ID of an image configuration (from DB) but directly it's configuration.

Where should we put the code for that than?

src/Image/ImageDimensions.php
ausi: For images with a ratio, but no absolute size. E.g. a SVG image with `width="100%" height="100%"`.
See the difference between `File->imageSize` and `File->imageViewSize` in contao/core#7882.

src/Image/ImageFactory.php
leofeyer: Not providing your e-mail address will help defending spam, because crawlers can find your e-mail address in the raw content of the file.

aschempp:
Now that the adapters are finally merged, I think we can continue with that?

ausi:
Yes we can :)
I will rebase the PR and continue working on it.

aschempp:
@ausi any progress on this? would be good to have a "final draft" at our meeting.

ausi:
@aschempp some progress :)

Basic functionality works now:

```php
$image = $this->get('contao.image.image_factory')->create($path, $size);
$picture = $this->get('contao.image.picture_factory')->create($path, $size);
```

Next steps:

* [ ] Implement caching rules and respect `contao.image.bypass_cache`
* [ ] Add more tests
* [ ] Add backwards compatibility layer

discordier:
Please add events, we should not introduce hooks in new code.

Toflar:
Looks great! Will add some comments :)

src/Image/PictureConfigurationItem.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * Gets the resize configuration.
+     *
+     * @return ResizeConfiguration
+     */
+    public function getResizeConfig()
+    {
+        if (null === $this->resizeConfig) {
+            $this->setResizeConfig(new ResizeConfiguration);
```
Toflar: We always add the `()` in our CS :)

src/Image/ResizeCalculator.php
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * Resizes an Image object
+     *
+     * @param ResizeConfiguration $config        The resize configuration
+     * @param ImageDimensions     $dimensions    The image dimensions
+     * @param ImportantPart       $importantPart The important part
+     *
+     * @return ResizeCoordinates The resize coordinates
+     */
+    public function calculate(
```
Toflar: It would be great to split this up in more methods.

src/Image/ResizeConfiguration.php
```diff
@@ -0,0 +1,165 @@
...
+     * @var int
+     */
+    private $height = 0;
+
+    /**
+     * @var string
+     */
+    private $mode = self::MODE_CROP;
+
+    const MODE_CROP = 'crop';
```
Toflar: Class constants are usually at the very top of the class, no matter to what var they belong :)

src/ImagineSvg/Image.php
```diff
@@ -0,0 +1,461 @@
...
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\ImagineSvg;
```
Toflar: I'd vote for putting all the image related classes into a general `Image` namespace.

ausi: So the namespace should be `Contao\CoreBundle\Image\ImagineSvg`?

Toflar: I think so, yes. I just prefer having all the image related stuff into its own namespace :)

src/ImagineSvg/Imagine.php
```diff
@@ -0,0 +1,139 @@
...
+ * @author Martin Auswöger <martin@auswoeger.com>
+ */
+class Imagine extends AbstractImagine
+{
+    /**
+     * @throws RuntimeException
+     */
+    public function __construct()
+    {
+        if (!extension_loaded('dom')) {
```
Toflar: We could add that requirement to the `composer.json` :-)

src/Image/PictureConfigurationItem.php
ausi: Done.

src/Image/ResizeConfiguration.php
ausi: Done.

src/ImagineSvg/Imagine.php
ausi: Should we? Should’t it just be a known limitation that without `dom` you cannot resize SVG images?

src/ImagineSvg/Image.php
ausi: I thought too deep nesting levels aren’t great.
@contao/developers do you agree on the namespace `Contao\CoreBundle\Image\ImagineSvg`?

aschempp: Why would it not be `class SvgImage` in `Contao\CoreBundle\Image` namespace?

src/ImagineSvg/Imagine.php
Toflar: Then it should be a `suggest` :-) But it is a requirement for Contao 3 as well, so why would we drop that now? :)

src/ImagineSvg/Image.php
ausi: IMO it would be better to have the whole Imagine-SVG stuff in one namespace.

aschempp: Are there multiple svg-related classes?

ausi: https://github.com/madeyourday/contao-core-bundle/tree/8bf084d9b6527890c95f43af8591a63cc20fd1bc/src/ImagineSvg

aschempp: Aah, so these are actually our SVG drivers for imagine? They don't belong in the bundle at all imho, they're a library. But that's a difficult topic :D

ausi: Correct. They are an implementation of the Imagine interfaces. We can split them into its own package some time.

ausi:
This should be ready for comments now :)

One thing I’d like to discuss before merging is, what should be the Symfony-way alternative to the image hooks `executeResize` and `getImage`? In my project I can now easily replace one of the services to add custom behavior, but how can I do that as a Contao bundle?

Note to myself: After this gets merged I should take a look at some other image related feature requests:

* Add support for `w` descriptors contao/image#1
* Add percentage important part to Image sizes contao/core#8360
* Image Size alias contao/core#7520
* CSS classes for image sizes contao/core#8202
* Overridable predefined sizes contao/core#8171

src/DependencyInjection/Compiler/AddRootDirPass.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+class AddRootDirPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $container->setParameter(
+            'contao.root_dir',
```
leofeyer: This code does not what the phpDoc comment says. Also, we should not add a `contao.root_dir` parameter.

src/Image/Resizer.php
```diff
@@ -0,0 +1,148 @@
...
+            @trigger_error(
+                'Using the executeResize and getImage hooks has been deprecated and will no longer work in Contao 5.0. '
+                    . 'Use the resize service instead.',
+                E_USER_DEPRECATED
+            );
+
+            $this->legacyImage = null;
+            $legacyPath = $image->getPath();
+
+            if (strpos($legacyPath, TL_ROOT . '/') === 0 || strpos($legacyPath, TL_ROOT . '\\') === 0) {
```
leofeyer: Shouldn't this be `TL_ROOT . DIRECTORY_SEPARATOR`?

src/DependencyInjection/Compiler/AddRootDirPass.php
ausi: We need a way to inject `"%contao.root_dir%/%contao.image.target_path%"` into the resizer service. Should I use the `%kernel.root_dir%` instead?

src/Image/Resizer.php
ausi: IMO the path could possibly contain mixed directory separators like `C:\path\to\webspace/files/image.jpg`.

src/Resources/contao/languages/en/explain.xlf
```diff
@@ -78,7 +78,7 @@
...
+        <source>If the sizes attribute is not defined, this setting simply defines the pixel densities you want to support. The dimensions of the images are adjusted automatically. E.g. &lt;code&gt;1x, 1.5x, 2x&lt;/code&gt; creates the following HTML code:&lt;br&gt;&lt;code&gt;&amp;lt;img srcset="img-a.jpg 1x, img-b.jpg 1.5x, img-c.jpg 2x"&amp;gt;&lt;/code&gt;&lt;br&gt;&lt;br&gt;If the sizes attribute is defined, the same image sizes get generated but width descriptors are used for the srcset attribute. E.g. a 200 pixel wide image with the densities &lt;code&gt;200w, 300w, 400w&lt;/code&gt; creates the following HTML code:&lt;br&gt;&lt;code&gt;&amp;lt;img srcset="img-a.jpg 200w, img-b.jpg 300w, img-c.jpg 400w"&amp;gt;&lt;/code&gt;&lt;br&gt;&lt;br&gt;For more information about the srcset attribute please visit &lt;a href="http://www.w3.org/html/wg/drafts/html/master/embedded-content.html#attr-img-srcset" target="_blank"&gt;w3.org&lt;/a&gt;.</source>
```
leofeyer: Shouldn't `with the densities &lt;code&gt;200w, 300w, 400w&lt;/code&gt;` be `widths` instead of `densities`?

src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1506,8 +1506,14 @@ public static function addImageToTemplate($objTemplate, $arrItem, $intMaxWidth=n
 
 		try
 		{
-			$src = \Image::create($arrItem['singleSRC'], $size)->executeResize()->getResizedPath();
-			$picture = \Picture::create($arrItem['singleSRC'], $size)->getTemplateData();
+			$src = \System::getContainer()->get('contao.image.image_factory')->create(TL_ROOT . '/' . $arrItem['singleSRC'], $size)->getUrl(TL_ROOT);
```
leofeyer: We are passing an absolute path (`TL_ROOT . '/' . $arrItem['singleSrc']`) to the `create()` method to strip it again in the `getUrl()` method?

src/Resources/contao/languages/en/explain.xlf
ausi: I think “width descriptors” is the term used by the HTML specification. Should I change it to “with the width descriptors `200w, 300w, 400w`”?

src/Resources/contao/library/Contao/Controller.php
ausi: Yes, IMO that’s the right way. The image factory doesn’t know about `TL_ROOT` and the resizer only knows where to store the image files.

src/DependencyInjection/Compiler/AddRootDirPass.php
leofeyer: This is not so easy. I have created a working implementation, but I need to discuss this with you on Mumble.

src/Image/Resizer.php
leofeyer: Yes, but we had several problems with exactly this in the core. It depends on where the path comes from, which in this case is `$image->getPath()`.

src/Resources/contao/languages/en/explain.xlf
leofeyer: You cannot change it anymore, because I have already merged the PR. I'll do it.

src/Resources/contao/library/Contao/Controller.php
leofeyer: I see.

src/Image/Resizer.php
discordier: We had the same issues in composer-client, we should use normalized pathes everywhere.

leofeyer:
Merged in 9b6694f3fa93ce8a9373d77ea10f510bcba53322.

Issue #261, opened 2015-05-15
Move the ContaoFramework class to its own namespace


leofeyer:
I'd like to discuss moving the Contao framework related resources into their own namespace:

```
src/
    Framework/
        ContaoFramework.php
        FrameworkAwareTrait.php
```

@contao/developers Please vote +1 if you like it.

aschempp:
We've discussed this already. :-1: from me

aschempp:
As a follow up, we can't move the classes in the RC phase, this would be for Contao 5 then…

discordier:
:-1: from me as well. Especially given the fact that we want to remove the framework over the next versions.

leofeyer:
Meanwhile changed through #346.

PR #346, opened 2015-09-15
[RTM] Added a general adapter factory

Toflar:
So this is the first attempt for the general adapter factory we discussed last week.

Open questions:
* [ ] It is in the nature of things that the adapter classes cannot be tested so I cannot write more unit tests for this, really. How do we ensure it's the way we want it to be before merging it? Should @ausi rebase and use it in his Image PR?


src/Adapter/Adapter.php
```diff
@@ -0,0 +1,81 @@
...
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\Adapter;
+
+/**
+ * A general Adapter class building a wrapper around any class that is not
+ * Unit testable/mockable. It only delegates method calls to the specified
```
aschempp: You should describe this is for static methods.

src/Adapter/Adapter.php
```diff
@@ -0,0 +1,81 @@
...
+        $method->setAccessible(true);
+
+        if ($method->isStatic()) {
+
+            return $method->invokeArgs(null, $arguments);
+        }
+
+        $instance = $this->factory->createInstance($this->class, $this->args);
+
+        return $method->invokeArgs($instance, $arguments);
```
aschempp: I guess you wanted to use `$instance` here. However, I disagree with the adapter handling non-static methods.

src/Adapter/AdapterFactory.php
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Creates a new instance of a given class.
+     *
+     * @param string $class Fully qualified class name.
+     * @param array $args Constructor arguments.
+     *
+     * @return mixed
+     */
+    public function createInstance($class, $args = [])
```
leofeyer: It should only be called `create` IMHO.

Toflar: We agreed on `createInstance` on the phone call. That's what my notes said :)

src/Adapter/Adapter.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Calls any method of the given class.
+     *
+     * @param string $name
+     * @param array  $arguments
+     *
+     * @return mixed
+     */
+    public function __call($name, array $arguments = [])
```
leofeyer: Are we sure that `__call()` is triggered if a method is called statically? I thought that's what `__callStatic()` is for.

ausi: The method is not called statically on the adapter.

leofeyer: I see :)

Toflar: That's the whole idea behind this adapter... :D

ausi:
Is there a need for calling non-static protected methods on the Adapter? Didn’t we agree to change them to `public` if we need to call them from outside?

Toflar:
Yeah, wasn't sure. So we could omit `ReflectionMethod` then. So it would simply look like this, right?

```php
public function __call($name, array $arguments = [])
{
    return call_user_func_array($this->class . '::' . $name, $arguments);
}
```

ausi:
Yes, I think so. And `[$this->class, $name]` instead of `$this->class . '::' . $name` IMO.

Toflar:
Not according to http://php.net/manual/en/function.call-user-func-array.php :) Fixed in 159564a539784da7ee1b5089425a22eed7984d04

ausi:
`['MyClass', 'myStaticMethod']` is of type `callable` according to the [PHP documentation](http://php.net/manual/en/language.types.callable.php), but I don’t know if it is the common convention.

src/Adapter/Adapter.php
```diff
@@ -0,0 +1,70 @@
...
+    private $factory;
+
+    /**
+     * Adapter constructor.
+     *
+     * @param string                    $class
+     * @param array                     $args Constructor arguments.
+     * @param AdapterFactoryInterface   $factory
+     */
+    public function __construct($class, $args = [], AdapterFactoryInterface $factory)
```
ausi: `$args` and `$factory` can be removed then.

src/Adapter/AdapterFactory.php
```diff
@@ -0,0 +1,75 @@
...
+    public function getAdapter($class, $args = [])
+    {
+        $cacheKey = md5($class . ';' . json_encode($args));
+
+        if (isset($this->adapterCache[$cacheKey])) {
+
+            return $this->adapterCache[$cacheKey];
+        }
+
+        $this->adapterCache[$cacheKey] = new Adapter($class, $args, $this);
```
ausi: And this method can be smiplified to `$this->adapterCache[$class] = new Adapter($class);`.

Toflar:
Fixed.

src/Adapter/AdapterFactory.php
leofeyer: According to my notes, we agreed on `getInstance` and `getAdapter`. But now that you have named the method `createInstance`, you might as well just use the more common name.

leofeyer: Ah, just realize that it should be `getInstance()`, because the `getAdapter()` method is actually the `create()` method.

leofeyer: So actually, the methods should be called `create()` and `getInstance()` or `getObject()`.

Toflar: No, both wrong. `createInstance` creates one. `getAdapter` gets the adapter and creates one if not yet there. But `getInstance` is certainly wrong because it creates one every time it's called.

leofeyer: Ok, then `createInstance()`. But `getAdapter()` is wrong for sure. Most factory classes I know have a `create()` or a `factor()` method, none of them has a `get*()` method.

aschempp: I'm not saying @leofeyer is wrong, but we discussed this in the call to what @Toflar says. Both are not ideal to me, but we should stick to the things we said in the call (so the current code is correct).

I don't know if one can call `__construct` to create an instance of a class?

ausi:
I added more tests and fixed a bug in Toflar/core-bundle#2

src/Adapter/AdapterFactory.php
ausi: Calling `__construct` doesn’t work. I added a test and a fix in Toflar/core-bundle#2

Toflar:
Merged.

src/Adapter/AdapterFactory.php
```diff
@@ -0,0 +1,72 @@
...
+     * Creates a new instance of a given class.
+     *
+     * @param string $class Fully qualified class name.
+     * @param array $args Constructor arguments.
+     *
+     * @return mixed
+     */
+    public function createInstance($class, $args = [])
+    {
+        $method = '__construct';
```
ausi: @Toflar I forgot to remove this line :)

ausi:
Did I put `LegacyClass` and `LegacySingletonClass` in the right place?

Should we prefix the class names with `Contao\` to prevent the usage of the adapter with non-Contao classes?

aschempp:
> Should we prefix the class names with Contao\ to prevent the usage of the adapter with non-Contao classes?

We discussed that but limiting to the `Contao\` namespace would prevent other developers from using the adapter to unit test their legacy stuff.

src/Adapter/AdapterFactory.php
Toflar: Fixed in b438e2d356a0abe8add90930e2d57bb14a2c5136 :)

Toflar:
Can we get this merged soon so I can work with the develop branch? I want to unit test my components relying on the `Image` class ;-)

ausi:
+1
I need it in #342 too

leofeyer:
I will try.

leofeyer:
We haven't agreed on how to call the methods yet.

**getAdapter()**

A factory is supposed to create a class. It is therefore name `<Class>Factory` and has a `create()` method which returns a `<Class>` object.

This means that the `AdapterFactory` creates an `Adapter` object. However, the method which is executing `new Adapter()` is currently called `getAdapter()`, which is wrong. It should be called `create()`.

** createInstance()**

The naming is correct, however, I don't understand how this method is relevant for the Adapter factory. It is just a general method which tries to instantiate objects. Where is the connection?

aschempp:
Trying to repeat what we discussed in the call:

### getAdapter()

The adapter is only created if it is not already available. Adapters are "shared", similar to the container. It's `$container->get('…')` and not `$container->create('…')` to get a service.

### createInstance()

This is a general factory for all legacy classes. Instead of writing `new FilesModel()` we should do `$adapterFactory->createInstance('FilesModel', […])`. If you create a mock of the adapter factory, you can return a mock of `FilesModel` on that call.

leofeyer:
I know what we have discussed, but we were wrong. Therefore we need to discuss again.

aschempp:
> we were wrong

We're not wrong, we just have different opinions. It would be useful to share yours ;-)

leofeyer:
I have. See https://github.com/contao/core-bundle/pull/346#issuecomment-145798003.

aschempp:
Didn't I just explain why we chose that? Why we have `createInstance()` for example?

leofeyer:
Still the method is not related to the `AdapterFactory` class as I wrote above and therefore does not belong there.

aschempp:
Where else would you put it? A separate service that always needs to be injected too?

leofeyer:
Don't know. Let's discuss it next week on Mumble.

leofeyer:
@ausi and @aschempp have shown me some examples of how to use the new feature during our Mumble call:

```php
// alt
new \Contao\Image('foo.jpg');
// neu
$this->framework->createInstance('Contao\Image', ['foo.jpg'])

// alt
Contao\Image::create('foo.jpg');
// neu
$this->framework->getAdapter('Contao\Image')->create('foo.jpg');

$config = $this->framework->getAdapter('Config')->getInstance();
```

```php
class Foobar
{
    public function test()
    {
        $filesRepo = $this->framework->getAdapter('FilesModel');
        $filesModel = $filesRepo->findByPk(5);
    }

    public function test2()
    {
        $image = $this->framework->createInstance('Contao\Image', […]);
    }
}
```

leofeyer:
As discussed on Mumble on October 14th, 2015, we want to add the methods to the `ContaoFramework` class. To preserve BC, we are going to add a new `ContaoFramework` and `ContaoFrameworkInterface` class in the `Framework` namespace and deprecate the existing ones to be removed in Contao 5.

This will also fix #261.

src/ContaoFramework.php
```diff
@@ -34,7 +36,8 @@
  * @author Dominik Tomasi <https://github.com/dtomasi>
  * @author Andreas Schempp <https://github.com/aschempp>
  *
- * @internal
```
leofeyer: The class is still internal, isn't it?

src/ContaoFramework.php
```diff
@@ -167,6 +176,56 @@ public function initialize()
...
 
     /**
+     * Creates a new instance of a given class.
+     *
+     * @param string $class Fully qualified class name.
+     * @param array $args Constructor arguments.
+     *
+     * @return mixed
+     */
+    public function createInstance($class, $args = [])
```
leofeyer: The new methods should only be added to the new `ContaoFramework` class, shouldn't they?

Toflar:
So I basically rewrote everything. You can pretty much forget about all the discussion that took place before because I think you need to have a fresh mind to check the code again. The basic principles, however, remained the same:

* `ContaoFramework` is now in the `Framework` namespace
* The old `ContaoFramework` class in still exists, but simply extends the new one. This is for BC reasons if somebody did a type hint against the class instead of the interface or instantiated it on their own. Both of which are wrong, so in my opinion we could also remove this class.
* The `ContaoFrameworkInterface` is now in the `Framework` namespace as well. It extends the old one and adds the two new methods `getAdapter()` and `createInstance()`.
* The old `ContaoFrameworkInterface` still exists. This is needed for BC reasons because type hinting against this was the correct way to go.
* The `Adapter` class was renamed to `GeneralAdapter` and is not `final` anymore. I did this with the fact in mind that we might want to provide a `registerAdapter($class, $adapterClass)` method in the framework. This would allow e.g. to provide a special adapter class (instead of the `GeneralAdapter`) for e.g. `System` by calling `$framework->registerAdapter('System', 'Contao\\CoreBundle\\Framework\\Adapters\\SystemAdapter')`. What do you think? 
* I removed all the dependencies on the old `ConfigAdapter` class.
* I fixed all the unit tests.

leofeyer:
> * The old `ContaoFramework` class in still exists, but simply extends the new one. This is for BC reasons if somebody did a type hint against the class instead of the interface or instantiated it on their own. Both of which are wrong, so in my opinion we could also remove this class.

This is wrong for sure, because now the old class also has the new methods. The old class has to remain unchanged and the new class has to extend the old one!

Toflar:
Disagree. The interface defines the methods that have to be present, not the class itself. So it doesn't matter at all if those methods are present in the old one too now. It's better the way I did it because we can just delete the old one without moving things around in Contao 5 or as I pointed out, even now.

src/Framework/ContaoFramework.php
```diff
@@ -0,0 +1,478 @@
...
+     * @param int                       $errorLevel    The PHP error level
+     */
+    public function __construct(
+        RequestStack $requestStack,
+        RouterInterface $router,
+        SessionInterface $session,
+        $rootDir,
+        CsrfTokenManagerInterface $tokenManager,
+        $csrfTokenName,
+        $config = null,
```
aschempp: Defining a default value and have the next parameter expect a value is just wrong. Removing the type hint should be enough to allow `null`

Toflar: What? It must be optional that's why it's `null` by default. I don't get your statement.

aschempp:
Here are my findings:

 1. The `ConfigAdapter` is not BC, and it's not possible to convert it to an adapter. Some of the methods in `Config` are static and some are not, you can't call them on the adapter.
 2. I dislike the idea to support different adapters. There should/must never be a need for it in my opinion.
 3. I thought about moving everything to the new class in my draft but couldn't figure out how. The only problem I see is if people required the `ContaoFramework` class instead of the interface. The service would no longer be an instance of that class. Not sure if that's something we should worry about though.
 4. Because of No. 2, I dislike the `GeneralAdapter` name and I find the sub-namespace unnecessary.
 5. We're changing the constructor arguments for `PrettyErrorScreenListener` which is not `@internal`. Not sure if that's a BC problem?

src/Framework/ContaoFramework.php
aschempp: The argument is not optional, as you must always set the `$errorLevel` variable that follows it…

Toflar: Yeah? so `null, $errorLevel`, where's the problem?

aschempp: Exactly. So why is there a default value on `$config` if you always need to pass `null` to set the error level? I'm not even sure that's supported ;-)

Toflar: What the heck does `null` have to do with the error level????

aschempp: http://php.net/manual/de/functions.arguments.php
Beispiel No 5:

Toflar: I cannot change the order due to BC?

aschempp: I know :D 
But you should not set a default value. Just removing the type hint (as you already did) is enough to allow for null.

leofeyer:
Same problem as in the other ticket: We have already discussed an implementation and now we are starting from scratch again. I'm not against it, because I am interested in the best solution of course, but it means we need to discuss again :(

Toflar:
It's 100% the same as we discussed. The adapter factory is gone in favor of the the ContaoFramework, the ContaoFramework is moved to the Framework namespace and we must keep BC which is what I tried. It's **exactly** what we discussed.

aschempp:
The framework implementation is the same (just moved). But the `Adapter` and `ConfigAdapter` are not …

aschempp:
Btw, can the new interface not contain the same methods as the old one? Which would make it clearer that the old is obsolete and we could simply remove the file later without needing to copy stuff?

Toflar:
> Btw, can the new interface not contain the same methods as the old one? Which would make it clearer that the old is obsolete and we could simply remove the file later without needing to copy stuff?

Yes it can. I could move it.

aschempp:
> Yes it can. I could copy it.

:+1: for that from me then.

Toflar:
> But the Adapter and ConfigAdapter are not …

I had to adjust them because the ConfigAdapter used to have an `initialize()` instead of `getInstance()` which is what you tried to fix with additional methods and I just extended the general adapter. It's the same functionality just better imho :)

Toflar:
> for that from me then.

Done in 532b54500c6609e41b1307d63365afcddf53f0bd.

aschempp:
I agree my solution with the methods was not nice, so +1 for that rewrite. But I would just keep the `ConfigAdapter` as it was, there's no reason to rewrite it to the new class. Just ignore the `$config` parameter to the framework and always create an internal adapter. Then we don't need to discuss potential implementations of a deprecated class ^^

Toflar:
> there's no reason to rewrite it to the new class

Yeah, well. I just wanted to make sure it still works with the new framework class when you inject the old ConfigAdapter but I agree this should never be the case. I'll restore it.

src/Framework/ContaoFramework.php
Toflar: I will, it shouldn't be allowed to override the $config anyway as this would cause `$this->config->getInstance()` to fail if somebody still passes the deprecated `ConfigAdapter`. I'm trying to fix this but there's a phpunit bug which prevents me from fixing it atm :D

aschempp: Whats the problem? It should be as simple as ignoring `$config` And alwasy call `getAdapter()`?

Toflar: Yes but I have to adjust the unit tests to mock the getAdapter() call and there's a bug in phpunit that prevents me from doing that.

Toflar: Done in 340a7eb40b6e077ec5e11c096a3ba91af5b5742d.

Toflar:
Okay, now I'm really done with this!

leofeyer:
Merged in e936124092e71c9dc1a7f8f72147e09ba76badf7.

leofeyer:
See https://github.com/contao/core-bundle/pull/378#issuecomment-151489593 for a follow-up discussion which occurred when working with the new adapters.

PR #348, opened 2015-09-16
Follow symlinks when searching for installed files

aschempp:
I used my draft of the composer plugin to install legacy packages into system modules. However, the .htaccess files were not found due to this missing flag. I can't think of any side effects with enabling this "by default".


Toflar:
:+1: I just stumbled across this issue as well and my solution was the same without knowing from your ticket :)

leofeyer:
Merged in b504e5d3561aef9ea43344eb4faf875b38b63ee4.

PR #349, opened 2015-09-16
Fix DcaExtractor when no database.sql file was found

aschempp:
`$blnFromFile` is set to true when one filed in the DCA does not contain an SQL definition. Then the extractor tried to load files and failed, resulting in the DCA file details not being parsed either.


leofeyer:
Merged in e7d93bf6dc4268d1aa21852a1556e451f249a891.

PR #350, opened 2015-09-25
[WIP] InsertTags rework

Toflar:
First of all, this is `WIP` and far from complete but I would like to make this `RFC` so we can discuss about the general issues and approaches before I continue to work on this.
### Prologue (why?)

Contao has a very poor caching mechanism at the moment as it only supports caching per page and not per module/content element aka controller. Obviously now that we have Symfony under the hood, we can leverage ESI to improve caching of partials. One of the main problems we have is that we have an `OutputFromCacheListener` which always initialises the framework to check for server side caches and to replace InsertTags on it. As long as we do have the current InsertTag handling, we will always have to boot the framework and replace those for every request.
### My findings

So there are actually several problems here. First of all, I think we should move the responsibility for caching settings from the user to the developers. Developers should think about caching and performance whenever they develop a piece of code right from the beginning and they will know if and how their e.g. front end module or content element can be cached and when those caches can/must be invalidated. There's absolutely no point in providing caching settings in the page settings. The solution to this is to consider every part of a single page as a controller. Contao actually provides a very good foundation for this because there are basically four ways to include content on a page:
- Article
- Front end module
- Content element
- InsertTags

So if all of them are rendered using ESI, every single one of them can define how long it can be cached on its own. Symfony provides superbe integration and a simple ESI proxy so we can use this.

My Pull Request focusses on InsertTags only for now. To leverage ESI we have to move from e.g. `{{email::test@test.com}}` which now returns an encoded `<a href="mailto:test@test.com>test@test.com</a>` to return something like `<esi:include src="http://domain.com/_controller:replaceInsertTags........">`.

This means that our InsertTags have to move from returning a replaced string only to a proper `Response` as they're now controllers. The same applies for our InsertTag flags as e.g. `refresh` would basically translate to a `Response` with `Cache-Control: private` header.

So I set out to rewrite everything. I've invented interfaces for both, the InsertTags as well as the InsertTag flags and I wrote collections and a compiler pass to register them easily so that we can provide a way how one should write InsertTags and flags from now on and just keep both a `LegacyInsertTag` and `LegacyInsertTagFlag` to keep compatibility. Then I thought about having at least two parsers, one responsible for parsing the InsertTag to a string (`StringParser`) and one to an ESI include tag (`EsiParser` not coded yet). 

But there is a big issue with the old hooks for both, InsertTags as well as flags because they now get parameters we will likely not be able to pass etc. Not quite sure how to proceed here and it's a lot of work so I would like to have your opinion about the issues and how we should proceed. I think converting our 4 ways to include content to controllers and leverage ESI and proper HTTP caching will give Contao 4 an incredible performance boost but rewriting the whole InsertTag handling was probably a bit ambitious :-D


aschempp:
I think this is a topic we need to discuss in an upcoming meeting as it's pretty complex. However, I would love to have the reverse proxy available in Contao 4.1, which would be rater simple if we can work around the non-caching insert tag in the output buffer. @contao/developers what do you think?

leofeyer:
@Toflar Can we close this ticket?

Toflar:
It is one of many PR's trying to bring a new way to define modules, content elements and also insert tags to Contao. As it seems like we've finally found a way, we might think about porting the same concept to insert tags too one day. I'll close it, as I have the branch on my fork anyway.

PR #353, opened 2015-10-01
[WIP] Added backwards compatibility for Database_Result etc.

aschempp:
Some (old) extensions still rely on classes like `Database_Result`, especially for method arguments. In Contao 3, the `class_alias` method was called directly in the class file, but this would not allow to trigger a deprecated warning.

### DIFF
--- a/src/Resources/contao/config/ide_compat.php
+++ b/src/Resources/contao/config/ide_compat.php
@@ -40,7 +40,6 @@ namespace {
 	class BackendFile extends \Contao\BackendFile {}
 	class BackendHelp extends \Contao\BackendHelp {}
 	class BackendIndex extends \Contao\BackendIndex {}
-	class BackendInstall extends \Contao\BackendInstall {}
 	class BackendMain extends \Contao\BackendMain {}
 	class BackendPage extends \Contao\BackendPage {}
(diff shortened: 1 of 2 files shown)


Toflar:
1. Isn't this actually a bug fix?
2. Shouldn't the deprecated warning say you should use the Doctrine DBAL `default_connection` service instead and stop type hinting against any Contao `Database` class or subclass anyway?

aschempp:
1. it sure is a bugfix
2. dunno about that. It would require to rewrite everything, whereas "my" hint would require to change one word only…

I'm still investigating this though, I do have a system where this problem occurs but somehow the patch here did not fix it (don't know why yet).

leofeyer:
> 1. it sure is a bugfix

So, which bug does it fix? :)

aschempp:
The BC break!

leofeyer:
The `Database_Result` class has already been deprecated in Contao 3 and was removed in Contao 4. We certainly won't add a BC layer for Contao 2, will we?

aschempp:
Well if that's true then we don't need this at all…

aschempp:
Closing this. Code should be updated, as @leofeyer pointed out we don't need BC for Contao 2 :D

PR #355, opened 2015-10-04
[WIP] Rewrite of the module wizard

aschempp:
Follow-up on https://github.com/contao/core/issues/7972

As discussed here's a rewrite of the module wizard. I decided to rewrite the Widget from scratch and use a Twig template. This also required some helper functions, so there's probably a lot to discuss.

The old widget obviously needs to remain in the namespace, but we could remove it's functionality and simply use a class alias / `x extends y`. I don't see a BC break here.

@contao/developers let me know your thoughts :-)


aschempp:
No feedback on the topic?

aschempp:
This certainly needs updates. Got

> Using an instance of "Twig_Function_Method" for function "tabindex" is deprecated. Use Twig_SimpleFunction instead.

^^

leofeyer:
As discussed in Mumble on April 28th, this change is a mixture between legacy code and new code, which we cannot really befriend with.

Issue #302, opened 2015-06-15
Generate a random token on installation


aschempp:
See https://github.com/Incenteev/ParameterHandler/issues/78 about a solution with the IncenteevParameterHandler

leofeyer:
What do you mean?

aschempp:
The default kernel secret is `ThisTokenIsNotSoSecretChangeIt`. We should replace the default with a random value so the user does not need to type something manually (but still can if necessary).

leofeyer:
As discussed on Mumble on August 13th, this is a good idea.

leofeyer:
Will you be working on this issue for Contao 4.1 or shall I change the milestone to 4.x.x?

aschempp:
see #358

PR #358, opened 2015-10-09
[RTM] Added ScriptHandler method to generate random secret

aschempp:
This adds the necessary script handler to generate the random secret for https://github.com/contao/core-bundle/issues/302

The random secret must only be generated if the `parameters.yml` file does not exist. This is because the `IncentiveeParametersHandler` will update/overwrite existing parameters if a respective environment variable is available.

Using this method means the user is not asked for a random secret on installation, but it can obviously still be changed in the `parameters.yml`.

### DIFF
--- a/src/Composer/ScriptHandler.php
+++ b/src/Composer/ScriptHandler.php
@@ -42,6 +42,28 @@ class ScriptHandler
     }
 
     /**
+     * Sets environment variable for random secret on installation.
+     *
+     * @param Event $event The event object
+     */
+    public static function generateRandomSecret(Event $event)
+    {
+        $extra = $event->getComposer()->getPackage()->getExtra();
+
+        if (!isset($extra['incenteev-parameters']) || !isset($extra['incenteev-parameters']['file'])) {
+            return;
+        }
+
+        $file = $extra['incenteev-parameters']['file'];
+
+        if (is_file($file)) {
+            return;
+        }
+
+        putenv('CONTAO_RANDOM_SECRET=' . md5(uniqid(mt_rand(), true)));
+    }
+
+    /**
      * Executes a command.
      *
      * @param string $cmd   The command
--- a/src/Composer/ScriptHandler.php
+++ b/src/Composer/ScriptHandler.php
@@ -50,17 +50,11 @@ class ScriptHandler
     {
         $extra = $event->getComposer()->getPackage()->getExtra();
 
-        if (!isset($extra['incenteev-parameters']) || !isset($extra['incenteev-parameters']['file'])) {
+        if (!isset($extra['incenteev-parameters']) || !static::canGenerateSecret($extra['incenteev-parameters'])) {
             return;
         }
 
-        $file = $extra['incenteev-parameters']['file'];
-
-        if (is_file($file)) {
-            return;
-        }
-
-        putenv('CONTAO_RANDOM_SECRET=' . md5(uniqid(mt_rand(), true)));
+        putenv('CONTAO_RANDOM_SECRET=' . bin2hex(random_bytes(32)));
     }
 
     /**
@@ -90,5 +84,31 @@ class ScriptHandler
         if (!$process->isSuccessful()) {
             throw new \RuntimeException('An error occurred while executing the "' . $cmd . '" command.');
         }
+    }
+
+    /**
+     * Validates that we can generate a random secret.
+     * We need to make sure at least one file is defined in the config but none of the files exist.
+     *
+     * @param array $config
+     *
+     * @return bool
+     */
+    private static function canGenerateSecret(array $config)
+    {
+        // "incenteev-parameters" config can be an array of files.
+        if (!isset($config['file'])) {
+            $result = false;
+
+            foreach ($config as $v) {
+                if (is_array($v) && isset($v['file']) && !is_file($v['file'])) {
+                    $result = true;
+                }
+            }
+
+            return $result;
+        }
+
+        return !is_file($config['file']);
     }
 }
--- a/src/Composer/ScriptHandler.php
+++ b/src/Composer/ScriptHandler.php
@@ -21,6 +21,8 @@ use Symfony\Component\Process\Process;
  */
 class ScriptHandler
 {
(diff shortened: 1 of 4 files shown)


aschempp:
See https://github.com/contao/standard-edition/pull/24/files for the necessary changes in the root `composer.json` file

src/Composer/ScriptHandler.php
```diff
@@ -42,6 +42,28 @@ public static function generateSymlinks(Event $event)
...
+            return;
+        }
+
+        $file = $extra['incenteev-parameters']['file'];
+
+        if (is_file($file)) {
+            return;
+        }
+
+        putenv('CONTAO_RANDOM_SECRET=' . md5(uniqid(mt_rand(), true)));
```
leofeyer: Is this secure? Other processes might be able to read the environment variables, too.

discordier:
Only when executed within the same shell process and therefore belong to the same user. IMO we can assume no malicious process to be running here. If there would be, reading the environment will be the least significant problem.

leofeyer:
>> `putenv('CONTAO_RANDOM_SECRET=' . md5(uniqid(mt_rand(), true)));`

> Is this secure? Other processes might be able to read the environment variables, too.

@ausi What's your opinion on this one?

Toflar:
@stof Can you maybe shed some light on this? Is this the way to go or is there a security issue with this? The goal is to provide a convenience method to auto generate a secret. You can obviously still adjust it.

src/Composer/ScriptHandler.php
stof: you can never change the environment of a parent process. So this environment variable will only be accessible to the current process (running Composer) and subprocesses launched from this process (subsequent composer scripts, or ``git`` commands used by the installer)

stof: not however that your implementation does not generate a strong random secret though

Toflar: Of course not, it's just convenience ;)

aschempp: @stof thanks for the feedback! Any suggestion for a stronger random secret?

leofeyer: We should probably use the Symfony security component to generate the token.

stof:
I see a drawback here though. The Incenteev ParametersHandler considers that the environment variable contains the uptodate version of the parameter, and will replace the value coming from an existing parameters.yml file (the goal is to be able to update env variables to change the config for the next deployment, without having to delete the file).
This means that your parameter set based on your ``CONTAO_RANDOM_SECRET`` env variable will be updated for each composer run (otherwise you have a big issue with the randomness of your random secret). And this can cause many issues. For instance, changing the secret used by the remember_me token in Symfony would cause all existing remember_me tokens to become invalid (as their signature won't match using the new secret). The secret should generally not change for each deployment.

src/Composer/ScriptHandler.php
stof: @aschempp ``random_bytes()`` on PHP 7, and the polyfill for PHP 5 (Symfony is deprecatign its SecureRandom class in 2.8 in favor of this new API)

aschempp:
The variable is only set if no `parameters.yml` exists, so it will never be updated ;-)

leofeyer:
As discussed on Mumble on October 14th, we want to use `random_bytes` with the [compatibility layer](https://github.com/paragonie/random_compat).

src/Composer/ScriptHandler.php
Toflar:
```php
bin2hex(random_bytes(32));
```

Read about the polyfill but forgot about it again, thx for the hint @stof :)

stof: be careful. bin2hex does not exist in PHP 5.3

bytehead: True. The requirement from `composer.json` is anyway `"php": ">=5.4.0",`.

ausi: `bin2hex` is there since PHP 4 IMO, `hex2bin ` is 5.4 :)

aschempp:
PR is updated

src/Composer/ScriptHandler.php
```diff
@@ -69,4 +85,30 @@ function ($type, $buffer) use ($event) {
...
+     */
+    private static function canGenerateSecret(array $config)
+    {
+        // "incenteev-parameters" config can be an array of files.
+        if (!isset($config['file'])) {
+            $result = false;
+
+            foreach ($config as $v) {
+                if (is_array($v) && isset($v['file']) && !is_file($v['file'])) {
+                    $result = true;
```
leofeyer: Why don't you simply `return true;` here?

aschempp: Because there could be a second file that does exist (and then we must not set the variable)

leofeyer: Then you should probably return `false` if the condition matches.

leofeyer:
Where did you enable the compat layer? You have to include the `lib/random.php` file at some point, haven't you?

src/Composer/ScriptHandler.php
leofeyer: What I mean is:

```php
foreach ($config as $v) {
    if (is_array($v) && isset($v['file']) && is_file($v['file'])) {
        return false; // if there is at least one existing file, we must not generate a secret, so return false
    }
}

return true;
```

Or am I missing something here?

stof:
> Where did you enable the compat layer? You have to include the lib/random.php file at some point, haven't you?

This is handled when loading the composer autoloader, thanks to the config of the package

src/Composer/ScriptHandler.php
aschempp: You're partially right. But we also don't want to generate the variable if there is no file defined in the array, where in your example we would return `true`

leofeyer: Ok, thanks for checking. The PR is RTM to me then.

aschempp:
Ok, I have updated the file check again based on @leofeyer's input and added full unit tests for the new method.

leofeyer:
Merged in 2c2fea1dad449c9b5f59b92b90b45f30dcbee505.

PR #361, opened 2015-10-14
Make sure the security user is from Contao

aschempp:
We must not handle the session for non-Contao users in Symfony.

### DIFF
--- a/src/EventListener/UserSessionListener.php
+++ b/src/EventListener/UserSessionListener.php
@@ -12,6 +12,7 @@ namespace Contao\CoreBundle\EventListener;
 
 use Contao\BackendUser;
 use Contao\FrontendUser;
+use Contao\User;
 use Doctrine\DBAL\Driver\Connection;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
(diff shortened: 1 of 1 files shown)


leofeyer:
Should be:

```php
if (!$this->hasUser() || !$this->isContaoMasterRequest($event)) {
    return;
}

$userObj = $this->getUserObject();

if (!$userObj instanceof User) {
    return;
}

$session = $this->getUserObject()->session;
…
```

aschempp:
Would obviously be the same ;-)

leofeyer:
No, my version only has one `getUserObject()` call, yours has two.

aschempp:
I did read "shouldn't it be" instead of "should be" ^^
Changed it now.

bytehead:
:+1:

leofeyer:
You didn't remove it from line 64 though :)

src/EventListener/UserSessionListener.php
```diff
@@ -84,6 +85,10 @@ public function onKernelResponse(FilterResponseEvent $event)
 
         $user = $this->getUserObject();
 
+        if (!$this->getUserObject() instanceof User) {
```
bytehead: why not `if (!$user instanceof User) {` ?

leofeyer: See https://github.com/contao/core-bundle/pull/361#issuecomment-148026917

aschempp:
Sorry, my fault ^^. Fixed it

bytehead:
It's still not what @leofeyer wants to have ^^

leofeyer:
Merged in b196d1b0b353276207cc0a13ba9a873e0c2841ed.

bytehead:
:+1:

PR #362, opened 2015-10-14
Lower priority of pretty error screen listener

aschempp:
The Twig listener is actually in priority `-128` (see https://github.com/symfony/symfony/blob/2.8/src/Symfony/Bundle/TwigBundle/Resources/config/twig.xml#L153 / https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpKernel/EventListener/ExceptionListener.php#L75)

The security component is handling user access redirects and other stuff on priority 0, which is prevented by our current priority `32`.

/cc @bytehead

### DIFF
--- a/src/Resources/config/listener.yml
+++ b/src/Resources/config/listener.yml
@@ -23,11 +23,14 @@
 # Exception listener
 #
 # The priority of the exception listeners must be higher than the one of the
-# Twig exception listener (defaults to 0).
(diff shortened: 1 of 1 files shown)


leofeyer:
Merged in 7c6232b8c443fc641361853ffb21245325720fd1.

bytehead:
Thanks! :+1:

Issue #321, opened 2015-07-29
Validator::isEmail() follow-up


leofeyer:
This is a follow-up ticket to contao/core-bundle#313.

@discordier It seems we do not have to IDNA encode at all if we use `filter_var($email, FILTER_VALIDATE_EMAIL) === $email`. Should we replace the "url" validation with `filter_var()` as well?

leofeyer:
> It seems we do not have to IDNA encode at all

Ok, I was wrong about this. Still the question about converting the "url" regex remains.

discordier:
I am unsure but thought about it before. I am currently unaware if Al new tld validate and what changes to this filter have been made since php 5.3.
For Contao 4 I'd say yes but I think we will most likely not back port it to Contao 3

leofeyer:
There is a general problem with both `FILTER_VALIDATE_EMAIL` and `FILTER_VALIDATE_EMAIL`: they do not support Unicode characters.

```php
// ASCII e-mail
filter_var("john@smith.com", FILTER_VALIDATE_EMAIL); // john@smith.com

// Unicode e-mail
filter_var("jöhn@smith.com", FILTER_VALIDATE_EMAIL); // false

// ASCII URL
filter_var("https://www.facebook.com/pages/page/156324134399840", FILTER_VALIDATE_URL); // https://www.facebook.com/pages/page/156324134399840

// Unicode URL
filter_var("https://www.facebook.com/pages/Наша-Україна/156324134399840", FILTER_VALIDATE_URL); // false
```

Our current `Validator::isValidUrl()` method supports Unicode URLs, pretty much due to contao/core#6402 I guess.

@contao/developers What shall we do?

aschempp:
I suppose PHP expects IDNA-decoded values?

leofeyer:
The local part of an e-mail address is never IDNA encoded.

leofeyer:
As discussed on Mumble on October 14th, we should keep our existing URL validation and revert the changes regarding the e-mail validation. We explicitly want to support Unicode characters in the local name.

leofeyer:
I have added the regex used by the PHP `FILTER_VALIDATE_URL`:

```php
[...]

PR #367, opened 2015-10-15
[RFC] Support unicode characters in the local part of an e-mail address

leofeyer:
This is the follow-up PR for #321.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Version 4.1.0-beta1 (2015-XX-XX)
 --------------------------------
 
 ### New
+Added a more accurate e-mail validation with unicode support (see #367).
+
+### New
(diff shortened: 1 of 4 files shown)


ausi:
The regular expression is very complex, but if the source is trusted it should be OK.

AFAIK the old regular expression didn’t allow `"`, `<` or `>`, if that’s the case we should check if the emails get correctly escaped everywhere.

leofeyer:
The source is trusted – it is the regex used by [PHP itself](https://github.com/leofeyer/php-src/blob/master/ext/filter/logical_filters.c#L601). I'm not 100% sure about my modifications, although the unit tests say that they are correct. Also see my PR here: php/php-src#1577

leofeyer:
Merged in 76b64fe8e57dc10b1667818cd5e5c2a402c010a4.

PR #369, opened 2015-10-16
[RFC] Add more information when throwing PageNotFoundException

aschempp:
Related to https://github.com/contao/core/issues/8009

### DIFF
--- a/src/Resources/contao/controllers/FrontendIndex.php
+++ b/src/Resources/contao/controllers/FrontendIndex.php
@@ -66,7 +66,7 @@ class FrontendIndex extends \Frontend
 		elseif ($pageId === false)
 		{
 			$this->User->authenticate();
-			throw new PageNotFoundException('Page not found');
+			throw new PageNotFoundException('Page not found. (URL: "' . \Environment::get('base') . \Environment::get('request') . '", Referer: "' . \Environment::get('httpReferer') . '")');
 		}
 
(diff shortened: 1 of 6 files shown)


leofeyer:
The purpose of contao/core#8009 was to get rid of some log entries, not to double them by logging them in `tl_log` and the `app/logs/prod.log` :)

aschempp:
I know. Now you simply have to remove `$this->log()` and the full message will be preserved ;-)

aschempp:
I have rebased this on the latest develop branch.

leofeyer:
Since we are already logging the exceptions, we should remove the `$this->log()` calls entirely IMHO. But let's discuss it before you spend more time working on it.

leofeyer:
As discussed in Mumble on April 28th, we want to add the additional information such as `url` and `referer` to the log context.

leofeyer:
Sadly, it does not work. The [line formatter](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Formatter/LineFormatter.php#L136) does not consider any additional data that the extension might have. So we are back at adding the context to the exception message, but that does not look too good:

Toflar:
Since when is the formatting of an exception relevant?

leofeyer:
Implemented as discussed in 0f8f51b5950ce7af4fa2812de9a547650a356c83.

PR #373, opened 2015-10-21
Correctly use the driver to check for enabled lock

aschempp:
The app config can be adjusted to use another driver, so we should not check for the file directly.


leofeyer:
Merged in 36da0df21a55011bbdb300c10f9cb16f2e18ada9.

PR #374, opened 2015-10-22
[RFC] Extracting the request token handling from the framework

Toflar:
Here's the first draft. Open questions:

* See my `@todo`. This was copied from the original source but shouldn't this target route be different?
* Same question as we had several times: Am I allowed to change the constructor of `ContaoFramework`? We really need to make a decision here once and for all.
* Is the naming for `contao.request_token.routes` okay?
* Where should the default settings for `contao.request_token.routes` go?
* Does the listener have to have any special priority?
* Isn't removing the request handling from the `ContaoFramework` a BC break? It automatically checked in Contao 4 but doesn't in Contao 4.1 anymore unless you enable it for your route...

And of course the unit tests are missing but I'll add them once we got the questions sorted.

### DIFF
--- a/src/EventListener/RequestTokenListener.php
+++ b/src/EventListener/RequestTokenListener.php
@@ -1,0 +1,126 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\Exception\AjaxRedirectResponseException;
+use Contao\CoreBundle\Exception\InvalidRequestTokenException;
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+
+/**
+ * Validates Contao request tokens.
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+class RequestTokenListener
+{
+    /**
+     * @var ContaoFrameworkInterface
+     */
+    private $framework;
+
+    /**
+     * @var array
+     */
+    private $routeNames;
+
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * @var CsrfTokenManagerInterface
+     */
+    private $tokenManager;
+
+    /**
+     * @var string
+     */
+    private $csrfTokenName;
+
+    /**
+     * Constructor.
+     *
+     * @param ContaoFrameworkInterface  $framework
+     * @param array                     $routeNames
+     * @param RouterInterface           $router
+     * @param CsrfTokenManagerInterface $tokenManager
+     * @param string                    $csrfTokenName
+     */
+    public function __construct(
+        ContaoFrameworkInterface $framework,
+        array $routeNames,
+        RouterInterface $router,
+        CsrfTokenManagerInterface $tokenManager,
+        $csrfTokenName
+    ) {
+        $this->framework = $framework;
+        $this->routeNames = $routeNames;
+        $this->router = $router;
+        $this->tokenManager = $tokenManager;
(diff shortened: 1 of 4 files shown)


src/EventListener/RequestTokenListener.php
```diff
@@ -0,0 +1,126 @@
...
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        if (!$this->framework->isInitialized()) {
+            return;
+        }
+
+        $request = $event->getRequest();
+
+        if (null === $request
+            || 'POST' !== $request->getRealMethod()
```
aschempp: Maybe we should use [`Request::isMethodSafe()`](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Request.php#L1468) so we also protect `PUT` etc.?
If not we should at least use `Request::isMethod('POST')`

src/EventListener/RequestTokenListener.php
```diff
@@ -0,0 +1,126 @@
...
+
+        throw new InvalidRequestTokenException('Invalid request token. Please reload the page and try again.');
+    }
+
+    /**
+     * Sets the REQUEST_TOKEN constant
+     */
+    private function setConstant()
+    {
+        // Deprecated since Contao 4.0, to be removed in Contao 5.0
```
aschempp: should probably use `@deprecated` on the method instead of a comment?

aschempp:
Looks pretty good to me. The only problem I see is that the request token check is no longer done if the framework is initialized later than the kernel request.

Where will the `routeNames` come from?

Toflar:
> Where will the routeNames come from?

That's part of my questions??

src/EventListener/RequestTokenListener.php
Toflar: For a private method nobody can use? Yeah, right...

aschempp:
I think `contao.request_token.routes` will not work, because it's an array. It can't be extended by third-party bundles. Why not use route attributes for that (similar to the scope)?

src/EventListener/RequestTokenListener.php
leofeyer: PhpStorm will only mark the method as deprecated with the `@deprecated` annotation.

src/EventListener/RequestTokenListener.php
```diff
@@ -0,0 +1,126 @@
...
+        $this->setConstant();
+
+        $token = new CsrfToken($this->csrfTokenName, $request->request->get('REQUEST_TOKEN'));
+
+        if ($this->tokenManager->isTokenValid($token)) {
+            return;
+        }
+
+        if ($request->isXmlHttpRequest()) {
+            // @todo is contao_backend really the right solution here?
```
leofeyer: Certainly not, especially because the request might be a front end request.

leofeyer:
> * Same question as we had several times: Am I allowed to change the constructor of `ContaoFramework`? We really need to make a decision here once and for all.

Yes, you are. First, the class is marked `@internal` and second, we have just added a new class for Contao 4.1, which we can change as much as we want until the RC release. And then the class is still marked as `@internal` :)

> * Is the naming for `contao.request_token.routes` okay?
> * Where should the default settings for `contao.request_token.routes` go?

I'd strongly prefer a route attribute.

> * Does the listener have to have any special priority?

I don't think so.

> * Isn't removing the request handling from the `ContaoFramework` a BC break? It automatically checked in Contao 4 but doesn't in Contao 4.1 anymore unless you enable it for your route...

It is not a BC break. We have not had any routes other than Contao routes in Contao 3; but now we have and so we have to adjust the request handling.

src/EventListener/RequestTokenListener.php
Toflar: Yeah but that's what we currently have in the develop branch. It was not my invention that's why I ask :-D

Toflar:
> It is not a BC break. We have not had any routes other than Contao routes in Contao 3; but now we have and so we have to adjust the request handling.

But we had them in 4.0.

leofeyer:
Sure, and that is the problem in the first place. We should have handled it in Contao 4.0 already, but since we did not think about it, we are now fixing it.

leofeyer:
As discussed on Mumble, the request token check should only be activated if there is a route attribute (to be added). It must not be a separate listener though.

Toflar:
Follow up on https://github.com/contao/core-bundle/pull/389.

PR #375, opened 2015-10-23
[RFC/WIP] Fix Environment::get('script')

bytehead:
I stumpled upon this issue, while using the `contao/core-bundle` as a Symfony bundle and running the application on the built-in php server via `php app/console server:start` on `127.0.0.1:8000`. It does not happen on a normal vhost-configuration with apache / nginx.

My url: `http://127.0.0.1:8000/kontakt/agbs.html`

[Contao\Environment.php:450](https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/library/Contao/Environment.php#L450): returns `string '' (length=0)` instead of `kontakt/agbs.html`. In this case `static::get('script')` and `static::get('request')` return exact the same (`kontakt/agbs.html` and that gets replaced with `''`). 

`static::get('script')` should return the script name, not the request (in this case it would be `app_dev.php` on my machine).

If you use `http://127.0.0.1:8000/app_dev.php/kontakt/agbs.html` it actually works correctly and returns `kontakt/agbs.html`.

I tried to fix this by using `pathinfo()` to get the script name. I've also added an `IndexFreeEnvironmentTest` class, which runs well. I also tested it on normal vhost configuration within a subdirectory and without.

I'm not sure, if it's the correct way to fix this issue. I even don't know, if there is some introduced issues now. I didn't test it under `cgi-fcgi` and `fpm-fcgi` environments.

Comments welcome - I would like to use Contao during the development on the PHP built-in webserver.

### DIFF
--- a/src/Resources/contao/library/Contao/Environment.php
+++ b/src/Resources/contao/library/Contao/Environment.php
@@ -425,7 +425,7 @@ class Environment
 	 */
 	protected static function script()
 	{
-		return preg_replace('/^' . preg_quote(static::get('path'), '/') . '\/?/', '', static::get('scriptName'));
+		return preg_replace('/^' . preg_quote(static::get('path'), '/') . '\/?/', '', pathinfo(static::get('scriptFilename'), PATHINFO_BASENAME) ?: static::get('scriptName'));
 	}
 
 
--- a/tests/Contao/IndexFreeEnvironmentTest.php
+++ b/tests/Contao/IndexFreeEnvironmentTest.php
@@ -1,0 +1,203 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\Test\Contao;
+
+use Contao\CoreBundle\Test\TestCase;
+use Contao\Environment;
+use Contao\System;
+
+/**
+ * Tests the Environment class.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class IndexFreeEnvironmentTest extends TestCase
+{
+    /**
+     * {@inheritdoc}
+     */
+    public static function setupBeforeClass()
+    {
+        Environment::reset();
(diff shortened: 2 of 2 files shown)


aschempp:
Your explanation sounds like a bug in the PHP server? Does the PHP server support the `.htaccess` rules? Or how does it know to server from `app.php` ?

bytehead:
When I start the webserver with `php app/console server:start` it runs in the dev-environment. You could force it to run in prod-environment with `php app/console server:start --env=test`.

bytehead:
For completion: it has the same behaviour in the prod-environment as in dev.

aschempp:
Does the Symfony request handle it correctly? What do you get for `$request->getBaseUrl()` or `$request->getBasePath()` or the fragment name?

aschempp:
So the most simple solution would be to use Symfony, right? If they both return the same result?

bytehead:
That would be a way I prefer ;-)
I try to come up with some refactoring.

leofeyer:
Does this mean we have to map more methods in the `Environment` class to the Symfony request?

aschempp:
It would make sense to me instead of working around issues with "new" code in "old" classes…

leofeyer:
This actually seems to be an issue with the built-in web server. Even Symfony returns the "wrong" value without the `app_dev.php` fragment:

```php
// http://127.0.0.1:8000/text-elements.html
echo $this->container->get('request_stack')->getCurrentRequest()->getScriptName(); // /text-elements.html

// http://127.0.0.1:8000/app_dev.php/text-elements.html
echo $this->container->get('request_stack')->getCurrentRequest()->getScriptName(); // /app_dev.php
```

I am not familiar with the built-in web server, but since it automatically rewrites the request to `app.php`, I guess it can be configured somehow.

leofeyer:
The configuration is done here: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Bundle/FrameworkBundle/Resources/config/router_dev.php

leofeyer:
See https://github.com/symfony/symfony/pull/16352.

bytehead:
Thanks for investigating!

PR #376, opened 2015-10-23
[RTM] Solution for event listeners

aschempp:
The solution is so simple, I can't believe we did not come up with it earlier:
`System::import()` and `System::importStatic()` simply allow to pass a service name as class name.

**Example for hooks:**
```php
// config.php
$GLOBALS['TL_HOOKS']['replaceInsertTags'][] = ['app.event_listener.foo', 'replaceInsertTags'];

// foo service
public function replaceInsertTags(/** ... */) {
    // just a regular function with the original arguments
}
```

This solves all issues with passing by reference etc. It also automatically enables event listener for callbacks in `DC_Table` as well as any third-party extensions that use the *correct* way of calling a hook.

**Other effects / advantages:**
In unit tests, I can now set a service called `Database` into the mock container. Which means that mocked services can be used for everyone who does `System::importStatic()`. I think this will be a great advantage to unit test legacy stuff.

Replaces #204

### DIFF
--- a/src/Resources/contao/library/Contao/System.php
+++ b/src/Resources/contao/library/Contao/System.php
@@ -152,7 +152,18 @@ abstract class System
 
 		if ($blnForce || !isset($this->arrObjects[$strKey]))
 		{
-			$this->arrObjects[$strKey] = (in_array('getInstance', get_class_methods($strClass))) ? call_user_func(array($strClass, 'getInstance')) : new $strClass();
+			if (static::getContainer()->has($strClass))
+			{
+				$this->arrObjects[$strKey] = self::getContainer()->get($strClass);
+			}
+			elseif (in_array('getInstance', get_class_methods($strClass)))
+			{
+				$this->arrObjects[$strKey] = call_user_func(array($strClass, 'getInstance'));
+			}
+			else
+			{
+				$this->arrObjects[$strKey] = new $strClass();
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


discordier:
Now the configuration compiler is missing which compiles tagged services into an ad-hoc populated TL_HOOKS array and you should be done.

aschempp:
I thought about this, but it won't work. You can't register hooks without the framework being booted. And you can't *correctly* sort the hooks etc.

aschempp:
The hooks need to be registered. It does not really matter if you do that in YAML or the existing `config.php`

leofeyer:
I like the idea a lot. We would still dispatch normal events for every **new** "hook" that we add, right?

aschempp:
Absolutly

leofeyer:
@bytehead /cc

leofeyer:
This is `RTM` now, isn't it? The title still says `RFC`.

aschempp:
There wasn't much feedback, but to me it's RTM…

leofeyer:
I have noticed that you are using the concept in your insert tag PRs, that's why I'm asking :)

bytehead:
Sorry for the delay, I didn't had the time to study this PR. But it looks good to me :+1:

leofeyer:
Merged in 4d4818b1a24dd7cfa0eeb19f434e0275ca23521c.

PR #378, opened 2015-10-24
[RTM] Use the RequestToken class to validate the Contao request token

leofeyer:
As discussed on Mumble, we should use the `RequestToken` class to validate the Contao request token to restore support for `disableRefererCheck`, `BYPASS_TOKEN_CHECK` and `requestTokenWhitelist`. This PR includes the necessary changes.

### DIFF
--- a/src/Framework/ContaoFramework.php
+++ b/src/Framework/ContaoFramework.php
@@ -17,14 +17,13 @@ use Contao\CoreBundle\Exception\AjaxRedirectResponseException;
 use Contao\CoreBundle\Exception\IncompleteInstallationException;
 use Contao\CoreBundle\Exception\InvalidRequestTokenException;
 use Contao\Input;
+use Contao\RequestToken;
 use Contao\System;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Routing\RouterInterface;
-use Symfony\Component\Security\Csrf\CsrfToken;
-use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 
 /**
  * Initializes the Contao framework.
@@ -47,11 +46,6 @@ class ContaoFramework implements ContaoFrameworkInterface
     private $router;
 
     /**
-     * @var CsrfTokenManagerInterface
-     */
-    private $tokenManager;
-
-    /**
      * @var SessionInterface
      */
     private $session;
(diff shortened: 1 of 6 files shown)


Toflar:
Why did you not use the adapter here now?

leofeyer:
For what? The class is not injected, therefore we cannot mock it. It would be cool though, if I could inject the adapter, but that's not possible, is it?

leofeyer:
I mean this:

```yml
services:
    contao.framework.request_token_adapter:
        class: Contao\CoreBundle\Framework\Adapter
        public: false
        arguments:
            - "Contao\\RequestToken"
```

But as far as I remember, this is exactly the use case we do **not** want, right?

aschempp:
You should alway inject the `ContaoFramework` and call `getAdapter()`!

leofeyer:
This is something we should discuss.

aschempp:
That was the whole point of adapters ???

leofeyer:
Maybe we were wrong. I had to remove the test for the `ContaoFramework::validateRequestToken()` method, because I could not mock the `RequestToken` class. Injecting the adapter would solve the problem.

This might be an edge case though, which only applies to the `ContaoFramework` class though.

aschempp:
What was the problem? Your test should mock the ContaoFramework to return a mock adapter (if necessary) ;-)

leofeyer:
How is that supposed to work when testing the `ContaoFramework` class itself? Am I supposed to inject the `ContaoFramework` class into the `ContaoFramework` class???

aschempp:
Hmm, no that won't work. I guess we can't test the request token *inside* the framework class. But we can test the new listener?

leofeyer:
> I guess we can't test the request token inside the framework class.

We could though, if we injected the adapter.

aschempp:
Dunno if I like that idea… guess we should really discuss it

leofeyer:
Thank you.

leofeyer:
I know what we have discussed, but now that I have worked with the new feature and experienced problems, I think it should be all different.

 * We should use `new Adapter()` instead of `$framework->getAdapter()`.
 * We should completely remove the `getAdapter()` and `createInstance()` methods.
 * Instead, we should use `$instance = (new Adapter())->createInstance()` if we need an instance.

This way we would have to inject each adapter separately instead of injecting the framework, which however has the **advantage** that if one of the old classes is replaced, we only have to exchange the adapter with the new service.

aschempp:
> if one of the old classes is replaced, we only have to exchange the adapter with the new service.

why do you mean by that?

leofeyer:
As discussed on Mumble, the current solution with the `ContaoFramework` class being the adapter factory is the best and we want to keep it.

leofeyer:
Merged in 8df561f509efb0e25212bbae37dc475d15c6fc7e.

Issue #244, opened 2015-05-04
Use the kernel.terminate event for the cronjobs


leofeyer:
Handle the registered cronjobs upon the `kernel.terminate` event (see contao/core#6976).

leofeyer:
See #380.

PR #380, opened 2015-10-26
[RFC] Use the kernel.terminate event for the command scheduler

leofeyer:
This is the follow-up ticket for #244. It includes the necessary changes to trigger the command scheduler upon the `kernel.terminate` event instead of using an Ajax request.

### DIFF
--- a/src/EventListener/CommandSchedulerListener.php
+++ b/src/EventListener/CommandSchedulerListener.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\Config;
+use Contao\CoreBundle\Controller\FrontendController;
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+/**
+ * Registers the Contao front end and back end session bags.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class CommandSchedulerListener
+{
+    use ContainerAwareTrait;
+
+    /**
+     * @var ContaoFrameworkInterface
+     */
+    private $framework;
+
+    /**
+     * Constructor.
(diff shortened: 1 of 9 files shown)


src/EventListener/CommandSchedulerListener.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\Config;
+use Contao\CoreBundle\Controller\FrontendController;
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+/**
+ * Registers the Contao front end and back end session bags.
```
aschempp: :see_no_evil:

aschempp:
I don't think we should use the controller to trigger the action. Is there a reason for that?

leofeyer:
We need the controller to trigger the command scheduler via a real cron job, so I figured the simplest solution would be to just run the controller. I have tried it with a sub-request first, but I did not get it to work. Maybe you will?

aschempp:
Well it should actually be some sort of service. Then we could have a controller or a console command or the terminate event…

leofeyer:
It's legacy code. We should not waist too much time refactoring it.

aschempp:
Valid point, but then we should probably trigger the `FrontendCron` legacy controller directly

leofeyer:
Ok. I'll change it.

leofeyer:
Changed. Much better now. :+1:

aschempp:
This: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/controllers/FrontendCron.php#L32-L33 might cause problems…

leofeyer:
Why?

aschempp:
Because the variables will already be set when running the terminate event?

leofeyer:
You must be right, still it works on my system (website and unit tests).

leofeyer:
Changed in 0b11661aee5bd4d434e6acd15763f15955ab71b4.

aschempp:
Probably because the errors are "silently ignored" because the response has been sent. But that depends on your server (CGI) etc.

leofeyer:
Merged in c0edaa828a035f38e24836aead62a4b92a8e0d69.

PR #381, opened 2015-10-26
[RTM] Allow to configure the image dir and if to cache images

aschempp:
This allows to define a custom image caching path in the application configuration.

Additionally, I can define per environment if I want images to be cached. My problem is that currently the dev environment does not cache images at all. I have > 50 images on one page, which makes using the dev controller almost unbearable.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -21,6 +21,21 @@ use Symfony\Component\Config\Definition\ConfigurationInterface;
 class Configuration implements ConfigurationInterface
 {
     /**
+     * @var bool
+     */
+    private $debug;
+
+    /**
+     * Constructor.
+     *
+     * @param bool $debug If kernel debug mode is enabled
+     */
+    public function __construct($debug)
+    {
+        $this->debug = (bool) $debug;
+    }
+
+    /**
      * Generates the configuration tree builder.
      *
      * @return TreeBuilder The tree builder
@@ -54,6 +69,12 @@ class Configuration implements ConfigurationInterface
                         })
                         ->thenInvalid('%s')
                     ->end()
+                ->end()
+                ->booleanNode('image_cache')
+                    ->defaultValue(!$this->debug)
+                ->end()
+                ->scalarNode('image_dir')
+                    ->defaultValue('assets/images')
                 ->end()
                 ->scalarNode('csrf_token_name')
                     ->cannotBeEmpty()
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -101,6 +101,7 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->children()
                 ->arrayNode('image')
+                    ->addDefaultsIfNotSet()
                     ->children()
                         ->booleanNode('bypass_cache')
                             ->defaultValue($this->debug)
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -10,6 +10,7 @@
 
 namespace Contao\CoreBundle\DependencyInjection;
 
+use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 
@@ -70,12 +71,6 @@ class Configuration implements ConfigurationInterface
                         ->thenInvalid('%s')
                     ->end()
                 ->end()
-                ->booleanNode('image_cache')
-                    ->defaultValue(!$this->debug)
-                ->end()
-                ->scalarNode('image_dir')
-                    ->defaultValue('assets/images')
-                ->end()
                 ->scalarNode('csrf_token_name')
                     ->cannotBeEmpty()
                     ->defaultValue('contao_csrf_token')
@@ -91,6 +86,31 @@ class Configuration implements ConfigurationInterface
             ->end()
         ;
 
+        $this->addImageSection($rootNode);
+
         return $treeBuilder;
+    }
+
+    /**
+     * Configures the `contao.image` section.
+     *
+     * @param ArrayNodeDefinition $rootNode
+     */
+    private function addImageSection(ArrayNodeDefinition $rootNode)
+    {
+        $rootNode
+            ->children()
+                ->arrayNode('image')
+                    ->children()
+                        ->booleanNode('bypass_cache')
+                            ->defaultValue($this->debug)
+                        ->end()
+                        ->scalarNode('target_dir')
+                            ->defaultValue('assets/images')
+                        ->end()
+                    ->end()
+                ->end()
+            ->end();
+        ;
     }
 }
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -106,7 +106,7 @@ class Configuration implements ConfigurationInterface
                         ->booleanNode('bypass_cache')
                             ->defaultValue($this->debug)
                         ->end()
-                        ->scalarNode('target_dir')
+                        ->scalarNode('target_path')
                             ->defaultValue('assets/images')
                         ->end()
                     ->end()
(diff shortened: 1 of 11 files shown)


src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -45,6 +46,17 @@ public function getAlias()
     /**
      * {@inheritdoc}
      */
+    public function getConfiguration(array $config, ContainerBuilder $container)
+    {
+        // Will add the resource to the container
+        parent::getConfiguration($config, $container);
+
+        return new Configuration($container->getParameter('kernel.debug'));
```
leofeyer: What's this for?

leofeyer:
Don't know if I like this. The purpose of the "dev" mode actually is not to use the cache.

src/DependencyInjection/ContaoCoreExtension.php
aschempp: To pass kernel parameters to the configuration (see http://symfony.com/doc/current/cookbook/configuration/using_parameters_in_dic.html)

aschempp:
I disagree. The Symfony dev container is cached too, and only rebuilt if some of the resources change. But that's obviously all configurable, that's what dependency configuration is for ;-)

Toflar:
I very much like this :+1:

leofeyer:
Why do we need to define a custom image cache path? What's wrong with `assets/images`?

aschempp:
In my current platform, I use deployment releases. The images should be shared across releases to prevent rebuilding hunderts of images after each release. The assets folder is automatically generated by the Contao installer, therefore it can't be shared across releases.

leofeyer:
Then we'd also have to make the CSS and JS folder configurable, wouldn't we?

aschempp:
We should make everything configurable now that we have a configuration container, but that's a ongoing task ;-)

leofeyer:
Well, I would prefer to keep `assets/css`, `assets/images` and `assets/js`. Convention over configuration.

discordier:
Big thumbs up for making everything configurable.

leofeyer:
I have read through [the docs](http://symfony.com/doc/current/cookbook/configuration/using_parameters_in_dic.html) and I now understand `ContaoCoreExtension:54`. However, I still think it is wrong to add the configuration files as resources. I have neither seen this done in Symfony.

aschempp:
Maybe @bytehead knows?

bytehead:
I don't know if it's best practice. I've seen it rarely before in bundles as well.
@leofeyer: what's your misgiving on adding the config files as resources?

leofeyer:
@bytehead If it were right, why does Symfony not do it? Do you know any bundles doing it?

aschempp:
Here's a manual about the topic: http://symfony.com/doc/current/components/config/caching.html

I know why it must not be done, and I have updated my PR: The `YamlFileLoader` already does it for us: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php#L46

bytehead:
Ah, good catch! I was searching for this :smile:

leofeyer:
The tests are still failing though :)

aschempp:
PR is rebased and unit tests are fixed. Also added a new test for the missing piece in the `Configuration` class.

src/DependencyInjection/Configuration.php
```diff
@@ -55,6 +70,12 @@ public function getConfigTreeBuilder()
                         ->thenInvalid('%s')
                     ->end()
                 ->end()
+                ->booleanNode('image_cache')
```
leofeyer: I don't like the naming of the variable. What about `rebuild_image_cache` or `recreate_images` or `bypass_image_cache`?

src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -45,6 +46,17 @@ public function getAlias()
     /**
      * {@inheritdoc}
      */
+    public function getConfiguration(array $config, ContainerBuilder $container)
+    {
+        // Will add the resource to the container
+        parent::getConfiguration($config, $container);
```
leofeyer: I though you had updated your PR? Why is this line still in then?

leofeyer:
A quick search for `assets/images` returned more than 20 results. You have to adjust those occurrences as well if the image path is configurable.

src/DependencyInjection/ContaoCoreExtension.php
aschempp: That's still necessary. We don't need to add the YAML resources, but still need to add the `Configuration` class.

src/DependencyInjection/Configuration.php
aschempp: I was thinking about adding a sub-namespace for images. Will be useful for @ausi's PR too…

aschempp:
I have renamed the parameters, because I think it will be more future-proof, especially with https://github.com/contao/core-bundle/pull/342. 
Let me know what you think, then I can update the other places where `assets/images` is hardcoded.

src/DependencyInjection/Configuration.php
```diff
@@ -70,6 +86,32 @@ public function getConfigTreeBuilder()
...
+     * @param ArrayNodeDefinition $rootNode
+     */
+    private function addImageSection(ArrayNodeDefinition $rootNode)
+    {
+        $rootNode
+            ->children()
+                ->arrayNode('image')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->booleanNode('bypass_cache')
```
leofeyer: No hard feelings, but this ain't any better. I suggested `bypass_image_cache`, because the flag is only bypassing the image cache. We also have an internal cache, a search results cache, a page cache, a script cache and the Symfony cache. Unless you are planning to bypass them all if the flag is enabled, the naming is too unspecific.

aschempp: Did you notice the `image` array node? The configuration will look like this:

```yaml
contao:
    image:
        bypass_cache: true
        target_dir: "foo/bar"
    prepend_locale: false
```

The container parameter will be `contao.image.bypass_cache`.

leofeyer: I did not notice it, sorry.

src/DependencyInjection/Configuration.php
```diff
@@ -70,6 +86,32 @@ public function getConfigTreeBuilder()
...
+                            ->defaultValue($this->debug)
+                        ->end()
+                        ->scalarNode('target_dir')
```
leofeyer: I would prefer `cache_dir` here.

src/DependencyInjection/ContaoCoreExtension.php
leofeyer: Can you point me to the Symfony docs where this is explained?

aschempp: http://symfony.com/doc/current/components/config/caching.html

The point is to execute the original method (https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/DependencyInjection/Extension/Extension.php#L91) which will not call the constructor itself but only add the resource.

leofeyer:
@aschempp If `cache_dir` instead of `target_dir` is ok with you, you could continue to work on this PR. We need to release the RC version soon (this week), so time presses.

aschempp:
I called it that way because that's in the `Image` class: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/Image.php#L264

It should actually be `target_path` …

leofeyer:
No, that's something else. You can specify a target path in the image class if you do not want the image to be created in the cache directory.

aschempp:
It's the target path where to place the image. Same as for us, ours is just the default. I don't like to call it `cache_xxx` because it's not a cache, it's where the resized images are placed. Disabling cache with the flag here would not prevent the files from being placed in the so-called `cache_dir`, which can't be correct.

leofeyer:
True.

aschempp:
Shall I change the variable to `target_path` ?

leofeyer:
No. We need to discuss :(

leofeyer:
My comment should have been in another ticket.

leofeyer:
As discussed on Mumble on November 5th, the setting shall be called `target_path`.

aschempp:
Variable renamed and all paths update.

leofeyer:
Merged in cde2b02f45fa28b2c435ebf269ea30478c74c706.

PR #383, opened 2015-10-26
Make sure Contao is installed before warming cache

aschempp:
I had trouble to install Contao because composer is running the scripts and `cache:clear` by default. If the database does not yet contain Contao, it will result in an exception that `tl_page` is not found.

Not sure if this should even go into 4.0…

### DIFF
--- a/src/Cache/ContaoCacheWarmer.php
+++ b/src/Cache/ContaoCacheWarmer.php
@@ -18,6 +18,7 @@ use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
 use Contao\DcaExtractor;
 use Contao\PageModel;
 use Doctrine\DBAL\Driver\Connection;
+use Doctrine\DBAL\Exception\TableNotFoundException;
 use Symfony\Component\Config\FileLocator;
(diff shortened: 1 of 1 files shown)


bytehead:
:+1:

leofeyer:
The only method which uses the database is `getLanguagesInUse()`. The check should only be executed there, because everything else will build nicely even if the installation has not been completed yet.

leofeyer:
Do you want to update your PR or do you want me to change it? I'd do something like this:

```php
try {
    $statement = $this->connection->prepare($query);
    $statement->execute();
} catch (TableNotFoundException $e) {
    return array('en');
}
```

leofeyer:
I wonder why we are not adding `en` by default in the `getLanguagesInUse()` method. Since it is always loaded as fallback language, we should always build the language cache.

aschempp:
Not true. We're also using `PageModel`

leofeyer:
You are right.

leofeyer:
Merged in ef20d495387ff00a4d78f2d1722cdd5f23460adf.

leofeyer:
The tests are failing :(

PR #384, opened 2015-10-26
Fixed Contao considering case in file search and the whole path

Toflar:
Two fixes here:

* When I search for `dsc` it did not find `61891_DSC.jpg` because the `preg_match` was case-sensitive.
* The `preg_match` was executed on the absolute path instead of the relative path.

### DIFF
--- a/src/Resources/contao/widgets/FileSelector.php
+++ b/src/Resources/contao/widgets/FileSelector.php
@@ -405,7 +405,7 @@ class FileSelector extends \Widget
 
 				if ($for != '')
 				{
-					if (!preg_match('/' . str_replace('/', '\\/', $for) . '/', $folders[$f] . '/' . $v))
+					if (!preg_match('/' . str_replace('/', '\\/', $for) . '/', $folders[$f] . '/i' . $v))
 					{
 						--$countFiles;
 					}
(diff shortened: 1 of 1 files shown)


leofeyer:
Can you please also adjust the `DC_Folder::generateTree()` method?

leofeyer:
As discussed on Mumble, two things need to be adjusted:

 * The file manager should have filter menus (files, folders, both).
 * The search menu of the file and page picker needs to be adjusted to return only resources matching the name. To open a found folder, it should be clickable as in the file manager or page manager.

leofeyer:
> The file manager should have filter menus (files, folders, both).

We were wrong about this, because what we want is not a filter but a search modifier. The difference is that the filter can also be applied without a search query (e.g. show only JPG images) and the search modifier not.

Also, a type based filter (show only files or show only folders) is not possible with a tree structure.

aschempp:
> Also, a type based filter (show only files or show only folders) is not possible with a tree structure.

It would only be possible with a search, right? And *only folders* would be possible, but not *only files*

leofeyer:
Only folders would be possible.

leofeyer:
Here is the corresponding PR: https://github.com/contao/core-bundle/pull/392

Toflar:
**Searching** and **Displaying** are two completely different things. Of course I can search for files only but obviously the path they are contained in (so the parent folders) need to be shown.

leofeyer:
That's not logically at all. In fact, **every** folder would be shown when "filtering" for files, except empty folders.

leofeyer:
As @aschempp said, only folder filtering would make sense.

leofeyer:
Closed in favor of #392.

PR #387, opened 2015-10-27
PHP7 compatibility

leofeyer:
See http://php.net/manual/en/migration70.incompatible.php

### DIFF
--- a/src/Framework/ContaoFramework.php
+++ b/src/Framework/ContaoFramework.php
@@ -416,7 +416,7 @@ class ContaoFramework implements ContaoFrameworkInterface
     {
         if (isset($GLOBALS['TL_HOOKS']['initializeSystem']) && is_array($GLOBALS['TL_HOOKS']['initializeSystem'])) {
             foreach ($GLOBALS['TL_HOOKS']['initializeSystem'] as $callback) {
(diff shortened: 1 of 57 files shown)


leofeyer:
Take the following example:

```php
System::importStatic($callback[0])->$callback[1]();
```

We have to options to fix this:

```php
// Requires PHP 5.5+ (Contao 4.1)
System::importStatic($callback[0])->($callback[1])();

// Compatible with PHP 5.4+ (Contao 3.5)
System::importStatic($callback[0])->{$callback[1]}();
```

In Contao 4.1, we could use the new syntax. However, I don't know if we care about PHP7 compatibility for Contao 3.5?

discordier:
PHP 5.6 Support timeline: Support until 28 Aug 2016 (security 28 Aug 2017)
Contao 3.5 Support timeline: 01 Jun 2018 (security 01 Jun 2019)

So yes, we must support PHP7 for Contao 3.5

leofeyer:
:+1:

leofeyer:
PHP7 support for Contao 3.5 has been added in contao/core@5cc52edd9291784907ba1806101bb95f6ef253dc.

PR #389, opened 2015-10-28
[RTM] Feature/disable request token route attribute

Toflar:
I'm using the adapters for the `RequestToken` class too now so I can properly mock everything. This allowed me to adjust the tests for valid and invalid tokens which are now properly tested.

Also, I implemented the convenience method for `mockContaoFramework()` so you can easily tell what you would like to get when you call `getAdapter()`:

```php
$expectedAdapters = [
    'Contao\RequestToken' => $requestTokenMock,
    'OtherAdapter' => $otherAdapterMock
];

$this->mockContaoFramework(null, null, $expectedAdapters);
```

After that, I was able to implement the new feature so we can enable the request token check via route/request settings.


leofeyer:
First glance looks good to me. Are there any open questions?

Toflar:
Not for me.

leofeyer:
Then you should change the title to `RTM` I guess :)

bytehead:
Really nice - we just stuck here with the request token because of a datatrans response :+1:

aschempp:
I would prefer an attribute name that is related to Contao. `_contao3_request_token_check` is a bit long but just as a hint… Also, in Symfony it's not a *request token* but a *csrf token*, right? (there's also *user token*)

leofeyer:
How about `_contao_csrf`?

Toflar:
> How about _contao_csrf?

The word "enable" is missing there for me.

aschempp:
Other options:
 - `_contao_token`
 - `_validate_contao_token`
 - `_contao_csrf_validation`
 - `_contao_csrf_protection`

As *CSRF* stands for *Cross-Site-Request-Forgery*, `_contao_csrf` does not describe that we're protecting against it…

leofeyer:
Neither do all of your suggestions :)

My favorite would be `_contao_csrf_protection`, although it is quite longish. Why does `_csrf_protection` not suffice? This is how the setting is called in Symfony.

leofeyer:
IMHO we should create a listener to validate the request token and make the listener check if the Contao 3 framework is booted and the request attribute is set. This will solve the problem with the framework only booting once as well as the master and sub request difficulty. And it even fulfills @discordier's wish that it shall be possible to disable the check programmatically.

aschempp:
I'm repeating myself:
The framework can be bootet at any point, for example in a Controller. For legacy reasons, that's the only place the request token should be checked, because it must not be checked if the framework is not booted.

Just a quick reminder regarding the framework having state: It already has, `isInitialized()` is also a state in that sense ;-)

leofeyer:
> because it must not be checked if the framework is not booted.

That's why the listener checks `$framework->isInitialized()`.

aschempp:
> That's why the listener checks $framework->isInitialized().

Which will not be true if the framework is booted in a controller but the listener checks in `kernel.request`

leofeyer:
The listener can be called at any time before the response is sent. It could be `kernel.controller` or even `kernel.response`.

aschempp:
No, we also discussed that. The post data would already be processed by the time `kernel.response` is executed, and that's wrong.

leofeyer:
True.

leofeyer:
Then the listener must not check the framework and just rely on the route attribute.

aschempp:
But the `REQUEST_TOKEN` only applies to the Contao 3 framework/code…

leofeyer:
The `AddToSearchIndex` and `OutputFromCache` listeners also only apply to the Contao 3 framework and still we have created listeners.

leofeyer:
As discussed on Mumble on November 5th, we eventually want the route attribute and to keep the check in the `ContaoFramework` class. Also, we want the feature to be _disabled_ by default (3 pros, 2 cons) and it shall be called `_token_check`.

Toflar:
Adjusted.

bytehead:
Looks good to me!
Even that we ditch *security first*, because the contao request token should not be enabled by default on non contao routes.

leofeyer:
> because the contao request token should not be enabled by default on non contao routes

@bytehead They would not have been. The check is only executed if the Contao 3 framework is initialized, therefore non Contao routes would never be affected.

The only problem is when you have a non Contao route and you boot the framework to do something; then the request token would be validated. On the other hand, when you boot the framework in one of your routes, does it not make it a Contao route? :)

aschempp:
strip-trouver-le-nom-de-variable-english650-final

leofeyer:
I have seen this too today ^^

leofeyer:
Merged in 7bb9430945415b897a616321f819d2601c3ab560.

bytehead:
@leofeyer That's actually true as well. But just because I get some contao frontend modules (the reason for booting the framework) as kind of *partials* on a non-contao route doesn't make the route a route that should be handled by contao imho. But the configurable solution now perfectly covers that need :)
Thanks to @Toflar for the PR!

discordier:
@bytehead that's exactly what I was always saying. :)

PR #391, opened 2015-10-29
Fix buffered query result

qzminski:
@aschempp and me received an error message when using `app/console cache:warmup`

```
[Doctrine\DBAL\Exception\DriverException]
An exception occurred while executing 'SELECT tl_page.* FROM tl_page WHERE tl_page.type='root'':
SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.
```

### DIFF
--- a/src/Cache/ContaoCacheWarmer.php
+++ b/src/Cache/ContaoCacheWarmer.php
@@ -332,7 +332,7 @@ class ContaoCacheWarmer implements CacheWarmerInterface
     private function isCompleteInstallation()
     {
         try {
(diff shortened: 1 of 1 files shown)


leofeyer:
I have fixed this already in 7cee14d5a2097e95e59325a79268c91384672a1b.

aschempp:
:+1:

Issue #393, opened 2015-10-29
Use config param upload_path upon installation


DanielSchwiperich:
The installation routine creates some empty dirs like the 'files'  folder.

The files folder can be overwritten via the `upload_path` param in the config.yml
`InstallCommand::addEmptyDirs()` is using an array where the entry for the upload path is fixed and not using the config param.

```
    private $emptyDirs = [
        'files',
        'system',
        'templates',
        'web/system',
    ];
``` 

I think `InstallCommand::addEmptyDirs()` should merge in the param from configuration class and drop it from the fixed array definition.

aschempp:
:+1: 

Applies to the `files` path only though because the other folders are only there for backwards compatibility and their name is unchangeable.

leofeyer:
Closed in favor of #394.

PR #394, opened 2015-10-29
[WIP] Use config param upload_path upon installation #393

DanielSchwiperich:
remove 'files' path from $emptyDirs Array
merge upload_path into addEmptyDirs process
refs #393

**This only is for the core-bundle part. See below**

The install.php at the `contao/standard-edition` does not load the config.yml and so it will always fall back to 'files' during web installation.

Running `php app/console contao:install` works now as expected and creates the upload_path dir configured in config.yml

### DIFF
--- a/src/Command/InstallCommand.php
+++ b/src/Command/InstallCommand.php
@@ -92,11 +92,22 @@ class InstallCommand extends AbstractLockedCommand
      */
     private function addEmptyDirs()
     {
-        $this->emptyDirs[] = $this->getContainer()->getParameter('contao.upload_path');
-
-        foreach ($this->emptyDirs as $path) {
+        foreach ($this->getEmptyDirs() as $path) {
             $this->addEmptyDir($this->rootDir . '/' . $path);
         }
+    }
+
+    /**
+     * Get static and config based empty dirs
+     *
+     * @return array $emptyDirs
+     */
+    private function getEmptyDirs()
+    {
+        $emptyDirs = $this->emptyDirs;
+        $emptyDirs[] = $this->getContainer()->getParameter('contao.upload_path');
+
+        return $emptyDirs;
     }
 
     /**
--- a/src/Command/InstallCommand.php
+++ b/src/Command/InstallCommand.php
@@ -40,7 +40,6 @@ class InstallCommand extends AbstractLockedCommand
      * @var array
      */
     private $emptyDirs = [
-        'files',
         'system',
         'templates',
         'web/system',
@@ -93,6 +92,8 @@ class InstallCommand extends AbstractLockedCommand
      */
     private function addEmptyDirs()
     {
+        $this->emptyDirs[] = $this->getContainer()->getParameter('contao.upload_path');
+
         foreach ($this->emptyDirs as $path) {
             $this->addEmptyDir($this->rootDir . '/' . $path);
         }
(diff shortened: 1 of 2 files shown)


aschempp:
You should not modify the service variable because it would be added twice on a subsequent call. How about adding the variable in the constructor only once?

aschempp:
Ah constructor does not work because the container is added using setter injection. I guess you should create a method variable then…

DanielSchwiperich:
Ok I see. Would cause no problem since in the addEmptyDir Method the path is checked for existence

Instead of a method variable I then would like to use a getter Method for reusability in the future. Even if it's a bit overkill for private access.
Otherwise there could be the `files` part completely missing when accessing emptyDirs or you've to double access the container config in every method.

```
/**
     * Adds the empty directories.
     */
    private function addEmptyDirs()
    {
        foreach ($this->getEmptyDirs() as $path) {
            $this->addEmptyDir($this->rootDir . '/' . $path);
        }
    }

    /**
     * Get static and config based empty dirs
     *
     * @return array $emptyDirs
     */
    private function getEmptyDirs()
    {
        $emptyDirs = $this->emptyDirs;
        $emptyDirs[] = $this->getContainer()->getParameter('contao.upload_path');

        return $emptyDirs;
    }
```

aschempp:
looks like a good plan to me :+1:

DanielSchwiperich:
Updated and did a test run on a docker contao instance

The testsuite also needs to be updated it looks like. Shall I do this?

leofeyer:
> The testsuite also needs to be updated it looks like. Shall I do this?

Yes please, otherwise I cannot merge the PR.

DanielSchwiperich:
I feel the need to explain my commit on the unit test 34c59aa

I first only wanted to add the param via
`$container->setParameter('contao.upload_path', $this->getRootDir() . '/files');`
but it doesn't seemed right to me (would have been the easiest way to make the test work)

the app/console part also has the config available and is currently only triggered via composer.
So I added the `ContaoCoreExtension` Part to be able to test against the default setting for `upload_path` and also if the setting is overwritten.

I'm still in the early learning phase with symfony so hopefully I did not something dumb.

Not related to the PR but what is the prefered contribution workflow for PR if the target branch (develop) has changed in the meantime?
This time I did nothing, (lucky no updated files do conflict as far as I see)
Should I have merged the upstream, or rebased, ... ?

aschempp:
I do understand your thoughts and I almost agreed. But then I noticed that you are now testing the `ContaoCoreExtension` in your code. Each test class should only test one class, which means it would be totally ok to assume the container parameter is present and correct…

aschempp:
Regarding the develop branch, you CAN rebase and force-push, but that's only useful if a lot happened and there would be merge conflict. It does not harm though…

aschempp:
Unfortunately time is running out for this in Contao 4.1. I have created a new PR at #400 and applied changes we discussed in the Core team.

DanielSchwiperich:
I was just on it and did the fix on the tests regarding your comments. 
had to rebase because there were changes made to the InstallCommand and Test Classes

Sorry for the delay from me.I'm on vacation this week

leofeyer:
@DanielSchwiperich No problem. Thank you for your help.

I'm closing this ticket in favor of #400.

Issue #320, opened 2015-07-28
Add route for 'contao/main.php?act=error'


aschempp:
This should be available as a route, e.g. `backend_error` which probably links to `/contao/error`

leofeyer:
This would also require to add a `BackendError` controller, wouldn't it?

aschempp:
Obviously, every route has a controller. But I don't think we need a controller in the legacy code?

leofeyer:
Then you cannot display the back end. Do you want to use a generic error page?

aschempp:
You mean because of the HTML5 template?

leofeyer:
No, I mean because of all the variables which are set in the `BackendMain` controller.

aschempp:
Which ones would affect this? I mean if we have a custom route, it does nothing but show the error information, right? Except for the `be_main` template with logout actions etc. Dunno if we need to keep that (but that was your question, right?) :D

leofeyer:
> but that was your question, right?

Exactly :D

aschempp:
Both would be ok to me. I wonder if we need an error page actually… because we should throw exceptions instead of redirecting to an error page…

aschempp:
Also see https://github.com/contao/core/issues/5851

leofeyer:
As discussed on Mumble, we should add a `BackendException` class and throw it together with the log message. Our exception and pretty error listeners should then handle the exception.

leofeyer:
Here's the proof of concept: #395

Please let me know if this is the way to go, then I'll complete the PR.

leofeyer:
Closed in favor of #395.

PR #395, opened 2015-10-30
[WIP] Use exceptions in the back end

leofeyer:
This is the PR for #320. Right now it is just a proof of concept to see if this is the way to go. You can test it by editing a member and manually entering an invalid ID in the URL.


aschempp:
Looks fairly good to me. I don't really like the `BackendException` name, because it's very unspecific. The error is also not necessarily a bad request, it could also be denied access or similar.

Maybe we can come up with something general, like a pretty error screen that shows the exception message if a backend user is logged in but does not for frontend users?

leofeyer:
You are right, the `BackendException` is only an example. We should probably have four classes to handle 400 (bad request), 403 (access denied), 404 (not found) and 500 (internal server error).

aschempp:
Hmm, these exceptions already exist? I would rather simply show the exception message (in our pretty error screen template) if a backend user is logged in. That could be useful in multiple cases.

leofeyer:
I am trying to use the same exceptions in both the back end and the front end. But now the system tries to load the custom 403 and 404 error pages from the site structure in the back end.

How would you determine if we are in the back end or front end in the first place? Using the deprecated scope?

leofeyer:
@aschempp How would you determine if we are in the back end or front end in the listener?

aschempp:
The only way I can think of is the security token or firewall name/scope.

aschempp:
`$container->get('security_context')->getToken()->getUser() instanceof BackendUser`

(freely out of my head). Dunno if we have access to the firewall "name"

aschempp:
and even the name could be changed, so dunno if there are firewall attributes…

leofeyer:
I have updated the PR. I think this is the way to go, don't you think?

src/Resources/contao/dca/tl_article.php
```diff
@@ -536,23 +534,13 @@ public function checkPermission()
...
-							$error = true;
-						}
+						throw new Contao\CoreBundle\Exception\AccessDeniedException('Not enough permissions to ' . Input::get('act') . ' ' . (strlen(Input::get('id')) ? 'article ID ' . Input::get('id') : ' articles') . ' on page ID ' . $id . ' or to paste it/them into page ID ' . $id . '.');
 					}
 				}
 			}
 		}
-
-		// Redirect if there is an error
-		if ($error)
```
aschempp: You should remove the `$error` variable too

aschempp: Btw. wasn't there a reason this was handled later instead of instantly redirecting?

aschempp:
Here are the things I don't like:

aschempp:
I don't like that you're using the same template to render an exception and to render the general error. It makes the template much more complex for no reason, and it also breaks existing template overrides.
It would be much better to add a `backend` template and not run through all the code to try to find matching strings and then ignore them because an exception is set in the template…

src/Resources/contao/dca/tl_article.php
leofeyer: Nope.

aschempp: ^^

leofeyer:
You must be kidding me! 6 days ago, you came up with the idea to use one template for both the front end and the back end (I had a `backend.html.twig` template in my initial draft):

> like a pretty error screen that shows the exception message if a backend user is logged in but does not for frontend users

And now you do not like it anymore?

> I don't like that you're using the same template to render an exception and to render the general error.

aschempp:
I'm sorry, but I said it's incorrect to use the same exception! Not the same template…

leofeyer:
Where did you say this?

aschempp:
> Looks fairly good to me. I don't really like the BackendException name, because it's very unspecific. The error is also not necessarily a bad request, it could also be denied access or similar.

leofeyer:
This comment is pretty outdated. There is no more `BackendException` class.

aschempp:
Well it answered your question ;-)

leofeyer:
So, are there any real problems with the PR?

src/Resources/views/Error/backend.html.twig
```diff
@@ -0,0 +1,9 @@
+{% extends "@ContaoCore/Error/layout.html.twig" %}
+{% block title %}
+    {{ statusName }}
+{% endblock %}
+{% block matter %}
+    <p>{{ exception }}</p>
+{% endblock %}
+{% block howToFix %}{% endblock %}
```
aschempp: You should simply override the `main` block in the parent template

src/Resources/views/Error/layout.html.twig
```diff
@@ -136,17 +136,17 @@
...
                             {% block main %}
                                 {% set _block = block('matter') %}
-                                {% if _block is not empty %}
+                                {% if _block|trim is not empty %}
                                     <h2>{{ error.matter }}</h2>
                                     {{ _block|raw }}
                                 {% endif %}
                                 {% set _block = block('howToFix') %}
-                                {% if _block is not empty %}
+                                {% if _block|trim is not empty %}
```
aschempp: Then these changes are no longer necessary

src/EventListener/PrettyErrorScreenListener.php
```diff
@@ -130,13 +139,24 @@ private function handleException(GetResponseForExceptionEvent $event)
                 break;
 
             default:
-                $this->logException($exception);
-                $this->renderTemplate('error', 500, $event);
-                break;
+                $this->renderErrorScreenByException($event);
```
aschempp: These changes are not really part of this PR, I'm unable to tell what the changes will cause. I think it will try to load language strings which is not intended here, but I'm not sure. Maybe @xtra needs to have a look.

discordier:
@aschempp I'm discordier over here. ;)

src/Resources/views/Error/layout.html.twig
leofeyer: They are not, but they should be applied anyway. It is best practice to trim the return value in conjunction with `not empty`.

src/EventListener/PrettyErrorScreenListener.php
leofeyer: They are part of the PR. We are now using the `renderErrorScreenByException()` method as fallback.

leofeyer:
Merged in 6083fd981e49e320f335d2f24403a60a563f7aaa.

PR #397, opened 2015-11-02
[RTM] Load the random_compat library

aschempp:
see https://github.com/contao/standard-edition/pull/27 for the complementary PR

Unfortunately, Composer does not autoload the `files` when the scripts are executed. This results in `random_bytes` not being available to our random secret generator. We have to manually load the necessary library…

### DIFF
--- a/src/Composer/ScriptHandler.php
+++ b/src/Composer/ScriptHandler.php
@@ -56,7 +56,9 @@ class ScriptHandler
             return;
         }
 
-        putenv(static::RANDOM_SECRET_NAME . '=' . bin2hex(random_bytes(32)));
+        if (static::loadRandomCompat($event)) {
+            putenv(static::RANDOM_SECRET_NAME . '=' . bin2hex(random_bytes(32)));
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Composer/ScriptHandler.php
```diff
@@ -109,4 +111,37 @@ private static function canGenerateSecret(array $config)
...
+        if (null === $package) {
+            return false;
+        }
+
+        $path     = $composer->getInstallationManager()->getInstaller('library')->getInstallPath($package);
+        $autoload = $package->getAutoload();
+
+        if (!empty($autoload['files'])) {
+            foreach ($autoload['files'] as $file) {
+                /** @noinspection PhpIncludeInspection */
```
leofeyer: You must not push IDE specific tags.

aschempp: Removed it

leofeyer:
This approach works, but it seems kind of cumbersome to me. We could also use a different method to generate the secret if `random_bytes` is not available, couldn't we?

aschempp:
none if which is secure… and the token is all about security.

leofeyer:
That's not all true. We could use `openssl_random_pseudo_bytes()` – the OpenSSL extension is already required for the extension repository and the Live Update, therefore chances are high that it is available.

aschempp:
https://paragonie.com/blog/2015/07/how-safely-generate-random-strings-and-integers-in-php

> Update: openssl_random_pseudo_bytes() might not be a great solution either!

leofeyer:
Ok.

leofeyer:
Fixed in 66af4420dba33c0014f7e331532a7022ad2d4770.

aschempp:
Why did you not merge my PR. Your implementation will result in a potential exception during installation, instead of simply not generating a random token if not possible.

leofeyer:
Potential exception?

aschempp:
If the method `random_byte` is not found. We're assuming the `random_compat` library always uses autoload files…

leofeyer:
If the library is not loaded correctly, I want an exception to be thrown so I notice it. Don't you agree?

aschempp:
Not during the installation where we're only providing a little helper for the random secret. Not generating it in this case would be totally fine to me, the user would just need to add it.

leofeyer:
I disagree for two reasons:

1. Since the feature would fail silently, we would probably not notice that we have to fix it.
2. If the user needs to enter the secret, they will most likely enter a short and weak secret, which we do not want (and which was the reason for adding the feature in the first place).

The way it is now, we would notice immediately if the library would no longer use autoload files.

PR #400, opened 2015-11-06
[RTM] Use config param upload_path upon installation 

aschempp:
Follow-up on #394


leofeyer:
Merged in 9ad5ec1fac1ddaf10d241be9605adac30335b950.

Issue #283, opened 2015-05-23
Maintenance mode buggy


aschempp:
Because the maintenance mode is part of the Contao framework
- the message is not shown if the framework is not initialized
- it is loaded very late and therefore any error in the system could result in the message not being shown

I can think of these solutions:
 1. adjust the entry point (`app.php` only) to trigger the maintenance mode very early
 2. Add a kernel listener for the maintenance mode and move the flag to `parameters.yml`
 3. Also change the exception listener to show maintenance regardless of the kernel error to prevent the generic error message from bubbling up.

aschempp:
Ok I was wrong with that. The maintenance message is shown in the frontend controller. To me, a maintenance mode should not be bound to a route though…

Toflar:
https://github.com/lexik/LexikMaintenanceBundle

leofeyer:
> To me, a maintenance mode should not be bound to a route though…

The maintenance mode was FE only ever since. If it were enabled in the back end too, you could not longer work; and the maintenance mode is e.g. enabled if you install an extension or run the Live Update, which both require back end access.

aschempp:
The maintenance bundle mentioned by @Toflar does fully support excluding routes or url paths like the backend. We should really consider that for a next minor version.

leofeyer:
+1

leofeyer:
Will you be working on this for Contao 4.1?

aschempp:
It's not assigned to me, that's why I didn't stumble upon it. But I just recently stumbled upon it again while reading about OroCRM (they use it and also have bundle autoloading ;-) ).

aschempp:
What I dislike about the bundle is that it requires Symfony to be fully operational (kernel boot & request/response handling). The original Contao implementation was triggered very early. I wonder if we should build a workaround for that?

leofeyer:
Do we need to change it at all?

bytehead:
Regarding to this issue, I have another question:
When [this exception](https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/controllers/FrontendIndex.php#L46) is thrown, there is no scope set.

I have an extended `PrettyErrorScreenListener` in use (because of the fact, that I use the `contao/core-bundle` as a bundle besides for example the `FosUserBundle` and I have to let the exception bubble up, if it's not a `ContaoMasterRequest`, which actually checks for the scope:
```php
protected function isContaoMasterRequest(KernelEvent $event)
{
[...]

PR #402, opened 2015-11-06
[RTM] Bypass the maintenance mode if logged in to the back end

leofeyer:
This is the follow-up PR to #283.


aschempp:
I intentionally did not name the parameter something with session, because it is not related to the (Symfony) session. In regards to the new image settings, maybe it should be `contao.user.bind_to_ip`?

The config would then make sense:

```yaml
contao:
    user:
        bind_to_ip: true
```

aschempp:
there could potentially be more user configuration in the future

leofeyer:
You cannot bind a user to an IP address. Only a session.

src/Resources/contao/library/Contao/Config.php
```diff
@@ -513,6 +513,7 @@ protected static function loadParameters()
 		}
 
 		$GLOBALS['TL_CONFIG']['debugMode'] = $container->getParameter('kernel.debug');
+		$GLOBALS['TL_CONFIG']['disableIpCheck'] = $container->getParameter('contao.security.disable_ip_check');
```
aschempp: You can move this to the array above now that we don't need inversion. Dunno why `debugMode` is not up there…

aschempp:
Hmm. Actually we do not need the path validation in the maintenance bundle! Our backend routes should have the respective attribute ?

src/Resources/contao/library/Contao/Config.php
leofeyer: Fixed in da4ebe63f09d9edad4bc6d32729e372e398cf9c4.

leofeyer:
Besides the fact that this would be more work, it would also not apply to custom back end routes added by third party developers.

aschempp:
That's true, but it would allow for more fine-grained control. It would also mean it still perfectly works if I rename the backend route path.

leofeyer:
If you rename the back end route, you can adjust the Lexik config as well ;)

leofeyer:
Also, was it not you saying that using route attributes is magic that should be avoided? This is a case where we could avoid it.

leofeyer:
Your suggestion is actually independent from this PR; if we should decide to use the route attribute, we'll have to adjust the standard edition. I have therefore merged this PR in cd2ebbe02f32a43dab2740abf11c100c2aa954c9.

PR #403, opened 2015-11-06
Catch all exceptions from the database

aschempp:
Otherwise an exception because of invalid access details etc. would not be catched.


leofeyer:
Merged in 16c881649ae67655073627fccf4001a906289450 (you forgot to remove the use statement).

PR #404, opened 2015-11-09
Never load valid classes from service container

aschempp:
I've got the following error message:

> Attempted to call an undefined method named "getData" of class "Symfony\Component\HttpFoundation\Session\Session".

This is because `System::import('Session')` would load the Symfony service instead of the legacy class.


leofeyer:
Isn't the container case-sensitive?

aschempp:
Unfortunately not, I thought so too… https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/DependencyInjection/Container.php#L262-L265

PR #415, opened 2015-11-17
[RTM] Added a DBAL field type for UUIDs

aschempp:
Our current UUID field types are not supported by Doctrine. I started using ORM and the binary column was not usable because it returns a resource by default.
I think Contao core bundle should provide a type to handle it's default fields.

### DIFF
--- a/src/ContaoCoreBundle.php
+++ b/src/ContaoCoreBundle.php
@@ -15,9 +15,8 @@ use Contao\CoreBundle\DependencyInjection\Compiler\AddPackagesPass;
 use Contao\CoreBundle\DependencyInjection\Compiler\AddResourcesPathsPass;
 use Contao\CoreBundle\DependencyInjection\Compiler\AddSessionBagsPass;
 use Contao\CoreBundle\DependencyInjection\Compiler\DoctrineMigrationsPass;
+use Contao\CoreBundle\DependencyInjection\Compiler\DoctrineTypesPass;
 use Contao\CoreBundle\DependencyInjection\ContaoCoreExtension;
-use Contao\CoreBundle\Doctrine\DBAL\Types\BinaryStringType;
-use Doctrine\DBAL\Types\Type;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
@@ -43,23 +42,9 @@ class ContaoCoreBundle extends Bundle
     /**
      * {@inheritdoc}
      */
-    public function boot()
-    {
-        if (!Type::hasType(BinaryStringType::NAME)) {
-            Type::addType(BinaryStringType::NAME, BinaryStringType::class);
-        }
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function build(ContainerBuilder $container)
     {
         parent::build($container);
-
-        if (!Type::hasType(BinaryStringType::NAME)) {
-            Type::addType(BinaryStringType::NAME, BinaryStringType::class);
-        }
 
         $container->addCompilerPass(
             new AddPackagesPass($container->getParameter('kernel.root_dir').'/../vendor/composer/installed.json')
@@ -69,5 +54,6 @@ class ContaoCoreBundle extends Bundle
         $container->addCompilerPass(new AddResourcesPathsPass());
         $container->addCompilerPass(new AddImagineClassPass());
         $container->addCompilerPass(new DoctrineMigrationsPass());
+        $container->addCompilerPass(new DoctrineTypesPass());
     }
 }
--- a/src/ContaoCoreBundle.php
+++ b/src/ContaoCoreBundle.php
@@ -16,6 +16,8 @@ use Contao\CoreBundle\DependencyInjection\Compiler\AddResourcesPathsPass;
 use Contao\CoreBundle\DependencyInjection\Compiler\AddSessionBagsPass;
 use Contao\CoreBundle\DependencyInjection\Compiler\DoctrineMigrationsPass;
 use Contao\CoreBundle\DependencyInjection\ContaoCoreExtension;
+use Contao\CoreBundle\Doctrine\DBAL\Types\BinaryStringType;
+use Doctrine\DBAL\Types\Type;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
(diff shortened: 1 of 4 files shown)


leofeyer:
Why are you adding the type in the bundle class? Shouldn't it be in a compiler pass?

aschempp:
The Compiler pass is to manipulate the container. The type is not registered in the container. Same as `Bootup::initAll();`

aschempp:
This should be for 4.2 ;-)
I'm working on more Doctrine related stuff, maybe I'll find a bridge between our models and Doctrine ORM :sunglasses:

leofeyer:
Why did you omit the following method? 

```php
public function getSqlDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
{
    return sprintf('BINARY(%d) COMMENT \'(DC2Type:uuid)\'', $fieldDeclaration['length']);
}
```

Not that I knew what it does, it just is in the class you have linked :)

aschempp:
That's what I meant with *We also need to verify the SQL comment.*

There is a method to tell Doctrine to add the comment, which I implemented. The `getSqlDeclaration` does nothing else than add the comment, but it omits a few other checks in the parent method (e.g. use `BLOB` if the field is too long (which should never be the case with UUID, but anyway…))

leofeyer:
Is this PR still relevant? As far as I remember, we have decided against adding your Doctrine implementation.

Toflar:
No matter what implementation of doctrine, the types will be needed anyway.

leofeyer:
Ok, then @aschempp will you complete the PR before the release candidate?

aschempp:
I'm not sure about that. We did find some potential problems with this, but if I remember correctly that was related to my Doctrine integration.

I guess everyone who want's to use Doctrine entities in their modules and has a relation to a file might need this.

leofeyer:
So are you going to complete it?

leofeyer:
As discussed in Mumble on April 28th, we want to add this to version 4.2. @aschempp will check if the PR is up to date and then add the `RTM` flag.

aschempp:
Pull request updated and added full test coverage.

src/Doctrine/DBAL/Types/UuidArrayType.php
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * @return string
+     */
+    public function convertToPhpValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value || !is_string($value)) {
+            return null;
+        }
+
+        $value = @unserialize($value);
```
leofeyer: Shouldn't we be using our `deserialize()` logic here, which does not allow serialized objects?

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/helper/functions.php#L192

discordier: As this is for data originating from the database and not for use by developers I think we can save the overhead.

leofeyer: The data we are running through `deserialize()` is also originating from the database, isn't it?

aschempp: Theoretically, but `deserialize` is a method from the old framework which is not (necessarily) available here. Also, if you would use Doctrine serialized array storage, the same problem would be the case.

Otherwise we would need to add a `secure_array` type that uses the same code as our deserialize method…?

leofeyer: > Theoretically, but `deserialize` is a method from the old framework which is not (necessarily) available here.

It is not available here ;)

> Also, if you would use Doctrine serialized array storage, the same problem would be the case.

That's a good point.

> Otherwise we would need to add a `secure_array` type that uses the same code as our deserialize method…?

No, it would be to specific for Contao then IMHO.

aschempp: It might be available if the framework is booted. Or we could boot the framework. But we know that's somewhat wrong :D

A secure Doctrine array (one that does not allow objects) is not specific for Contao. I know Joomla had similar security issues with unserializing object (from the session though).

leofeyer: You mean you would replace the `array` type with the `secure_array` type? I certainly would prefer that.

aschempp: No I would not replace it, because that would potentially break third-party bundles. I would just add a `secure_array` type that people should use in their entities. But I wonder if these things should be part of `contao/core-bundle`…

leofeyer: > But I wonder if these things should be part of `contao/core-bundle`…

True. :frowning:

leofeyer: As discussed in Mumble on May 4th, we want to provide an `secure_array` type, which integrates the [https://github.com/contao/core-bundle/blob/master/src/Resources/contao/helper/functions.php#L192](deserialize check). The `uuid_array` type should then inherit from the `secure_array` type.

aschempp:
PR is updated, added new `secure_array` type

Toflar:
Looks good to me 👍

leofeyer:
As briefly discussed in Slack, we want to postpone merging this PR, because we want to use the [ramsey/uuid](https://github.com/ramsey/uuid) library for the UUID conversion. A quick POC shows that the library provides the required methods.

Also, there is a Doctrine field type already ([ramsey/uuid-doctrine](https://github.com/ramsey/uuid-doctrine)) which we might be able to use.

leofeyer:
@Toflar also mentioned that we should prefix our custom types to prevent collisions.

aschempp:
I just tested ramsey/uuid-doctrine in my application, and the `binary_uuid` type works flawlessly. It does not support arrays though, and that's gonna be a problem whenever one needs to store a checkbox selection in a file tree.

leofeyer:
>  It does not support arrays though

Then we should go for a custom implementation, I guess.

aschempp:
Difficult to say. First of all, the core does not use Doctrine ORM, so it's difficult to say what we should support. Also, it's fairly easy to add whatever you need to your application, because this is still pretty application specific.

I just run across the issue that ramsey/uuid-doctrine is actually not using strings in PHP, but a custom `Uuid` object that has a `__toString` method. This works in most contexts, but not in all.

Regardless of what we implement, a Doctrine type that internally converts between binary and string UUID automatically is a behaviour change to the current implementation. All we actually want is `blob` and `binary` that is not converted to a resource stream… not really related to UUID actually 😂

aschempp:
I have updated the PR after realising what we actually need in a real-work scenario. A UUID type that automatically converts from binary to string will not work with any current implementation, as for example searching the DB would need the binary value and not the string.

What we actually need is just a regular binary field, but it must not be stored in a resource but simply a PHP string. That's what this PR provides.

PS: yes, unit tests are missing and it needs to be rebased. But RFC ;-)

leofeyer:
LGTM.

leofeyer:
Can you please give me push access to this pull request so I can make the CS fixes?

leofeyer:
The tests are failing with `--prefer-lowest`: https://travis-ci.org/contao/core-bundle/jobs/194957297

aschempp:
The tests are fixed. I was actually correct, I just forgot a third place ;-)

PR #422, opened 2015-11-23
[RFC] Extract the framework initialization logic

leofeyer:
This PR extracts the framework initialization logic from the `ContaoFramework` class.

The `ContaoFramework` class now does more than just initializing the Contao 3 framework, therefore the initialization logic should not be in the class anymore. I have created a `FrameworkInitializer` service instead, which is injected only if the `initialize()` method is called.

All the changes are internal, so the API remains exactly the same as before.

@contao/developers Any objections?

### DIFF
--- a/src/Framework/ContaoFramework.php
+++ b/src/Framework/ContaoFramework.php
@@ -10,70 +10,25 @@
 
 namespace Contao\CoreBundle\Framework;
 
-use Contao\ClassLoader;
-use Contao\Config;
-use Contao\CoreBundle\ContaoCoreBundle;
-use Contao\CoreBundle\Exception\AjaxRedirectResponseException;
-use Contao\CoreBundle\Exception\IncompleteInstallationException;
-use Contao\CoreBundle\Exception\InvalidRequestTokenException;
-use Contao\Input;
-use Contao\RequestToken;
-use Contao\System;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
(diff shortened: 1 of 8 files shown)


aschempp:
What's the point in changing this? To me, the Contao framework class is "everything" related to the old framework. Initializing the framework and using the adapters are very much related to each other in the sense that they will "provide access" to the old framework.

In your refactoring, people can (and probably will) start injecting the initializer, which does not make sense if I understand your approach correctly.

Once everything is refactored, the whole `Framework` namespace should be dropped and we should not spend time on improving that code base.

leofeyer:
First of all, the work has already been done, so the time has already been spent. Second of all, the code is much cleaner now and the `ContaoFramework` class does not need any dependencies if it is used to just create adapters or object instances.

IMO, it is wrong to inject 6 dependencies if none of them is used (e.g. because you just need an adapter).

> Once everything is refactored, the whole `Framework` namespace should be dropped

Yes, and this will also include the `FrameworkInitializer`.

aschempp:
> First of all, the work has already been done, so the time has already been spent.

Very bad reason to include something just because work has already been done :P

So, where's the difference now compared to an adapter factory, which we had in the first place?
Please remember, whenever you need an adapter, you always need to make sure the framework is initialized. So the injections are always necessary because you can't use the adapters without the framework being booted.

leofeyer:
I still think it is wrong to have it all in one class. But never mind, I'll close the PR.

PR #424, opened 2015-11-26
Fixed wrong quoting in findInSet() database helper

Toflar:


leofeyer:
It should not be necessary to quote the field name unless you are using a reserved keyword.

aschempp:
The docblock is misleading. The *key* is actually the word you're searching for, that's why most of the time a question mark will be supplied as key.

leofeyer:
The key is the column name, not a word you are searching for.

Toflar:
Just see what DC_Table does for eval->csv filters and compare to Contao 3.5 :)

leofeyer:
Merged in 890878c5c39bb6116c20e3f543815185e607ca7e.

PR #427, opened 2015-12-07
[RFC] Allow to insert Twig templates in Contao templates

aschempp:
Just a quick idea while stumbling through the code. What do you think?

/cc @Toflar @discordier

### DIFF
--- a/src/Resources/contao/library/Contao/TemplateInheritance.php
+++ b/src/Resources/contao/library/Contao/TemplateInheritance.php
@@ -258,6 +258,12 @@ trait TemplateInheritance
 	 */
 	public function insert($name, array $data=null)
 	{
+		if (pathinfo($name, PATHINFO_EXTENSION) == 'twig')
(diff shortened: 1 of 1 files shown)


Toflar:
I guess this should work. However, I wonder

a) if mixing is a good thing?
b) if supporting only `insert()` and not the general use of `.twig` is a good idea?

aschempp:
The general use of Twig is already supported (though not a template editor). Also, it obviously does not support the opposite (insert Contao template from Twig).

discordier:
It should work indeed. I wondered if `pathinfo`copes with something like `'@ContaoCore/Error/error.html.twig'` but it does.
```
php -r "var_dump(pathinfo('@ContaoCore/Error/error.html.twig', PATHINFO_EXTENSION));"
string(4) "twig"
```
However I wonder how many will make use from it as it is only supporting imported templates.

PR #430, opened 2015-12-10
Fix CURRENT_ID and correctly map keys in all cases

aschempp:
I first noticed that `CURRENT_ID` is no longer working in the current session.

While attempting to add that, I also noticed that my previous PR to fix `popupReferer` didn't actually handle all cases…


leofeyer:
Merged in 939448c6dcb14f0c6d791cf529e8470cb8175347.

PR #431, opened 2016-01-02
Fixing contao.org-searchform in layout.html.twig

ADoebeling:


xchs:
We could also rename the German search page `suche.html` to `search.html`. Then you get redirected to the right search form according to the browser's language setting.

discordier:
eh... @issue-bot. Do not!!!!!
This PR was correct here and perfectly valid.
The template in core-bundle is wrong.

@leofeyer plz reopen.

leofeyer:
I see. This one is tricky, because before the relaunch, the URL was perfectly fine and the language was added automatically depending on the user's language. Now that we have individual aliases in each language, we need to think about another solution.

leofeyer:
I have adjusted the server-side script so it handles the new page aliases.

PR #434, opened 2016-01-13
Webprofiler redesign

aschempp:
rewrote the profiler for Symfony 2.8. Not sure if it still works/looks good with 2.7, but we should update the dependencies anyway.

bildschirmfoto 2016-01-13 um 22 41 01

### DIFF
--- a/src/DataCollector/ContaoDataCollector.php
+++ b/src/DataCollector/ContaoDataCollector.php
@@ -216,9 +216,12 @@ class ContaoDataCollector extends DataCollector
 
         $this->data['summary'] = [
             'version' => $this->getContaoVersion(),
-            'layout' => $this->getLayoutName(),
             'framework' => $framework,
             'models' => $modelCount,
+            'frontend' => isset($GLOBALS['objPage']),
+            'preview' => BE_USER_LOGGED_IN === true,
+            'layout' => $this->getLayoutName(),
+            'template' => $this->getTemplateName(),
         ];
     }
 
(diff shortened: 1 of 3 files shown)


Toflar:
:+1:

leofeyer:
Do you want to upgrade to Symfony 2.8 for Contao 4.1, too?

discordier:
Yep. Would make things more easy. It not already `~2.8 | ~3.0` if we could get rid of the scopes.

aschempp:
No I would only make it for 4.2. 4.1 should still be marked as Compat with 2.7

aschempp:
Btw we actually need Symfony 2.8.3 or 3.0.2 due to https://github.com/symfony/symfony/pull/16961

aschempp:
Oh well… it can only be 2.8 as we're still using scopes and not supporting Symfony 3…

discordier:
We somehow must get rid of them. :(

leofeyer:
> We somehow must get rid of them. :(

Indeed!

aschempp:
Well *theoretically* we could add a lot of if/else statements to the container scope calls. If a bundle needs scopes, it would need to require Symfony 2.8. But I think that's dangerous (because people will only require `contao/core-bundle`). And it also means we need to add an alternative to scopes, and I still disagree with a route attribute ^^

leofeyer:
I am adding this to the topic list of the dev meeting next week.

DanielSchwiperich:
Like @aschempp wrote if the scopes get dropped an alternative is needed. 
I think this could be done without breaking the public API. Keeping the ScopeAwareTrait for example, mark it as deprecated and modify the methods using the new *concept* 

I think there are enough use cases where *scopes* (being it symfony scopes or not) are useful especially to give an extension to distinguish if the current process or request is frontend, backend, cli whatever. 

To this moment I didn't study why the scopes were dropped in 3.0 (will do on the weekend) so there may be something fundamental wrong in my posting. Sorry then, pls ignore :smiley:

leofeyer:
Merged in 94db00b88ec14871cc8ebd40ff7066bbabf8c0b8.

PR #438, opened 2016-01-21
[WIP] Writing DCA files in the YAML notation

Toflar:
I found that for me personally, it's way easier to write a DCA in YAML instead of PHP. I'm way faster and I make less typos. Here's a proposal for Contao 4.2, obviously it's still missing tests and stuff but I wanted to hear your feedback about it :-)

Also, that way we could decide that starting from Contao 4.2 (or whatever version) using PHP DCA files is deprecated. This would give the users enough time to slowly migrate to using YAML files which in contrast to our php files can be cached someday :-)

### DIFF
--- a/src/Config/Loader/YamlDcaFileLoader.php
+++ b/src/Config/Loader/YamlDcaFileLoader.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2015 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\Config\Loader;
+
+use Symfony\Component\Config\Loader\Loader;
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * Reads YAML files and returns the content.
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+class YamlDcaFileLoader extends Loader
+{
+    /**
+     * Reads the contents of a YAML file.
+     *
+     * @param string      $file A YAML file path
+     * @param string|null $type The resource type
+     *
+     * @return string The PHP code without the PHP tags
+     */
+    public function load($file, $type = null)
+    {
+        $table = basename($file, '.yml');
+        $data = var_export(Yaml::parse(file_get_contents($file)), true);
+        return "\$GLOBALS['TL_DCA']['$table'] = array_replace_recursive($data, \$GLOBALS['TL_DCA']['$table']);\n";
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports($resource, $type = null)
+    {
+        return 'yml' === pathinfo($resource, PATHINFO_EXTENSION);
+    }
+}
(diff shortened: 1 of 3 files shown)


aschempp:
:+1: 
In addition to caching, we can also apply transformations to a later format of DCA definitions, which is not possible with the current array.

src/Config/Loader/YamlDcaFileLoader.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param string      $file A YAML file path
+     * @param string|null $type The resource type
+     *
+     * @return string The PHP code without the PHP tags
+     */
+    public function load($file, $type = null)
+    {
+        $table = basename($file, '.yml');
+        $data = var_export(Yaml::parse(file_get_contents($file)), true);
+        return "\$GLOBALS['TL_DCA']['$table'] = array_replace_recursive($data, \$GLOBALS['TL_DCA']['$table']);\n";
```
aschempp: I think `$data` should be the second argument to `array_replace_recursive`?

aschempp: Also need to make sure `$GLOBALS['TL_DCA']['$table']` is an array.

Toflar:
And we'll have to find a translator solution for all the labels which are now references.

fiedsch:
+1 for the proposal

concerning the labels: could specifying the "path" below TL_LANG and using Symfony's expression language be an option as you always have references there!

Toflar:
The EL does not evaluate PHP code, it provides its own sandboxed parser. That's the whole point about it ;-) So EL is not going to help us here, I guess.

fiedsch:
I know. Of course the method thats reads the yaml code woul have to be extended to use the 

label:  tl_article.edit

to eventually evaluate the matching component of the $GLOBALS language array. This is where I meant that expression language could come in handy.

discordier:
@fiedsch: using the path below will not work without also specifying the domain.
Hence here we have the same problem we stumbled over when implementing [contao-community-alliance/translator](https://github.com/contao-community-alliance/translator).

[Here](https://github.com/contao-community-alliance/translator/blob/9eb806d53151551696e4aa283b9b5da395e929ad/src/ContaoCommunityAlliance/Translator/Contao/LangArrayTranslator.php#L67) we have to treat 'languages', 'default', 'modules' etc. domains different from table domains, as they each have different sub pathes.

The key `MSC.titleFormat` originates from `default.xlf`, whereas `tl_article.edit` originates from `tl_article.xlf`.
To make things even worse, the referenced language strings are not strings at all but arrays for fields (name and description) and strings for other things (like above mentioned titleFormat).

We therefore will have to alter the dca in YAML to accept `title` and `description` instead of `label` and translate it like this when loading the DCA:
```php
foreach ($data['fields'] as &$field) {
  $field['label'] = array(
    $this->translator->$translator->trans($field['title'], array(), $table),
    $this->translator->$translator->trans($field['description'], array(), $table),
);
}
```
Right after: https://github.com/contao/core-bundle/pull/438#discussion-diff-50380510R35

The operation buttons are a bit tougher to accomplish, as they have placeholders and therefore are harder to accomplish.

The other solution however would be to add a custom message resource to the translator which will hold all language strings we define in this table.

Both approaches obviously will break the current way of defining language resources by reference (which is bad habit anyway now we have a proper translation component at hand).

leofeyer:
> obviously it's still missing tests and stuff

@Toflar Will you add the missing stuff until the release candidate?

Toflar:
No, because it won't work without further adjustments, like we've discussed. Labels need to be translated using the translator. Everything that uses references will break etc.

m-vo:
And because I referenced it, something useful at least: Couldn't we make this work as an alternative approach? Then we could declare using it as a 'best practice' and extensions could use it. (The core obviously only for new feaatures.)

PR #440, opened 2016-01-26
[RTM] Allow to set the cron job for purging image cache

aschempp:


src/DependencyInjection/Configuration.php
```diff
@@ -91,6 +91,9 @@ public function getConfigTreeBuilder()
                         ->scalarNode('target_path')
                             ->defaultValue('assets/images')
                         ->end()
+                        ->scalarNode('purge_cache')
+                            ->defaultValue('monthly')
```
discordier: I'd suggest to add here.
```php
  ->validate()
    ->ifNotInArray(['monthly', 'weekly', 'daily', 'hourly', 'minutely'])
      ->thenInvalid('Id mapper incorrect!')
    ->end()
```

aschempp: :+1:

leofeyer:
Ok, but the naming is wrong. It should be something like `image.purge_frequency`, don't you think?

discordier:
Dunno. However I have no objections against either naming.

aschempp:
be aware that it can also be `false` to disable purging. I guess I came up with that name because the other property is called `bypass_cache`. `purge_frequency` does not relate to the cache…

discordier:
Which would ultimatively lead to:
```
contao.image.cache.path
contao.image.cache.purge_frequeny
contao.image.cache.bypass
```
instead of the current options...

aschempp:
I would not add a third part

leofeyer:
I don't recall what we have discussed. Are we going for `purge_frequency`?

aschempp:
can't really remember, but you also said something about `never` instead of `false`

discordier:
I voted against never because of the way they do it at symfony, however, they also have a sub-key "enabled" for each section which will magically get set via `ArrayNodeDefinition::canBeEnabled()` for `false`, `true` and `null` values on the root key.
IMO we should try to mimic the symfony configuration layout where ever possible.

Sadly this interferes with the present keys.

aschempp:
How would `enabled` help in our case?

discordier:
```yaml
contao:
  # [...]
  image:
    cache:
      # (default true)
      - enabled:
      # (default: default path)
      - path:
      #  (default: monthly or so, can be set to false)
      - purge:
      # (drop this, as is the same as enabled == false or setting the whole key to false)
      - bypass:
```

So you can do:
```yaml
contao:
  # [...]
  image:
    cache: false
```
And have the whole caching disabled.
The very same as with `validation` in symfony `framework_bundle`

What I am trying to say is, we should push our config values all into hierarchical domains of components.
i.e. the image domain within Contao might also add support for other drivers or something like that in the future and moving them later on is harder than having a domain from the beginning.

aschempp:
First of all, `enabled` is used by Symfony framework bundle to integrate components. I don’t think that applies here.
Secondly, your suggestion would not make sense at all, because my proposal is only about the cronjob.

discordier:
1. IMO the definition of image cache applies as component in the widest meaning.
2. So, then we are adding this as key `contao.image.purge_frequeny` now and move all cache related items over to a cache section later?

I simply wanted to promote that it makes more sense to group related config keys together, if that is nonsense to you, fine for me. I just wanted to say it before we end up having unorganized config keys at 4.5 or so.

leofeyer:
I like `never`, but this does not solve the naming problem :)

leofeyer:
Except the naming problem, I don't really understand why we need this feature in the first place?

If you want to change, add or remove a cron job, you have to adjust the `$GLOBALS['TL_CRON']` array. This PR only adds a special routine for one particular job, instead of an abstract solution for all of them. This does not seem good practice.

aschempp:
I partially agree. It adds new features to a part we're already rewriting. We already added new image configuration, and @ausi is working on rewriting it to Imagine. So it partially makes sense. Iterating over `$GLOBALS['TL_CRON']` is a pain in the ass…

However, I would prefer to completely rewrite cronjobs, but that might be a difficult topic…

leofeyer:
> Iterating over `$GLOBALS['TL_CRON']` is a pain in the ass…

Then we should maybe simplify this:

```php
$GLOBALS['TL_CRON'] = array
(
	'monthly' => array
	(
		'purgeImageCache' => array('Automator', 'purgeImageCache')
	),
	'weekly' => array
	(
		'generateSitemap' => array('Automator', 'generateSitemap'),
		'purgeScriptCache' => array('Automator', 'purgeScriptCache'),
		'purgeSearchCache' => array('Automator', 'purgeSearchCache')
	),
	'daily' => array
	(
		'purgeTempFolder' => array('Automator', 'purgeTempFolder'),
		'checkForUpdates' => array('Automator', 'checkForUpdates')
	),
	'hourly' => array(),
	'minutely' => array()
);
```

The change should be 100% backwards compatible and it allows you to use `isset()` and `unset()`.

discordier:
Can't we make it a service registry and keep the global array only for bc to obtain and generate tagged services?

aschempp:
I was rather looking at sollutions like https://github.com/Cron/Symfony-Bundle :-)

leofeyer:
As discussed in Mumble on April 21st, we want to implement the [simple solution](https://github.com/contao/core-bundle/pull/440#issuecomment-205849188) since `$GLOBALS['TL_CRON']` is legacy and will be reworked anyway.

leofeyer:
Implemented in aac4dcec97919f66fb6262a20e75bcc9a24c0cb3.

ausi:
The image purging was removed from TL_CRON now in 9f43f022235324570493bb1bb79c15c40a363f8c. Wouldn’t it be better if we purge the cache automatically but only remove files that didn’t were accessed for at least 30 days or so?

leofeyer:
We should discuss this in a separate ticket.

PR #441, opened 2016-02-01
Sort file symlinks by path depth before symlinking

Toflar:
The problem in the current implementation is the fact that it relies on the finder to return a list of files from top to bottom (parents to nested). However, apparently we cannot rely on this for every filesystem. That's why we need to sort them first to have top level paths first, expanding down to more nested paths.

Only after sorting the found paths, we can filter them for parents already symlinked.

### DIFF
--- a/src/Command/SymlinksCommand.php
+++ b/src/Command/SymlinksCommand.php
@@ -218,7 +218,6 @@ class SymlinksCommand extends AbstractLockedCommand
             }
 
             $parents[] = $file->getRelativePath();
-            $parents   = array_unique($parents);
         }
 
         return $result;
--- a/src/Command/SymlinksCommand.php
+++ b/src/Command/SymlinksCommand.php
@@ -22,6 +22,7 @@ use Symfony\Component\Finder\SplFileInfo;
  * Symlinks the public resources into the /web directory.
  *
  * @author Leo Feyer <https://github.com/leofeyer>
+ * @author Yanick Witschi <https://github.com/toflar>
  */
 class SymlinksCommand extends AbstractLockedCommand
 {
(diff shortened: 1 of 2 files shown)


src/Command/SymlinksCommand.php
```diff
@@ -175,41 +178,68 @@ private function symlink($source, $target)
...
+                        $file->getRelativePath(),
+                        $parent
+                    )
+                );
+
+                unset($result[$k]);
+            }
+
+            $parents[] = $file->getRelativePath();
+            $parents   = array_unique($parents);
```
ausi: Is this line necessary?

Toflar: No, but makes `in_array()` calls faster if the amount of files to check increases :)

ausi: I don’t see a case where the same path can be added twice, IMO this call never removes an entry.

Toflar: Hmm...true. Removed in cb446d0020626c2cd535e0db0efb2e89e639df68,

aschempp:
Why not simply apply a sorting callback over the `iterator_to_array` and keep the rest as-is?

Toflar:
Because the finder can sort already and the existing filter function was far to complicated.

leofeyer:
> However, apparently we cannot rely on this for every filesystem

Which filesystem are you using that returns a "wrong" order?

Toflar:
ext4, Ubuntu 14.04...I don't know why the order is different to be honest but we have to ensure it's correct no matter what the filesystem delivers.

DanielSchwiperich:
We stumbled over this problem too in a current project. 
Thought it's not allowed by convention to have sub folder made public if there is already a public parent.

Was in a Ubuntu Docker Container I think (aufs / Union Filesystem over ext4).

aschempp:
Also see https://github.com/contao/core/issues/6391

leofeyer:
Merged in a6e331d5d0a2fe8fb6942b8a5ccae802044efc68.

PR #435, opened 2016-01-15
Allow services for backend module callbacks

aschempp:

PR #444, opened 2016-02-12
[RFC] CSV import for options form field

qzminski:
Added the CSV import feature for the `options` field in the form generator. Also introduced the CSV Import utility which is now used also by `ListWizard` and `TableWizard` widgets.


Toflar:
Thank you for the PR Kamil! I would like to make some general comments on what I don't like about the architecture of this PR. I think, if you added the unit tests - that are mandatory to make it into the core-bundle - already, you would have noticed some of the problems :-)

1. General question: I think there are many CSV libraries out there and I wonder why we don't make use of one of them?
2. There are a few violations of separation of concerns. Especially, the architecture of `CsvImportUtil` feels very strange to me:
    
    a) Why would it need the framework? `Contao\Message` should be adapted so it can handle Symfony's flash messages, which would immediately drop the need of that one.
    b) Why would it need the framework part 2: You don't need `System::setCookie()`. Firstly, we set cookies on the request now, not on `System` anymore and then why is it the task of the `CsvImportUtil` to set cookies?
    c) Why does the `CsvImportUtil` throw `RedirectResponseException`'s? Redirecting is certainly not a task of a "csv import util" but belongs to a controller :-)

A good indicator of wrong dependencies are the constructor arguments. Why should a csv import utility care about a service named `tokenStorage`?

Imho, you have a good start here but what you need to do is to split up the tasks of your `CsvImportUtil` even more:

1. The real CSV handling stuff, I doubt you need to reinvent the wheel here and you can maybe use a third party library from packagist.
2. The `view` component which is responsible to generate your back end views.

qzminski:
Just as a word for introduction, the class is currently designed to handle only the backend CSV import that is commonly used for different widgets in Contao.

1. Why would you like to use the CSV library if reading a CSV file is a matter of three lines of code using [fgetcsv](https://secure.php.net/manual/pl/function.fgetcsv.php) function?

2a. Indeed ```Contao\Message``` should be adapted to Symfony's flash messages but at the time being it is not. Once it is adapted, I agree it should be used instead.

2b. That was just copy-pasted from the old method. Can go the controller.

2c. Yeah, that should probably go to the controller.

The ```tokenStorage``` is passed to the class so it can determine the current user file uploader. It can belong to controller as well which will tell which ```FileUpload``` instance should the class use. 

I'll introduce those changes soon. Controller will grow a little bigger than it currently is but that shouldn't be a problem huh? @aschempp ?

aschempp:
I haven't completely reviewed the code yet, but I remember we discussed the `FileUpload` instance should be passed to the utility class, which would probably make `$tokenStorage` unnecessary.

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,223 @@
...
+    {
+        $uploader = $this->getDefaultUploader();
+
+        if ($uploader === null) {
+            return '';
+        }
+
+        $csvImport = new CsvImportUtil(
+            $this->connection,
+            new Message(),
```
aschempp: Should use `$this->framework->createInstance('Contao\Message');`

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,223 @@
...
+     */
+    public function importOptionWizard(DataContainer $dc)
+    {
+        $uploader = $this->getDefaultUploader();
+
+        if ($uploader === null) {
+            return '';
+        }
+
+        $csvImport = new CsvImportUtil(
```
aschempp: looks like this code is duplicated several times, it should be moved to a method then.

src/Resources/contao/classes/Backend.php
```diff
@@ -413,7 +413,7 @@ protected function getBackendModule($module)
 		// Custom action (if key is not defined in config.php the default action will be called)
 		elseif (\Input::get('key') && isset($arrModule[\Input::get('key')]))
 		{
-			$objCallback = new $arrModule[\Input::get('key')][0]();
+			$objCallback = System::importStatic($arrModule[\Input::get('key')][0]);
```
aschempp: Looks like a duplicate of https://github.com/contao/core-bundle/pull/435 and should be removed once that's merged.

src/Controller/BackendCsvImportController.php
Toflar: Of course not.

Toflar:
> 2a. Indeed Contao\Message should be adapted to Symfony's flash messages but at the time being it is not. Once it is adapted, I agree it should be used instead.

Here you go: https://github.com/contao/core-bundle/pull/445 :smile:

src/Controller/BackendCsvImportController.php
aschempp: ?

Toflar: All the methods are static??

aschempp: Ah well, so you mean we should use `getAdapter` :D

src/Controller/BackendCsvImportController.php
qzminski: I am not sure because then I would need to create the method that accepts at least 5 arguments (DataContainer, field name, data callback, submit label and allowed separators).

qzminski:
I have introduced some improvements. However, I am not sure how to move ```RedirectResponseException``` to the controller. Should then controller listen on util class if it throws a regular ```\Exception``` and if so trigger ```RedirectResponseException```? Any suggestions?

src/Controller/BackendCsvImportController.php
aschempp: Not really, except for the uploader the arguments are always identical? I'm only talking about creating the instance.

aschempp:
> I have introduced some improvements. However, I am not sure how to move RedirectResponseException to the controller. Should then controller listen on util class if it throws a regular \Exception and if so trigger RedirectResponseException? Any suggestions?

I'm not entirely sure. But maybe it should throw an "generic" exception including the error message that is catched by the controller. The controller would then add the flash message, and it would do the reload. For the other redirect, the `run()` method could return true/false (or just nothing, really) and the controller could then also do a redirect. It always needs to redirect if `run()` succeeds (which feels kinda strange…)

qzminski:
I have done some improvements, please review them.

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,250 @@
...
+            try {
+                $csvImport->run($dc->table, $fieldName, $dc->id);
+            } catch (CsvImportErrorException $e) {
+                // Add an error message and reload the page on import failure
+                $this->flashBag->add('error', $e->getMessage());
+                throw new RedirectResponseException($this->requestStack->getCurrentRequest()->getRequestUri());
+            }
+
+            // Set the backend offset cookie
+            $this->framework->getAdapter('Contao\System')->setCookie('BE_PAGE_OFFSET', 0, 0);
```
discordier: I wonder if we really have to reset the Cookie.
If not, we only need to inject the max filesize below and are completely independent from the framework in this class, which I consider a goal for every new class in Contao.

aschempp: I guess that's just how it currently behaves…

qzminski: Can't we reset the cookie in other way, i.e. not through Contao framework? Create the ```RedirectResponse``` with appropriate cookie which would then be used as constructor argument in ```ResponseException```?

aschempp: Pretty sure you can (and should!) add the cookie to the response :-)

aschempp:
I think we should fall back to `FileUpload` if no token or user is available. Also, it should be possible to set the uploader script.

src/Controller/BackendCsvImportController.php
qzminski: Done.

qzminski:
@aschempp uploader fallback set in the ```CsvImportUtil``` itself or in the backend controller? Frankly I would put it in the utility class so it behaves just similar to the ```getTemplate()``` method (returns fallback if no template was set).

aschempp:
I think this can be updated to RFC or RTM, what do you think?

qzminski:
@contao/developers - request for feedback

leofeyer:
RTM = ready to merge
RFC = request for comments

I have changed the title accordingly :)

aschempp:
*bump*

src/Resources/contao/classes/Backend.php
Toflar: Can you remove this, Kamil?

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,243 @@
...
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+
+/**
+ * Handles the backend CSV import.
+ *
+ * @author Kamil Kuzminski <https://github.com/qzminski>
+ */
+class BackendCsvImportController
```
Toflar: Controllers in Symfony use `Action` as method suffix. I think they should all use that too here.

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,243 @@
...
+     * @param DataContainer $dc
+     * @param string        $fieldName
+     * @param string        $submitLabel
+     *
+     * @return string
+     *
+     * @throws RedirectResponseException
+     * @throws ResponseException
+     */
+    private function runDefaultRoutine(CsvImportUtil $csvImport, DataContainer $dc, $fieldName, $submitLabel)
```
Toflar: This should certainly not be `private`. Think of another developer trying to add an import wizard for their own widget, how will they be able to use this? I think it should rather look like this:

```php
class AbstractBackendCsvImportController
{
    protected function createImportInstance() {}
    protected function runDefaultRoutine(CsvImportUtil $csvImport, DataContainer $dc, $fieldName, $submitLabel) {}
}

TableWizardCsvImportBackendController extends AbstractBackendCsvImportController
{
    public function importTableWizard() {}
```

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * Create the CSV import instance
+     *
+     * @return CsvImportUtil
+     */
+    private function createImportInstance()
+    {
+        return new CsvImportUtil(
+            $this->connection,
+            $this->flashBag,
```
Toflar: The flashbag is part of the request itself, why is this a separate argument?

qzminski: Is it? I also included it in the service constructor.

Toflar: Doesn't make sense here, I guess. In our import util it never makes sense to add a message to a flashbag that is not related to the current session of the request (which you pass as argument too). I don't think it ever occurs that you want to write to a different flashbag than the one of the request you pass in the constructor too?

Toflar:
As I stated on February 15th already, Unit Tests are missing. I know that not everything of it will be testable. Especially the parts where we are working in legacy code. However, the controllers as well as the import util is new code and if it's not unit testable, something must be wrong. We need tests here :)

src/Controller/BackendCsvImportController.php
aschempp: I feel its still better for unit tests and dependency injection. Otherwise you would need to rely on the mocking the request object with a flash bag, right? I did have the same thoughts with the session wich is also available from `$request->getSession()`

qzminski: I have changed that but one thing that concerns me is that the ```$request->getSession()``` returns ```SessionInterface``` which does not imply the method ```getFlashBag()```. Although it works.

qzminski:
@Toflar can you please review if the last commits cover our Skype discussion conclusions?

src/Resources/config/services.yml
```diff
@@ -1,4 +1,16 @@
 services:
+    contao.controller.abstract_csv_import:
+        class: Contao\CoreBundle\Controller\AbstractCsvImportController
+        abstract: true
+        calls:
```
Toflar: As these are mandatory, you must use constructor injection rather than setter injections. Setter injections are for optional dependencies.

src/Resources/contao/classes/Backend.php
```diff
@@ -413,8 +415,25 @@ protected function getBackendModule($module)
...
+			if (!class_exists($className) && $container->has($className)) {
+				$request = $container->get('request_stack')->getCurrentRequest();
+				$request->attributes->set('id', $dc->id);
+				$request->attributes->set('table', $dc->table);
+
+				/** @var Response $response */
+				$response = $container->get($className)->$methodName($request);
+
+				$this->Template->main .= $response->getContent();
+			} else {
```
Toflar: I'd add a `// BC` comment here and maybe even a `trigger_error()` for deprecated usage as there's now a proper way of implemnting this :)

Toflar:
Cool! Now it's a proper controller. However, we don't take advantage of it yet. You call them directly using the service instead of using it as a controller and pass it through the kernel. I guess we should enable the fragments listener and take advantage of the `_fragment` route and add some http cache headers.

src/Resources/config/services.yml
qzminski: I have changed that but left the setters to allow service overriding by child classes.

Toflar:
Important note, this contains a security issue as long as 

```
framework:
    fragments: { path: /_fragment }
```
is not present in the `config.yml`.

aschempp:
Guess we should add a PR for that on `contao/standard-edition`? Or do you think we need to validate for that security issue?

Toflar:
If this gets merged, we'll add it to standard, simple as that.

qzminski:
Yanick have you discussed the thing we were concerned about on the Contao call? Was it passing the request object around or something...

Toflar:
I have no idea. We did not discuss, because it does not have the up for discussion label. But I cannot recall what it was. Was it because of the referer?

src/Controller/AbstractCsvImportController.php
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * Set the database connection
+     *
+     * @param Connection $connection
+     */
+    public function setConnection($connection)
+    {
+        $this->connection = $connection;
```
aschempp: Why are there setter methods for the dependencies that must be already passed as constructor arguments?

src/Resources/config/services.yml
```diff
@@ -27,6 +27,18 @@ services:
         tags:
             - { name: kernel.cache_warmer }
 
+    contao.controller.abstract_csv_import:
```
aschempp: I would not create an abstract service for this just because there's an abstract class. It's pretty unlikely that the abstract service will ever be used again.

src/Resources/config/services.yml
```diff
@@ -27,6 +27,18 @@ services:
...
+        class: Contao\CoreBundle\Controller\AbstractCsvImportController
+        abstract: true
+        arguments:
+            - "@database_connection"
+            - "@contao.framework"
+            - "@security.token_storage"
+
+    contao.controller.backend_csv_import:
```
aschempp: Should this service be `public: false`?

qzminski:
The issue we have now is that the view/controller is rendered by the fragment handler. That means the URI of the request object will become something like ```_fragment?id=1&key=option...``` whereas the current real URL is ```contao?do=form&table=tl_form_field&id=1&act=edit&key=option...```. Currently it harms the upload form as its action points to the ```_fragment...``` URL.

https://github.com/qzminski/contao-core-bundle/blob/feature/options-import/src/Util/CsvImportUtil.php#L116

aschempp:
That could "easily" be solved by using the master request instead of the current request.

Toflar:
Nope, a fragment is always a master request already.

aschempp:
Really? Sounds weird, but maybe that's why we're already passing the `url` around?

src/Resources/contao/classes/BackendModule.php
```diff
@@ -17,6 +17,9 @@
  * @property string $table
  *
  * @author Leo Feyer <https://github.com/leofeyer>
+ *         
+ * @deprecated This class is deprecated and will be removed in Contao 5.0. Use
+ *             custom controllers instead.
```
aschempp: Should also trigger a deprecated notice in file or in constructor.

src/Resources/contao/classes/Backend.php
```diff
@@ -411,10 +414,40 @@ protected function getBackendModule($module)
...
+				$objCallback = System::importStatic($className);
+				$this->Template->main .= $objCallback->$methodName($dc);
+			} elseif (isset($arrModule['controllers'][\Input::get('key')])) {
+				// Use a controller
+				$container = static::getContainer();
+
+				/** @var Request $request */
+				$request    = $container->get('request_stack')->getCurrentRequest();
+				$controller = $arrModule['controllers'][\Input::get('key')][0];
+				$strategy   = ($arrModule['controllers'][\Input::get('key')][1]) ?: 'inline';
```
aschempp: What do we need the strategy for? To allow for something like ESI?

qzminski: @Toflar can you remember why was it introduced?

Toflar: Just read the symfony docs...

src/Util/CsvImportUtil.php
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    /**
+     * Get the template
+     *
+     * @return Template
+     */
+    public function getTemplate()
+    {
+        if ($this->template === null) {
```
aschempp: we use yoda conditions (`null === $this->template`)

src/Controller/AbstractCsvImportController.php
qzminski: AFAIR to allow overriding of the services in child classes as this is an abstract class.

src/Util/CsvImportUtil.php
```diff
@@ -0,0 +1,428 @@
...
+     * Generate the separators for the template
+     *
+     * @return array
+     */
+    protected function generateSeparators()
+    {
+        $parsed = [];
+        $mapper = $this->getSeparatorsMapper();
+
+        foreach ($this->separators as $separator) {
```
aschempp: That looks weird. I think you should iterate over the `$mapper` array, otherwise custom mappers added in `getSeparatorsMapper` would not be added to `$parsed`

src/Controller/AbstractCsvImportController.php
aschempp: well the abstract class needs them as constructor arguments, I don't think you would need to override the dependencies at runtime…?

src/Resources/contao/classes/Backend.php
aschempp: I know what strategies are, but I don't think we need to support ESI or hinclude here…?

src/Controller/AbstractCsvImportController.php
qzminski: What if I would need a different connection for the child class? We could obviously take out the abstract class from the services list but then every child class would need to define those services again (DRY). They are like the default ones here with an option to override by using setters.

aschempp: I don't understand. You mean you would not call the parent constructor in a child class? If that is possible and would work, then the dependencies should not be in the constructor anyway (because they are optional). But I don't think that's the case?

src/Util/CsvImportUtil.php
qzminski: But then I would also have to make the ```getSeparatorsMapper()``` protected, right? It doesn't make much sense to keep it private in such case.

src/Util/CsvImportUtil.php
qzminski: Yeah, I don't feel like changing everything now for no reason. We'll keep that in mind for the future though.

src/Resources/config/services.yml
qzminski: I don't know if controllers should be private but I guess not. According to the docs ```Simply said: A service can be marked as private if you do not want to access it directly from your code.``` but I am not sure if it relates to the controllers.

src/Controller/AbstractCsvImportController.php
qzminski: I am not sure how exactly the sf2 works behind the scenes with it but you can certainly override a dependency if needed and that's why I have made it an abstract service. The abstract class needs certain dependencies to work so they are set by default in a constructor, still giving you the option to override them setters though.

https://symfony.com/doc/current/components/dependency_injection/parentservices.html#overriding-parent-dependencies

src/Resources/config/services.yml
aschempp: You're right, it needs to be public because the legacy code does not use real dependency injection.

src/Util/CsvImportUtil.php
aschempp: Yeah, that's a bit of complex code here. Are you intending to allow overriding the mapper array in child classes? Maybe the `$separator` property should simply already be an array of these (so it can completely be overriden using `setSeparators`?

src/Controller/AbstractCsvImportController.php
aschempp: Ok, so this relates to my question whether we need the abstract service at all. In the example, they're overriding a setter injection, not a constructor injection. That feels weird because the container would need to build service1 for constructor injection and then immediately throw it away by overriding it with service2. I think a compiler pass would be the right thing to solve this if really necessary.

Toflar: Basically, the abstract controller is here because of `runDefaultRoutine()` which will be pretty much the same for all csv importers in the back end :) However, they don't need to allow for setter injection, imho.

aschempp: I totally agree with the abstract class, but an abstract service definition is not really necessary imho?

src/Resources/contao/classes/Backend.php
Toflar: Most likely not, that's why `inline` is default. But why would we want to limit it?

src/Controller/AbstractCsvImportController.php
Toflar: I don't see any comment about the service definition here? Cannot follow.

aschempp: https://github.com/contao/core-bundle/pull/444/files/ace4aa3982e6d402fc324f5037bab02ac1a05191#r68539105

src/Resources/config/services.yml
Toflar: Ah different comment here :D Yeah it's pretty unlikely that I, myself, would use this service definition when writing my own service. Not needed imho, indeed. Makes ready the services.yml just more complicated for no real reason.

src/Util/CsvImportUtil.php
qzminski: Changed in 26aaf53.

src/Util/CsvImportUtil.php
```diff
@@ -241,20 +236,61 @@ public function setCallback(callable $callback)
...
+            }
+        }
+
         $this->separators = $separators;
     }
 
     /**
+     * Set default separators
+     */
+    private function setDefaultSeparators()
```
aschempp: this should be done in the constructor to always set a default value for the property

src/Resources/contao/classes/Backend.php
aschempp: Because adding it also means we need to actively support it, e.g. using ESI. And we just don't know that…

aschempp:
@contao/developers please review

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,356 @@
...
+        }
+
+        foreach ($files as &$file) {
+            $extension = pathinfo($file, PATHINFO_EXTENSION);
+
+            if ('csv' !== $extension) {
+                throw new \RuntimeException(sprintf($GLOBALS['TL_LANG']['ERR']['filetype'], $extension));
+            }
+
+            // Drop TL_ROOT when we got rid of FileUpload which uses it
```
Toflar: Should probably be prefixed by `TODO: `

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,356 @@
...
+                $data = $this->fetchData($uploader, $request->request->get('separator'), $callback);
+            } catch (\RuntimeException $e) {
+                $request->getSession()->getFlashBag()->add($e->getMessage());
+
+                return new RedirectResponse($request->getUri(), 303);
+            }
+
+            $this->connection->update(
+                $table,
+                [$field => serialize($data)],
```
Toflar: This doesn't quite feel right to me. What if my widget stores the data elsewhere? Then I cannot use this method anymore, right? Can we maybe abstract this a little? Maybe using a `callable` like in other methods here?

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,356 @@
...
+     */
+    private function prepareTemplate(Request $request, FileUpload $uploader, $allowLinebreak = false)
+    {
+        /** @var BackendTemplate|\stdClass $template */
+        $template = new BackendTemplate('be_csv_import');
+
+        $template->formId      = $this->getFormId($request);
+        $template->backUrl     = $this->getBackUrl($request);
+        $template->action      = $request->getRequestUri();
+        $template->fileMaxSize = $this->framework->getAdapter(Config::class)->get('maxFileSize');
```
Toflar: Just for the record: This is valid, as we're requiring php 5.5.

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,356 @@
...
+    /**
+     * Returns the back button and redirect URL.
+     *
+     * @param Request $request
+     *
+     * @return string
+     */
+    private function getBackUrl(Request $request)
+    {
+        return str_replace('&key='.$request->query->get('key'), '', $request->getRequestUri());
```
Toflar: Won't work if there's only `key` as parameter like e.g. `?key=`. This is new code, I think we should make use of the power the request class gives us (getting all params from query and unsetting `key`).

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,356 @@
...
+     * @return string
+     *
+     * @throws \RuntimeException
+     */
+    private function getDelimiter($separator)
+    {
+        $separators = $this->getSeparators(true);
+
+        if (!isset($separators[$separator])) {
+            throw new \RuntimeException($GLOBALS['TL_LANG']['MSC']['separator'][1]);
```
Toflar: Should be an `\InvalidArgumentException`.

src/Controller/BackendCsvImportController.php
aschempp: First of all, the methods are currently private, so they are not re-usable. But why would you use that method if you can just write those few lines yourself if they work differently?

src/Controller/BackendCsvImportController.php
aschempp: There will never be only `key` as parameter, because that would not work in the backend. I do understand the "do it perfectly", but I think that's over-engineered for this case. Very likely the request query parameters to not match what's actually in the URL…

src/Controller/BackendCsvImportController.php
aschempp: 🤔 two issues with that:
 1. `\InvalidArgumentException` is a `LogicException`, which is not really correct for user input.
 2. We're currently using `RuntimeException` to add flashbag messages, so that would not work anymore…

src/Controller/BackendCsvImportController.php
aschempp: that TODO would most likely live for years 😂

src/Controller/BackendCsvImportController.php
Toflar: Fine for me.

src/Resources/contao/widgets/ListWizard.php
```diff
@@ -115,117 +118,21 @@ public function generate()
 	 * @param DataContainer $dc
 	 *
 	 * @return string
+	 *
+	 * @throws \Exception
+	 * @throws \Contao\CoreBundle\Exception\RedirectResponseException
+	 *
+	 * @deprecated Deprecated since Contao 4.3 to be removed in 5.0. Use the BackendCsvImportController instead.
```
ausi: Shouldn’t this be “Use the contao.controller.backend_csv_import service instead”?

aschempp: Changed in https://github.com/contao/core-bundle/pull/444/commits/7ba5e323248780ad1d729674c7de0cce3d1677af

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,356 @@
...
+     * @param callable    $callback
+     * @param string      $table
+     * @param string      $field
+     * @param int         $id
+     * @param string|null $submitLabel
+     * @param bool        $allowLinebreak
+     *
+     * @return Response
+     */
+    protected function importFromTemplate(
```
leofeyer: CS violation

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,356 @@
...
+     * Returns an array of separators for the template.
+     *
+     * @param bool $allowLinebreak
+     *
+     * @return array
+     */
+    private function getSeparators($allowLinebreak = false)
+    {
+        $separators = [
+            self::SEPARATOR_COMMA     => [
```
leofeyer: CS violation

src/Controller/BackendCsvImportController.php
leofeyer: The comment can be removed. It is missing the `TODO: ` prefix and I guess we are all aware of the `TL_ROOT` deprecation.

leofeyer:
I actually don't understand why we need this change. Now we have rewritten two of many legacy widgets – why not all of them? We have had the exact same discussion in #355 and this was the summary:

> As discussed in Mumble on April 28th, this change is a mixture between legacy code and new code, which we cannot really befriend with.

What makes this PR different?

src/Controller/BackendCsvImportController.php
ausi: We can use the `contao.root_dir` parameter here.

leofeyer:
@qzminski Can you please give me push access for this PR?

https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

qzminski:
Just did.

leofeyer:
Thanks. 😉

src/Resources/contao/templates/backend/be_csv_import.html5
```diff
@@ -1,7 +1,18 @@
 <div id="tl_buttons">
     <a href="<?= ampersand($this->backUrl) ?>" class="header_back" title="<?= specialchars($GLOBALS['TL_LANG']['MSC']['backBTTitle']) ?>" accesskey="b"><?= $GLOBALS['TL_LANG']['MSC']['backBT'] ?></a>
 </div>
-<?= $this->message ?>
```
leofeyer: Why are you not using `Message::generate()` here? It does effectively the same as your code.

leofeyer:
I have replaced

```php
<?php if ($this->messages): ?>
    <div class="tl_message">
        <?php foreach ($this->messages as $type => $messages): ?>
            <?php foreach ($messages as $message): ?>
                <p class="tl_<?= $type ?>"><?= $message ?></p>
            <?php endforeach; ?>
        <?php endforeach; ?>
    </div>
<?php endif; ?>
```

with

```php
<?= Message::generate() ?>
```

which does exactly the same, doesn't it? Or am I missing something?

src/Controller/BackendCsvImportController.php
```diff
@@ -0,0 +1,360 @@
...
+     * @param string|null $submitLabel
+     * @param bool        $allowLinebreak
+     *
+     * @return Response
+     */
+    protected function importFromTemplate(callable $callback, $table, $field, $id, $submitLabel = null, $allowLinebreak = false)
+    {
+        $this->framework->initialize();
+
+        $request = $this->requestStack->getCurrentRequest();
```
leofeyer: `$request` can be `null` and we are not handling the case. What would be the right thing to do in this case?

aschempp: throw an exception? We can't handle that case, it cannot happen in the Contao backend.

qzminski:
I think the only difference with ```Message::generate()``` is that it gets the flash messages from the ```session``` top level service, whereas the initial code takes those messages from the session bound to the request object we use in controller. But I guess that session instance is always the same.

Also previously we fetched the messages by getting ```all()``` of them from flash bag whereas ```Message``` takes only those of Contao scope. But that should not be an issue here either.

Thus I'm fine with that change, just also drop line 214 in that case:

https://github.com/qzminski/contao-core-bundle/blob/1023d3c798d71bd037dc43e0eb0542f8ef7df03a/src/Controller/BackendCsvImportController.php#L214

aschempp:
Be aware that there is only one session. Subrequests will inherit the main session from the master request, so I guess @leofeyer is right with the changes.

leofeyer:
Big thanks @qzminski, @aschempp and @Toflar. This was a huge PR. 😄

PR #445, opened 2016-02-15
[RTM] Use symfony's flashbag instead of our own Message class

Toflar:


aschempp:
:+1: but you should not change coding style for legacy classes. Also, why not trigger the deprecated message if the class is loaded instead of for every method?

Toflar:
> Also, why not trigger the deprecated message if the class is loaded instead of for every method?

Because all methods are static.

> but you should not change coding style for legacy classes.

phpStorm did and I do not care about it that much really.

aschempp:
> Because all methods are static.

Just add the deprecated warning into the file before the class definition. As soon as the class is loaded by the class loader, no matter for what reason, the message will be triggered.

Toflar:
Yeah, but I prefer my solution so I'll leave it as is for now :smile:

discordier:
:+1: on the overall solution.

:-1: on changing code style on otherwise unchanged lines. It makes the PR hard to read as the change set is much bigger than it needs to be.

Toflar:
Fixed and sqashed.

discordier:
Thank you very much. It's much cleaner now.

aschempp:
Did you search the code if somewhere we're using `$_SESSION['MESSAGES']` directly? Pretty sure that developers do, maybe we should add an `ArrayAccess` mapping?

Toflar:
I think we should not support that. `Message` was introduced long ago and people should've used this since long time ago. I'm very much for providing BC if we only supported only the array modification in Contao 3 but not if we had a central class that was responsible for managing this.

> Did you search the code if somewhere we're using $_SESSION['MESSAGES'] directly?

No, unfortunately they're also not grouped in `MESSAGES`. They just use e.g. `$_SESSION['TL_RAW']`.

discordier:
I second that. Using the array was in Contao 2.x. In Contao 3 it was the Messages class and now in Contao 4 it is the FlashBag.

leofeyer:
The PR looks RTM to me.

@Toflar Can we agree on removing the deprecation notice in all methods that call `static::add()`? Otherwise the error is triggered twice (once in the initial method and once in `add()`).

Toflar:
> @Toflar Can we agree on removing the deprecation notice in all methods that call static::add()? Otherwise the error is triggered twice (once in the initial method and once in add()).

Fuck, did not notice that! Of course, yes! Feel free to change it if you want to merge it now, I'm working on the package manager :D

leofeyer:
Will do :smile:

aschempp:
I actually think the deprecation message should be at the top of the `Message` class like we do for other classes. https://github.com/contao/core-bundle/blob/c317f24c91a98e7bbfa9e4c10fe0033e6e171729/src/Resources/contao/library/Contao/String.php

discordier:
@aschempp No, the deprecation should be at every call as otherwise you will only get one notice for the first offender and therefore will never realize that your extension is using the class as well.
Imagine `isotope` being loaded first and calling the messages. Then you will get the deprecation message and when `MetaModels` calls the very same, I will never realize it as the class has already been loaded.

aschempp:
Neither will you with each call, because you will just get the same message repeatedly...

discordier:
But i can follow the stack trace where it originated from...?

leofeyer:
Merged in 1f695a4f4615b41255314f6370082903a5781d3d.

However, I had to remove all deprecation notices in 546f68b9a1bb87fb554f3b8fac631ef7afc72c68, because we must not deprecate a class if there is no alternative. And I don't see an alternative to the `generate()` and `generateUnwrapped()` methods. Do you?

Toflar:
I don't see why this was ever a task of `Message`. Adding HTML is the task of a template or wherever one consumes these messages. The deprecated messages MUST remain and the core bundle has to stop using them.

leofeyer:
That would mean a lot of duplicated code. We would have to copy and paste the `generate()` code everywhere the method is used. Not a good idea.

Toflar:
I see. I think making it a twig template and do a `<?= $this->twig->render('messages'); ?>` or so instead should be fine then. A service is a bit overkill imho.

leofeyer:
You mean `System::getContainer()->get('twig')->render('messages')`, right?

But the template would need to have access to `Message::getTypes()`, the session object and the `$_POST` variable. How would you solve this? Pass them every time the template is rendered?

Toflar:
> But the template would need to have access to Message::getTypes(), the session object and the $_POST variable. How would you solve this? Pass them every time the template is rendered?

Why? It's all in the flashbag now? :)

leofeyer:
No, it is not. Read the code :)

Toflar:
?

leofeyer:
[This](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/Message.php#L149-L188) is the code we have to rebuild in the template.

Toflar:
Yes, I think this should be possible using twig now?

Something like this (untested):

```twig
{% for flashMessage in app.session.flashbag.get('contao.frontend.raw') %}

    {{ flashMessage }}

{% endfor %}
```

I don't think that `$_POST` check is needed anymore but if you still need it just use

```twig
{% if app.request.request.all() is empty %}
```

leofeyer:
And how about the message types?

Toflar:
You get them one after the other? `contao.frontend.raw`, `contao.frontend.error` etc. You can also use variables in twig templates and loop over them to make it shorter :)

leofeyer:
Here's a Twig template that works:

```twig
{% set unwrapped = unwrapped|default(false) %}
{% set scope = scope|default(app.request.attributes.get('_scope')) %}
{% set types = ['error', 'confirm', 'new', 'info', 'raw'] %}

{% if not unwrapped %}
    <div class="tl_message">
{% endif %}

{% for type in types %}
    {% for message in app.session.flashbag.get('contao.' ~ scope ~ '.' ~ type) %}
        {% if type == 'raw' %}
            {{ message }}
        {%  else %}
            <p class="tl_{{ type }}">{{ message }}</p>
        {%  endif %}
    {% endfor %}
{% endfor %}

{% if not unwrapped %}
    </div>
{% endif %}
```

However, I don't like that I need to define the types array in the template. Any idea?

Toflar:
Why not?

Toflar:
They have been hardcoded ever since... see https://github.com/Toflar/core-bundle/blob/4c75b3b983a6b783c31dddc3c6a740c5d8b7d9cb/src/Resources/contao/library/Contao/Message.php#L226

leofeyer:
But the method can be overwritten in a child class to change the types.

See #490 for the follow-up PR.

PR #449, opened 2016-02-23
[RTM] Use monolog handler to write to Contao `tl_log` table

aschempp:
TODO:

* [x] Unit tests
* [x] PhpDocs
* [x] anonymize IP


leofeyer:
I really like this. However, it does not support dynamic categories (like `newsletter_<id>`), does it?

aschempp:
It does, but they will currently be sent to the log file.

aschempp:
PR is updated, please review @contao/developers 

The `config.yml` should now look something like this afaik (untested):

```yaml
# Monolog configuration
monolog:
    handlers:
        contao:
            type: service
            id:   contao.monolog.handler
        main:
            type:         fingers_crossed
            action_level: error
            handler:      nested
        nested:
            type:      rotating_file
            max_files: 10
            path:      "%kernel.logs_dir%/%kernel.environment%.log"
            level:     info
        console:
            type: console
```

Because the Contao handler is first, it will first try to log everything that has a `ContaoContext` to the table. If that fails, it will continue down the stack. It would also mean that Contao logs will **not** be placed in the files by default, but I think that's correct. It will be written to file if DB connection does not work or bubbling is activated.

Also, we're still adding channels for the Contao action. That means people could still add a custom handler for a channel (e.g. `contao_error`) to redirect those away from the table and into an external error reporting system. 

Pretty cool, eh? 😎

Toflar:
Yeah, looks cool!! Imho that's a very neat solution that should be finished :)

aschempp:
I've updated the PR with a slight adjustment I came across wile working on my presentation: The logger channel duplicates the `ContaoContext` action. Therefore, the `ContaoContext` action is now optional too and will be determined from logger channel if necessary. Example:

**Good:**
```php
System::getContainer()
    ->get('monolog.logger.contao_error')
    ->error(
         'Fehlermeldung',     
        ['contao' => ContaoContext::create( __METHOD__ )] 
    )
;
```

> Result: Logs to channel `contao_error` with level *error*. `tl_log.action` = `ERROR`


```php
System::getContainer()
    ->get('monolog.logger.contao_newsletter')
    ->debug(
         'Send email to mail@example.com', 
        ['contao' => ContaoContext::create( __METHOD__ , 'newsletter_17')] 
    )
;
```

> Result: Logs to channel `contao_newsletter` with level *debug*. `tl_log.action` = `newsletter_17`

**Bad:**
```php
System::getContainer()
    ->get('logger')
    ->info(
         'Fehlermeldung',     
        ['contao' => ContaoContext::create( __METHOD__ , ContaoContext::ERROR)] 
    )
;
```

> Result: Logs to channel `app` with level *info*. `tl_log.action` = `ERROR`

aschempp:
PR updated again. I realized that using channels for different "error levels" or categories is just wrong. In Monolog, each bundle is supposed to use it's own channel. We would work again this by using categories as channels, because people would be obliged to use our channels instead of one for their bundle.

aschempp:
I have added full unit tests for `ContaoTableProcessor`. The `ContaoContext` does not really make sense to be tested imho, as it's a stupid data container. For the `ContaoTableHandler`, I'm not sure if we can/should test it as everything is database-storage related.

aschempp:
I have slightly adjusted the implementation again to follow the standards for processors and formatters. Tested everything again and works perfectly.

The only problem I see is in our standard edition. If the database logging fails for some reason (e.g. table does not yet exist), the log entry is bubbling up to the next handler. Unfortunately, that's our "main" handler which does not log anything as long as there is no severity in the process. Not sure if that's something we should fix?

src/Monolog/ContaoTableHandler.php
```diff
@@ -0,0 +1,195 @@
...
+    {
+        if (null !== $this->statement) {
+            return true;
+        }
+
+        if (null === $this->container || !$this->container->has($this->dbalServiceName)) {
+            return false;
+        }
+
+        try {
```
leofeyer: The try-catch block feels wrong in a method called "canWriteToDb". Why is this method needed in the first place? Could the statement not be prepared in the `write()` method, too, which is already wrapped in a try-catch block?

leofeyer:
Except for my comment above, the PR looks good to me.

aschempp:
What about my latest comment?

src/Monolog/ContaoTableHandler.php
aschempp: The `try-catch` in `canWriteToDb` is to catch potential errors when creating the statement (which is only done once). The other is on write errors. Just be aware that `write()` does not have a return statement, so we can't easily move all code from `canWriteToDb` into the `write` method (which must be there for the parent abstract…)

leofeyer:
I don't think we need to handle this case. The DB connection not yet being available implies that the application has not yet been completely installed, doesn't it?

aschempp:
Yeah, or something else failing, like the database server being unavailable. It would have been a cool feature if logging would then still work (by writing to a file). Also, if we ever move the logging to real files, most of the logs would never be written because our application ignores everything except on errors?

leofeyer:
Do we really need to worry about that right now? Because that's definitely something we should discuss, but not before we are moving the logging IMHO.

aschempp:
fine with me :-)

leofeyer:
Merged in e6f51e88a22667e53539a3bad5a89e7ee979f066.

PR #450, opened 2016-02-23
[RTM] Allow sub-subpalettes in DC_Table

aschempp:


leofeyer:
I like it. It does not yet support "edit all". Are you going to complete it?

leofeyer:
Are you? :smile:

aschempp:
You can assign it to me, but idk when I'll find the time…

leofeyer:
It should be no more than 10 minutes work, you have told me :smile:

leofeyer:
There is also a bug in the implementation. If the inner palette is closed, the outer one can be opened and closed as usual. However, if the inner palette is open and the outer one is closed, the outer palette is not loaded correctly when opened again (page reload required).

See https://www.dropbox.com/s/z0hrq5k1kcv9tig/contao-palettes.mov?dl=0

aschempp:
I have fixed the issue 😎

leofeyer:
So is the pull request RTM now?

aschempp:
It worked in all cases I tested. But please someone else test it :D

Toflar:
Is there any place where it might be useful in the core bundle itself?

leofeyer:
Probably yes. I needed it once but then implemented a workaround. Don't remember though.

aschempp:
This should be tagged for 4.2 again, right?

leofeyer:
Merged in ca2b1ce43f310b8df56a7f2c407a1b87458eda4f.

PR #453, opened 2016-03-01
Fixed the flash messages not being rendered in DC drivers

qzminski:
The only parameter in ```Message::generate()``` is now scope.

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/Message.php#L129


leofeyer:
Merged in 0d4f39403bee580384ad478328193152fc1f1842. Thanks @qzminski :)

PR #455, opened 2016-03-03
[WIP] Autoloading of system/modules

aschempp:
This is a very early draft to show autoloading of `system/modules` using https://github.com/mmoreram/symfony-bundle-dependencies

The great thing about this is that it allows for full flexibility for developers while providing easy installation for regular users. If you do not want bundle autoloading at all, simply do not use the trait. If you do not want autoloading of `system/modules`, simply do not load the `ContaoModuleAutoloadBundle`.

For a regular user though, this means using composer to install a legacy extension (like currently possible) will **automatically** enable it like in all previous Contao versions. I think that's a big step forward to the acceptance of Contao 4.
### Testing

The following changes to your installation are required to test this. I have added **one** module to `system/modules` and it was automatically loaded (see tasks about multiple modules).
1. Adjust your `composer json`
   
   ``` json
   "require": {
       "contao/core-bundle": "dev-bundle-dependencies as 4.2.9999"
   },
   "repositories": [
       {
           "type": "vcs",
           "url": "https://github.com/aschempp/contao-core-bundle.git"
       }
   ],
   "minimum-stability": "dev",
   "prefer-stable": true
   ```
2. Adjust your `AppKernel.php`
   
   ``` php
   {
       global $loader;
   
       if ($loader instanceof \Composer\Autoload\ClassLoader) {
           $loader->addPsr4('Contao\CoreBundle\HttpKernel\Bundle\\', $this->getCacheDir() . '/contao/bundles');
       }
   
       $bundles = [
           'Lexik\Bundle\MaintenanceBundle\LexikMaintenanceBundle',
           'Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle',
           'Contao\CoreBundle\ContaoCoreBundle',
           'Contao\CalendarBundle\ContaoCalendarBundle',
           'Contao\CommentsBundle\ContaoCommentsBundle',
           'Contao\FaqBundle\ContaoFaqBundle',
           'Contao\ListingBundle\ContaoListingBundle',
           'Contao\NewsBundle\ContaoNewsBundle',
           'Contao\NewsletterBundle\ContaoNewsletterBundle',
           'Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleAutoloadBundle',
       ];
   
       return $this->getBundleInstances($this, $bundles);
   }
   ```
3. Update composer dependencies (obviously)
### Open questions / tasks
- [x] `ContaoModuleBundle` should parse `autoload.ini` to add dependencies of a module. Probably restore our drafts from early 2015 to parse the ini-file dependencies (see https://github.com/contao/core-bundle/tree/feature/symfony2/src/Autoload).
- [x] Discuss if we should keep the core bundles (like `FrameworkBundle`) in the `AppKernel`. They **must** be in `ContaoCoreBundle` though, otherwise the order of loading would not be correct.
- [x] Check if the current dependency resolver library does correctly resolve multiple times the same class with different constructor arguments.

### DIFF
--- a/src/DependencyInjection/Compiler/AddResourcesPathsPass.php
+++ b/src/DependencyInjection/Compiler/AddResourcesPathsPass.php
@@ -41,7 +41,7 @@ class AddResourcesPathsPass implements CompilerPassInterface
         $rootDir = dirname($container->getParameter('kernel.root_dir'));
 
         foreach ($container->getParameter('kernel.bundles') as $name => $class) {
-            if ('Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle' === $class) {
+            if (is_a($class, 'Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle', true)) {
                 $paths[] = sprintf('%s/system/modules/%s', $rootDir, $name);
             } elseif (null !== ($path = $this->getResourcesPathFromClassName($class))) {
                 $paths[] = $path;
--- a/src/HttpKernel/Bundle/ContaoModuleAutoloadBundle.php
+++ b/src/HttpKernel/Bundle/ContaoModuleAutoloadBundle.php
@@ -11,7 +11,6 @@
 namespace Contao\CoreBundle\HttpKernel\Bundle;
 
 use Mmoreram\SymfonyBundleDependencies\DependentBundleInterface;
-use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 use Symfony\Component\HttpKernel\KernelInterface;
 
@@ -28,6 +27,12 @@ final class ContaoModuleAutoloadBundle extends Bundle implements DependentBundle
     public static $cacheDir;
 
     /**
+     * @var ModuleBundleGenerator
+     */
+    public static $generator;
+
+
+    /**
      * @inheritdoc
      */
     public static function getBundleDependencies(KernelInterface $kernel)
@@ -36,42 +41,19 @@ final class ContaoModuleAutoloadBundle extends Bundle implements DependentBundle
             static::$cacheDir = $kernel->getCacheDir() . '/contao/bundles';
         }
 
-        $kernelDir  = $kernel->getRootDir();
-        $bundles    = [];
+        if (!static::$generator instanceof ModuleBundleGenerator) {
+            static::$generator = new ModuleBundleGenerator();
+        }
 
-        $generator = new ModuleBundleGenerator();
-        $generator->generateBundles(static::$cacheDir, $kernelDir);
+        $rootDir = $kernel->getRootDir();
+        $bundles = [];
 
-        foreach (static::getContaoModules(dirname($kernelDir)) as $module) {
-            $bundles[] = sprintf('Contao\CoreBundle\HttpKernel\Bundle\%sModuleBundle', Container::camelize($module));
+        static::$generator->generateBundles(static::$cacheDir, $rootDir);
+
+        foreach (static::$generator->getContaoModules($rootDir) as $module) {
+            $bundles[] = ModuleBundleGenerator::convertModuleToClass($module);
         }
 
         return $bundles;
-    }
-
-    /**
-     * Find Contao modules in system/modules that are not marked as skippable
-     *
-     * @param string $contaoRoot
-     *
-     * @return array
-     */
-    private static function getContaoModules($contaoRoot)
-    {
-        $modules = [];
-
-        foreach (scandir($contaoRoot . '/system/modules') as $dir) {
-            if ('.' === $dir || '..' === $dir) {
-                continue;
-            }
-
-            if (is_dir($contaoRoot . '/system/modules/' . $dir)
-                && !file_exists($contaoRoot . '/system/modules/' . $dir . '/.skip')
-            ) {
-                $modules[] = $dir;
-            }
-        }
-
-        return $modules;
     }
 }
--- a/src/HttpKernel/Bundle/ContaoModuleAutoloadBundle.php
+++ b/src/HttpKernel/Bundle/ContaoModuleAutoloadBundle.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2016 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\HttpKernel\Bundle;
+
+use Mmoreram\SymfonyBundleDependencies\DependentBundleInterface;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * Automatically register all legacy modules as dependencies.
+ *
+ * @author Andreas Schempp <https://github.com/aschempp>
+ */
+final class ContaoModuleAutoloadBundle extends Bundle implements DependentBundleInterface
+{
+    /**
+     * @inheritdoc
+     */
+    public static function getBundleDependencies(KernelInterface $kernel)
+    {
+        $kernelDir  = $kernel->getRootDir();
+        $bundles    = [];
+
+        foreach (static::getContaoModules(dirname($kernelDir)) as $module) {
+            $bundles[] = ['Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle', [$module, $kernelDir]];
+        }
+
+        return $bundles;
+    }
+
+    /**
+     * Find Contao modules in system/modules that are not marked as skippable
+     *
+     * @param string $contaoRoot
+     *
+     * @return array
+     */
+    private static function getContaoModules($contaoRoot)
+    {
+        $modules = [];
+
+        foreach (scandir($contaoRoot . '/system/modules') as $dir) {
+            if ('.' === $dir || '..' === $dir) {
+                continue;
+            }
+
+            if (is_dir($contaoRoot . '/system/modules/' . $dir)
+                && !file_exists($contaoRoot . '/system/modules/' . $dir . '/.skip')
+            ) {
+                $modules[] = $dir;
+            }
+        }
+
+        return $modules;
+    }
+}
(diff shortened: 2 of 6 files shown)


aschempp:
@contao/developers no feedback?

src/HttpKernel/Bundle/ContaoModuleAutoloadBundle.php
```diff
@@ -0,0 +1,64 @@
...
+{
+    /**
+     * @inheritdoc
+     */
+    public static function getBundleDependencies(KernelInterface $kernel)
+    {
+        $kernelDir  = $kernel->getRootDir();
+        $bundles    = [];
+
+        foreach (static::getContaoModules(dirname($kernelDir)) as $module) {
```
leofeyer: Does this mean that every module using this class has **every** non-skippable module as dependency?

leofeyer:
I have commented the file changes. Why do we need the `ContaoModuleAutoloadBundle` class?

src/HttpKernel/Bundle/ContaoModuleAutoloadBundle.php
aschempp: > every module using this class

What do you mean by that?

aschempp:
> Why do we need the `ContaoModuleAutoloadBundle` class?

Because we can't have the core module add these as dependencies, that would result in the dependencies being loaded **before** core bundle.

leofeyer:
Let's discuss this in the next call :)

leofeyer:
https://github.com/contao/core-bundle/tree/feature/symfony2

aschempp:
I have update the pull request. After investigating the options of the same bundle with different constructor argument, I decided to go with the bundle class generator. That's the only option which will work without completely rewriting `mmoreram/symfony-bundle-dependencies`. I think it would be wrong to create our own solution for bundle dependencies.

So the loading works perfectly as expected. Unfortunately, the resolving of `mmoreram/symfony-bundle-dependencies` does not seem to work correctly on multiple dependencies. I'll need to create a PR for `mmoreram/symfony-bundle-dependencies` with our known resolving code, which would profit everyone.

Plz let me know your thoughts @contao/developers

src/HttpKernel/Bundle/ContaoModuleBundle.php
```diff
@@ -35,5 +55,84 @@ public function __construct($name, $rootDir)
         if (!is_dir($this->path)) {
             throw new \LogicException('The module folder "system/modules/' . $this->name . '" does not exist.');
         }
+
+        if (file_exists($this->path . '/.skip')) {
+            throw new \RuntimeException('The module "system/modules/' . $this->name . '" as been disabled.');
```
Toflar: `has`

src/HttpKernel/Bundle/ContaoModuleBundle.php
```diff
@@ -35,5 +55,84 @@ public function __construct($name, $rootDir)
...
+
+        if (false === $autoload || !isset($autoload['requires']) || !is_array($autoload['requires'])) {
+            return $requires;
+        }
+
+        return array_unique(array_merge($requires, $autoload['requires']));
+    }
+
+
+    private static function convertModuleToClass($name)
```
Toflar: Comment missing.

Toflar:
Lots of CS stuff but I guess this is POC anyway :) Remarks:

- Wouldn't it be better to generate the classes when the container is built? Or does this not work?
- As we were also thinking about generating classes in the future maybe for doctrine entities, why not usehttps://github.com/zendframework/zend-code for generating the class, as this is properly unit tested etc.?


> I'll need to create a PR for mmoreram/symfony-bundle-dependencies with our known resolving code, which would profit everyone.

Yes!

aschempp:
> Wouldn't it be better to generate the classes when the container is built? Or does this not work?

That won't work as the bundles are loaded before the container or even the kernel is ready ;-)

Toflar:
> That won't work as the bundles are loaded before the container or even the kernel is ready ;-)

Makes sense :-D But currently the classes are recreated on every request. Don't you think that's something a command should do? Just like `contao:install`, `contao:symlink` we'd have a `contao:dump-dependency-classes` or so?

src/DependencyInjection/Compiler/AddResourcesPathsPass.php
```diff
@@ -41,7 +41,7 @@ private function getResourcesPath(ContainerBuilder $container)
         $rootDir = dirname($container->getParameter('kernel.root_dir'));
 
         foreach ($container->getParameter('kernel.bundles') as $name => $class) {
-            if ('Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle' === $class) {
+            if (is_a($class, 'Contao\CoreBundle\HttpKernel\Bundle\ContaoModuleBundle', true)) {
```
leofeyer: `is_a()` is deprecated in favor of the `instanceof` operator and will trigger an `E_STRICT` warning.

leofeyer:
> After investigating the options of the same bundle with different constructor argument, I decided to go with the bundle class generator.

When we discussed it, however, we explicitly decided against class generation.

src/DependencyInjection/Compiler/AddResourcesPathsPass.php
Toflar: Not true. `is_a()` is not deprecated, neither will `instanceof` work on a string.

aschempp: It has been un-deprecated in PHP 5.3 or so…

aschempp:
> When we discussed it, however, we explicitly decided against class generation.

I'm very well aware of that, but it does not work the other way. See my comment in https://github.com/mmoreram/symfony-bundle-dependencies/pull/11#issuecomment-213040727, I very much doubt that will happen.

src/DependencyInjection/Compiler/AddResourcesPathsPass.php
leofeyer: > It has been un-deprecated in PHP 5.3 or so…

You are right.

> neither will `instanceof` work on a string.

I don't understand this argument. `is_a()` requires an object as first argument, so it will not work on a string, either.

aschempp: Just read the docs and you will understand ;-)
http://php.net/is_a

leofeyer: I have read it and I still don't understand @Toflar's comment. The first argument has to be an object, hasn't it?

aschempp: The third method argument is set to `true` in my code
> **allow_string**
If this parameter set to FALSE, string class name as object is not allowed. This also prevents from calling autoloader if the class doesn't exist.

leofeyer: Ahhhhh 😄

leofeyer:
As discussed in Mumble on April 28th, @aschempp will create a new PR for `mmoreram/symfony-bundle-dependencies`. Once this PR is merged, he will complete this PR.

src/HttpKernel/Bundle/ModuleBundleGenerator.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    private function getContaoModules($modulesDir)
+    {
+        $modules = [];
+
+        foreach (scandir($modulesDir) as $dir) {
+            if ('.' === $dir || '..' === $dir) {
+                continue;
+            }
+
```
ausi: To prevent syntax errors and security issues, we should only allow directories with certain characters IMO, like:

```php
if (preg_match('/[^a-z0-9_-.]/i', $dir)) {
	continue;
}
```

src/HttpKernel/Bundle/ModuleBundleGenerator.php
```diff
@@ -0,0 +1,103 @@
...
+                $modules[] = $dir;
+            }
+        }
+
+        return $modules;
+    }
+
+    private function createBundleClass($module, $cacheDir, $rootDir)
+    {
+        $className = Container::camelize($module) . 'ModuleBundle';
```
ausi: `Container::camelize()` doesn’t support dashes, maybe use something like this:

```php
$className = Container::camelize(strtr($module, ['-' => '_'])) . 'ModuleBundle';
```

src/HttpKernel/Bundle/ModuleBundleGenerator.php
```diff
@@ -0,0 +1,103 @@
...
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * Generated bundle class for Contao module in system/modules/$module
+ */
+class $className extends ContaoModuleBundle
+{
+    public function __construct()
+    {
+        parent::__construct('$module', '$rootDir');
```
ausi: `$rootDir` could contain special characters like `'` which would break this code. IMO starting this method with:

```php
$rootDir = var_export($rootDir, true);
```

and replacing this line with:

```php
parent::__construct('$module', $rootDir);
```

should be safer.

src/HttpKernel/Bundle/ModuleBundleGenerator.php
```diff
@@ -0,0 +1,103 @@
...
+
+        $modules = $this->getContaoModules(dirname($rootDir) . '/system/modules');
+
+        foreach ($modules as $module) {
+            $this->createBundleClass($module, $cacheDir, $rootDir);
+        }
+    }
+
+    /**
+     * Find Contao modules in system/modules that are not marked as skippable
```
ausi:
```diff
- Find Contao modules in system/modules that are not marked as skippable
+ Find Contao modules in system/modules
```

src/HttpKernel/Bundle/ModuleBundleGenerator.php
aschempp: we're not doing anything like that since ever, I don't think we should start adding limitations on the folder name?
see https://github.com/contao/core/blob/master/system/modules/core/library/Contao/ModuleLoader.php#L104

ausi: If we want to support all characters in module directories we should normalize it for the conversion into a class name before camelizing it, e.g.:

```php
preg_replace('/[^a-z0-9]+/i', '_', $module);
```

aschempp:
I have update this PR and changed @ausi's recommendations.

leofeyer:
I guess we can close this PR now that we have the Contao manager?

aschempp:
I guess so too 😎

PR #462, opened 2016-03-21
Fix the pretty error screen exception handling

qzminski:
Consider the following code:

```php
// NewsNotFoundException.php
class NewsNotFoundException extends PageNotFoundException {}

// NewsListModule.php
if ($newsModel === null) {
    throw new NewsNotFoundException();
}
```

Normally this would result in a pretty 500 screen message because you checked strictly for the exception class name. I think you should check for the instanceof instead so we can define our own exceptions that extend ```PageNotFoundException``` and thus result in pretty 404 screen message.

Basically the issue was in ```ExceptionConverterListener``` but I have also found similar snippet in ```PrettyErrorScreenListener```.

### DIFF
--- a/src/EventListener/ExceptionConverterListener.php
+++ b/src/EventListener/ExceptionConverterListener.php
@@ -51,14 +51,14 @@ class ExceptionConverterListener
     public function onKernelException(GetResponseForExceptionEvent $event)
     {
         $exception = $event->getException();
-        $class = get_class($exception);
 
-        if (!isset($this->mapper[$class])) {
-            return;
-        }
-
-        if (null !== ($httpException = $this->convertToHttpException($exception, $this->mapper[$class]))) {
-            $event->setException($httpException);
+        foreach ($this->mapper as $origin => $dest) {
+            if ($exception instanceof $origin) {
+                if (null !== ($httpException = $this->convertToHttpException($exception, $dest))) {
+                    $event->setException($httpException);
+                }
+                break;
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


Toflar:
:+1: Definitely want to see that feature but it needs unit test coverage :)

leofeyer:
Isn't this a bugfix actually?

discordier:
AFAIR in my initial PR I was walking all class parents.
Don't remember the reason why this was changed.

leofeyer:
Your code is still there:

https://github.com/contao/core-bundle/blob/master/src/EventListener/PrettyErrorScreenListener.php#L218-L221

It just does not handle this particular case.

leofeyer:
But indeed, the changes to the `PrettyErrorScreenListener` class should not be necessary, because `renderErrorScreenByException()` bubbles up through the exception chain and will eventually find the parent exception. @qzminski Why did you change the class?

leofeyer:
Ah, forget my question. I have just realized that the method only checks the previous exceptions and not the parent classes.

leofeyer:
Merged in 45c5e1a91b34b20095acb0826b1627e2b2ebe21f.

discordier:
> Your code is still there:
> https://github.com/contao/core-bundle/blob/master/src/EventListener/PrettyErrorScreenListener.php#L218-L221
> But it only walks through the exception chain and does not check parent classes.

No, it isn't. My original is here:
https://github.com/discordier/core-bundle/blob/f747359e80f92047de2e0557648ac92f1a5dca58/src/EventListener/ExceptionListener.php#L273

In that state it was still rendering and handling of exceptions mangled but the template lookup was in there.

Issue #437, opened 2016-01-21
Deprecated the ClassLoader


aschempp:
`Contao\ClassLoader` should be deprecated and removed in Contao 5. We should use the Composer class loader, if necessary with a filemap.

Should be an easy-pick.

leofeyer:
+1

leofeyer:
It is not that easy a pick as we thought.

Using a class map implies that the classes are no longer aliased to the global namespace, which will effectively break everything. All content element and module classes extend from `\ContentElement` or `\Module` instead of `Contao\ContentElement` and `Contao\Module`. Many type hints expect a `\Model` instead of a `Contao\Model` and so on.

The class loader should be deprecated to be removed in Contao 5, but I don't think there is an alternative in Contao 4.

leofeyer:
Of course we could add a legacy class loader which automatically adds the global aliases when needed :)

aschempp:
The `ClassLoader` must be deprecated for all extensions. Noone should ever use it. We obviously need to keep it for BC right now.

leofeyer:
I have created a POC in #466.

leofeyer:
The PR #466 has been merged in 42661a616297b030f2e0952e7eaeca77ae8579a4.

antoinedavid:
Hi,
With contao 4.2.4 the composer autoloader dont generate the good path or i miss something.

Here my composer.json :
`{
    "name": "addictic/orphanews-bundle",
    "type": "contao-bundle",
    "description": "Addictic orphanews bundle",
    "license": "LGPL-3.0+",
    "authors": [
        {
            "name": "Addictic",
            "homepage": "https://www.addictic.fr"
        },
        {
            "name": "Antoine DAVID",
            "email": "antdvd@outlook.com"
        }
    ],
    "require": {
        "php": ">=5.5.0",
        "symfony/framework-bundle": "~2.8|~3.0",
        "contao/core-bundle": "^4.2.3"
    },
    "require-dev": {
		"doctrine/data-fixtures": "1.0.*",
		"doctrine/dbal": "~2.3",
		"doctrine/orm": "~2.3",
		"doctrine/doctrine-bundle": "~1.6"
    },
    "conflict": {
        "contao/core": "*"
    },
    "autoload": {
        "psr-4": {
            "Addictic\\OrphanewsBundle\\": "contao/orphanews-bundle/src/"
        },
        "classmap": [
            "src/Resources/contao"
        ],
        "exclude-from-classmap": [
[...]

PR #466, opened 2016-03-29
[RTM] Use the Composer autoloader to load the Contao classes

leofeyer:
This PR implements #437. @contao/developers Do we want it this way?


aschempp:
That could work. We should also add a `@deprecated` to the `ClassLoader` to make sure people stop using it.

Toflar:
And what about the `trigger_error()`?

leofeyer:
I have not yet added the deprecation stuff, because I was not sure if you like the implementation. So do you want me to complete the PR?

Toflar:
I think it looks good, yes.

aschempp:
How will it behave if someone is currently overriding the global name of a class?

leofeyer:
The PR is RTM now. Please review again.

leofeyer:
Merged in 42661a616297b030f2e0952e7eaeca77ae8579a4.

Issue #451, opened 2016-02-29
Use instanceof instead of !== null


leofeyer:
<a href="https://github.com/aschempp"><img src="https://avatars.githubusercontent.com/u/1073273?v=3" align="left" width="42" height="42"></img></a> [Comment](https://github.com/contao/core/pull/8238#discussion_r54377445) by @aschempp
February 29th, 2016, 07:57 GMT

A general idea. If we would write `($objTarget = $this->objModel->getRelated('jumpTo')) instanceof \PageModel`, it would be more safe, we would not need the `@var` hint and work exactly the same.

DanielSchwiperich:
Not directly related: What is the reason contao uses like in @aschempp example global namespace so often?
Afaik this is considered bad practice

aschempp:
Global namespaces were introduced in Contao 3 but are deprecated since Contao 4. They ARE bad practice, but we need to support them in the core for BC reasons.

leofeyer:
We are no longer supporting them in Contao 4, though (see [DEPRECATED.md](https://github.com/contao/core-bundle/blob/master/DEPRECATED.md#contao-class-loader)).

DanielSchwiperich:
But they are still used a lot and I see no reason why (can't see how it would affect BC)

Just 2 quick examples: 
https://github.com/contao/core-bundle/blob/master/src/Resources/contao/classes/Ajax.php
`class Ajax extends \Backend`
`$objSessionBag = \System::getContainer()->get('session')->getBag('contao_backend');`
(losts more in this file)

https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/classes/FrontendTemplate.php
`class FrontendTemplate extends \Template`
`if ($blnCheckRequest && \Input::hasUnusedGet())`

I saw it also in a recent commit:
https://github.com/contao/core-bundle/commit/a4656c5ac35810e4bc0a61346a08003fc58f7980
But you did fix this in a later commit. 

So is it just legacy code which will be refactored as time goes by? 
Would you accept external PRs for it? I could let our trainee work in it. I think it would be a good job for someone to learn the core and also doing something productive (we would pre check before the PR for sure)

aschempp:
As far as I remember, we did not want to rewrite all of the legacy code (yet) because it would create a lot of noise in the commit messages (e.g. `git blame`). But I like the idea of having a trainee get something done :-)

[...]

PR #467, opened 2016-03-30
[RTM] Use instanceof to check the return value of Model::getRelated()

leofeyer:
This PR implements #451. @contao/developers Please review.


aschempp:
According to what I saw in Symfony code, we should not add brackets around the `instanceof` statement :smile:

leofeyer:
It works without, because `instanceof` has a higher precedence than `!`, but the PHP manual suggest to use the more explicit statement: http://php.net/manual/en/language.operators.type.php#example-144

leofeyer:
Merged in a10a30783645f1270306d843071d833b67b29475.

Issue #442, opened 2016-02-04
Do not store non-configuration values into `localconfig.php`


Toflar:
Configuration values must be static. If these files get updated by a cronjob or whatever, it makes deployment a pain because you want to be able to track the changes of configuration files in a VCS like git.

The current version of the Contao bundle stores e.g. the latest version of Contao that is available into the `localconfig.php` (the key is named `latestVersion`). But this information is not static so it either belongs to the database or a temporary file. 
A quick search resulted in another example where we make that mistake: `installCount`.

These two non-configuration parameters must be removed from the `localconfig.php`. In the case of `latestVersion` I'd even vote to get rid of it completely but I'm open for alternative proposals.

discordier:
I am totally with you on this one. The configuration must be and remain static.
I seem to remember that we had this discussion during Ahrweiler-2nd-edition and came to the conclusion that we ultimately want to get rid of localconfig at all and only want parameters in the future.

However on the topic of the `latestVersion`, I do not see any benefit of this at all aside from confused customers which get the warning in the backend that their Version is not the most recent one. This bears especially the problem that NO checks are done if upgrading will break everything (with ER2) or is impossible (with composer). It simply warns that the core is not up to date and therefore is misleading at best and issuing false assumptions at worst.
Therefore I also vote for getting rid of it completely.

leofeyer:
We can easily store `latestVersion` in the database. But `installCount` will be difficult, because we cannot safely assume that there is a database connection in the install tool.

Toflar:
Imho both could be stored into the tmp folder instead of the localconfig.php. If the tmp folder gets purged, install count would be reset and the new version available message would temporarily be gone until the next cron runs. Both not critical? I don't see why we should pollute the DB.

qzminski:
In my opinion informations like ```installCount``` or ```latestVersion``` or any other of this kind should not be stored in the database but in a separate file that is accessible independently from the system status (e.g. whether it is installed or not; whether it has db connection or not). That control file would not be maintained in the repository and would be for internal app's usage only.

The ```localconfig.php``` file should be in the end moved to the ```parameters.yml```.

I am also for dropping the ```latestVersion``` or at least not displaying it in the backend (perhaps a sf2 debug bar would be more appropriate place?).

aschempp:
Yeah. But we should use a generic Doctrine cache service instead of a custom implementation ;-)

leofeyer:
[...]

PR #468, opened 2016-03-31
[RTM] Use the Doctrine cache to store the latest version

leofeyer:
This PR is an enhancement to #442.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "psr/log": "~1.0",
         "twig/twig": "~1.20",
         "doctrine/doctrine-bundle": "^1.5.2",
+        "doctrine/doctrine-cache-bundle": "~1.3",
         "doctrine/annotations": "^1.2.7",
         "doctrine/cache": "^1.3.2",
         "doctrine/common": "^2.5.1",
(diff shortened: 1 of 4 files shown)


aschempp:
We don't need the cache bundle, we just need the library

leofeyer:
As discussed, we need the bundle to have the console commands available. E.g.

```
app/console doctrine:cache:clear contao.cache
```

leofeyer:
Merged in aa5449a1becabc3bf7eb545edb4c4ed8e26b9625 then.

discordier:
> As discussed, we need the bundle to have the console commands available. E.g.

This does belong to the application then, not the bundle, the bundle does not need nor know about the commands of another bundle.

leofeyer:
It is a bit strange if the bundle suggest running a console command and at the same time does not require the bundle providing the command, isn't it?

discordier:
Not really, I mean for these occasions there exists a complete "suggests" section in the `composer.json`.
I suggest you do something but I do not require you to do it. You can still simply remove the cache dir by hand if you want to completely omitting the "bloat" of the bundle which might cause conflicts with another bundle or package.

You should require stuff you use classes from and suggest packages that bring more user experience etc.

aschempp:
As suggested in Slack, I think we should actually put a link to the manual, which can describe the available options in-depth.

leofeyer:
@discordier's argument is about the Composer requirement.

discordier:
Exactly, but now I wonder to which "options" @aschempp is referring to.

Issue #236, opened 2015-04-24
Remove the automated logout module


Toflar:
It's wrong that this is a front end module because it doesn't make sense. It should be a custom page type.

aschempp:
Not sure, it will be handled by security anyway later on (see #122)

Toflar:
Yeah, whatever - just not a FE module :D

leofeyer:
Will you be working on this issue for Contao 4.1 or shall I change the milestone to 4.x.x?

aschempp:
I don't think this would be correct to add in 4.1. We can't remove the frontend module in a minor release. And adding a page type would probably not be the correct solution in regards of the Symfony security component.

leofeyer:
We would not have to remove the front end module. We could also just add the new page type and mark the front end module as deprecated.

We should discuss the Symfony context though.

leofeyer:
I have created a PR: #469

PR #469, opened 2016-04-05
[RFC] Add a logout page as an alternative to the logout module

leofeyer:
This PR implements #236. @contao/developers Please review.

(The missing page icon will be added automatically when we merge the back end icons branch.)

### DIFF
--- a/src/Resources/contao/pages/PageLogout.php
+++ b/src/Resources/contao/pages/PageLogout.php
@@ -1,0 +1,60 @@
+<?php
+
+/**
+ * Contao Open Source CMS
+ *
+ * Copyright (c) 2005-2016 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao;
+
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Response;
+
+
+/**
+ * Provide methods to handle a logout page.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class PageLogout extends \Frontend
+{
+
+	/**
+	 * Return a redirect response object
+	 *
+	 * @param PageModel $objPage
+	 *
+	 * @return Response
+	 */
+	public function getResponse($objPage)
+	{
+		// Set last page visited
+		if ($objPage->redirectBack)
+		{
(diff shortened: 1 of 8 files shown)


qzminski:
Just out of curiosity, why do you still use the legacy code for the new features?

leofeyer:
How else would you add a new page type?

qzminski:
I do not particularly mean the logic behind it but the coding style and not using the DIC.

leofeyer:
Because it is legacy code. New legacy code :smile:

DanielSchwiperich:
regarding legacy code, it maybe would a good idea to not use global namespace for new classes like also mentioned here https://github.com/contao/core-bundle/issues/451#issuecomment-190123917 by yourself that they are deprecated ;)

not sure if that was also considered *legacy code* by @qzminski

qzminski:
The new legacy code works but is a short-term solution. When the time comes the ```tl_page``` class will have to be converted into ```Contao\DataContainer\PageContainer``` (or whatever you call it) and thus you will have to rework even more than just the old Contao 3 code.

As @DanielSchwiperich has pointed out the ```PageLogout``` could also be moved to the correct namespace already. No need for legacy here as it is a new feature with a new class. The page types registry of course has to remain as is in this PR because it's irrelevant here.

Why not do it now and just slightly adjust (e.g. dependencies) in the future if needed? The logic should not change after all, right? Same thing about CS and variable type-prefixes ```$objPage```.

src/Resources/contao/pages/PageLogout.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+class PageLogout extends \Frontend
+{
+
+	/**
+	 * Return a redirect response object
+	 *
+	 * @param PageModel $objPage
+	 *
+	 * @return Response
```
qzminski: I think it should be ```RedirectResponse``` to be precise.

leofeyer:
Please accept the fact that we will be adding new features to the old code. Reworking the code is a long term goal and not every feature can be postponed until then.

> the ```PageLogout``` could also be moved to the correct namespace already.

If we knew today what the "correct namespace" was, I would do it. But right now, we don't even know how pages will be handled in the future, so there is no way of knowing the "correct namespace".

leofeyer:
Merged in 6712cb9eaaf84135cd93b78702633d06ac566f67.

PR #471, opened 2016-04-08
[RTM] Extract bundle code

aschempp:
There are various places where the current core bundle that is still related to other bundles that must not necessarily be installed.

**Corresponding PRs:**
NewsBundle: https://github.com/contao/news-bundle/pull/4
CalendarBundle: https://github.com/contao/calendar-bundle/pull/3
FaqBundle: https://github.com/contao/faq-bundle/pull/2

If we can quickly agree on the general approach (listener with DI for hook callback etc.) I will continue migration other stuff like insert tags.

### DIFF
--- a/src/Resources/contao/classes/Backend.php
+++ b/src/Resources/contao/classes/Backend.php
@@ -728,31 +728,6 @@ abstract class Backend extends \Controller
 							  ->execute($intPid);
 				break;
 
-			case 'tl_news':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT jumpTo FROM tl_news_archive WHERE id=(SELECT pid FROM tl_news WHERE id=?))")
-							  ->execute($intPid);
-				break;
-
-			case 'tl_news_archive':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT jumpTo FROM tl_news_archive WHERE id=?)")
-							  ->execute($intPid);
-				break;
-
-			case 'tl_calendar_events':
-				$objPage = $db->prepare("SELECT * FROM tl_page WHERE id=(SELECT jumpTo FROM tl_calendar WHERE id=(SELECT pid FROM tl_calendar_events WHERE id=?))")
-							  ->execute($intPid);
-				break;
-
-			case 'tl_calendar':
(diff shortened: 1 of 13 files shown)


Toflar:
Fine for me! Except for using our own `Database` class. This should not be done anymore.

aschempp:
Unfortunately the hook expects a `Database\Result` as return value, so there's no way around that. I don't think we should refactor the core class just for that.

aschempp:
I've extracted all remaining functionality. There are now several translations still in `core-bundle`, but afaik they will automatically be dropped on Transifex sync.

leofeyer:
Merged in 24167cf3871be7365274acc47f68079ba015d0e5.

@aschempp There are still some places where the core uses extension code:

* https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/library/Contao/InsertTags.php#L522-L530
* https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/dca/tl_member.php#L744-L755

I figure you left them in because you did not want to add new hooks, am I right?

aschempp:
Certainly true for the insert tag. We might be able to move the newsletter stuff by using a `save_callback`? Must have not noticed that…

leofeyer:
The `last_update` insert tag is a fail by design anyway, because it only supports three tables out of five or six possible ones and it is not extendable. We should fix this. On the other hand, the insert tag seems pretty useless to me in the first place ;)

Toflar:
It is pretty useless anyway, let's face it. Even if e.g. Isotope was able to modify this through an event, what's the point in showing the last modification date on a page that does not display any Isotope data? The insert tag might be used completely out of context of the last modification date. We can as well just drop it/let it show the last modification date of the articles only.

aschempp:
I'm pretty sure it was added because people want to show that. And they are currently using it, so we should not simply remove. But I would deprecated it for Contao 5 and not fix anything for now…

leofeyer:
Fixed the newsletter sync callback in f94f556e880025cf1aec9de697965cc250ad7143 and contao/newsletter-bundle@32976e97d9e895968d8d036f29ab0b6330ce822d.

aschempp:
Why are you adding code to the legacy classes and not create a correct listener? Want me to PR?

leofeyer:
It was not worth the effort IMHO. But feel free to create a PR :)

PR #473, opened 2016-04-08
[RTM] Recursively replace insert tags

leofeyer:
This PR implements contao/core#8172. It also allows us to remove the "nested insert tags" routines.

### DIFF
--- a/src/Resources/contao/library/Contao/InsertTags.php
+++ b/src/Resources/contao/library/Contao/InsertTags.php
@@ -52,7 +52,7 @@ class InsertTags extends \Controller
 			return \StringUtil::restoreBasicEntities($strBuffer);
 		}
 
-		$tags = preg_split('/{{(([^{}]*|(?R))*)}}/', $strBuffer, -1, PREG_SPLIT_DELIM_CAPTURE);
+		$tags = preg_split('/{{([^{}]+)}}/', $strBuffer, -1, PREG_SPLIT_DELIM_CAPTURE);
 
 		$strBuffer = '';
 
@@ -60,7 +60,7 @@ class InsertTags extends \Controller
 		static $arrItCache;
 		$arrCache = &$arrItCache[$blnCache];
 
-		for ($_rit=0, $_cnt=count($tags); $_rit<$_cnt; $_rit+=3)
+		for ($_rit=0, $_cnt=count($tags); $_rit<$_cnt; $_rit+=2)
 		{
 			$strBuffer .= $tags[$_rit];
 			$strTag = $tags[$_rit+1];
@@ -69,12 +69,6 @@ class InsertTags extends \Controller
 			if ($strTag == '')
 			{
 				continue;
-			}
-
-			// Run the replacement again if there are more tags (see #4402)
-			if (strpos($strTag, '{{') !== false)
-			{
-				$strTag = $this->replace($strTag, $blnCache);
 			}
 
 			$flags = explode('|', $strTag);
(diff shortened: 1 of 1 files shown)


leofeyer:
This is RTM to me. @Toflar What about you?

src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -52,15 +52,20 @@ public function replace($strBuffer, $blnCache=true)
 			return \StringUtil::restoreBasicEntities($strBuffer);
 		}
 
-		$tags = preg_split('/{{(([^{}]*|(?R))*)}}/', $strBuffer, -1, PREG_SPLIT_DELIM_CAPTURE);
+		$tags = preg_split('/{{([^{}]+)}}/', $strBuffer, -1, PREG_SPLIT_DELIM_CAPTURE);
+
+		if (count($tags) < 2)
+		{
+			return \StringUtil::restoreBasicEntities($strBuffer);
```
leofeyer: @Toflar This new return statement actually prevents endless loops now, so we do no longer need the nesting level counter.

Toflar: Ah true, why did I not think about this^^

Toflar:
Looks good, I think. Maybe @fritzmg can try wether this code solves his problem. Moreover, are there maybe other places where we can now skip replacing insert tags because they are recursive now?

leofeyer:
> Moreover, are there maybe other places where we can now skip replacing insert tags because they are recursive now?

I bet there are. Searching …

leofeyer:
Unexpectedly, I did not find any places where the `replaceInsertTags()` call can be omitted.

fritzmg:
> Maybe @fritzmg can try wether this code solves his problem.

Well actually I didn't experience a problem (directly) - @aschempp had the following problem originally: https://github.com/contao/core/issues/7097#issuecomment-126189860

I reinvestigated this because I was curious about whether insert tags are not replaced recursively in the first place and because it would have solved some other problems that were reported in the past by other users.

Toflar:
Ah, true. Then imho this should be fine.

leofeyer:
Merged in c05c02a0dc79015f8be1f7be19c6608cdb3af289.

leofeyer:
Unfortunately, it does not work if there are non-cacheable insert tags. `count($tags)` will be greater than 2 in this case and we are back in the endless loop. 😞

leofeyer:
Luckily, it was easy to fix: ab4434da67136c15ed28cb02cca869249199dcb7

aschempp:
Doesn’t that mean we will not recursive replace insert tags if the content is not cached?

fritzmg:
Hm, consider the following example:

Some content somewhere:
```
{{insert_content::123}}
```

Content of content element 123:
```
{{insert_content::124}}
```

When the page is generated for the cache, `{{insert_content::123}}` will be replaced by `{{insert_content::124}}`. But that insert tag will not be replaced anymore, if I understand the current code correclty, thus it will land in the cache and it also will not be replaced for the frontend output during cache generation, I think?

I think the only way to properly implement recursive insert tag replacing while allowing cacheable insert tags is with the checksum method that @Toflar originally thought up [here](https://github.com/contao/core/issues/8172#issuecomment-167105203).

aschempp:
As I pointed out in https://github.com/contao/core/issues/8172#issuecomment-185710174, we don't even need a checksum. Haste simply compares if the execution of `replaceInsertTag` gives the same result as before and if so then stops the loop…

leofeyer:
But this would mean an additional wrapper method, wouldn't it?

aschempp:
Why? Just check original content passed to the method with the return content before calling again?

leofeyer:
Fixed in 824b012eaefbabcba9cac59881bf27f6bfcdffca.

aschempp:
I would still do the `strpos($strBuffer, '{{') !== false` to prevent the whole replace thing to run again if not necessary.

leofeyer:
Re-added in 976a568834b8b6f065caf3ae643497cddefda480.

PR #475, opened 2016-04-22
Fix twig runtime error

bytehead:
This commit fixes a twig runtime error. This error gets displayed instead of the contao pretty error view:

`Key "hint" for array with keys "error, matter, errorOccurred, howToFix, errorFixOne, more, errorExplain" does not exist in "@ContaoCore/Error/layout.html.twig" at line 162`

### DIFF
--- a/src/Resources/views/Error/layout.html.twig
+++ b/src/Resources/views/Error/layout.html.twig
@@ -159,7 +159,11 @@
     </div>
     <div id="footer">
         <div class="wrap">
-            {% block hint %}{{ error.hint|format(template)|raw }}{% endblock %}
+            {% block hint %}
+                {% if error.hint is defined %}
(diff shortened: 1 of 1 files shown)


leofeyer:
Should we not debug why `error.hint` can be undefined in the first place?

bytehead:
Because it's probably never set: https://github.com/contao/core-bundle/blob/master/src/EventListener/PrettyErrorScreenListener.php#L278-L301

bytehead:
`error.hint` should have a message like this: `To customize this notice, create a custom template file named <em>%</em>.`

leofeyer:
Fixed in 48420924e2e684651f979774ea288ad7e9999016. We need to back port the changes if there should be a version 4.1.4 (not planned).

PR #480, opened 2016-05-04
[RFC] Url generator

aschempp:
Per our discussion, the `UrlGenerator` is supposed to replace `Controller::generateFrontendUrl`.

It allows for simple but flexible use of Contao urls (or pages actually).

 1. `$urlGenerator->generate('foo');`

    > Result: `/foo.html`

 2. `$urlGenerator->generate('index');`

    > Result: `/`

 3. `$urlGenerator->generate('foo/{items}', ['items' => 'bar']);`

    > Result: `/foo/items/bar.html` if `items` **is not** in `$GLOBALS['TL_AUTO_ITEM']`
    > Result: `/foo/bar.html` if `items` **is** in `$GLOBALS['TL_AUTO_ITEM']`

 4. `$urlGenerator->generate('foo/{items}', ['items' => 'bar', 'auto_item' => 'items']);`

    > Result: `/foo/bar.html` (regardless of `$GLOBALS['TL_AUTO_ITEM']`)

 5. `$urlGenerator->generate('foo/{items}/{article}', ['items' => 'bar', 'article' => 'test');`

    > Result: `/foo/bar/article/test.html` (assuming `items` **is** in `$GLOBALS['TL_AUTO_ITEM']`)

 6. `$urlGenerator->generate('foo/{items}.html', ['items' => 'bar', 'article' => 'test');`

    > Result: `/foo/bar.html?article=test` (assuming `items` **is** in `$GLOBALS['TL_AUTO_ITEM']`)

 7. Example with `PageModel`
     ```php
    $page = \Contao\PageModel::findWithDetails(/* ... */);

    $urlGenerator->generate(
        $page->alias ?: $page->id, 
        [
            '_locale' => $page->rootLanguage,
            '_domain' => $page->domain,
            '_ssl' => (bool) $page->rootUseSSL,
        ]
    );
    ```

**ToDo**
 - [x] Handle host and schema
 - [ ] Add docblocks with extensive info 
 - [ ] More unit tests ?

### DIFF
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -15,6 +15,7 @@ use Contao\CoreBundle\Exception\AjaxRedirectResponseException;
 use Contao\CoreBundle\Exception\PageNotFoundException;
 use Contao\CoreBundle\Exception\RedirectResponseException;
 use League\Uri\Components\Query;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 
 /**
@@ -1103,8 +1104,20 @@ abstract class Controller extends \System
 			$arrParams['_locale'] = $objPage->rootLanguage;
 		}
 
+		// Add the domain if it differs from the current one (see #3765 and #6927)
+		if ($blnFixDomain)
+		{
+			$arrParams['_domain'] = $arrRow['domain'];
+			$arrParams['_ssl'] = (bool) $arrRow['rootUseSSL'];
+		}
+
 		$strUrl = $objUrlGenerator->generate($strAlias, $arrParams);
-		$strUrl = substr($strUrl, strlen(\Environment::get('path')) + 1);
+
+		// Remove path from absolute URLs
+		if (0 === strpos($strUrl, '/'))
+		{
+			$strUrl = substr($strUrl, strlen(\Environment::get('path')) + 1);
+		}
 
 		// Decode sprintf placeholders
 		if (strpos($strParams, '%') !== false)
@@ -1116,12 +1129,6 @@ abstract class Controller extends \System
 			{
 				$strUrl = str_replace('%25' . $v, '%' . $v, $strUrl);
 			}
-		}
-
-		// Add the domain if it differs from the current one (see #3765 and #6927)
-		if ($blnFixDomain && !empty($arrRow['domain']) && $arrRow['domain'] != \Environment::get('host'))
-		{
-			$strUrl = ($arrRow['rootUseSSL'] ? 'https://' : 'http://') . $arrRow['domain'] . \Environment::get('path') . '/' . $strUrl;
 		}
 
 		// HOOK: add custom logic
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -1078,19 +1078,9 @@ abstract class Controller extends \System
 			}
 		}
 
-		$objRouter = \System::getContainer()->get('router');
+		$objUrlGenerator = \System::getContainer()->get('contao.routing.url_generator');
 		$arrParams = [];
-		$strRoute = 'contao_frontend';
-
-		// Correctly handle the "index" alias (see #3961)
-		if ($arrRow['alias'] == 'index' && $strParams == '')
-		{
-			$strRoute = 'contao_index';
-		}
-		else
-		{
-			$arrParams['alias'] = ($arrRow['alias'] ?: $arrRow['id']) . $strParams;
-		}
+		$strAlias = ($arrRow['alias'] ?: $arrRow['id']) . $strParams;
 
 		// Set the language
 		if ($strForceLang != '')
@@ -1113,7 +1103,7 @@ abstract class Controller extends \System
 			$arrParams['_locale'] = $objPage->rootLanguage;
 		}
 
-		$strUrl = $objRouter->generate($strRoute, $arrParams);
+		$strUrl = $objUrlGenerator->generate($strAlias, $arrParams);
 		$strUrl = substr($strUrl, strlen(\Environment::get('path')) + 1);
 
 		// Decode sprintf placeholders
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -4,7 +4,6 @@ namespace Contao\CoreBundle\Routing;
 
 use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
-use Symfony\Component\Routing\Matcher\UrlMatcherInterface;
 use Symfony\Component\Routing\RequestContext;
 
 /**
@@ -27,8 +26,8 @@ class UrlGenerator implements UrlGeneratorInterface
     /**
      * Constructor.
      *
-     * @param UrlMatcherInterface $router
-     * @param bool                $prependLocale
+     * @param UrlGeneratorInterface $router
+     * @param bool                  $prependLocale
      */
     public function __construct(UrlGeneratorInterface $router, $prependLocale)
     {
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -2,9 +2,7 @@
 
 namespace Contao\CoreBundle\Routing;
 
-use Symfony\Component\Routing\Exception\InvalidParameterException;
 use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
-use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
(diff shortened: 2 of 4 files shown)


qzminski:
Looks great but can you explain the behavior in 5th example? Why is ```{article}``` converted to ```article/test``` and simply not ```test```?

src/Routing/UrlGenerator.php
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @return array
+     *
+     * @throws MissingMandatoryParametersException
+     */
+    private function prepareParameters($alias, array $parameters)
+    {
+        $hasAutoItem = false;
+        $autoItem = (isset($GLOBALS['TL_AUTO_ITEM']) && is_array($GLOBALS['TL_AUTO_ITEM'])) ? $GLOBALS['TL_AUTO_ITEM'] : [];
+        $autoItem = array_key_exists('auto_item', $parameters) ? [$parameters['auto_item']] : $autoItem;
```
leofeyer: This seems weird. Is it

```php
if (array_key_exists('auto_item', $parameters)) {
    $autoItem = [$parameters['auto_item']];
} elseif ((isset($GLOBALS['TL_AUTO_ITEM']) && is_array($GLOBALS['TL_AUTO_ITEM']))) {
    $autoItem = $GLOBALS['TL_AUTO_ITEM'];
} else {
    $autoItem = [];
}
```

?

src/Routing/UrlGenerator.php
```diff
@@ -0,0 +1,121 @@
...
+
+        $parameters['alias'] = preg_replace_callback(
+            '/\{([^\}]+)\}/',
+            function ($matches) use ($alias, &$parameters, $autoItem, &$hasAutoItem) {
+                $param = $matches[1];
+
+                if (!isset($parameters[$param])) {
+                    throw new MissingMandatoryParametersException(
```
leofeyer: Should be `MissingMandatoryParameterException` (no *s* after *Parameter*), shouldn't it?

tests/Routing/UrlGeneratorTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->assertEquals('contao_frontend', $this->generator(false, 0)->generate('foobar'));
+        $this->assertEquals('contao_frontend', $this->generator(true, 0)->generate('foobar'));
+        $this->assertEquals('contao_frontend', $this->generator(false, 0)->generate('foobar/test'));
+    }
+
+    public function testWithoutParameters()
+    {
+        $this->assertEquals('foobar', $this->generator()->generate('foobar')['alias']);
+        $this->assertEquals('foobar/test', $this->generator()->generate('foobar/test')['alias']);
+        $this->assertEquals('foobar/article/test', $this->generator()->generate('foobar/article/test')['alias']);
```
leofeyer: Is this syntax (`->generate(…)['alias']`) supported in PHP 5.5?

leofeyer:
As for your first TODO: we must not only handle host and scheme, but all four reference types:

```php
interface UrlGeneratorInterface extends RequestContextAwareInterface
{
    /**
     * Generates an absolute URL, e.g. "http://example.com/dir/file".
     */
    const ABSOLUTE_URL = 0;

    /**
     * Generates an absolute path, e.g. "/dir/file".
     */
    const ABSOLUTE_PATH = 1;

    /**
     * Generates a relative path based on the current request path, e.g. "../parent-file".
     *
     * @see UrlGenerator::getRelativePath()
     */
    const RELATIVE_PATH = 2;

    /**
     * Generates a network path, e.g. "//example.com/dir/file".
     * Such reference reuses the current scheme but specifies the host.
     */
    const NETWORK_PATH = 3;
```

Toflar:
Aaaah, this is heaven! Thank you for working on it!

src/Routing/UrlGenerator.php
aschempp: would result in the same, yeah

aschempp:
@qzminski regarding point 5, this is the default behavior of Symfony `UrlGenerator`. If you pass additional parameters that are not used in the Url, they will be appended as query parameters.

aschempp:
So that explained point 6 😂 
Point 5: you can only have one auto_item, that's why the second one is ignored.

tests/Routing/UrlGeneratorTest.php
aschempp: Its supported in PHP 5.4 ;-)
http://php.net/manual/en/language.types.array.php#example-102

src/Routing/UrlGenerator.php
aschempp: That exception does not exist, the interface is defining it as being thrown:
http://api.symfony.com/2.3/Symfony/Component/Routing/Generator/UrlGeneratorInterface.html#method_generate

src/Routing/UrlGenerator.php
aschempp: Improved in 1ad5a836cb121e54fd6e513fd51f72fe5bd6a5f1

src/Routing/UrlGenerator.php
leofeyer: I see.

tests/Routing/UrlGeneratorTest.php
leofeyer: Cool 😎

aschempp:
PR updated and added a 7th example

src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1113,8 +1104,20 @@ public static function generateFrontendUrl(array $arrRow, $strParams=null, $strF
...
+		// Add the domain if it differs from the current one (see #3765 and #6927)
+		if ($blnFixDomain)
+		{
+			$arrParams['_domain'] = $arrRow['domain'];
+			$arrParams['_ssl'] = (bool) $arrRow['rootUseSSL'];
+		}
+
+		$strUrl = $objUrlGenerator->generate($strAlias, $arrParams);
+
+		// Remove path from absolute URLs
```
leofeyer: Shouldn't we use the `RELATIVE_PATH` constant here and provided the path in the context? Then we would not have to use `substr()`.

aschempp: We don't want `RELATIVE_PATH`, it would return something like `../../foobar.html` (depending on your current page).

Toflar:
As it is a specific Contao service anyway, can we add a helper method that accepts a `PageModel`?

aschempp:
I thought about that too, but its rather useless if the pagemodel already has a method for that 😂

Toflar:
Yeah, you're right. So we need an 8th example 😆

src/Resources/contao/library/Contao/Controller.php
leofeyer: > depending on your current page

Exactly. And if we make sure that the base path is in the context, it will be relative to it. Just like we want it :)

https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L286

Toflar:
Question: Can't we (or shouldn't we) deprecate `Controller::generateFrontendUrl()` then?

DanielSchwiperich:
Not sure I get this right. You say 
> UrlGenerator is supposed to replace Controller::generateFrontendUrl.

But the implementation shown in this PR is not replacing it, it's be used within. 

So is the goal to replace it in Contao 5 and now *only* change the inner workings of `Controller::generateFrontendUrl` ?

An option to add custom logic to it like now with `generateFrontendUrl` - Hook will remain, right?

aschempp:
Imho it's supposed to replace it. You're right it does not (yet) fire events, but all the other stuff in `Controller::generateFrontendUrl` is actually not necessary to build a URL if you're working with Symfony. See example 7 above.

leofeyer:
There is another problem: the router appends `?alias=index` on the home page.

src/Routing/UrlGenerator.php
```diff
@@ -0,0 +1,196 @@
...
+     *
+     * @return string
+     */
+    public function generate($name, $parameters = [], $referenceType = self::ABSOLUTE_PATH)
+    {
+        $route = 'index' === $name ? 'contao_index' : 'contao_frontend';
+        $parameters = is_array($parameters) ? $parameters : [];
+
+        // Store original request context
+        $context = $this->getContext();
```
leofeyer: Why don't we just clone the context object?

aschempp:
Not anymore 😎

src/Routing/UrlGenerator.php
aschempp: Because I want to change the object that is references in multiple places

leofeyer:
As discussed in Mumble on May 12th, the PR can be merged. Additionally, we want to deprecate the `Controller::getFrontendUrl()` method and copy the path fixes, the sprintf fixes and the hook to the `PageModel` class. In the `PageModel` class, the hook shall trigger a deprecation warning.

leofeyer:
Merged in 901ce97411d7954a5a51655d1e2313fee89be62b. I'll add a PR for the other changes.

Issue #454, opened 2016-03-03
Allow for multiple autologins per frontend user / member


DanielSchwiperich:
At the moment the autologin cookie hash is saved to the tl_member entry.
This way the autologin feature is only available for one device / session and gets overwritten each time a user logs in from a different device / browser (with autologin checkbox enabled). 

If we bind the autologin feature to a session (would maybe need fumbling with session expiration and ip validation) or would give it its own scope with a 1->N relation we could allow for multiple autologins like many other systems do. 

If the core team is for it, we would try to provide a PR for it (if you do not prefer to do it yourself for sure)

Toflar:
We wanted to change the whole thing to use the Symfony auth system anyway. Maybe @aschempp can shed some light and you can join forces here?

aschempp:
Well, Symfony does have it's own remember-me functionality (see http://symfony.com/doc/current/cookbook/security/remember_me.html). This is however not currently usable in Contao. We discussed to migrate the backend login to Symfony authentication first, but I can't say if that will make it into 4.2…

leofeyer:
Changing the database schema to an 1->N relation should be fairly easy. @contao/developers Why would we not want to do it?

aschempp:
Adding a new table just for the autologin value? And then drop it in Contao 5 (or before)?

leofeyer:
Using the Symfony authentication in the front end will be quite tricky to implement (as you have said yourself) and we don't know if we can accomplish it before Contao 5. Therefore making a few changes to support the feature in Contao 4 seems reasonable to me.

aschempp:
Supporting it is not a unsolvable problem, but BC is. That however depends on a discussion wether `tl_session` should be considered "API"...

leofeyer:
True. Although we could leave the table untouched and just add an `tl_session_additional_remember_me_tokens` table – not necessarily with this name thoug :)

aschempp:
Yeah. Appart from the fact that remember me is not related to tl_session 😂

What I find more interesting is that Symfony does not need a table for that at all…

DanielSchwiperich:
So the question is do you / we implement a interim solution until we go full symfony authentication correct?

I would then be for a easy solution. The current implementation of the autologin feature is not related to the session at all. Only the FE_AUTO_LOGIN cookie is checked and if a match is found, I'm logged in. 
[...]

PR #481, opened 2016-05-08
[RFC] Symfony RememberMe Service Implementation

DanielSchwiperich:
This is the PR for Issue #454 

The current status is completely working but I need Feedback before going further
## What is this PR for:

Providing autologin functionality across different devices for a frontend user. The current implementation only allows autologin for one device.
## Questions
1. **Make the Implementation optional, parallel or replace?** We could add _Allow autologin from multiple devices_ to tl_settings. See https://github.com/contao/core-bundle/issues/454#issuecomment-217706318 for infos on it.  
   I would vote for optional in combination with parallel for the most user comfort with deprecation warning for Contao 5. (Switching from normal autologin to multi-autologin would not break autologin cookies)
2. **How to handle the _cookie lifetime_ param?** Since the RememberMeService needs this setting at startup and cannot be changed later I had to put it in the DI Container for now. There the `$GLOBALS['TL_CONFIG']['autologin']` is not available atm. So we either had to 
   - remove it from tl_settings
   - write it back to the parameters.yml (bad practice?) 
   - load the localconfig.php with default.php (as fallback) in DI / CoreExtension (also bad practice?)
   - write it to another, dynamic configuration.yml and load in  DI / CoreExtension
## Further Implementation Infos
- Currently the rememberMeService lives beside the current autologin implementation (which is no problem, they do not harm each other in any way)
- The contao.logger you see in the services.yml / logger class is for development. The Symfony rememberMe Service class can _only_ handle a psr-3 compatible logger, so I made a quick wrapper for it. Will be removed 
- I had to extend the Contao UserProvider to make it compatible to the remember service. The current implementation does not allow to really load a user. I'm sure this was done on purpose by @aschempp but don't know why
- The rememberMeService requires The User Object to implement the `UserInterface`. I moved that part to the `User` class instead of the `FrontendUser`, so this has not to be changed for future auth implementations where the backend is also involved.
- The authenticate method diff looks bigger than it is. I had to move one part and restructure a little bit to avoid code duplication
- Code is not very pretty atm. Are there any Code / Style Conventions? Didn't found any info on it
## How To test
- Setup a Contao Installation with a login form and autologin enabled
- Create a member user
- Login to the frontend
- Delete the FE_AUTH and FE_AUTOLOGIN cookie, reload the site (currently the rememberMeService uses it's own Cookie FE_AUTOLOGIN_MULTI). See the user should still be logged in. 
## Todos
- [ ] Write Unittests
- [x] Implementation of Question 1,2
- [ ] Code Style

---

~~I had a to rush a little bit when I was half through the PR, because guests arrived for BQ, but I wanted to get out to you. Will reread that later and _optimize_~~ <- done :smile:


discordier:
> Make the Implementation optional, parallel or replace?

I also vote for optional here.

> How to handle the cookie lifetime param?

IMO we should move it to parameters, we already agreed in the past that we want to move everything from `localconfig.php` to `parameters.yml` eventually when we are touching the area where it applies.
The way of accessing config values via `$GLOBALS['TL_CONFIG']` is not supported in Contao 4 and all access must be done either via DI or via `\Config::get()` calls. Therefore it is IMO totally fine to push it to a config parameter.

For your additions to the user classes and the user provider.
I wonder if it now would make sense to implement a real user provider, user class (implementing the UserInterface) and Authenticator (implementing the current authenticate functionality) and only delegate from the legacy classes.
This would allow us to use all the authentication mechanisms of Symfony.
Until now there was no need to implement something like this but as you have to alter the current implementation anyway, maybe it is time to do so. IMO the implementation of this PR will be slightly slimmer then, as it does not need to change implementation (aside from activating).

On your general question about code style, we are mostly using the symfony code style for new code and the "Contao legacy style" in legacy code.

DanielSchwiperich:
Thanks for feedback @discordier 
Ok I'll continue and make it optional.

I agree with you that the authentication system needs an overhaul. Also regardings stuff like auth providers, events instead of hooks ( + more events like fire an event if a user gets blocked) etc. 
But please not **now** and not in **this** PR :wink: (This would be something for Contao 5 I think, since it would be hard to not break BC)
My additions to the `UserProvider` and the `User` are pretty simple / trivial. Would like to hear if @aschempp  is ok with it, since he implemented it originally (at least that's the doc block saying, didn't check history)

DanielSchwiperich:
I made the remember_me service optional and by default off. Can be set in the backend via  the security panel.

What about translations. I don't need to provide all, do I? :fearful:  

I'm not so happy to have the expiration value set in the parameters.yml / DI for the multi-autologin session and in tl_settings for the single-autologin sessions. 
This does not feel right. I think they should always be the same. But we cannot remove it from tl_settings without breaking BC. And I cannot use the value from tl_settings since the it is not available on container / DI  setup in `ContaoCoreExtension::loadInternal` :disappointed: 

Will do the Unittests on the weekend probably.

aschempp:
Translations are usually managed on Transifex? 

I think we need to move it to the container, we did the same for several other parameters. It will be BC in the sense that we write container parameters to `$GLOBALS['TL_CONFIG']`, but if adjusted the setting in `localconfig.php` will not be applied for the multi-login, which is fine for me.

We're only adding things to the parameters.yml that must be adjusted on each installations. Things that are usually the same can simply be bundle config options with default values (see `url_suffix` for example).

DanielSchwiperich:
yeah Transifex is a chapter I completely missed / skipped until now. Did take a look, still not sure if it's ok / enough whats in the PR or if it's even wrong?

Regarding the expiration value, I'll replace the current `tl_setting` / `$GLOBALS['TL_CONFIG']`  autologin value then in this PR like done with `url_suffix`. 
thanks for feedback

DanielSchwiperich:
So the `autologin` setting has gone from the backned and will be autofilled with the container param. 

I've to say it works really well so far. 
Only Unittests are missing now. 

@leofeyer Do you want me to use the `Config` class from global Namespace to be on par with the rest of the class, leave it combined, or should I adjust the other usages in this class, since it's now now BC anymore to contao 3?

src/DependencyInjection/Configuration.php
```diff
@@ -99,6 +99,23 @@ public function getConfigTreeBuilder()
...
+                            ->children()
+                                ->scalarNode('cookie_name')
+                                    ->cannotBeEmpty()
+                                    ->defaultValue('FE_AUTO_LOGIN_MULTI')
+                                ->end()
+                                ->integerNode('cookie_lifetime')
+                                    ->min(0)
+                                    ->defaultValue(7776000)
+                                ->end()
+                                ->scalarNode('formfield_name')
```
leofeyer: Why does this need to be configurable?

leofeyer:
Why did you make the new feature optional? We should not provide a flag to enable or disable it but simply replace the existing routines.

leofeyer:
Haha, I just realised that this was one of your initial questions. 😄 

> Make the Implementation optional, parallel or replace?

Replace. Definitely.

> How to handle the cookie lifetime param?

A container parameter seems fine to me.

src/DependencyInjection/Configuration.php
aschempp: Probably just best practice because Symfony allows to configure field names of the form login as well.

aschempp:
this probably conflicts with #559

bytehead:
I don't know if it there are some future problems, if you merge this PR and then I try to adjust/implement this changes to #559?

leofeyer:
We should definitely consider #559. But we should also merge this PR, shouldn't we? 😄

DanielSchwiperich:
I alreaday mentioned in #559 that there may some overlapping. 
I choosed to make the feature optional because @discordier voted for this on my question (and it was the only answer at that point 😉 ) 
But sure I can change that. I was for optional because it changes the behaviour and theoretically people could rely on the current one. I have no idea why anyone only wants to have autologin just for one device nut who knows. So it depends on if you consider the current implementation as *bugged* or  not. You say 😸 

Just let me know know if I can / should finish the implementation (unit tests are also missing right now and I have to adjust compatibility with current release). Will be happy to do so.

leofeyer:
Definite +1 from me!

leofeyer:
@contao/developers What about you?

Toflar:
I definitely like to see a remember me setting! But also, I like to see the whole auth being processed via Symfony so we can push this further and add OAuth etc. Unfortunately, I cannot invest my time here because there's too much other stuff I want to work on. So in the best case @DanielSchwiperich and @bytehead can come up with a complete PR that handles the BE and FE auth via Symfony and provides a remember me function too :)

leofeyer:
As discussed at the Contao Camp:

* We want to replace the current implementation instead of having it optional.
* We do not want to add any configuration parameters. Things should be read from the Sf configuration or the default values.

The only remaining conflict is that this PR makes the `FrontendUser` class implement the `UserInterface` and #559 makes the `UserModel` class implement the `UserInterface`.

src/Resources/contao/library/Contao/User.php
```diff
@@ -673,6 +675,47 @@ public function isMemberOf($id)
 		return false;
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	public function getRoles()
+	{
+		return [];
```
aschempp: This should be the same roles as defined in the Token

aschempp:
So yesterday I took a few hours to look into this, as we wanted to merge it into Contao 4.3 as a bug fix. I fixed several little issues (like removing the config options), and then digged again into Symfony security. I re-found what I implemented two years ago (for #559), and unfortunately also found issues with this implementation that cannot be easily fixed.

As you know, this implementation no longer requires a database table to hold the identification key, as Contao currently does with `tl_member.autologin`. The authentication is solely based on the cookie which is encrypted with the `kernel.secret`. The issue with this is, if someone steals your cookie, there is no way to revoke the cookie.

In the current implementation, a device is automatically *revoked* if another device is authenticated. This would no longer be possible, every device that once has that cookie will login forever. I don't think that's the solution we want, as it would lower security of the application.

There is a solution build into Symfony. Instead of using the `TokenBasedRememberMeService` we would need to use the `PersistentTokenBasedRememberMeServices`. That requires a new database table though, and implementation would not be done in a few hours.

A good read on the topic of *persistent tokens* can be found at https://paragonie.com/blog/2015/04/secure-authentication-php-with-long-term-persistence
A quick summary is at http://stackoverflow.com/questions/244882/what-is-the-best-way-to-implement-remember-me-for-a-website#answer-244907

leofeyer:
> Instead of using the `TokenBasedRememberMeService` we would need to use the `PersistentTokenBasedRememberMeServices`. That requires a new database table though, and implementation would not be done in a few hours.

Still I think this is the right way to go. The current implementation is not secure (enough).

aschempp:
Absolutely, but not in 4.3 rc imho...

leofeyer:
Since #685 is almost ready to merge, I'm closing this PR in favor of #685.

PR #485, opened 2016-05-10
[RFC] Twig!

aschempp:
This PR adds Twig support for all extension developers. It allows to use Twig templates for everything in Contao, but it does not infer with the existing template system. Mixing YAML and PHP templates just won't work consistently and it won't be BC.

This PR is pretty simple:
```php
$container->get('contao.twig')->render('ce_text.twig');
```

The `contao.twig` engine differs in two things from the regular one.
 1. automatic output encoding is disabled
 2. adds automatic template loading

Also, the templates can be edited in the regular Contao backend.

### DIFF
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -108,6 +108,12 @@ abstract class Controller extends \System
 			foreach ($arrCustomized as $strFile)
 			{
 				$strTemplate = basename($strFile, strrchr($strFile, '.'));
+
+				if (pathinfo($strTemplate, PATHINFO_EXTENSION) == '.twig')
+				{
+					$strTemplate = basename($strFile, strrchr($strFile, '.'));
+				}
+
 				$arrTemplates[$strTemplate][] = $GLOBALS['TL_LANG']['MSC']['global'];
 			}
 		}
@@ -140,14 +146,12 @@ abstract class Controller extends \System
 							{
 								$strTemplate = basename($strFile, strrchr($strFile, '.'));
 
-								if (!isset($arrTemplates[$strTemplate]))
+								if (pathinfo($strTemplate, PATHINFO_EXTENSION) == '.twig')
 								{
-									$arrTemplates[$strTemplate][] = $objTheme->name;
+									$strTemplate = basename($strFile, strrchr($strFile, '.'));
 								}
-								else
-								{
-									$arrTemplates[$strTemplate][] = $objTheme->name;
-								}
+
+								$arrTemplates[$strTemplate][] = $objTheme->name;
 							}
 						}
 					}
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -52,6 +52,11 @@ abstract class Controller extends \System
 	 */
 	public static function getTemplate($strTemplate, $strFormat='html5')
 	{
+		if ($strFormat != 'html5')
+		{
+			trigger_error('Template formats are deprecated', E_USER_DEPRECATED);
+		}
+
 		$arrAllowed = trimsplit(',', strtolower(\Config::get('templateFiles')));
 		array_push($arrAllowed, 'html5'); // see #3398
 
@@ -108,11 +113,6 @@ abstract class Controller extends \System
 			foreach ($arrCustomized as $strFile)
 			{
 				$strTemplate = basename($strFile, strrchr($strFile, '.'));
-
-				if (pathinfo($strTemplate, PATHINFO_EXTENSION) == '.twig')
-				{
-					$strTemplate = basename($strFile, strrchr($strFile, '.'));
-				}
 
 				$arrTemplates[$strTemplate][] = $GLOBALS['TL_LANG']['MSC']['global'];
 			}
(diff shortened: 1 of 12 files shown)


bytehead:
😍

src/Resources/contao/library/Contao/Controller.php
```diff
@@ -139,15 +139,7 @@ public static function getTemplateGroup($strPrefix)
...
-
-								if (!isset($arrTemplates[$strTemplate]))
-								{
-									$arrTemplates[$strTemplate][] = $objTheme->name;
-								}
-								else
-								{
-									$arrTemplates[$strTemplate][] = $objTheme->name;
-								}
+								$arrTemplates[$strTemplate][] = $objTheme->name;
```
discordier: Was the old code originally maybe supposed to look like:
```php
if (!isset($arrTemplates[$strTemplate]))
{
	$arrTemplates[$strTemplate] = array();
}
$arrTemplates[$strTemplate][] = $objTheme->name;
```
This way no notice would be issued when the array does not exist. IMO it would be good if we would reduce the amount of notices to an absolute minimum.
Therefore, when you want to touch this file, please in a manner where no notice is issued, otherwise this chunk can be omitted as unrelated to the PR.

discordier:
First off, good that you are tackling this issue but I have a bad feeling in the guts if this approach is right.
To me it feels like we are going the wrong way here.
IMO we should decode the arrData in the legacy Template classes and pass it to the normal twig engine then.
Using the approach in this PR, we will teach developers to use contao.twig now for rendering (and therefore they will have to alter their code and templates) and in Contao 5 we will drop output encoding in the Input and Widget classes and therefore also drop the (then obsolete) contao.twig engine which will force them to alter their code and templates yet again.

aschempp:
The point here is also about to get people used to Twig. We discussed that decoding is not possible…

Toflar:
Decoding is very well possible. Just not automatically done, I agree. You cannot just decode everything. However, instead of a different twig environment, you could provide a `decodeContaoEntities` twig filter. If users want to output raw data, they would need to write something like this:

```twig
{{ page.description|decodeContaoEntities|raw }}
```

This would force people to think about input and output encoding and make sure they understand why this is. In Contao 5, they would need to drop both filters the `decodeContaoEntities` as well as the `raw` which will result in correct output encoded content.
I think it's kind of a decision how we want to go about that. It's not about the technical implementation, really.

* Either we provide a different twig environment and users won't really think about it and just go on like they do now. Advantages: People don't have to worry or understand anything and can familiarize with twig. Disadvantages: People won't understand the `raw` filter because it does not have any effect in Contao, they won't understand input vs. output encoding and won't really understand why there is a templating language (because to be honest: automatic output encoding is **the** major feature of a templating language).

* or we provide a filter that decodes our encoded input which will force people to combine this filter and the `raw` filter. Advantages: People familiarize with output encoding and will be prepared for the switch in Contao 5 because they understand. Disadvantages: Well, they have to add two filters for pretty much any Contao template variable.

I think it's really just this. Both approaches have advantages and disadvantages but we have to decide for one. Do you agree? I mean on the fact that we have these two options, we can decide afterwards.

Toflar:
Obviously we could simplify the `|decodeContaoEntities|raw` by making it `|contaoRaw` or so.

aschempp:
The problem with both of your approaches is we would also need to build a complete user-friendly management (template editor) as well as think about the issue that templates in the default twig engine is always related to each bundle and cannot be overridden by another extension. Our custom environment keeps all things currently familiar but allows extension developers to use twig. 

That could also be added as a third party bundle, but that obviously wont have the same adoption.

leofeyer:
I like the solution. However I would adjust the template loader so it finds the Twig templates automatically instead of adding ugly code to the `autoload.php` file of each extension. @aschempp Do you think that's possible?

aschempp:
@leofeyer everything's possible, but I tried not to add new code to the old classes. I'm just following the best practice how to register templates for old Contao stuff. I can't see a reason to write more than what I did in `autoload.php`.

leofeyer:
> I can't see a reason to write more than what I did in `autoload.php`.

I can. It would mean a lot of unnecessary code duplication.

At least let us add a `TemplateLoader::addTwigTemplates()` method.

aschempp:
Why, what's the difference of that method?

leofeyer:
Maybe I got your code wrong. Do I need to add the stuff that you have added to the `autoload.php` file to the `autoload.php` file of my extensions? Or is it meant to be in the core autoload file only?

aschempp:
That's only necessary in the core! contao/core-bundle will register ALL templates known to the contao.twig.loader, which means all other bundles!

leofeyer:
Ah ok, then there is no code duplication of course :)

I would still prefer to have this in the `TemplateLoader` class and not in an auto-generated file.

leofeyer:
As discussed in Mumble on April 12th, we want to merge the PR. However, we also need an ACE highlighting and cache invalidation for the prod environment.

leofeyer:
Unit tests are missing.

aschempp:
Changed environment to enable auto-reload.

Regarding ACE, we can probably use https://github.com/heygrady/ace-mode-twig

src/Resources/contao/library/Contao/Controller.php
leofeyer: I have just fixed this in f77e263edb99b025e9e4729d1dc1cd40fe490eaf.

ausi:
I’d like to remove `$options['autoescape'] = false;`. Developers should either decode the data before passing it to the template or using `{% autoescape false %}` in the template.

leofeyer:
See #488 for the follow-up.

PR #487, opened 2016-05-13
[RFC] Deprecate the generateFrontendUrl() method

leofeyer:
This is the follow-up PR to #480. As discussed, I have deprecated the `getFrontendUrl()` method and copied some logic to the `PageModel::getFrontendUrl()` class.

### DIFF
--- a/src/Resources/contao/models/PageModel.php
+++ b/src/Resources/contao/models/PageModel.php
@@ -11,6 +11,7 @@
 namespace Contao;
 
 use Contao\CoreBundle\Exception\NoRootPageFoundException;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/Resources/contao/models/PageModel.php
```diff
@@ -981,11 +1005,63 @@ public function getFrontendUrl($strParams=null, $strForceLang=null)
...
+			array
+			(
+				'_locale' => $this->rootLanguage,
+				'_domain' => $this->domain,
+				'_ssl' => (bool) $this->rootUseSSL,
+			),
+			UrlGeneratorInterface::ABSOLUTE_URL
+		);
+
+		$strUrl = $this->applyLegacyLogic($strUrl, $strParams);
```
Toflar: Not sure but doesn't this change behaviour. Before you got a relative URL in the hook and now you get an absolute one?

leofeyer: You are right. But does that make sense?

Toflar: Nope. People could not know if getAbsoluteUrl() or getFrontendUrl() was called in the hook. Now they would but I guess we should not introduce this "new feature" in Contao 4 if it's deprecated anyway.

leofeyer: There is no other way I guess if we want to use the router with the `UrlGeneratorInterface::ABSOLUTE_URL` constant. Otherwise we had to add the scheme, host and path manually.

aschempp: If I understand correctly, the domain was already in `Controller::generateFrontendUrl` before the hook if the last option to the method was `true`?

leofeyer: Only if the host was different from the current host.

aschempp: Even when using `getAbsoluteUrl`?

leofeyer: Yes, because the absolute URL was only added afterwards.

Toflar:
👍

leofeyer:
Merged in 7de6b51ef9cf64d21eb031cd0283824e2b8cf21d.

PR #488, opened 2016-05-13
[RFC] Twig! 2

leofeyer:
This is the follow-up PR to #485. I have applied the CS fixes and moved the Twig auto-registration into the `TemplateLoader::initialize()` method. The unit tests are still missing.

As @ausi mentioned, we should re-enable the `autoescape` option, so people can decide whether they want to decode the database content before passing it to the template or to use the `|raw` filter in their Twig templates.

@contao/developers Is it ok if I change the code accordingly?


leofeyer:
There is one disadvantage with the approach: If people copy a Twig template into the `templates/` folder to modify it, they are overriding a template in `src/Resources/contao/templates`. This makes it impossible for us to support Twig templates in `src/Resources/views` later on, because Contao would not know whether to override the Symfony Twig template or the Contao Twig template.

How does someone edit or override a Twig template in `src/Resources/views` in the back end? Is this possible at all because of the compile cache?

aschempp:
`src/Resources/views` is not related to our Twig engine, our templates are always in `src/Resources/contao/templates`.

As I said, I disagree with the `autoescape` thing. Contao data is input encoded, so we must not apply output encoding. People (including us) will do it wrong, we're already struggling with that in existing cases (see https://github.com/isotope/core/issues/1637 for an example).

aschempp:
Just thinking about it, can the template loader handle a `ce_text.html5` and `ce_text.twig` at the same time?

PR #490, opened 2016-05-17
[WIP] Deprecate the Message class

leofeyer:
This is the follow-up PR to #445.

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -1,6 +1,20 @@
 Deprecated features
 ===================
 
+### Message class
+
+The `Message` class has been deprecated in Contao 4.2 and will be removed in
+Contao 5.0. Use the session and twig service instead:
+
+```php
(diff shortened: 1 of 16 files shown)


leofeyer:
As discussed, this should rather be a service with an interface defining the methods to add and render messages.

aschempp:
I wasn't part of the discussion, does that mean I can currently ignore the PR?

aschempp:
Just as a quick note, `TL_MODE` would not work here as it does not equal `frontend` and `backend`.

leofeyer:
You can ignore the PR. I have to rework it.

Toflar:
That's still something that would be helpful, no?

leofeyer:
Of course. I just have not had the time yet.

leofeyer:
I'm closing this in favor of #1211.

PR #491, opened 2016-05-17
[RFC] Store the full URI in the search results

leofeyer:
This is the follow-up PR to contao/core#8331.

@fritzmg Does it solve your issue?

@Toflar Is `varchar(255)` long enough to store the full URI?

### DIFF
--- a/src/Resources/contao/classes/Frontend.php
+++ b/src/Resources/contao/classes/Frontend.php
@@ -627,7 +627,7 @@ abstract class Frontend extends \Controller
 				if ($blnIndex)
 				{
 					$arrData = array(
-						'url'       => \Environment::get('relativeRequest'),
+						'url'       => \Environment::get('base') . \Environment::get('relativeRequest'),
 						'content'   => $objResponse->getContent(),
 						'title'     => $objPage->pageTitle ?: $objPage->title,
 						'protected' => ($objPage->protected ? '1' : ''),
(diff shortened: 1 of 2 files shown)


Toflar:
Hard to say if 255 are enough. But imho that question is the same if it's a relative URL :D Most domains are not 250 characters long ;)

Toflar:
I'd leave it as is for now.

leofeyer:
Merged in 3d03fe0bf2f894fa992470229e4649f177945c49. I have also patched contao.org to use absolute URLs and it seems to work like a charm.

Toflar:
That's a bugfix then. Should be backported to 3.5.

leofeyer:
Yes. Will do if no problems occur.

fritzmg:
> @fritzmg Does it solve your issue?

I can only test with Contao 3.5 right now, but if it's the same as doing
```php
'url' => \Environment::get('uri')
```
in [/system/modules/core/classes/FrontendTemplate.php#L321](https://github.com/contao/core/blob/3.5.12/system/modules/core/classes/FrontendTemplate.php#L321), then yes, it should work.

leofeyer:
Yes, that is the same. In Contao 4, `get('uri')` will include the script name though (e.g. `app_dev.php`) and `get('relativeRequest')` will not, that's why I have used the latter.

leofeyer:
Back ported in 3e6dc79e8914b714878f3a76ec882fb6c7338b7e.

PR #515, opened 2016-06-10
Handle the `useAutoItem` option in the URL generator

### LINKED ISSUE #489: UrlGenerator and auto_item disabled
Just thinking about it, can we still disable `auto_items` in Contao 4? And if so, I think the UrlGenerator would ignore that?

leofeyer:
This PR fixes #489.

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -10,6 +10,8 @@
 
 namespace Contao\CoreBundle\Routing;
 
+use Contao\Config;
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
 use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RequestContext;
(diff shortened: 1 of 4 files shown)


src/Routing/UrlGenerator.php
```diff
@@ -142,13 +158,13 @@ function ($matches) use ($alias, &$parameters, $autoItems, &$hasAutoItem) {
                 $value = $parameters[$param];
                 unset($parameters[$param]);
 
-                if (!$hasAutoItem && in_array($param, $autoItems, true)) {
-                    $hasAutoItem = true;
-
-                    return $value;
+                if (!$config->get('useAutoItem') || $hasAutoItem || !in_array($param, $autoItems, true)) {
```
aschempp: Are you sure this is correct? The auto item should only be used if the config flag is true

leofeyer: The if condition now returns `$param.'/'.$value` :)

aschempp: You're right, my fault

leofeyer:
Merged in 6efa96281d5ab20ada7258143f6bdbef858557ac.

PR #524, opened 2016-06-22
Local exceptions for PaletteManipulator

aschempp:
I've run into the problem that DcaManipulator throws an `InvalidArgumentException`, which is catched by the `DcaLoader` because `FileLocator` throws the same.

1. we should always use local exceptions
2. Added logging if an exception is catched

### DIFF
--- a/src/Resources/contao/library/Contao/DcaLoader.php
+++ b/src/Resources/contao/library/Contao/DcaLoader.php
@@ -85,7 +85,10 @@ class DcaLoader extends \Controller
 					include $file;
 				}
 			}
-			catch (\InvalidArgumentException $e) {}
+			catch (\InvalidArgumentException $e)
+			{
+				\System::getContainer()->get('monolog.logger.contao')->notice($e->getMessage(), ['exception' => $e]);
+			}
 		}
 
 		// HOOK: allow to load custom settings
--- a/src/DataContainer/PaletteManipulator.php
+++ b/src/DataContainer/PaletteManipulator.php
@@ -10,6 +10,8 @@
 
 namespace Contao\CoreBundle\DataContainer;
 
+use Contao\CoreBundle\Exception\PalettePositionException;
+use Contao\CoreBundle\Exception\PaletteNotFoundException;
 use Contao\StringUtil;
 
 /**
@@ -84,7 +86,7 @@ class PaletteManipulator
      *
      * @return static The object instance
      *
-     * @throws \InvalidArgumentException If $position or $fallbackPosition is invalid
+     * @throws PalettePositionException If $position or $fallbackPosition is invalid
      */
     public function addField(
         $name,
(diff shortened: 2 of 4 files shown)


aschempp:
@leofeyer ping

leofeyer:
I guess you will have to re-create the PR against an infinite branch. I cannot reopen it. :cry:

aschempp:
Or you just merge it into 4.2.2? ;-)

leofeyer:
I can't. It's a feature.

aschempp:
well then merge it to develop…

leofeyer:
Ok, will do.

src/Resources/contao/library/Contao/DcaLoader.php
```diff
@@ -85,7 +85,10 @@ public function load($blnNoCache=false)
 					include $file;
 				}
 			}
-			catch (\InvalidArgumentException $e) {}
+			catch (\InvalidArgumentException $e)
+			{
+				\System::getContainer()->get('monolog.logger.contao')->notice($e->getMessage(), ['exception' => $e]);
```
leofeyer: Why did you add this here? It does not seem to be related to the PR.

aschempp: Yes it is, that's point 2 in my comment

leofeyer: But this will also log exceptions for every bundle or module that does not have a `dca/` folder! What exactly do you want to log?

aschempp: It would log if no such file was found in any `dca` folder, right? Which should theoretically not happen because we would not find the table…

My issue was that an `InvalidArgumentException` was thrown by a DCA file (in this case `PaletteManipulator` but it could be anything), and there was no way to figure out what went wrong and why my file was not loaded.

Maybe we should create a PR for Symfony to throw a `ResourceNotFoundException` in the next version?

leofeyer: > Maybe we should create a PR for Symfony to throw a `ResourceNotFoundException` in the next version?

The Symfony `FileLocator` class throws the `\InvalidArgumentException`, therefore we cannot change this (except by replacing the class). But the following is wrong anyway:

```php
class PaletteNotFoundException extends \InvalidArgumentException
```

All palette errors are `\LogicException`s and if the palette exceptions extended these, they would not be caught by the file locator routine.

aschempp: Well we can't change that for BC reasons, extending from `\InvalidArgumentException` is BC.

aschempp: And I think specifying an invalid palette is an `InvalidArgumentException` ;-)

leofeyer: Both wrong IMHO.

> Well we can't change that for BC reasons, extending from `\InvalidArgumentException` is BC.

The `PaletteNotFoundException` class is only added in this PR, therefore there cannot be a BC break.

> And I think specifying an invalid palette is an `InvalidArgumentException` ;-)

Not necessarily. An invalid argument is something that is passed at runtime. A configuration error in a file (in this case a DCA file) is a logic error.

aschempp: > The PaletteNotFoundException class is only added in this PR, therefore there cannot be a BC break.

But previously it threw an `InvalidArgumentException`, so we must stay compatible with that if someone already catched it.

leofeyer: I see.

PR #525, opened 2016-06-22
[RFC] Do not spam the system log

leofeyer:
This PR will change the log level of the "No active page for page ID" message from "error" to "general", which stops the app from logging them in the `app/logs/prod-*.log` file.

This is a security enhancement, which shall prevent storage overflow attacks.

@contao/developers Any objections?

### DIFF
--- a/src/Resources/contao/controllers/FrontendIndex.php
+++ b/src/Resources/contao/controllers/FrontendIndex.php
@@ -135,9 +135,6 @@ class FrontendIndex extends \Frontend
 		// Throw a 404 error if the page could not be found
 		if ($objPage === null)
 		{
-			$this->User->authenticate();
-			$this->log('No active page for page ID "' . $pageId . '" (' . \Environment::get('base') . \Environment::get('request') . ')', __METHOD__, TL_GENERAL);
-
 			throw new PageNotFoundException('Page not found: ' . \Environment::get('uri'));
 		}
 
@@ -207,9 +204,6 @@ class FrontendIndex extends \Frontend
 		// Check wether the language matches the root page language
 		if (\Config::get('addLanguageToUrl') && \Input::get('language') != $objPage->rootLanguage)
 		{
-			$this->User->authenticate();
-			$this->log('No active page for page ID "' . $pageId . '" and language "' . \Input::get('language') . '" (' . \Environment::get('base') . \Environment::get('request') . ')', __METHOD__, TL_ERROR);
-
 			throw new PageNotFoundException('Page not found: ' . \Environment::get('uri'));
 		}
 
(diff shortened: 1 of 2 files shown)


aschempp:
I don't think that can be correct. What you changed will change the backend logging "color", but the file log results from the exception being thrown.

leofeyer:
Huh? What exactly do you mean?

aschempp:
The log message in `app/logs` results from the exception being thrown and being logged by either our pretty error screen handler or the Symfony error handler. Not related to `$this->log()` at all.

leofeyer:
Ok, so we have fixed the monolog configuration in 46f63c77298830b665846caed0623ebeff66743f and contao/standard-edition@1ef8948db84c0b466a27a97dbcf829db794c0227.

Still I would like to merge this PR and change the level of the "No active page for page ID" error from ERROR to INFO, because it is not an application error if search bots etc. request non-existing URLs. In fact, there is not even much the webmaster can do about it, so we might not log this at all?

Toflar:
I think I'd not log this at all. The information does not provide any added value because it is basically reproducing the `access.log` of the server itself. If you want to monitor for old pages still being looked for you should be checking the server logs and add redirects on server level. Basically it's all related to the server, not the CMS.

aschempp:
I might agree with the reasons, but its something that people are very used to. Maybe it would make sense to add flags for different logging features? Like `contao.logging.page_not_found = true` and then we could add others like login logout etc?

leofeyer:
> Maybe it would make sense to add flags for different logging features?

I don't think this is necessary.

DanielSchwiperich:
We are very used to it, and would love to see it gone or at least not appear in the error channel

aschempp:
@DanielSchwiperich thats a confusing statement 😂 
Does "used to it" not mean you like the feature?

DanielSchwiperich:
No we are used to see those messages in the log and they always feel a bit like spam and make sometimes difficult to identify real errors. 

Auf deutsch: Wir sind gewöhnt diese Meldungen im Log zu sehen. Lieben tun wir sie nicht. Im Gegenteil ;)

leofeyer:
That says it all 😄

aschempp:
just don't forget thats just one personal opinion ;)

leofeyer:
An opinion shared by @Toflar, @bytehead and me. 😄

discordier:
Add me to the list. I really hate this log spam.

leofeyer:
Ok, so as most of us think that we should not log the message – should we change this in Contao 3.5, too?

fritzmg:
> Ok, so as most of us think that we should not log the message – should we change this in Contao 3.5, too?

I'd say yes, isn't it quite easy to spam a site running Contao with nonsense get requests on regular pages and nonsense page requests, to generate a lot of database queries and inserts in `tl_search`, `tl_search_index` and `tl_log`?

aschempp:
I would not change it in 3.5 just because we should not unnecessary alter what users are used to in that version. Would a Symfony 404 exception be stored in the apache server error log?

joergmold:
I hate these messages too. I would love to see them gone, also in Contao 3.5 ;-)

leofeyer:
Ok, so I have removed four log messages, which spam the system log rather than providing any valuable information. @contao/developers Can I merge the PR like this?

aschempp:
👍

Toflar:
👍🏻

discordier:
:+1:
However, the failing test on Windows should get fixed in a follow up PR.

leofeyer:
Merged in 1811b8b632cfa59a2e8f9f8f488249fe444f430a.

PR #529, opened 2016-06-28
[RTM] Password reset command

aschempp:
This allows to reset a backend user password from the command line.

**Todo:**
 - [x] Add unit tests

### DIFF
--- a/src/Command/ResetPasswordCommand.php
+++ b/src/Command/ResetPasswordCommand.php
@@ -13,6 +13,7 @@ namespace Contao\CoreBundle\Command;
 use Contao\Config;
 use Contao\Encryption;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Exception\RuntimeException;
 use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -45,6 +46,10 @@ class ResetPasswordCommand extends ContainerAwareCommand
      */
     protected function interact(InputInterface $input, OutputInterface $output)
     {
+        if (null === $input->getArgument('username')) {
+            throw new RuntimeException('You must enter a username on the command.');
+        }
+
         if (null !== $input->getOption('password')) {
             return;
         }
@@ -53,7 +58,7 @@ class ResetPasswordCommand extends ContainerAwareCommand
         $confirm  = $this->askForPassword('Please confirm the password:', $input, $output);
 
         if ($password !== $confirm) {
-            throw new \LogicException('Your passwords do not match');
+            throw new RuntimeException('Your passwords do not match');
         }
 
         $input->setOption('password', $password);
@@ -122,7 +127,7 @@ class ResetPasswordCommand extends ContainerAwareCommand
         $passwordLength = $config->get('passwordLength');
 
         if (strlen($password) < $passwordLength) {
-            throw new \UnderflowException(sprintf('Password must be at least %s characters.', $passwordLength));
+            throw new RuntimeException(sprintf('Password must be at least %s characters.', $passwordLength));
         }
 
         /** @var Encryption $encryption */
--- a/src/Command/ResetPasswordCommand.php
+++ b/src/Command/ResetPasswordCommand.php
@@ -1,0 +1,133 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2016 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\Command;
+
+use Contao\Config;
+use Contao\Encryption;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+
+/**
+ * Resets password for a Contao backend user.
+ *
+ * @author Andreas Schempp <https://github.com/aschempp>
+ */
+class ResetPasswordCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('contao:reset-password')
+            ->setDescription('Resets password for a Contao backend user.')
+            ->addArgument('username', InputArgument::REQUIRED)
+            ->addOption('password', 'p', InputOption::VALUE_REQUIRED)
+        ;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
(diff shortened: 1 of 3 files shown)


src/Command/ResetPasswordCommand.php
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * @inheritdoc
+     */
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        if (null === $input->getArgument('username')) {
+            throw new RuntimeException('You must enter a username on the command.');
+        }
+
+        if (null !== $input->getOption('password')) {
```
leofeyer: We should not allow to pass a clear-text password as argument. It will show up in the server's process list and thus is a security issue.

aschempp: I thought about this too, but it's a command feature to optionally allow it (see mysql command for example).

leofeyer: It shouldn't be allowed in the `mysql` command either. But ok, then let's keep it.

src/Command/ResetPasswordCommand.php
```diff
@@ -0,0 +1,144 @@
...
+ */
+class ResetPasswordCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('contao:reset-password')
```
leofeyer: I would call it just `contao:password`, because it does not really reset the password but rather allows to set a new one.

aschempp: well it could be `contao:change-password`, but I think `contao:password` is not clear enough. Or maybe we call it `contao:user:password` because we also want to add a `contao:user:create` and possibly the same for members?

Toflar: > Or maybe we call it contao:user:password because we also want to add a contao:user:create and possibly the same for members?

👍 

Regarding the command line option for the password: I agree we should keep it but we should maybe add a warning if the user uses it.

leofeyer: > well it could be `contao:change-password`, but I think `contao:password` is not clear enough.

The Unix command to set/change a password is also just `passwd`, but I know what you mean.

> Or maybe we call it `contao:user:password` because we also want to add a `contao:user:create` and possibly the same for members?

👍

aschempp: Command is renamed and warning added.

aschempp:
We might need to rewrite the password generator once we use Symfony security with #559

src/Command/UserPasswordCommand.php
```diff
@@ -0,0 +1,145 @@
...
+use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+
+/**
+ * Resets password for a Contao backend user.
```
xchs: "backend" → "back end"

src/Command/UserPasswordCommand.php
```diff
@@ -0,0 +1,145 @@
...
+class UserPasswordCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('contao:user:password')
+            ->setDescription('Change the password for a Contao backend user.')
```
xchs: "backend" → "back end"

aschempp: 👍 thanks, both fixed

leofeyer:
We should probably wait für @bytehead to complete #559 so we don't do the work twice.

leofeyer:
As discussed on Mumble, rewriting the password generator when switching to Symfony form auth is most likely not necessary. So @aschempp, you only need to complete the unit tests. 😄

aschempp:
Unit tests added. I don't know how to write tests for question input though…

tests/Command/UserPasswordCommandTest.php
```diff
@@ -0,0 +1,282 @@
...
+
+    public function mockContaoFramework(RequestStack $requestStack = null, RouterInterface $router = null, array $adapters = [])
+    {
+        $encryption = $this->getMock('Contao\CoreBundle\Framework\Adapter', ['hash'], ['Contao\Encryption']);
+        $encryption
+            ->expects($this->any())
+            ->method('hash')
+            ->willReturnCallback(
+                function ($password) {
+                    return Encryption::hash($password);
```
discordier: Change this to:
```php
return 'HA$HED-' . $password . '-HA$HED';
```

tests/Command/UserPasswordCommandTest.php
```diff
@@ -0,0 +1,282 @@
...
+            )
+        ;
+    }
+
+    /**
+     * @dataProvider usernamePasswordProvider
+     */
+    public function testDatabaseUpdate($username, $password)
+    {
+        $passwordConstraint = new \PHPUnit_Framework_Constraint_Callback(
```
discordier: Remove all of this and change to:
```php
$connection = $this->container->get('database_connection');
$connection
            ->expects($this->once())
            ->method('update')
            ->with('tl_user', 'HA$HED-' . $password . '-HA$HED', ['username' => $username])
            ->willReturn(1)
        ;
```

leofeyer:
Most of the exceptions should be `InvalidArgumentException` instead of `RuntimeException` according to Symfony. @aschempp Are you ok with me changing this?

aschempp:
What do you mean by "according to Symfony"? I agree some should be different, but a missing argument would probably be an `UnderflowException`?

leofeyer:
Most of the Symfony commands use an `InvalidArgumentException` if one of the arguments is invalid. 😄

aschempp:
So shall I change it for all cases?

discordier:
Underflow exceptions are for removing elements of an empty container.

IMO the following qualify:
- `InvalidArgumentException` - "Exception thrown if an argument does not match with the expected value." http://php.net/manual/en/class.invalidargumentexception.php
- `RuntimeException` - "Exception thrown if an error which can only be found on runtime occurs." http://php.net/manual/en/class.runtimeexception.php
- `\BadMethodCallException` - "Exception thrown if a callback refers to an undefined method or if some
arguments are missing." http://php.net/manual/en/class.badmethodcallexception.php

Of these I also tend to `InvalidArgumentException` as the passed value is clearly an invalid argument.

src/Command/UserPasswordCommand.php
```diff
@@ -0,0 +1,145 @@
...
+            ->addOption('password', 'p', InputOption::VALUE_REQUIRED)
+        ;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        if (null === $input->getArgument('username')) {
```
aschempp: well apparently the unit test can run into this case…

leofeyer:
Thank you @aschempp.

src/Command/UserPasswordCommand.php
discordier: Only because it is not attached to an application which would initialize the input I presume.

leofeyer: No, I have tested it and it also needs to be checked if called via `bin/console …`.

PR #531, opened 2016-06-28
[RFC] Always use the production URL to rebuild the search index

leofeyer:
I have noticed that rebuilding the search index adds the `app_dev.php/` fragment when I am in debug mode. This makes the process very slow, because all front end pages are also opened in debug mode.

This PR simply removes the `app_dev.php/` fragment from the search URLs, however, it might also be a good idea to always strip the fragment from `findSearchablePages()`? @contao/developers

### DIFF
--- a/src/Resources/contao/classes/RebuildIndex.php
+++ b/src/Resources/contao/classes/RebuildIndex.php
@@ -130,7 +130,8 @@ class RebuildIndex extends \Backend implements \executable
 			// Display the pages
 			for ($i=0, $c=count($arrPages); $i<$c; $i++)
 			{
-				$strBuffer .= '<span class="page_url" data-url="' . $arrPages[$i] . '#' . $rand . $i . '">' . \StringUtil::substr($arrPages[$i], 100) . '</span><br>';
+				$strUrl = str_replace('app_dev.php/', '', $arrPages[$i]);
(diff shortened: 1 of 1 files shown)


Toflar:
Why would rebuilding the search index in any other environment than production make any sense? :D

discordier:
I'm afraid, this is bad because you are now hardcoding removal of a fragment entirely unknown to the bundle.
The entry point script is in the application and the bundle must not expect to be run in said environment.
It would be better to disable the cache for non prod environment altogether (via config flag).

leofeyer:
@discordier You are missing the point. It is not about implementation (see my comment above) but about the general question if the script name ever makes sense in the URLs returned by `findSearchablePages()`.

aschempp:
Maybe we should simply disallow search indexing in non-prod environment?

Toflar:
> Maybe we should simply disallow search indexing in non-prod environment?

That's what I asked.

leofeyer:
Uhm, no. What a weird suggestion actually – `getSearchablePages()` also affects sitemaps and RSS feeds and it is a valid use case to rebuild these resources even in dev mode. Only it does not make any sense to add the script name in this case.

aschempp:
Keep in mind that the website can also be run with app.php in the URL, which is totally valid. And we actually don't know the production entry point name…

PR #532, opened 2016-06-29
Do not append the sent headers to the response if they already exist

qzminski:
I have encountered that bug when using the [FineUploader](https://github.com/terminal42/contao-fineuploader) extension with Contao 4.2. It uses the ```JsonResponse``` from Haste in the backend AJAX call which recently started using Symfony ```Response``` class:

https://github.com/codefog/contao-haste/blob/master/library/Haste/Http/Response/Response.php#L252

The main problem is that the Contao ```Ajax``` class sends the header to be ```text/html``` for every callback, even the custom ones:

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/classes/Ajax.php#L181

However the ```JsonResponse``` sets the header content type to be ```application/json```. This causes a serious problem as the ```MergeHttpHeaderListener``` class will append the ```text/html``` content-type header to the response's existing one, which results in having two content-type headers. And that will cause the Apache to terminate with an error.

### DIFF
--- a/src/EventListener/MergeHttpHeadersListener.php
+++ b/src/EventListener/MergeHttpHeadersListener.php
@@ -92,8 +92,10 @@ class MergeHttpHeadersListener
                 header_remove($name);
             }
 
-            // Do not replace existing headers as the response object has a higher priority
-            $response->headers->set($name, trim($content), false);
+            // Do not replace existing response headers
+            if (!$response->headers->has($name)) {
+                $response->headers->set($name, trim($content));
+            }
         }
 
         return $response;
--- a/tests/EventListener/MergeHttpHeadersListenerTest.php
+++ b/tests/EventListener/MergeHttpHeadersListenerTest.php
@@ -127,10 +127,42 @@ class MergeHttpHeadersListenerTest extends TestCase
         $response = $responseEvent->getResponse();
 
         $this->assertTrue($response->headers->has('FOOBAR'));
+        $this->assertSame('content', $response->headers->get('FOOBAR', null, false)[0]);
+    }
 
-        $allHeaders = $response->headers->get('FOOBAR', null, false);
+    /**
+     * Tests that headers are not appended.
+     */
+    public function testHeadersAreNotAppendedIfAlreadyPresentInResponse()
+    {
+        $response = new Response();
+        $response->headers->set('FOOBAR', 'content');
 
-        $this->assertSame('content', $allHeaders[0]);
-        $this->assertSame('new-content', $allHeaders[1]);
+        $responseEvent = new FilterResponseEvent(
+            $this->mockKernel(),
+            new Request(),
+            HttpKernelInterface::MASTER_REQUEST,
+            $response
+        );
+
+        /** @var ContaoFrameworkInterface|\PHPUnit_Framework_MockObject_MockObject $framework */
+        $framework = $this->getMock('Contao\CoreBundle\Framework\ContaoFrameworkInterface');
+
+        $framework
+            ->expects($this->once())
+            ->method('isInitialized')
+            ->willReturn(true)
+        ;
+
+        $headersCount = count($response->headers->get('FOOBAR', null, false));
+
+        $listener = new MergeHttpHeadersListener($framework);
+        $listener->setHeaders(['FOOBAR: new-content']);
+        $listener->onKernelResponse($responseEvent);
+
+        $response = $responseEvent->getResponse();
+
+        $this->assertTrue($response->headers->has('FOOBAR'));
+        $this->assertSame($headersCount, count($response->headers->get('FOOBAR', null, false)));
     }
 }


src/EventListener/MergeHttpHeadersListener.php
```diff
@@ -92,8 +92,10 @@ private function mergeHttpHeaders(Response $response)
                 header_remove($name);
             }
 
-            // Do not replace existing headers as the response object has a higher priority
-            $response->headers->set($name, trim($content), false);
```
leofeyer: Why is this necessary? The third argument should prevent replacing existing headers already.

qzminski: If you provide ```false``` as 3rd argument it will indeed prevent from replacing the header but it will add it to the existing collection instead.

Toflar: Not replacing it means it adds it. You end up having e.g. two `Content-Type` header which is illegal and will not work properly.

leofeyer:
Merged in acd13110f6c0b0e40870540e673c8aaf592d5f20.

leofeyer:
I had to revert this PR in a864da1e5aee09d3dc75b8d38c88b142625ad198. It actually breaks the front end login/logout and the front end preview. :cry:

leofeyer:
@Toflar /cc

leofeyer:
Here are the headers before your changes:

before

And here are the headers after your changes:

after

It seems that the cookie is first unset and then re-set. However, the changes only allow to set one cookie, therefore the cookie is only unset.

I don't know if we should allow additional cookie headers to be added to the response?

leofeyer:
This would work then:

```php
// Do not replace existing response headers
if ('Set-Cookie' === $name) {
    $response->headers->set($name, trim($content), false);
} elseif (!$response->headers->has($name)) {
    $response->headers->set($name, trim($content));
}
```

DanielSchwiperich:
This could also affect my PR for multilogin #454 

And wouldn't your proposal mean that only one cookie can be set. If I remember correct the fe_auth cookie is set on every request for logged in user resulting in that no other cookies can be set.

The set-cookie header is a special case where multiples are allowed: https://tools.ietf.org/html/rfc6265#section-4.1.2

> An origin server can include multiple
   Set-Cookie header fields in a single response.

aschempp:
I would go the opposite way and only forbid duplicate Content-Type headers.

leofeyer:
Fixed in 428fa2272c6d6ba5aae61595be8992af17e13a6a then.

Toflar:
I have to admit, I don't really like this solution. Most of the headers should not be set multiple times (`Server`, `Date`, `Location` etc. pp.). It is the minority that can be set multiple times (`Set-Cookie`, `Link`).

Unfortunately, there's no way of dynamically detecting which ones have to be treated in a special manner but hardcoding `Content-Type` just doesn't feel right to me. This is why I ask you to reconsider. 

I tried to outline my proposed solution:

```php
public function __construct(ContaoFrameworkInterface $contaoFramework, array $allowedMultipleHeaders = [])
{
    $this->contaoFramework = $contaoFramework;
    $this->setHeaders(headers_list());
    $this->allowedMultipleHeaders = $allowedMultipleHeaders;
 }

private function mergeHttpHeaders(Response $response)
{
    foreach ($this->getHeaders() as $header) {
        list($name, $content) = explode(':', $header, 2);
        if ('cli' !== PHP_SAPI) {
            header_remove($name);
        }

        // Simply add if allowed to be set multiple times
        if (in_array($name, $this->allowedMultipleHeaders)) {
            $response->headers->set($name, trim($content), false);
        } else {
            // Only replace if not already present on response object as this one has higher priority
            if (!$response->headers->has($name)) {
                $response->headers->set($name, trim($content));
            }
        }
    }
   
    return $response;
}
```


listener.yml:
```yml
    contao.listener.merge_http_headers:
        class: Contao\CoreBundle\EventListener\MergeHttpHeadersListener
        arguments:
            - "@contao.framework"
            - ['Set-Cookie', 'Link']
        tags:
            - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse, priority: 256 }
```

This would allow developers to easily extend the list for themselves and makes everything more maintainable.

aschempp:
What would happen in Contao 3 if someone calls the same `header()` multiple times? Does PHP know which headers can be set again?

discordier:
:+1: on the approach by @toflar

Toflar:
> Does PHP know which headers can be set again?

PHP doesn't have to know. PHP just passes the headers using the server API (SAPI) so it depends on the implementation I guess :)

discordier:
@aschempp AFAIR it is responsibility of the developer. PHP can not decide as headers are not "known". Everyone may invent/use own headers.

leofeyer:
I'm fine either way.

leofeyer:
Just let me know if you want me to change the implementation.

aschempp:
I see. I like @Toflar's idea, but I would still use a forbid-list. And the DI value should not be necessary by default,  the event listener should have a sane default value.

Toflar:
> but I would still use a forbid-list.

My list **is** a blacklist. The listener should merge headers but there's a list of headers that it is not allowed to merge. You want to have a white list for the headers that are allowed to be set on the response object but I guess that would be a list of 50+ headers because most can be set.

> And the DI value should not be necessary by default, the event listener should have a sane default value.

I can do that, no prob.

aschempp:
Uhm, your example contains `['Set-Cookie', 'Link']` which seems a whitelist?

Toflar:
Those are the headers the listener should ignore. The task of the listener is to merge. So it always merges except for that list which makes it rather a blacklist than a whitelist. But anyway, your approach is not feasible because the list would be incredibly long. As I said, most headers are not allowed to be set multiple times.

leofeyer:
It is a whitelist and it seems a proper solution to the problem IMHO.

leofeyer:
Changed accordingly in 30c3c574fd927d59df409b1a89da03fe665b4872.

DanielSchwiperich:
This implementation could break extensions since now all custom headers are also not allowed to be set multiple times. 
And since the whitelist is hardcoded there is no good way for a developer to just add his headers to the whitelist.

Toflar:
The whitelist can be extended, see additional commits on the hotfix branch.

PR #540, opened 2016-07-13
[RTM] Doctrine schema provider from DCA

aschempp:
This PR adds support for DCA files in Doctrine Migrations or the Doctrine schema manager in general. Looking for initial feedback ☀️ 

**Known Issues & Tasks:**
 - [x] Doctrine migrations seems to require ORM
 - [x] Doctrine does not support key with defined length (as we have in `tl_files.path`). See https://github.com/doctrine/doctrine2/issues/3553
 - [x] We should use a compiler pass to add the services only if Doctrine Migrations is installed (see TODO)
 - [x] Do not require ORM or migrations for install tool
 - [x] Add DCA information to Doctrine ORM schema
 - [x] Add unit tests


Toflar:
Very cool! Really looking forward to this! I guess it should be possible to get rid of the ORM, maybe the requirement is due to a bug (hard dependency) in one of the doctrine bundles but imho it should not be the case.

leofeyer:
Very cool! `tl_files.path` will be a problem, though.

> This is too specific a schema feature. Use EventListener to listen to events triggered from the MySQLPlatform to update the SQL accordingly.

Can we solve it this way as recommended in doctrine/doctrine2#3553?

aschempp:
> Can we solve it this way as recommended in doctrine/doctrine2#3553?

I really doubt it, but I did not thoroughly investigate it. Because the `Schema` from the current table does also not include the index size… but it sure is something we should investigate.

aschempp:
This is now ready for review (apart from unit tests).

How it works:
 1. In a "blank" Contao core installation, the install tool will now use Doctrine schema diff to determine the necessary database adjustments.
 2. If Doctrine Migrations (Bundle) is installed, the `doctrine:migrations:diff` command is now able to determine differences of DCA to database and generate respective migrations files.
 3. If ORM is installed, the ORM mapping information is extended with DCA information so that `doctrine:schema:update` and similar commands will know about DCA tables and update them accordingly. Be aware that this only works if there is at least one regular ORM entity.

aschempp:
I have updated/finalized the implementation and fixed the command issue. The best way I could come up with is to replace the default Migrations command if ORM is not active, because the command does not work at all anyway in this case.

One thing to mention (again): neither the command nor the install tool will work if ORM is enabled but no entity is found. But I think we can live with that?

Toflar:
Can we detect that and show a hint in that case?

aschempp:
Not really. In both cases (install tool and command line) there will be a `No mapping information available` exception.

leofeyer:
Looks good to me even with the "ORM without entities" restriction. Can be merged as soon as the unit tests have been added. :wink:

aschempp:
This is RTM now.

composer.json
```diff
@@ -52,7 +52,9 @@
         "monolog/monolog": "~1.11",
         "phpunit/phpunit": "~4.5",
         "satooshi/php-coveralls": "~0.6",
-        "symfony/phpunit-bridge": "~2.8|~3.0"
+        "symfony/phpunit-bridge": "~2.8|~3.0",
+        "doctrine/orm": "^2.4.8",
```
leofeyer: Why do we require `doctrine/orm` if the code is also supposed to work without ORM?

leofeyer: I found one [reference](https://github.com/contao/core-bundle/pull/540/files#diff-03e30d6a5962b13bb2fe1ad26aa7e4b5R20) to `Doctrine\ORM` though.

aschempp: It's a developer dependency, meaning we need it for the unit tests to test a case where ORM is available ;-)

Toflar: require_dev then?

discordier: This is in the unit test, therefore the dependency is ok in `require-dev`.
Normal installs will not get it pulled, as dev requirements are only pulled for the root project.
Therefore, everything good here. :)

discordier: It already is?

Toflar: Ah, didn't see that. Thought if @leofeyer is asking, it is in `requires`. :) All good then.

leofeyer:
Thank you @aschempp.

PR #544, opened 2016-07-20
ContaoFramework must be able to boot with a request without route

aschempp:


src/Framework/ContaoFramework.php
```diff
@@ -243,10 +243,14 @@ private function getRoute()
...
-            $this->request->attributes->get('_route'),
-            $this->request->attributes->get('_route_params')
-        );
+        try {
+            $route = $this->router->generate(
+                $this->request->attributes->get('_route'),
+                $this->request->attributes->get('_route_params')
+            );
+        } catch (\InvalidArgumentException $e) {
+            return null;
```
discordier: PHP Doc for return value must be changed to `@return string|null` also.

aschempp: question is if `null` is correct? Or if `console` should be `null` too?

leofeyer: I guess it should be `null` on the command line, too, then.

aschempp: I agree, but should that be considered a BC break? I very much doubt anyone uses `TL_SCRIPT` on the command line, as it's only necessary for the legacy `initialize.php`…

leofeyer: I also doubt that anyone uses `TL_SCRIPT` an the command line, but if you are worried about BC, we can just keep the return value.

leofeyer:
Merged in a0c5d6da39b498b1eff4a0df44d32995caa99cf9.

PR #545, opened 2016-07-21
Fix Content-Dispositon Header for downloads

christianromeni:
When set to null, the filename will not be passed in the content disposition.
Seems like Internet Explorer can't handle an empty filename and sets the filename to the last page parameter (on download.html it would be download.pdf instead of file_a.pdf)

### DIFF
--- a/src/Resources/contao/library/Contao/File.php
+++ b/src/Resources/contao/library/Contao/File.php
@@ -755,7 +755,7 @@ class File extends \System
 	 *
 	 * @throws ResponseException
 	 */
-	public function sendToBrowser($filename=null)
+	public function sendToBrowser($filename='')
 	{
 		$response = new BinaryFileResponse(TL_ROOT . '/' . $this->strFile);
 


Toflar:
The fix is correct. All the other browser just take the filename from the URL if you pass a nonsense `Content-Dispositon` header which sort of makes it look correct. However, I think this time I have to come to IE's defence as I think IE behaves correctly here. We provide a `null` filename because we pass `null` [here](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/File.php#L758): 

However, Symfony checks for `''` [here](https://github.com/symfony/http-foundation/blob/master/BinaryFileResponse.php#L160) where it would add the filename automatically.

This PR is RTM for the 4.2.2 hotfix branch 👍 .

leofeyer:
Merged in d0b6150931af6a7e6eaf8b8e6544e2dfb91e8b4c.

PR #547, opened 2016-07-25
Recreate system/tmp/.gitignore

bytehead:
Recreate `system/tmp/.gitignore` after purging the `system/tmp` folder.

Would it make sense to adapt [`rrdir`()](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/Files.php#L117) to prevent deleting `.gitignore` files?


bytehead:
This is not really needed.

PR #548, opened 2016-07-25
[RFC] Re-add title attribute for backend navigation group nodes.

sheeep:
The title attribute is still assigned to the `modules` array but somehow it is currently not used in the template. This PR addresses this issue by re-adding the title attribute ("_Collapse node_") to the node link.

This was removed in 62f6d52968a05c7d496c864bc3bb97346f4245fc.
See https://github.com/contao/core-bundle/commit/62f6d52968a05c7d496c864bc3bb97346f4245fc#diff-72faeb98eda3b28ecf6dedf7ae2ced0cL76

Cc: @discordier


leofeyer:
Re-adding this also requires to fix the tooltip position.

leofeyer:
Merged in ed778702bd00c6ba71d3db690c06ff9a886fe2bb.

Issue #484, opened 2016-05-10
HTTP/2 optimizations


Toflar:
This is a accumulative ticket collecting all the features/todo's for future versions of Contao to better support HTTP/2 to get the best out of it. The list is not final, I'll just keep updating it, every time something comes to my mind.

* [ ] Must be able to disable concatenation of files in the page layout.
* [ ] Add support for server hints (server push) using either `Link: <https://domain.tld/asset.jpg>; rel=prefetch` headers or `<link rel="prefetch" href="https://domain.tld/asset.jpg">` `<head>` tags. Also think about other strategies like `preload` etc.

leofeyer:
> Must be able to disable concatenation of files in the page layout.

This is actually not a requirement, is it? Just something we could stop doing if everyone used HTTP/2.

Toflar:
Well then we could also drop the domain sharding settings. But I think we should not drop it but have it configurable.

leofeyer:
We definitely must not drop it, because currently browsers only support HTTP/2 via HTTPS. And not everybody uses HTTPS (yet).

leofeyer:
Or was it servers that only support it via HTTPS?

Toflar:
You can strip the "currently", I think that won't change :)

aschempp:
Keep in mind the sharding settings are still necessary for CDNs

Toflar:
I don't think that's true. Can you elaborate?

aschempp:
Serving all static files from a CDN?

Toflar:
And how does this relate to our settings? Does not at all. Also, why would you serve the files via CDN but not the page HTML? This is complete nonsense.

aschempp:
Looks like you never had that usecase ;-)

Toflar:
Looks like you make up a use case that does not exist. That's why I asked you to elaborate.

aschempp:
Well I just used that, so it does exist. Dunno whats to explain on "serve static files from CDN"?

discordier:
It makes sense to provide stuff from cdn but not the dynamic content.
The usecase for having support for a cdn is real (my cdn is providing the files and I only reference them).
However, it is not possible to use the current approach of one global setting, as not all static files are available in the cdn (as Contao creates them but does not publish to the cdn), therefore I had to add my own config option and CDN handling anyway in the past.

Therefore, I do not see any sense in keeping the current approach (unless we have also a mechanism for pushing to the cdn) as sharding is not a cdn.

[...]

PR #549, opened 2016-07-26
[RFC] Make script combination optional

leofeyer:
Make script combination optional in case the website supports HTTP/2 (see #484). There is a [related PR for the installation bundle](https://github.com/contao/installation-bundle/pull/23), which also needs to be merged when merging this PR.

@contao/developers /cc

### DIFF
--- a/src/Resources/contao/dca/tl_layout.php
+++ b/src/Resources/contao/dca/tl_layout.php
@@ -97,7 +97,7 @@ $GLOBALS['TL_DCA']['tl_layout'] = array
 	'palettes' => array
 	(
 		'__selector__'                => array('rows', 'cols', 'addJQuery', 'addMooTools', 'static'),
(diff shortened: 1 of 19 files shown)


src/Resources/contao/dca/tl_layout.php
```diff
@@ -272,8 +272,17 @@
...
+			'eval'                    => array('tl_class'=>'w50'),
 			'sql'                     => "varchar(16) NOT NULL default ''"
 		),
+		'combineScripts' => array
+		(
+			'label'                   => &$GLOBALS['TL_LANG']['tl_layout']['combineScripts'],
+			'exclude'                 => true,
+			'inputType'               => 'checkbox',
+			'eval'                    => array('tl_class'=>'w50 m12'),
+			'sql'                     => "char(1) NOT NULL default ''"
```
discordier: I wonder if the default must be `1`.
As previous versions would combine everytime, it is a change in default behaviour when upgrading.

IMO we must at least document and communicate this prominently when it defaults to off.

leofeyer:
Merged in 3dabeef7be78bdb3917a6e6a407a777bf7700d7e and contao/installation-bundle@899a579cff03e36d224c4d107a0f58cae91cd673.

PR #550, opened 2016-07-26
Do not set any user cookies if none was sent as this breaks caching m…

Toflar:
…echanisms


leofeyer:
Needs to be fixed in the LTS version, too.

PR #553, opened 2016-07-26
Add CSS class to image sizes

ausi:
This pull request depends on #342 and adds a css class field to the image size settings.


ausi:
This PR was based on the wrong branch, follow up: #555

PR #554, opened 2016-07-26
[RTM] Add default image densities to theme settings

ausi:
This pull request depends on #342 and adds a default image densities field to the theme settings.

See contao/core#8171

### DIFF
--- a/src/Image/PictureFactory.php
+++ b/src/Image/PictureFactory.php
@@ -90,7 +90,7 @@ class PictureFactory implements PictureFactoryInterface
     {
         $attributes = [];
 
-        if (is_object($path) && $path instanceof ImageInterface) {
+        if ($path instanceof ImageInterface) {
             $image = $path;
         } else {
             $image = $this->imageFactory->create($path);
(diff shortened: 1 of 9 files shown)


leofeyer:
Ok, please rebase the PR. Not sure about the test changes. 😢

ausi:
Done.

src/Resources/contao/pages/PageRegular.php
```diff
@@ -88,6 +88,7 @@ protected function prepare($objPage)
 
 		/** @var ThemeModel $objTheme */
 		$objTheme = $objLayout->getRelated('pid');
+		\System::getContainer()->get('contao.image.picture_factory')->setDefaultDensities($objTheme->defaultImageDensities);
```
leofeyer: Should we check for `'' !== $objTheme->defaultImageDensities` here?

ausi: I don’t think so, because if the default densities are already set in the picture factory, we want to reset it to `''` if the theme doesn’t have it configured.

src/Image/PictureFactory.php
```diff
@@ -73,23 +79,33 @@ public function __construct(
...
+
+    /**
+     * {@inheritdoc}
+     */
     public function create($path, $size = null)
     {
-        if (is_array($size) && isset($size[2]) && 1 === substr_count($size[2], '_')) {
-            $image = $this->imageFactory->create($path, $size);
-            $config = new PictureConfiguration();
+        if (is_object($path) && $path instanceof ImageInterface) {
```
leofeyer: Is the `is_object()` check necessary before `instanceof`?

ausi: Fixed in 0f05cc69dd108915c8e404b4bdbaf245d866b133

PR #556, opened 2016-07-29
[RTM] Added EditorConfig for root and Contao 3 files

aschempp:


.editorconfig
```diff
@@ -0,0 +1,22 @@
...
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.php]
```
leofeyer: This could be simplified as `[*.{php,twig,yml}]`.

src/Resources/contao/.editorconfig
```diff
@@ -0,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+[*.php]
```
leofeyer: This could be simplified as [*.{css,js,php}].

leofeyer:
I have consolidated the files in 3153ac9b3f837841afafafc6d4487d5b8382caa1.

PR #557, opened 2016-08-02
Local exceptions for PaletteManipulator

aschempp:
Replaces #524

### DIFF
--- a/src/DataContainer/PaletteManipulator.php
+++ b/src/DataContainer/PaletteManipulator.php
@@ -10,6 +10,8 @@
 
 namespace Contao\CoreBundle\DataContainer;
 
+use Contao\CoreBundle\Exception\PalettePositionException;
+use Contao\CoreBundle\Exception\PaletteNotFoundException;
 use Contao\StringUtil;
 
 /**
@@ -84,7 +86,7 @@ class PaletteManipulator
      *
      * @return static
      *
-     * @throws \InvalidArgumentException
+     * @throws PalettePositionException
      */
     public function addField(
         $name,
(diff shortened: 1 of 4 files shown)


leofeyer:
As discussed, the nested try-catch blocks need to be added everywhere else where we are using the file locator.

leofeyer:
I think the code could be simplified as follows:

```php
// Now
$blnBubble = false;

try
{
	foreach (\System::getContainer()->get('contao.resource_locator')->locate('dca/' . $this->strTable . '.php', null, false) as $file)
	{
		try
		{
			include $file;
		}
		catch (\InvalidArgumentException $e)
		{
			$blnBubble = true;
			throw $e;
		}
	}
}
catch (\InvalidArgumentException $e)
{
	// Re-throw the InvalidArgumentException if it has originated from a DCA file
	if ($blnBubble)
	{
		throw $e;
	}
}
```

```php
// Then
try
{
	$files = \System::getContainer()->get('contao.resource_locator')->locate('dca/' . $this->strTable . '.php', null, false);
}
catch (\InvalidArgumentException $e)
{
	$files = null;
}

if ($files !== null)
{
	foreach ($files as $file)
	{
		include $file;
	}
}
```

@aschempp What do you think?

aschempp:
very good idea. If you assign an empty array instead of `null` you can even skip the `if` statement :-)

leofeyer:
Exactly what I though right now. 😄

leofeyer:
Merged in ef216e9f8442ce90fa8988a6a623a7a77ca1307e.

PR #558, opened 2016-08-02
[RFC] Harden authentication against session fixation

discordier:
This hardens the authentication against session fixation attacks.
Thanks to Rico Kaltofen for pointing it out.

Todo:
- [x] Check the symfony session migration parameter.

### DIFF
--- a/src/Resources/contao/library/Contao/User.php
+++ b/src/Resources/contao/library/Contao/User.php
@@ -10,6 +10,7 @@
 
 namespace Contao;
 
+use Symfony\Component\Security\Http\Session\SessionAuthenticationStrategy;
 
 /**
  * Authenticates and initializes user objects
@@ -567,6 +568,27 @@ abstract class User extends \System
 	 */
 	protected function generateSession()
 	{
+		// Regenerate a new session id to harden against session fixation attacks if migration is configured.
+		// This is mostly copy&paste from the default session authentication strategy by symfony.
+		$container = \System::getContainer();
+		switch ($strategy = $container->getParameter('security.authentication.session_strategy.strategy'))
+		{
+			case SessionAuthenticationStrategy::NONE:
+				break;
+
+			case SessionAuthenticationStrategy::MIGRATE:
+				// Never destroy the session, this is Contao special, as we have to preserve the current values.
+				$container->get('session')->migrate();
+				break;
+
+			case SessionAuthenticationStrategy::INVALIDATE:
+				$container->get('session')->invalidate();
+				break;
+
+			default:
+				throw new \RuntimeException(sprintf('Invalid session authentication strategy "%s"', $strategy));
+		}
+
 		$time = time();
 
 		// Generate the cookie hash


leofeyer:
@contao/developers Haven't we already agreed that this fix is wrong?

discordier:
Not more than the overall implementation.
The only issue that might arise is, that we are migrating a session where the config tells not to migrate which implies custom authentication but using that is not possible with the current state of Contao 4.
Support for this will come from @bytehead for 4.3.
Of course we can leave this unpatched in 4.2 and wait for 4.3 to get it properly done, I simply wanted to get the fix from 3.5 future ported to 4.2 as well.

To be honest, there were actually only two votes so far. @aschempp voted against patching and waiting for 4.3 and I voted for patching like done in this PR.
No vote from @leofeyer, @Toflar, @bytehead, @sheeep or @ausi yet, therefore it is still up for discussion.

Toflar:
If @bytehead is sure to implement full sf authentication support for **both**, back end as well as front end users, for Contao 4.3, I think we should not do anything about it. If he thinks it won't make it into 4.3 we should discuss this and I'd likely vote for merging it.

leofeyer:
There is now a PR for the Symfony authentication: #559

ausi:
If there are known BC breaks I’d vote against merging it in 4.2 and 3.5.

discordier:
> If there are known BC breaks I’d vote against merging it in 4.2 and 3.5.

So far there is no evidence that there are.

leofeyer:
> So far there is no evidence that there are.

That's not a valid argument, because there is no evidence that there are not, either.

aschempp:
As I said to @leofeyer, it's primarily a gut feeling that tells me this is causing side effects. No hard evidence on my side. But I also pointed out the possible attack vector is so unrealistic in my opinion that it's not worth the risk of breaking stuff (especially because we're switching to Symfony Security soon).

Toflar:
The session fixation vulnerability is listed on second position on the OWASP top 10 list. I think this should be proof of its importance. It's not something neglectable at all...

aschempp:
> Most session fixation attacks are web based, and most rely on session identifiers being accepted from URLs (query string) or POST data.

> https://en.wikipedia.org/wiki/Session_fixation

Sure, but in Contao it is not possible as a remote attack (because we disallow URL session), the session is (by default) fixed to an IP address and it will time out after a few minutes. I think that already prevents a realistic scenario. Anything involving manually walking up to a computer to attack an account is highly unrealistic.

discordier:
> Sure, but in Contao it is not possible as a remote attack (because we disallow URL session), the session is (by default) fixed to an IP address and it will time out after a few minutes. I think that already prevents a realistic scenario. Anything involving manually walking up to a computer to attack an account is highly unrealistic.

1. The "manually walking over" also is a Teamviewer session or the like.
2. For the timing out, this does not matter at all when the session id can predicted and inherited. I simply have to reload until you are logged back into the backend and I am back in (this is assuming a shared computer like a kiosk, library, university or anything like this for ease of example).
3. If anyone disables the IP binding in the Backend settings (which is a feature and will get the only possibility in 4.3 as symfony does not bind to ip however prevents session fixation), all your sessions are wide open, I only need you to log in using my somehow placed cookie in your browser and all subsequent session ids will be like I want them to be. The only thing I have to do is trick your browser to take my cookie (however I could do it, multiple possibilities, placing it in your browser by hand is only the most obvious one) and the session ID will never again(!) change.
4. As the non binding to ip and session fixation prevention will become the only supported way in 4.3, everything that will break due to this change will 100% also break by the merging of the #559 by @bytehead, as in there the session id will also migrate (as already stated above by you).

aschempp:
Aaaand… your long list of arguments/conditions shows why I find this unrealistic 😉

discordier:
> Aaaand… your long list of arguments/conditions shows why I find this unrealistic :wink:

If you say so... but you seem to ignore the fact that browsers might suffer from cross domain cookie injection or the like, yet again, above mentioned scenarios are only a simplification of a concrete example because the possibilities to inject a cookie to a victims machine are potentially endless, ranging from zero day exploits in any software to outdated software.
.

leofeyer:
As discussed in Mumble on August 11th, @discordier will rework the PR and check the Symfony session configuration.

discordier:
PR has been updated with fetching the session fixation parameter.
It now behaves exactly like the symfony session migration with the exception that we do not destroy the session upon migration.
This is changed to honor the request by @aschempp that existing session values must be preserved.

It works on my system like a charm, the session cookie is properly changed after login but preserved if the strategy is set to none. Invalidation was not possible in Contao before but is in Symfony and therefore is now also honored rather than throwing an exception for that also.

So please review again.

src/Resources/contao/library/Contao/User.php
```diff
@@ -567,6 +568,27 @@ public function save()
...
+		// Regenerate a new session id to harden against session fixation attacks if migration is configured.
+		// This is mostly copy&paste from the default session authentication strategy by symfony.
+		$container = \System::getContainer();
+		switch ($strategy = $container->getParameter('security.authentication.session_strategy.strategy'))
+		{
+			case SessionAuthenticationStrategy::NONE:
+				break;
+
+			case SessionAuthenticationStrategy::MIGRATE:
+				// Never destroy the session, this is Contao special, as we have to preserve the current values.
```
leofeyer: Why is migration "Contao special"?

leofeyer:
Merged in 7950d7fb1a778ab8b3b7642482fc3db2b4d355f2.

PR #559, opened 2016-08-03
[WIP] Symfony authentication

bytehead:
This is an early working draft to use the replace the current `simple_preauth` mechanism with the Symfony `form_login` authentication listener.

ToDos:
- [ ] Fix messages on `be_login.html5`
- [ ] Implement `$GLOBALS['TL_HOOKS']['postAuthenticate']` as an event
- [ ] Test frontend preview
- [ ] ~~Use `form_login` for `FrontendUser` (members) as well~~*
- [ ] ~~Implement role switcher / impersonate any `FrontendUser` for frontend preview~~*
- [ ] Write tests
- [ ] Code style

*I'll provide the Symfony authentication for FE users in a separate PR.


aschempp:
❤️ 

I think we should not add frontend yet, just replace the backend in 4.3. This will give use more time to see into the solution, as frontend is more complex to add and has more potential side effects.

DanielSchwiperich:
Please keep also in mind there's the open PR for Multilogin which introduces also some Symfony related changes to the authentication system. There may be some overlappings.
#481

src/Resources/contao/classes/BackendUser.php
```diff
@@ -154,39 +158,6 @@ public function __get($strKey)
 		return parent::__get($strKey);
 	}
 
-
-	/**
-	 * Redirect to the login screen if authentication fails
-	 *
-	 * @return boolean True if the user could be authenticated
-	 */
-	public function authenticate()
```
discordier: Sadly you must keep this method as it is part of the API and must continue to work as it does currently.

Toflar: Imho it does because the parent class handles it.

src/Resources/contao/library/Contao/User.php
```diff
@@ -256,58 +260,24 @@ public function getData()
...
-		}
+		/** @var TokenStorageInterface $tokenStorage */
+		$tokenStorage = $container->get('security.token_storage');
 
-		$time = time();
-		$container = \System::getContainer();
-		$session = $container->get('session');
-
-		// Validate the session
-		if ($objSession->sessionID != $session->getId() || (!$container->getParameter('contao.security.disable_ip_check') && $objSession->ip != $this->strIp) || $objSession->hash != $this->strHash || ($objSession->tstamp + \Config::get('sessionTimeout')) < $time)
```
discordier: Restricting the session to an IP must be kept somehow.

src/Resources/contao/controllers/BackendIndex.php
```diff
@@ -32,35 +45,12 @@ class BackendIndex extends \Backend
...
 
-		// Login
-		if ($this->User->login())
-		{
-			$strUrl = 'contao/main.php';
-
-			// Redirect to the last page visited
-			if (\Input::get('referer', true) != '')
-			{
-				$strUrl = base64_decode(\Input::get('referer', true));
```
discordier: Can we pass this information to the login route and redirect back post authentication?

src/Resources/contao/library/Contao/User.php
Toflar: No, we said Symfony doesn't so we won't either.

discordier: Which is an BC break then and the whole stuff has to get postponed to 5.0 then... I do not like this.

src/Resources/contao/classes/BackendUser.php
discordier: The redirect is missing in the parent class... in fact, the redirect is the only thing which made `BackendUser::authenticate()` almost unusable for the outside world, however this is an BC break if this method now will behave different.
We do not have to use it in the core anymore but it has to remain for existing extensions which call it and expect the redirect and script termination to happen.

src/Resources/contao/library/Contao/User.php
Toflar: This is not a BC break to me.

src/Resources/contao/controllers/BackendIndex.php
bytehead: I'll give it a shot :)

src/Resources/contao/classes/BackendUser.php
bytehead: The method is back in there - I've got it wrong in the first view. 
If we now use `BackendUser::authenticate()` we are compatible to all other backend entry points (see https://github.com/contao/core-bundle/pull/559/commits/58e6b65c9569fd70532332118dbdb7baac834bcf).

src/Resources/contao/controllers/BackendIndex.php
bytehead: This is probably done [automatically](http://symfony.com/doc/current/security/form_login.html#redirecting-after-success).

First tests worked ok. @discordier Is there any edge cases I could have missed?

discordier: The problem with the original symfony approach is, that it is stored in the session and not as URL parameter.
There might arise issues from this when opening a browser with multiple tabs.
The stored value in the session will be the one for the last tab that get's redirected to the login form then and cause all tabs to contain the same backend route after login.
We could implement it as stated here http://symfony.com/doc/current/security/form_login.html#control-the-redirect-url-from-inside-the-form
We would need to make our value in the template variable though as someone might reconfigure the application to use another parameter name (see config.yml in above link).

aschempp: @bytehead can you check my old implementation. Pretty sure I fixed it there, but I can't remember how exactly.

leofeyer:
@bytehead What's the status on this one?

src/Resources/contao/controllers/BackendIndex.php
bytehead: @aschempp didn't find anything. Do you have any hint?

bytehead:
@leofeyer still in progress - didn't find any time during holidays...

src/Resources/contao/controllers/BackendIndex.php
aschempp: Here you go: http://symfony.com/doc/current/security/form_login.html#control-the-redirect-url-from-inside-the-form

bytehead: I'm actually curious about your old implementation you've mentioned.

aschempp: That was the solution I remembered, but I could not find my implementation either.

aschempp:
I guess this was not supposed to be closed, but @bytehead and me wanted to create a new PR anyway, right?

bytehead:
Closed in favor of #685.

PR #561, opened 2016-08-07
Fix single-line doc comments

ausi:


leofeyer:
Merged in 99096cffcd89b2ac7be22df7371f4377ccab7d3b and complemented in d0048c1722f4c2c564b317223b12a3061f8d7e43.

ausi:
But you did the exact opposite in d0048c1722f4c2c564b317223b12a3061f8d7e43 ?

ausi:
AFAIK `/** @var` is correct and `/* @var` is incorrect.

leofeyer:
Hm, Doctrine uses `/* @var` and Symfony is using both, so I don't really know.

aschempp:
PhpStorm generates `/** @var`, I'd go with that.

leofeyer:
NetBeans recommends `/* @var` though: https://blogs.oracle.com/netbeansphp/entry/defining_variable_type_in_a

ausi:
[phpDocumentor](https://phpdoc.org/docs/latest/glossary.html) says:

> `/** This is a single line DocComment. */`

My editor only supports `/** @var`.

leofeyer:
The [PSR-5 draft](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#55-examples) (written by Mike himself) recommends `/** @var`, so I'll revert the changes.

leofeyer:
Reverted in af746de491d701bc4ec55296e994598f9f5fed17.

ausi:
PHP itself only understands `/**`:

```php
class Foo 
{
	/** @var string $foo */
	public $foo = 'foo';

	/* @var string $bar */
	public $bar = 'bar';
}

$reflector = new \ReflectionClass('Foo');
var_dump($reflector->getProperty('foo')->getDocComment()); 
// string(23) "/** @var string $foo */"
var_dump($reflector->getProperty('bar')->getDocComment());
// bool(false)
```

PR #562, opened 2016-08-08
[WIP] Feature/esi module

Toflar:
Finally got it! ESI is quite complicated so I'll put this PR on RFC because I think there are a few things to discuss but it is working like a charm on my development environment already!
## What ESI is

For the people not knowing about ESI: Please read this: https://en.wikipedia.org/wiki/Edge_Side_Includes

Basically, you split up the website into different parts. The main request contains an arbitrary number of `<esi src="...>"` tags which will be then evaluated and requested using a separate HTTP request. It is thus possible to e.g. cache your whole page for 2 days except for a shopping cart module.
## What you can now do thanks to this PR
- Configure caching settings for both, client and server cache per page (previously was a system setting so you could not distinguish).
- Configure caching settings for every single front end module.

It's thus now very easy to e.g. cache your whole page on the client side for 1 minute (= no requests to the server at all), 2 days on the server side (= the server does not need to generate any content for this page) **except** for module A in the right column, module B in the footer etc. 😎 😎 😎 
## What this PR does
- Replaces the current very basic HTML output cache by the Symfony reverse proxy. I have taken this chance to enhance the caching settings in `tl_page` to be different for either the client or the server (reverse proxy) cache.
- To support our "uncacheable" insert tags like `{{date}}` or `{{ua}}` or anything with the insert tag flag `uncached`, we now generate `<esi>` tags for them which will make sure, they are always re-rendered.
- Adds regular expression filtering settings to the `AddToSearchIndexListener` because it shall not index on `/_fragment` requests (or whatever you configured it to be).
- Adds ESI caching settings to every front end module. You can simply activate ESI and then choose from the various settings which I have included. Make sure you read the language files to understand why they are there and why they are useful. Together with @aschempp and @ausi  we have tried to think of many different use cases which is why the settings are there. E.g. the setting `ignore page information` can be useful if you know, your module (e.g. a simple `html` module) does not need any `global $objPage` information and you can thus improve the performance and cache usage (no `page` information in `<esi src="....">` url = not different cache files per page) etc. I have even added support for our own `ResponseException`.

screen shot 2016-08-08 at 17 47 51
## Changes needed in config.yml

``` diff
# Framework configuration
framework:
    ...
    translator: { fallbacks: ["%locale%"] }
+   esi: { enabled: true }
+   fragments: { enabled: true, path: /_fragment }
```

I'm really excited about this, hope you are too and looking forward to the comments 😄  🌞 
## ToDo's
- [ ] Unit tests
- [ ] Update routine (tl_settings.cache should be considered and tl_page entries should be adjusted accordingly)
- [x] Consider BE_USER_LOGGED_IN and FE_USER_LOGGED_IN for ESI requests. Do we really need Vary header support?
- [x] How to consider TL_VIEW Cookie?
- [x] How do we load `$objPage` correctly? --> `method_exists()` for page types.
- [x] Should we catch the `ResponseException`? No, listeners do this.
- [ ] Is the way to disable the search index listener for `_fragment` requests ok?
- [ ] Should ESI be enabled (current state) or disabled by default?
- [ ] What about Hybrids?
- [ ] Does Symfony ignore ESI tags generation on non-safe HTTP methods?
- [ ] Use `supportsEsi()` and make sure it is not enabled by default for non core modules.
- [ ] Modules must pass params to keep, not user.
- [ ] Root page must include vary header for `Accept-Language`.
- [x] Add check for `mobileLayout` in `FrontendTemplate`.
- [ ] Move `$GLOBALS` for JS, CSS etc. to `renderEsi()` for every module.

PS: Commit messages are mostly nonsense, I'll squash everything once it's ready anyway.


fenepedia:
Great! Contao 4 get more and more exiting features.

psren:
Clever move @Toflar
I also see the opportunity to include parts of the CMS into other systems.

e.g. if you use a shopsoftware you could include the content management parts via ESI and keep your controllers untouched.

Toflar:
> I also see the opportunity to include parts of the CMS into other systems.
> 
> e.g. if you use a shopsoftware you could include the content management parts via ESI and keep your controllers untouched.

You can do this ever since we moved to Symfony :D

aschempp:
> e.g. if you use a shopsoftware you could include the content management parts via ESI and keep your controllers untouched.

I don't think that will easily work, as ESI requests are protected and you can't really generate the secure URL from outside of Symfony…

Toflar:
> I don't think that will easily work, as ESI requests are protected and you can't really generate the secure URL from outside of Symfony…

ESI requests are not protected. Only fragment ESI requests which is what we use. That's why I'm saying: You could've done this ever since we moved to Symfony by adjusting your `config.yml` and adding `AppCache` to your `app.php` :) But anyway, this is off topic.

psren:
I thought of fragment ones out of others Bundles installed in the same Symfony Installation.
It's possible to use the existing Fragment Actions now without writing own actions.
Better Caching is the bigger part of this PR indeed.

src/Controller/EsiController.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public function __construct($framework)
+    {
+        $this->framework = $framework;
+    }
+
+    /**
+     * @param string $insertTag
+     */
+    public function renderInsertTag($insertTag)
```
discordier: For the sake of verbose code, this variable should rather be named `$buffer` as it is the buffer in which insert tags should get replaced.

Toflar: No it is not :) The variable contains e.g. `{{date::d.m.Y}}`. The place where it is called never contains anything else than just an insert tag.

discordier: Ah, sorry did not know that. Maybe if phpdoc was added I would have known. ;)
I guess it is fine like it is then.

Toflar: As it really **is** just the insert tag I thought adding php doc is useless here :)

src/Resources/contao/classes/Frontend.php
```diff
@@ -648,232 +648,4 @@ public static function indexPageIfApplicable(Response $objResponse)
...
 		}
 	}
-
-
-	/**
-	 * Check whether there is a cached version of the page and return a response object
-	 *
-	 * @return Response|null
-	 */
-	public static function getResponseFromCache()
```
discordier: I think we must keep this for BC reasons as:
```php
public static function getResponseFromCache()
{
    return null;
}
```

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -240,103 +239,6 @@ public function hasAuthenticatedBackendUser()
 
 
 	/**
-	 * Add the template output to the cache and add the cache headers
-	 */
-	protected function addToCache()
```
discordier: Unsure if this method must be kept as empty wrapper for bc reasons.

src/Resources/contao/classes/Frontend.php
Toflar: This was only for internal use anyway. I don't think we have to.

src/Resources/contao/dca/tl_settings.php
```diff
@@ -164,14 +164,6 @@
 			'inputType'               => 'text',
 			'eval'                    => array('rgxp'=>'url', 'tl_class'=>'w50')
 		),
-		'cacheMode' => array
```
discordier: Do we need an upgrade task in the install tool that will update all entries in `tl_page` if this setting was set to anything other than `none`?

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Same here but I'd say no.

src/Resources/contao/dca/tl_settings.php
Toflar: Yes, you are right. Adding this to the todo section in the issue.

src/Resources/contao/classes/Frontend.php
discordier: Well, from this point of view, 90% of the code base might be for internal use only. In the past we rejected event the slightest change in the API which could have been overridden/used by others and this includes all `public` and even most of the `protected` methods.
Keeping it as No-op and maybe commenting with "do not use, use ESI" in the phpDoc does not harm IMO but will not crash anything.

src/Resources/contao/classes/FrontendTemplate.php
discordier: We definitely loose a HOOK in here, so this may be a code flow change.
Granted, it should only be used to alter the cache key but might have been used for doing other stuff.
I still am undecided.

Toflar: You could only modify the cache key which is now simply irrelevant :)

bytehead: I use the hook in one of my extensions afaik.

discordier: > You could only modify the cache key which is now simply irrelevant :)

I could have done anything I want in there up to telling my coffee machine to brew a fresh cup, but as stated above, this is arguable and I am willing to sacrifice the hook.

src/Controller/EsiController.php
discordier: Well, at least I did not know until your clarification, therefore docs are not useless at all.
I agree, when you see this PR isolated, it does make sense as it is, just in context with the code of the `InsertTag` class (where the input is a whole buffer) it is not that intuitive.
So let's keep it as it is.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: > I use the hook in one of my extensions afaik.

What for?

Toflar:
> Consider BE_USER_LOGGED_IN and FE_USER_LOGGED_IN for ESI requests. Do we really need Vary header support?

Regarding this. At the moment, the main document is `private`  (so never cached by the reverse proxy) if a front end user is logged in or the front end preview is active. I think we should also add this to the `EsiController` methods. Agree?
The question is, whether we really need a setting for `Vary` headers. I mean, yes, it does make sense but are there really use cases for us? Let's face it, `Vary: Cookie` does not make sense. Cookie is so generic that we'll hardly ever have a cache hit. `Content-Type`? In a front end module? Not really, isn't it? But maybe yes, so we should better keep it? Undecided. WDYT?

> How to consider TL_VIEW Cookie?

Currently, this is not yet considered which it obviously has to. I can simply make it a request attribute. I could also make a random `key` request attribute which I calculate using `md5()`. This would allow us to add more stuff to that key calculation (like e.g. if a front end user is logged in). Point is this: Symfony is going to generate this as something like `<esi src="_fragment?_path=informationregardingthecontroller-and-containing-all-other-attributes-such-as-the-key-if-I-pass-it">` which will make sure, everytime the key is different, the cache entry is different too. So it is somehow something like a `Vary` header but it can be more specific than the `Vary` header :-)
I could even ask the module instance so it has the chance to decide on its own. WDYT?

Maybe easier on the phone...

src/Resources/contao/classes/FrontendTemplate.php
bytehead: I use it [here](https://github.com/1up-lab/contao-cache-warmup/blob/master/src/CacheWarmup/CacheWarmup.php#L219) to generate cache files for desktop and mobile [independent from a mobile layout](https://github.com/contao/core/blob/master/system/modules/core/classes/FrontendTemplate.php#L233-L244).

discordier:
> Maybe easier on the phone...

I'd say so too, let's discuss this on mumble.
However, I agree with you that it should simply be one (of many) request attributes.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: But how does this make sense? :D The mobile layout should be considered in the cache key?

bytehead: I don't really know. But for creating `.mobile` cache files, contao is checking  `$objPage->mobileLayout > 0` but not for getting them from cache: https://github.com/contao/core/blob/master/system/modules/core/controllers/FrontendIndex.php#L392

I never have a mobile layout defined in themes.

So at the end the system never creates cache files for mobile devices - but [checks their existence](https://github.com/contao/core/blob/master/system/modules/core/controllers/FrontendIndex.php#L397).

And with my hook Contao generates `.mobile` cache files anyway.

Toflar: Okay, but it's not a use case we have to consider. If somebody uses the mobile layout, the SF reverse proxy has to consider this but the hook is not relevant.

bytehead: Sure. The Contao cache gets unused anyway...

src/Resources/contao/classes/Frontend.php
Toflar: Agreed. Brought back in 2df0b7578cb0009758398d446845de7e6fe43f46.

Toflar:
I thought about the Vary header support again and I'd vote to keep it. Moreover, I've implemented the `cacheKey` solution and already used it to support `TL_VIEW` and it's working perfectly fine 😊   I've found a nice way to provide full control to the modules themselves, too. 😎

discordier:
A pretty neat idea and implementation, indeed.

src/Resources/contao/classes/FrontendTemplate.php
aschempp: The equivalent of this would be an event that allows to modify the `Vary` headers, or the response in general, right?

src/Controller/EsiController.php
```diff
@@ -0,0 +1,117 @@
...
+
+        try {
+            $result = $this->framework->createInstance('Contao\InsertTags')
+                ->replace($insertTag, false);
+            $response = new Response($result);
+            
+            // Never cache non cacheable insert tags
+            $response->setPrivate();
+
+        } catch (ResponseException $e) {
```
aschempp: is that even necessary? Our ResponseExceptionListener should handle them just fine?

src/Controller/EsiController.php
```diff
@@ -0,0 +1,117 @@
...
+        $sharedMaxAge = 0
+    ) {
+        $this->framework->initialize();
+
+        if (0 !== (int) $pageId) {
+            $page = $this->framework->getAdapter('Contao\PageModel')
+                        ->findWithDetails((int) $pageId);
+
+            if (null !== $page) {
+                $GLOBALS['objPage'] = $page;
```
aschempp: There is more to be done to generate a *global page object*. See https://github.com/isotope/core/blob/master/system/modules/isotope/library/Isotope/Frontend.php#L646

src/Controller/EsiController.php
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        // Never cache on front end preview
+        if (true === BE_USER_LOGGED_IN) {
+            $response->setPrivate();
+            return $response;
+        }
+
+        if ((int) $sharedMaxAge > 0) {
+            $response->setSharedMaxAge((int) $sharedMaxAge);
```
aschempp: This must not be done on a `ResponseException`. As said before, we probably shouldn't catch it?

src/DependencyInjection/Compiler/AddFragmentsSupportPass.php
```diff
@@ -0,0 +1,45 @@
...
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Adds the composer packages and versions to the container.
```
aschempp: wrong

src/DependencyInjection/Compiler/AddFragmentsSupportPass.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+class AddFragmentsSupportPass implements CompilerPassInterface
```
aschempp: this class name is rather strange for what it does. Wouldn't we be better of by just having a `disabled` method on the `AddToSearchIndexListener` which can be called by any controller? Or just the opposite which is only called in our frontend controller?

src/Resources/contao/classes/Frontend.php
```diff
@@ -649,231 +649,19 @@ public static function indexPageIfApplicable(Response $objResponse)
...
 	}
 
-
 	/**
 	 * Check whether there is a cached version of the page and return a response object
 	 *
 	 * @return Response|null
+	 *
+	 * @deprecated This method will always return null and does not have any effect
+	 *             anymore. Caching is achieved using the powerful Symfony reverse
```
aschempp: If to use the Sf reverse proxy is a configuration of the application (`contao/standard-edition`), it could very well be Varnish or a CDN. Something like `Caching is handled by the reverse proxy, depending on your application setup`.

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -414,4 +316,34 @@ public function getCustomSections($strKey=null)
...
+			return;
+		}
+
+		// If FE_USER_LOGGED_IN or BE_USER_LOGGED_IN every request is private
+		if (true === FE_USER_LOGGED_IN || true === BE_USER_LOGGED_IN) {
+			$response->setPrivate();
+			return;
+		}
+
+		// The client cache is always depending on the admin settings
```
aschempp: what do you mean by *admin settings* ?

src/Resources/contao/library/Contao/Controller.php
```diff
@@ -174,12 +174,13 @@ public static function getTemplateGroup($strPrefix)
...
-	 * @param mixed  $intId     A module ID or a Model object
-	 * @param string $strColumn The name of the column
+	 * @param mixed  $intId        A module ID or a Model object
+	 * @param string $strColumn    The name of the column
+	 * @param bool   $blnIgnoreEsi Whether ESI should be ignored
 	 *
 	 * @return string The module HTML markup
 	 */
-	public static function getFrontendModule($intId, $strColumn='main')
+	public static function getFrontendModule($intId, $strColumn='main', $blnIgnoreEsi = false)
```
aschempp: I think this should be reversed. For BC reasons, ESI should be disabled by default, and our core calls should enabled it if appropriate.

src/Resources/contao/modules/Module.php
```diff
@@ -456,4 +458,108 @@ public static function findClass($strName)
...
+	 * the "varyHeaders" array key in getEsiAttributes().
+	 *
+	 * @return string
+	 */
+	protected function getEsiCacheKey()
+	{
+		/** @var Request $request */
+		$request = \System::getContainer()->get('request_stack')->getCurrentRequest();
+
+		return md5($request->cookies->get('TL_VIEW'));
```
aschempp: Can't this produce 3 states (not set, `desktop` and `mobile`)?

aschempp:
Looks great! Can you provide the necessary changes (PR) to `contao/standard-edition` so we can test this in a real application?

Toflar:
It's 2 lines in the config.yml and they are part of the description. Why would I make a PR for 2 lines?

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Should be possible by the regular `kernel.response` listeners? No?

src/Controller/EsiController.php
Toflar: And probably more. Depends on the page type. I'll add this to the issue description.

src/Controller/EsiController.php
Toflar: Added this to the issue description as well.

src/DependencyInjection/Compiler/AddFragmentsSupportPass.php
Toflar: Thx, fixed in a6529367cf958e7dabb71dcec423b76f53a67877.

src/DependencyInjection/Compiler/AddFragmentsSupportPass.php
Toflar: Dunno what's best. Added to the issue description.

src/Resources/contao/classes/Frontend.php
Toflar: Fixed in 7752432cd69274061ab68cdbdea038f7cc2081b4.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: I've removed it because the comment is apparently confusing.

src/Resources/contao/library/Contao/Controller.php
Toflar: I disagree but I put it on the list.

src/Resources/contao/modules/Module.php
Toflar: That's correct. Fixed in 250e5177e3ed638b20d001080a061d7916be53ac.

src/Resources/contao/library/Contao/InsertTags.php
```diff
@@ -112,9 +114,14 @@ protected function doReplace($strBuffer, $blnCache)
 			// Skip certain elements if the output will be cached
 			if ($blnCache)
 			{
-				if ($elements[0] == 'date' || $elements[0] == 'ua' || $elements[0] == 'post' || $elements[0] == 'file' || $elements[1] == 'back' || $elements[1] == 'referer' || $elements[0] == 'request_token' || $elements[0] == 'toggle_view' || strncmp($elements[0], 'cache_', 6) === 0 || in_array('uncached', $flags))
+				if ($elements[0] == 'date' || $elements[0] == 'ua' || $elements[0] == 'post' || ($elements[0] == 'file' && !\Validator::isStringUuid($elements[1])) || $elements[1] == 'back' || $elements[1] == 'referer' || $elements[0] == 'request_token' || $elements[0] == 'toggle_view' || strncmp($elements[0], 'cache_', 6) === 0 || in_array('uncached', $flags))
```
leofeyer: Why did you change this? And isn't `$elements[1]` always a string UUID?

src/Resources/contao/dca/tl_page.php
```diff
@@ -520,6 +520,19 @@
 			'inputType'               => 'select',
 			'options'                 => array(0, 5, 15, 30, 60, 300, 900, 1800, 3600, 10800, 21600, 43200, 86400, 259200, 604800, 2592000),
 			'reference'               => &$GLOBALS['TL_LANG']['CACHE'],
+			'eval'                    => array('tl_class'=>'w50'),
+			'sql'                     => "int(10) unsigned NOT NULL default '0'"
+		),
+		'clientCache' => array
```
leofeyer: What's the point of having two different cache timeouts?

leofeyer: I do understand that you want to set `setMaxAge()` and `setSharedMaxAge()` with different values, but what is the point?

src/Resources/contao/library/Contao/InsertTags.php
Toflar: To improve the cache. `file` initially included template files only (that's why this is not cacheable). Later we introduced support for the UUID as a parameter but did not change this.

src/Resources/contao/dca/tl_page.php
aschempp: One is the client (browser) cache and the other is the reverse proxy cache time.

Toflar:
This is on hold because there is a lot of other stuff that needs to be done first, otherwise the subrequests cannot be handled properly.

src/Controller/EsiController.php
Toflar: Added in 7f13a07b9a45ac993865f6f933ec945bc08f9d41.

Toflar:
Closed in favour of https://github.com/contao/core-bundle/pull/700 and other PR's that have been merged standalone in the meantime :)

PR #563, opened 2016-08-10
[RTM] Allow Model::find() to return arrays instead of collections

aschempp:
this will make developer live soooo much easier by simply opting-in on arrays 😎 

Also solves https://github.com/contao/core/issues/6147

### DIFF
--- a/src/Resources/contao/library/Contao/Model.php
+++ b/src/Resources/contao/library/Contao/Model.php
@@ -1083,7 +1083,7 @@ abstract class Model
 
 		if ($objResult->numRows < 1)
 		{
-			return $arrOptions['return'] == 'Array' ? [] : null;
+			return $arrOptions['return'] == 'Array' ? array() : null;
 		}
 
 		$objResult = static::postFind($objResult);
--- a/src/Resources/contao/library/Contao/Model.php
+++ b/src/Resources/contao/library/Contao/Model.php
@@ -1083,7 +1083,7 @@ abstract class Model
 
 		if ($objResult->numRows < 1)
 		{
-			return null;
+			return $arrOptions['return'] == 'Array' ? [] : null;
 		}
 
 		$objResult = static::postFind($objResult);
@@ -1099,6 +1099,12 @@ abstract class Model
 			}
 
 			return static::createModelFromDbResult($objResult);
+		}
+		else if ($arrOptions['return'] == 'Array')
+		{
+			$objCollection = static::createCollectionFromDbResult($objResult, static::$strTable);
+
+			return $objCollection->getModels();
 		}
 		else
 		{
--- a/src/Resources/contao/library/Contao/Model.php
+++ b/src/Resources/contao/library/Contao/Model.php
@@ -1102,9 +1102,7 @@ abstract class Model
 		}
 		else if ($arrOptions['return'] == 'Array')
 		{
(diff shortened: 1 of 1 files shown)


leofeyer:
Except I like the implementation from contao/core#6147 a lot more than yours. Why return an array and not an empty collection?

leofeyer:
> BTW, as far as I used Doctrine in a current project, you simply get an array of models (or an empty array) from the repository. No collection or thelike. Not sure if there’s such a feature though ;-)

That's probably why, isn't it?

leofeyer:
On the other hand, you also said:

> I would not change it either. Version 4 will include Symfony support, so people looking for a new API can use Doctrine (if we do have a bridge until then ;-) ).

aschempp:
An empty collection is useless. The collection class in general is useless, it's basically a wrapper for developers still using the old `Contao\Database` api. There's no need for a collection, Doctrine simply returns an array of entities (or an empty array).

As we know, the Doctrine integration is not that simple. For existing Contao content we still need the existing models.


With the new code, instead of this
```php
$result = PageModel::findBy('pid', 15);

if (null !== $result) {
    foreach ($result as $model) {
        // do something
    } 
}
```

I can simply do that
```php
foreach (PageModel::findBy('pid', 15, ['result' => 'Array']) as $model) {
    // do something
}
```

src/Resources/contao/library/Contao/Model.php
```diff
@@ -1083,7 +1083,7 @@ protected static function find(array $arrOptions)
 
 		if ($objResult->numRows < 1)
 		{
-			return null;
+			return $arrOptions['return'] == 'Array' ? [] : null;
```
leofeyer: I am missing the `isset()` check here.

aschempp: We're not checking for that on line 1092 either, we never do in the "old" code…

src/Resources/contao/library/Contao/Model.php
```diff
@@ -1100,6 +1100,12 @@ protected static function find(array $arrOptions)
 
 			return static::createModelFromDbResult($objResult);
 		}
+		else if ($arrOptions['return'] == 'Array')
+		{
+			$objCollection = static::createCollectionFromDbResult($objResult, static::$strTable);
```
leofeyer: `return static::createCollectionFromDbResult($objResult, static::$strTable)->getModels();`?

src/Resources/contao/library/Contao/Model.php
leofeyer: True. (Also not good either.)

src/Resources/contao/library/Contao/Model.php
aschempp: 👍

src/Resources/contao/library/Contao/Model.php
leofeyer: Since this is "old" code, you need to replace `[]` with `array()`. Then its `RTM` IMO.

aschempp: done

leofeyer:
Merged in 37c5c04ba6e06fe30e1db9a8facffa6b65a85cf0. Too bad we cannot make this the default behavior. 😄

PR #565, opened 2016-08-11
[RTM] PHP Toolbox support

aschempp:
The [PHP Toolbox](https://plugins.jetbrains.com/plugin/8133) plugin adds customizable autocomplete additions to PhpStorm. Here's a simple draft for Contao `core-bundle`.
- Autocomplete and type detection of `ContaoFrameworkInterface::getAdapter()`
  
- Autocomplete for DCA and languages (currently hardcoded)
  
- Autocomplete for model options
  

These are just examples and if we come up with more ideas it's highly extendable 😎

### DIFF
--- a/.ide-toolbox.metadata.json
+++ b/.ide-toolbox.metadata.json
@@ -160,7 +160,32 @@
         },
         {
             "name": "contao_languagefiles",
-            "lookup_strings": ["default", "explain"]
+            "lookup_strings": [
+                "default",
+                "explain",
+                "tl_article",
+                "tl_content",
+                "tl_cron",
+                "tl_files",
+                "tl_form",
+                "tl_form_field",
+                "tl_image_size",
+                "tl_image_size_item",
+                "tl_layout",
+                "tl_log",
+                "tl_member",
+                "tl_member_group",
+                "tl_module",
+                "tl_page",
+                "tl_settings",
+                "tl_style",
+                "tl_style_sheet",
+                "tl_templates",
+                "tl_theme",
+                "tl_undo",
+                "tl_user",
+                "tl_user_group"
+            ]
         },
         {
             "name": "contao_model_options",
@@ -174,7 +199,7 @@
                     "type_text": "string|array"
                 },
                 {
-                    "lookup_string": "sorting",
+                    "lookup_string": "order",
                     "type_text": "SQL"
                 },
                 {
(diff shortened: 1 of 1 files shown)


leofeyer:
Nice. :v:

aschempp:
Is this something we want to add to the core, event though it's IDE specific?

If so, let's think of other places we could use autocomplete. And no, DCA probably wont work 😂

aschempp:
I just thought about this again, and I can't come up with additional ideas. I think this can be merged as-is, and we'll simply extend when we find something.

.ide-toolbox.metadata.json
```diff
@@ -0,0 +1,199 @@
...
+                "tl_templates",
+                "tl_theme",
+                "tl_undo",
+                "tl_user",
+                "tl_user_group"
+            ]
+        },
+        {
+            "name": "contao_languagefiles",
+            "lookup_strings": ["default", "explain"]
```
leofeyer: Should we add the other core-bundle language files here?

.ide-toolbox.metadata.json
```diff
@@ -0,0 +1,199 @@
...
+                    "type_text": "integer"
+                },
+                {
+                    "lookup_string": "offset",
+                    "type_text": "integer"
+                },
+                {
+                    "lookup_string": "return",
+                    "type_text": "Model|Collection|Array"
+                }
```
leofeyer: Here are the options I have found:

 * [ ] `table`
 * [x] `column`
 * [x] `value`
 * [x] `limit`
 * [x] `offset`
 * [x] `return` 
 * [ ] `order` (probably what you mean by `sorting`)

.ide-toolbox.metadata.json
```diff
@@ -0,0 +1,199 @@
...
+                    "method": "getAdapter",
+                    "type": "type"
+                }
+            ]
+        },
+        {
+            "provider":"contao_framework_classes",
+            "language":"php",
+            "signature":[
+                "Contao\\CoreBundle\\Framework\\ContaoFrameworkInterface:getAdapter"
```
aschempp: The provider is different. The first one allows subclasses of `Contao\System`, but some are not (like `System` itself and `Config`, maybe there's even more).

.ide-toolbox.metadata.json
aschempp: It would not harm, I was just too lazy in the draft :-)

.ide-toolbox.metadata.json
aschempp: Fixed in c8ecd792f9afa9b4795a171950921bbeb0a51229

PR #567, opened 2016-08-11
Add the String::toAscii() method

leofeyer:
This PR solves #560 by adding a locale aware version of `toAscii()`.

TODO

- [ ] Adjust all `Utf8::toAscii()` calls


leofeyer:
As tested and discussed, the PR does not solve the problem. We would need to use a library such as https://github.com/cocur/slugify instead.

PR #569, opened 2016-08-12
[WIP] Add the contao.slugify service

leofeyer:
TODO
- [ ] Move the "folderUrl" setting into the application configuration
- [ ] Set the slugify regex depending on the "folderUrl" setting
- [ ] Replace the `Utf8::toAscii()` calls where applicable

@contao/developers /cc


leofeyer:
However, why are we slugifying at all? All browsers can handle UTF-8 URLs meanwhile, so we might as well just leave `news/neue-größen-eingetroffen.html` as is.

leofeyer:
The PR required the following code to be added in the `app/config/config.yml`:

```yml
# Slugify
cocur_slugify:
    regexp: "#([^A-Za-z0-9/.]|-)+#"
```

This will add `/` and `.` to the list of allowed characters, because our aliases can be `index.de` or `folder/urls/supported`.

src/Resources/contao/classes/Theme.php
```diff
@@ -712,10 +712,8 @@ public function exportTheme(DataContainer $dc)
 		}
 
 		// Romanize the name
-		$strName = Utf8::toAscii($objTheme->name);
-		$strName = strtolower(str_replace(' ', '_', $strName));
-		$strName = preg_replace('/[^A-Za-z0-9._-]/', '', $strName);
-		$strName = basename($strName);
+		$strName = \System::getContainer()->get('contao.slugify')->slugify($objTheme->name);
+		$strName = str_replace(array('-/', '/-', '.-', '.'), array('', '', '-', ''), $strName);
```
leofeyer: This is nasty. We need the service to be aware of `Config::get('folderUrl')`, which means we first have to move the setting into the application configuration.

leofeyer:
I have updated the PR so it initializes the Slugify regex depending on the `contao.folder_urls` setting. However, it still feels wrong.

The initial problem is the Slugify service not providing a `setOption()` method IMO.

aschempp:
The ruleset stuff feels wrong. We're always activating rulesets, but they are never removed?

leofeyer:
The ruleset stuff *is* wrong. See https://github.com/cocur/slugify/pull/133 😄

aschempp:
I think this PR is outdated or am I wrong?

leofeyer:
You are wrong. Our changes have been merged but there has not been a release since then. As soon as a new version is released, I will complete the PR.

Toflar:
To be honest, I think it depends. It's something that should be a setting of a root page. If you think in German, somebody might want to have URI's like `foobar.de/über-uns.html` and an other one would like to have `foobar.de/ueber-uns.html`. It's a matter of taste. Also I think `prepend_locale` as well as `folderUrls` are both settings that should be a root page setting, not a system setting. There are other tickets where people would like to have the main language to be accessible via `www.domain.de` and all translations via `www.domain.de/{locale}`. If it was configurable in the root page settings, I guess this would be possible too then. So I think we should mark that as a feature for 4.5.

leofeyer:
If `folderUrls` was a root page setting, how would the resolver know whether to look for folder URLs or not?

aschempp:
I don't think that's currently possible, I guess the Symfony CMF-Router can solve this, but as we know that's not a simple task.

leofeyer:
I'm closing this in favor of #1016.

PR #570, opened 2016-08-16
[RTM] Fixed simple tokens "parser"

Toflar:
Follow up on https://github.com/contao/core/pull/8438 which contains unit tests so we can backport this into 3.5 with confidence.

### DIFF
--- a/src/Resources/contao/library/Contao/StringUtil.php
+++ b/src/Resources/contao/library/Contao/StringUtil.php
@@ -511,14 +511,26 @@ class StringUtil
 	 * @return string The converted string
 	 *
 	 * @throws \Exception If $strString cannot be parsed
+	 * @throws \InvalidArgumentException If $strString contains php code
 	 */
 	public static function parseSimpleTokens($strString, $arrData)
 	{
 		$strReturn = '';
 
-		// Remove PHP code from string. Not supported for security reasons.
-		$strString = preg_replace('/<script language="php">.*<\/script>/', '', $strString);
-		$strString = preg_replace('/<(\?|\%)\=?(php)?.*(\%|\?)>/', '', $strString);
+		// Do not allow php code for security reasons.
+		$containsPhp = function($value) {
+			return strpos($value, '<?php') !== false
+				|| strpos($value, '<?') !== false
+				|| strpos($value, '<%') !== false
+				|| strpos($value, '<%=') !== false
+				|| strpos($value, '<script language="php">') !== false
+				|| strpos($value, '<script language=\'php\'>') !== false;
+		};
+
+		if ($containsPhp($strString))
+		{
+			throw new \InvalidArgumentException('PHP Code is not allowed in subject string.');
+		}
 
 		$arrTags = preg_split('/({[^}]+})/', $strString, -1, PREG_SPLIT_DELIM_CAPTURE|PREG_SPLIT_NO_EMPTY);
 
--- a/src/Resources/contao/library/Contao/StringUtil.php
+++ b/src/Resources/contao/library/Contao/StringUtil.php
@@ -516,26 +516,37 @@ class StringUtil
 	{
 		$strReturn = '';
 
-		// Remove any unwanted tags (especially PHP tags)
-		$strString = strip_tags($strString, \Config::get('allowedTags'));
+		// Remove PHP code from string. Not supported for security reasons.
+		$strString = preg_replace('/<script language="php">.*<\/script>/', '', $strString);
+		$strString = preg_replace('/<(\?|\%)\=?(php)?.*(\%|\?)>/', '', $strString);
+
 		$arrTags = preg_split('/({[^}]+})/', $strString, -1, PREG_SPLIT_DELIM_CAPTURE|PREG_SPLIT_NO_EMPTY);
 
 		// Replace the tags
 		foreach ($arrTags as $strTag)
 		{
-			if (strncmp($strTag, '{if', 3) === 0)
+			if (strncmp($strTag, '{if', 3) === 0 || strncmp($strTag, '{elseif', 7) === 0)
 			{
-				$strReturn .= preg_replace('/\{if ([A-Za-z0-9_]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i', '<?php if ($arrData[\'$1\'] $2 $3): ?>', $strTag);
+				$strReturn .= preg_replace_callback(
+					'/{(if|elseif) ([^=!<>\s]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i',
+					function (array $matches) use ($arrData)
+					{
+						if (!array_key_exists($matches[2], $arrData))
+						{
+							System::log(sprintf('Tried to evaluate (%s statement) unknown simple token "%s".', $matches[1], $matches[2]), __METHOD__, TL_ERROR);
+							return sprintf('<?php %s (false): ?>', $matches[1]);
+						}
+
+						return sprintf('<?php %s ($arrData[\'%s\'] %s %s): ?>', $matches[1], addslashes($matches[2]), $matches[3], $matches[4]);
+					},
+					$strTag
+				);
 			}
-			elseif (strncmp($strTag, '{elseif', 7) === 0)
-			{
-				$strReturn .= preg_replace('/\{elseif ([A-Za-z0-9_]+)([=!<>]+)([^;$\(\)\[\]\}]+).*\}/i', '<?php elseif ($arrData[\'$1\'] $2 $3): ?>', $strTag);
-			}
-			elseif (strncmp($strTag, '{else', 5) === 0)
+			elseif (strncmp($strTag, '{else}', 6) === 0)
 			{
 				$strReturn .= '<?php else: ?>';
 			}
-			elseif (strncmp($strTag, '{endif', 6) === 0)
+			elseif (strncmp($strTag, '{endif}', 7) === 0)
 			{
 				$strReturn .= '<?php endif; ?>';
 			}
@@ -547,13 +558,27 @@ class StringUtil
 
 		// Replace tokens
 		$strReturn = str_replace('?><br />', '?>', $strReturn);
-		$strReturn = preg_replace('/##([A-Za-z0-9_]+)##/i', '<?= $arrData[\'$1\'] ?>', $strReturn);
-		$strReturn = str_replace("] ?>\n", '] . "\n" ?>' . "\n", $strReturn); // see #7178
+		$strReturn = preg_replace_callback(
+			'/##([^=!<>\s]+)##/',
+			function (array $matches) use ($arrData)
+			{
+				if (!array_key_exists($matches[1], $arrData))
+				{
+					System::log(sprintf('Tried to parse unknown simple token "%s".', $matches[1]), __METHOD__, TL_ERROR);
+					return '##' . $matches[1] . '##';
+				}
+
+				return $arrData[$matches[1]];
+			},
+			$strReturn
+		);
+		$strReturn = str_replace("]; ?>\n", '] . "\n"; ?>' . "\n", $strReturn); // see #7178
 
 		// Eval the code
 		ob_start();
 		$blnEval = eval("?>" . $strReturn);
-		$strReturn = ob_get_clean();
+		$strReturn = ob_get_contents();
+		ob_end_clean();
 
 		// Throw an exception if there is an eval() error
 		if ($blnEval === false)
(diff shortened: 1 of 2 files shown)


src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -547,13 +558,27 @@ public static function parseSimpleTokens($strString, $arrData)
...
+				return $arrData[$matches[1]];
+			},
+			$strReturn
+		);
+		$strReturn = str_replace("]; ?>\n", '] . "\n"; ?>' . "\n", $strReturn); // see #7178
 
 		// Eval the code
 		ob_start();
 		$blnEval = eval("?>" . $strReturn);
-		$strReturn = ob_get_clean();
```
leofeyer: Why did you change this? `ob_get_clean()` is the same as `ob_get_contents()` + `ob_end_clean()`.

src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -516,26 +516,37 @@ public static function parseSimpleTokens($strString, $arrData)
 	{
 		$strReturn = '';
 
-		// Remove any unwanted tags (especially PHP tags)
-		$strString = strip_tags($strString, \Config::get('allowedTags'));
+		// Remove PHP code from string. Not supported for security reasons.
+		$strString = preg_replace('/<script language="php">.*<\/script>/', '', $strString);
+		$strString = preg_replace('/<(\?|\%)\=?(php)?.*(\%|\?)>/', '', $strString);
```
leofeyer: Is this really secure?

src/Resources/contao/library/Contao/StringUtil.php
Toflar: Contao 3.5 code

src/Resources/contao/library/Contao/StringUtil.php
Toflar: Nope

src/Resources/contao/library/Contao/StringUtil.php
leofeyer: Yes, but this is Contao 4 and I have already changed every `ob_get_contents()` + `ob_end_clean()` to `ob_get_clean()` and I would appreciate if you would not reset this (unless required for the PR). 😄

src/Resources/contao/library/Contao/StringUtil.php
leofeyer: Then we should go for @ausi's suggestion and throw an exception if we detect a PHP tag, shouldn't we?

src/Resources/contao/library/Contao/StringUtil.php
Toflar: Restored.

src/Resources/contao/library/Contao/StringUtil.php
Toflar: Of course.

Toflar:
Updated the PR. Validates subject string and token data now. /cc @ausi

leofeyer:
Looks good to me. Don't forget to set it to `RTM` once @ausi has reviewed the changes.

src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -511,31 +511,64 @@ public static function toHtml5($strString)
...
-		// Remove any unwanted tags (especially PHP tags)
-		$strString = strip_tags($strString, \Config::get('allowedTags'));
+		// Do not allow php code for security reasons.
+		$containsPhp = function($value) {
+			return strpos($value, '<?php') !== false
+				|| strpos($value, '<?') !== false
+				|| strpos($value, '<%') !== false
+				|| strpos($value, '<%=') !== false
+				|| strpos($value, '<script language="php">') !== false
+				|| strpos($value, '<script language=\'php\'>') !== false;
```
ausi: I think this can be bypassed in several ways `<SCRIPT language="php">`, `<script language=php>`, `<script foo="bar" language="php">`, `<script language="php" >`. Maybe we should still `strip_tags()` after this check?

Toflar: > Maybe we should still strip_tags() after this check?

It would kill `<` comparisons.

src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -511,31 +511,64 @@ public static function toHtml5($strString)
...
 		$strReturn = '';
 
-		// Remove any unwanted tags (especially PHP tags)
-		$strString = strip_tags($strString, \Config::get('allowedTags'));
+		// Do not allow php code for security reasons.
+		$containsPhp = function($value) {
+			return strpos($value, '<?php') !== false
+				|| strpos($value, '<?') !== false
+				|| strpos($value, '<%') !== false
+				|| strpos($value, '<%=') !== false
```
ausi: IMO the checks for `<?php` and `<%=` can be removed because they will already be catched by `<?` and `<%`.

Toflar: Yes, and we could go for a `stripos()` for the script tags which does not yet catch the `<script foo="bar" language="php">` stuff although I don't know if php would even interpret this.

ausi: Looks like `<script foo="bar" language="php">` doesn’t get executed. But `<script  language  =  PHP  >` does for example (in PHP 5)

Toflar: As this has been deprecated long time ago, can we maybe disable this using `ini_set()`? Then we don't have to check for `<%` either.

ausi: The documentation says:

> In PHP 5, there are up to five different pairs of opening and closing tags available in PHP, depending on how PHP is configured. Two of these, `<?php ?>` and `<script language="php"> </script>`, are always available.

Toflar: See regexes here: https://wiki.php.net/rfc/remove_alternative_php_tags

src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -547,8 +580,21 @@ public static function parseSimpleTokens($strString, $arrData)
 
 		// Replace tokens
 		$strReturn = str_replace('?><br />', '?>', $strReturn);
-		$strReturn = preg_replace('/##([A-Za-z0-9_]+)##/i', '<?= $arrData[\'$1\'] ?>', $strReturn);
-		$strReturn = str_replace("] ?>\n", '] . "\n" ?>' . "\n", $strReturn); // see #7178
+		$strReturn = preg_replace_callback(
+			'/##([^=!<>\s]+)##/',
```
ausi: The `+` should be ungreedy in this case so that things like `##token1##,##token1##` get matched as two seperate tokens: `/##([^=!<>\s]+?)##/`

src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -511,31 +511,64 @@ public static function toHtml5($strString)
...
+		{
+			throw new \InvalidArgumentException('PHP Code is not allowed in subject string.');
+		}
+
+		// Validate replacement data
+		foreach ($arrData as $k => $v)
+		{
+			if ($containsPhp($v))
+			{
+				throw new \InvalidArgumentException(sprintf('PHP Code is not allowed in replacement data (Key "%s").', $k));
```
ausi: This check can be bypassed by splitting the php tag into two tokens like `##token1## ##token2##` with the data `['token1' => '<script', 'token2' => 'language="php"> badfunction(); </script>']`

Toflar: Right, so shall we even check for php code at all? :D

ausi: We should not use `eval()`, then we wouldn’t have to check for php code :D

But I think we could replace the tokens before the php check safely:

1. replace tokens
2. check for php-code
3. replace if-tags 
4. eval the code.

ausi: The only problem I see here, is that you could inject if-tags with the tokens then...

Toflar: I know we should use stuff like `token_get_all()` but that's unrealistic for now.

ausi: But how about using `token_get_all()` to check for PHP tags? If it returns only one token of type `T_INLINE_HTML` it should be safe, because it didn’t find an opening tag.

Toflar: Could work :)

src/Resources/contao/library/Contao/StringUtil.php
Toflar: Right! Fixed in 7142939924c468f78f0496dfcd701407f031fe2f.

src/Resources/contao/library/Contao/StringUtil.php
Toflar: Tried to fix it in 240c5e1d8fb52a6858afc953182b9c8899893bad. Obviously the tokenizer does not detect opening tags in PHP 7 because it's not part of the language anymore. Moreover, I guess it will not detect `<?` if short open tags are disabled but again, this is not dangerous because then it will also not interpret it in `eval()`. Unfortunately, the legacy testsuite is not run on travis and I did only test it on php7. Can you give it a shot @leofeyer and @ausi?

leofeyer: The legacy testsuite should run on [Appveyor](https://ci.appveyor.com/project/leofeyer/core-bundle/history), however I cannot find the build.

src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -511,31 +511,64 @@ public static function toHtml5($strString)
...
+				|| strpos($value, '<script language=\'php\'>') !== false;
+		};
+
+		// Validate subject string
+		if ($containsPhp($strString))
+		{
+			throw new \InvalidArgumentException('PHP Code is not allowed in subject string.');
+		}
+
+		// Validate replacement data
```
leofeyer: Is this really necessary? The replacement data comes from the application and not from the user.

Toflar: Not in Isotope, not in notification center, not in...... :)

src/Resources/contao/library/Contao/StringUtil.php
Toflar: Does it not consider PR's? Maybe we have to configure this?

leofeyer: I guess so.

bildschirmfoto 2016-08-23 um 17 44 10

Except it your branch does not have an `appveyor.yml` file, but that's not the case.

tests/Contao/StringUtilTest.php
```diff
@@ -0,0 +1,236 @@
...
+                'This <?= $var ?> is a test.',
+                false
+            ],
+            '(<?)' => [
+                'This <? var_dump() ?> is a test.',
+                version_compare(PHP_VERSION, '7.0.0', '>=')
+            ],
+            '(<%)' => [
+                'This <% var_dump() ?> is a test.',
+                version_compare(PHP_VERSION, '7.0.0', '>=')
```
ausi: Should be changed to something like `!in_array(strtolower(ini_get('asp_tags')), ['1', 'on', 'yes', 'true'])`

Toflar: Added in 9ba15ceb5375845bbac812b980e8c4ecfa2992fb.

Toflar:
Okay, `asp_tags` is now considered. I also tried if `short_open_tag` has any impact but that's not the case. Should be RTM now :)

ausi:
But we can still inject PHP code with the tokens as noted in https://github.com/contao/core-bundle/pull/570#discussion_r75648931

Toflar:
Fixed and covered by a unit test :)

tests/Contao/StringUtilTest.php
```diff
@@ -84,6 +84,21 @@ public function testParseSimpleTokensThrowsExceptionWhenTokensContainingPhp(arra
...
 
     /**
+     * Tests the parseSimpleTokens() method throws exception when tokens contain php
+     * code that is generated only after replacing the tokens.
+     *
+     * @expectedException \InvalidArgumentException
+     */
+    public function testParseSimpleTokensThrowsExceptionWhenTryingToCombineTokensForPhp()
+    {
+        StringUtil::parseSimpleTokens('This is ##open####open2###close# evil', [
```
ausi: Shouldn’t this be `##open####open2####close##`?

Toflar: of course, fixed.

leofeyer:
Merged in b68d28266bda83da2cf0cdf03df9d37c809b9464.

PR #572, opened 2016-08-26
[RFC] Log unknown insert tags

leofeyer:
This PR implements contao/core#8451. @contao/developers Please review.


leofeyer:
We should probably think about the log messages before merging this.

Toflar:
a) Does our Contao logger context automatically log the request?
b) What's wrong about the log messages?

leofeyer:
> What's wrong about the log messages?

Nothing, I just figured that `Unknown insert tag: %s` and `Unknown insert tag flag: %s` might be more compact and maybe better readable?

PR #575, opened 2016-08-30
[WIP] Feature/no more superglobals

discordier:
This PR tries to remove the dependency on the super global arrays `$_SERVER`, `$_GET`, `$_POST` and  `$_COOKIE`.

It does so by registering a new listener which is executed upon every request prior and after the request has been handled.

Things to do:
- [X] Make class `\Contao\Environment` independent of super globals.
- [X] Make class `\Contao\Input` independent of super globals (It still holds a dependency on `$_SESSION` which can not be solved in the current state - we need to discuss).
  - [X] Link the filtered request values to the super globals.
- [X] Make class `ValueAdapter` final so no one will tamper with it 71c861f.
- [X] Utilize `return self::$objRequest->isXmlHttpRequest();` in `Environment` d8bee92.
- [ ] Remove access to super globals from the core where possible (rewrite to use the request directly).
- [ ] Remove access to `Environment` and `Input` from the core where possible (rewrite to use the request directly).
- [x] Use scopes in `LegacyRequestValueSynchronizingListener` 2c27fb1.

Some words about the implementation:
I tried to keep the changes to an absolute minimum while maintaining the maximum of backwards compatibility, therefore the whole input filtering is still in the input class and hence still not unit tested.
I'd prefer to move the filter stuff into `Utils\InputFilter` or something like that as they are completely unrelated to the input itself (aside from caching), even as we intend to remove all of this for Contao 5 it would be better to move the XSS checks and the like into an own class as the `Input` class would become a pretty small delegator then.

So far I have not run into any bc breaks, if you find some, please comment.

Things to decide:

* Move enconding methods from `\Input` to `StringUtil`
  - [ ] `\Input::stripTags` (Strip HTML and PHP tags preserving HTML comments).
  - [ ] `\Input::decodeEntities` (Decode HTML entities - will be needed for converting the current DB values to unencoded values in Contao 5 when we switch to output encoding).
  - [ ] `\Input::preserveBasicEntities` (Preserve basic entities by replacing them with square brackets (e.g. `&amp;` becomes `[amp]` - comes in handy when working with the RTE and we have the restoreBasicEntities in `StringUtil` already so why not also this method?)).
  - [ ] `\Input::encodeSpecialChars` (Encode special characters which are potentially dangerous, these are: `[#<>()\=]`).
  - [ ] `\Input::encodeInsertTags` (Encode the opening and closing delimiters of insert tags - don't know if we still need this when we have fragments).

### DIFF
--- a/src/EventListener/AddToSearchIndexListener.php
+++ b/src/EventListener/AddToSearchIndexListener.php
@@ -12,7 +12,7 @@ namespace Contao\CoreBundle\EventListener;
 
 use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
 use Contao\Frontend;
-use Symfony\Component\HttpKernel\Event\PostResponseEvent;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
 
 /**
  * Adds a page to the search index after the response has been sent.
@@ -48,9 +48,9 @@ class AddToSearchIndexListener
     /**
      * Forwards the request to the Frontend class if there is a page object.
      *
-     * @param PostResponseEvent $event
+     * @param FilterResponseEvent $event
      */
-    public function onKernelTerminate(PostResponseEvent $event)
+    public function onKernelResponse(FilterResponseEvent $event)
     {
         if (!$this->framework->isInitialized()) {
             return;
--- a/src/EventListener/LegacyRequestValueSynchronizingListener.php
+++ b/src/EventListener/LegacyRequestValueSynchronizingListener.php
@@ -10,13 +10,13 @@
 
 namespace Contao\CoreBundle\EventListener;
 
+use Contao\CoreBundle\ContaoCoreBundle;
+use Contao\CoreBundle\Framework\ScopeAwareTrait;
 use Contao\CoreBundle\Request\ValueAdapter;
 use Contao\Environment;
 use Contao\Input;
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\HttpKernel\KernelEvents;
 
 /**
  * Sets the current request on the \Input and \Environment classes.
@@ -28,6 +28,8 @@ use Symfony\Component\HttpKernel\KernelEvents;
  */
 class LegacyRequestValueSynchronizingListener
 {
+    use ScopeAwareTrait;
+
     /**
      * @var RequestStack
      */
@@ -51,6 +53,11 @@ class LegacyRequestValueSynchronizingListener
     public function startRequest()
     {
         $request = $this->requestStack->getCurrentRequest();
+        // Only handle for Contao requests.
+        if (!$this->isContaoScope($request)) {
+            $this->handleForRequest(null);
+            return;
+        }
 
         if (!$request->attributes->has('_contao_value_adapter')) {
             $request->attributes->set('_contao_value_adapter', new ValueAdapter($request));
@@ -67,6 +74,11 @@ class LegacyRequestValueSynchronizingListener
     public function finishRequest()
     {
         $request = $this->requestStack->getParentRequest();
+        // Only handle for Contao requests.
+        if (!$this->isContaoScope($request)) {
+            $this->handleForRequest(null);
+            return;
+        }
 
         $this->handleForRequest($request);
     }
@@ -82,5 +94,36 @@ class LegacyRequestValueSynchronizingListener
     {
         Environment::setRequest($request);
         Input::setValueAdapter($request ? $request->attributes->get('_contao_value_adapter') : null);
+    }
+
+    /**
+     * Checks whether the request is a Contao request.
+     *
+     * @param Request $request The request to check.
+     *
+     * @return bool
+     */
+    protected function isContaoScope(Request $request = null)
+    {
+        return $this->isScope(ContaoCoreBundle::SCOPE_BACKEND, $request)
+            || $this->isScope(ContaoCoreBundle::SCOPE_FRONTEND, $request);
+    }
+
+    /**
+     * Checks whether the _scope attributes matches a scope.
+     *
+     * @param string  $scope   The scope to check.
+     *
+     * @param Request $request The request to check.
+     *
+     * @return bool
+     */
+    private function isScope($scope, Request $request = null)
+    {
+        if (null === $request || !$request->attributes->has('_scope')) {
+            return false;
+        }
+
+        return $request->attributes->get('_scope') === $scope;
     }
 }
--- a/src/EventListener/LegacyRequestValueSynchronizingListener.php
+++ b/src/EventListener/LegacyRequestValueSynchronizingListener.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2016 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\EventListener;
+
+use Contao\CoreBundle\Request\ValueAdapter;
+use Contao\Environment;
+use Contao\Input;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Sets the current request on the \Input and \Environment classes.
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ *
+ * @deprecated Deprecated since Contao 4.3, to be removed in Contao 5.0.
+ *             Use the request or request stack instead.
+ */
+class LegacyRequestValueSynchronizingListener implements EventSubscriberInterface
+{
+    /**
+     * @var RequestStack
+     */
+    private $requestStack;
+
+    /**
+     * Constructor.
+     *
+     * @param RequestStack $requestStack
+     */
+    public function __construct(RequestStack $requestStack)
+    {
+        $this->requestStack = $requestStack;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::REQUEST => array('startRequest', 10),
+            KernelEvents::FINISH_REQUEST => array('finishRequest', -10)
+        ];
+    }
+
+    /**
+     * Store the request in the Contao 3 classes.
+     *
+     * @return void
+     */
+    public function startRequest()
+    {
+        $request = $this->requestStack->getCurrentRequest();
+
+        if (!$request->attributes->has('_contao_value_adapter')) {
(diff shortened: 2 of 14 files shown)


src/EventListener/AddToSearchIndexListener.php
```diff
@@ -40,9 +40,9 @@ public function __construct(ContaoFrameworkInterface $framework)
     /**
      * Forwards the request to the Frontend class if there is a page object.
      *
-     * @param PostResponseEvent $event
+     * @param FilterResponseEvent $event
      */
-    public function onKernelTerminate(PostResponseEvent $event)
+    public function onKernelResponse(FilterResponseEvent $event)
```
Toflar: Can you maybe shed some light on why this is needed again? Because of the hooks that maybe use the Input class? B/c our own code could be changed to use the Request. And can you maybe also elaborate on why your request handling cannot be done on terminate? Is terminate not called for subrequests?

src/EventListener/LegacyRequestValueSynchronizingListener.php
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Sets the current request on the \Input and \Environment classes.
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ *
+ * @deprecated Deprecated since Contao 4.3, to be removed in Contao 5.0.
+ *             Use the request or request stack instead.
+ */
+class LegacyRequestValueSynchronizingListener implements EventSubscriberInterface
```
Toflar: Any reason this is an event subscriber whereas we use regular listeners everywhere else? Imho we should change this just to keep the consistency.

src/EventListener/LegacyRequestValueSynchronizingListener.php
```diff
@@ -0,0 +1,97 @@
...
+            KernelEvents::FINISH_REQUEST => array('finishRequest', -10)
+        ];
+    }
+
+    /**
+     * Store the request in the Contao 3 classes.
+     *
+     * @return void
+     */
+    public function startRequest()
```
Toflar: Imho the `ScopeAwareTrait` should be used here, otherwise it will be triggered for any non Contao route as well.

src/Request/ValueAdapter.php
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * This class is the encapsulation of the request values to keep bc with Contao 3.
+ *
+ * @author Christian Schiffler <https://github.com/discordier>
+ *
+ * @deprecated Deprecated since Contao 4.3, to be removed in Contao 5.0. Use values from the request instead.
+ *
+ * @internal
+ */
+class ValueAdapter
```
Toflar: Please make this class final just to make sure nobody's extending it :D

src/Resources/config/listener.yml
```diff
@@ -112,6 +112,13 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 20 }
 
+    contao.listener.request_synchronizer:
```
Toflar: Same here, probably just turn this into an event listener to keep consistency.

src/Resources/contao/library/Contao/Environment.php
```diff
@@ -332,7 +292,7 @@ protected static function httpXForwardedHost()
 	 */
 	protected static function ssl()
```
Toflar: I know you did follow exactly the 1:1 mapping here but I think we could safely rely on SF's `isSecure()` here.

src/EventListener/LegacyRequestValueSynchronizingListener.php
discordier: Because it needs to handle two events and has to do almost the same.
Changing it to EventListeners would require 3 classes (two handling the events and one abstract base for implementing the common setter code) in order to not trigger duplicate code checks on Scrutinizer etc.

This class is an example for what EventSubscribers have been invented for, see in Symfony code for further examples (they also reside in the EventListener namespace).

In the end, an subscriber is a listener that get's registered for multiple events.

Therefore from code size, memory footprint (only one instance) and speed (need to only load one class) I strongly favor an`EventSubscriber`.

src/Resources/contao/library/Contao/Environment.php
```diff
@@ -502,7 +448,7 @@ protected static function host()
 	 */
 	protected static function isAjaxRequest()
 	{
-		return @$_SERVER['HTTP_X_REQUESTED_WITH'] == 'XMLHttpRequest';
```
Toflar: Can use the `isXmlHttpRequest()` of `$objRequest` here directly.

src/EventListener/LegacyRequestValueSynchronizingListener.php
Toflar: > Because it needs to handle two events and has to do almost the same.
> Changing it to EventListeners would require 3 classes (two handling the events and one abstract base for implementing the common setter code) in order to not trigger duplicate code checks on Scrutinizer etc.

Whoot, no? Just register the same class for two tags:

```
tags:
    - { name: kernel.event_listener, event: kernel.request, method: startRequest }
    - { name: kernel.event_listener, event: kernel.response, method: finishRequest }
```

src/EventListener/AddToSearchIndexListener.php
discordier: The workflow was like the following:
1. `kernel.request` (master request) => `LegacyRequestValueSynchronizingListener` sets request.
2. `kernel.request` (sub request) => `LegacyRequestValueSynchronizingListener` sets request.
3. `kernel.response` => `LegacyRequestValueSynchronizingListener` sets parent request (master request).
4. `kernel.response` => `LegacyRequestValueSynchronizingListener` sets parent request (no request anymore as no parent).
5. `kernel.terminate` => `AddToSearchIndexListener` tries to access `Environment` which has no request anymore. :bomb: 

Now it is:
1. `kernel.request` (master request) => `LegacyRequestValueSynchronizingListener` sets request.
2. `kernel.request` (sub request) => `LegacyRequestValueSynchronizingListener` sets request.
3. `kernel.response` => `LegacyRequestValueSynchronizingListener` sets parent request (master request).
4. `kernel.response` => `AddToSearchIndexListener` accesses `Environment` which has the master request.
5. `kernel.response` => `LegacyRequestValueSynchronizingListener` sets parent request (no request anymore as no parent).

This is of course an intermediate change because as soon as I finish my work on the crawler, the whole `AddToSearchIndexListener` will get removed.

src/EventListener/LegacyRequestValueSynchronizingListener.php
aschempp: We agreed on using tags instead of listener methods because they are more obvious and overridable in container configuration. Please change this.

src/EventListener/AddToSearchIndexListener.php
Toflar: Perfect, thank you!

Toflar:
Looks very promising! Thank you!
I have a small comment on this:

> I'd prefer to move the filter stuff into Utils\InputFilter or something like that as they are completely unrelated to the input itself (aside from caching), even as we intend to remove all of this for Contao 5 it would be better to move the XSS checks and the like into an own class as the Input class would become a pretty small delegator then.

I think we should only move the stuff that might be still useful in 5. If there is only input encoding related stuff, we don't need to move it because we'll drop it anyway. If there are methods that might still be useful (any filter stuff that does not encode anything) we might want to extract that (cannot think of anything atm though).

src/EventListener/LegacyRequestValueSynchronizingListener.php
discordier: I refrained from using scopes at all for the time being because they were removed in Symfony 3.0 and we are yet to come up with a proper replacement.
I'll add a note above.

Toflar: ? I'm talking about the `_scope` request attribute which is not deprecated at all but the current method of checking whether it is a Contao route or not.

src/Request/ValueAdapter.php
discordier: Good point. Noted. :smiley_cat:

src/Resources/config/listener.yml
discordier: See remark above.

src/Resources/contao/library/Contao/Environment.php
discordier: If we agree to move away from the current `Environment` implementation, I will be eager to change this.
For the moment it is, as you stated, a 1:1 translation.

src/Resources/contao/library/Contao/Environment.php
discordier: Whoops, seems like I missed that one.

discordier:
> If there are methods that might still be useful (any filter stuff that does not encode anything) we might want to extract that (cannot think of anything atm though).

Candidates so far are:
* `\Input::stripTags` (Strip HTML and PHP tags preserving HTML comments).
* `\Input::decodeEntities` (Decode HTML entities - will be needed for converting the current DB values to unencoded values in Contao 5 when we switch to output encoding).
* `\Input::preserveBasicEntities` (Preserve basic entities by replacing them with square brackets (e.g. `&amp;` becomes `[amp]` - comes in handy when working with the RTE and we have the restoreBasicEntities in `StringUtil` already so why not also this method?)).
* `\Input::encodeSpecialChars` (Encode special characters which are potentially dangerous, these are: `[#<>()\=]`).
* `\Input::encodeInsertTags` (Encode the opening and closing delimiters of insert tags - don't know if we still need this when we have fragments).

src/EventListener/LegacyRequestValueSynchronizingListener.php
discordier: Changed in 80c0dfe, however, I do not remember that we made such a decision.
Have we documented somewhere that we do not want to use subscribers?

src/EventListener/LegacyRequestValueSynchronizingListener.php
discordier: Implemented in 2c27fb1

src/Resources/contao/library/Contao/Environment.php
discordier: Implemented in d8bee92

discordier:
Rebased and force pushed.

We should decide if we want to move the methods of the `Input` class to `StringUtils`. See comments above.

Toflar:
Cool, thanks!

> We should decide if we want to move the methods of the Input class to StringUtils. See comments above.

Yeah, I think we should move them. But imho we should have a helper method that provides full BC with the encoding stuff of the `Input` class. I mean I would like to switch from

```php
$value = Input::get('var');
```

to

```php
$value = StringUtil::whateverEncode($this->request->query->get('var'));
```

This will a) allow me to use DI and unit testing and b) ease transition to Contao 5 one day as I can simply remove the `StringUtil` call.

discordier:
That is exactly what I was proposing

Toflar:
Then that's a yes for me. #xfactor

Toflar:
I still think this PR is very valuable but as discussed on the phone the only relevant thing is mapping the `Environment` class. @discordier is going to update the PR :)

PR #576, opened 2016-08-30
[RTM] Feature/remove html cache

Toflar:
This is a spin-off PR of #562. It removes our own html cache and adds support for caching via proxies.

To make this work, you need to adjust your `config.yml`:

```diff
# Framework configuration
framework:
    ...
    translator: { fallbacks: ["%locale%"] }
+   esi: { enabled: true }
+   fragments: { enabled: true, path: /_fragment }
```


src/EventListener/AddToSearchIndexListener.php
```diff
@@ -48,6 +48,12 @@ public function onKernelTerminate(PostResponseEvent $event)
             return;
         }
 
+        // Hard coded support for _fragment URIs that should not get indexed.
+        // TODO: remove this listener completely and work towards a real crawler.
+        if (preg_match('/_fragment/', $event->getRequest()->getPathInfo())) {
```
aschempp: Maybe the hardcoded `_fragment` could be injected from the container?

Toflar: I could, yes. Just didn't do because it's going to be dropped anyway.

Toflar: Added in 7f13a07b9a45ac993865f6f933ec945bc08f9d41.

discordier:
@Toflar Would you fix the unit tests or do you want a PR against your PR? :)

Toflar:
I'll fix them once we decided how to proceed on the caching issues. /cc @leofeyer

discordier:
I was just asking because it was labeled as RTM. :)

leofeyer:
The problem is that pages with a mobile layout cannot be cached anymore if we merge this.

It will not affect me, because I never use mobile layouts, but I cannot estimate how many people are using the feature. 

Also, I understand that having two different layouts for the same page can never be handled by any kind of reverse proxy, so this seems like a feature we should deprecate and remove in Contao anyway.

Toflar:
> Also, I understand that having two different layouts for the same page can never be handled by any kind of reverse proxy, so this seems like a feature we should deprecate and remove in Contao anyway.

It can be handled if we provide the means to find out what header to vary on based on the cookie. I'll work on this because we need it for authenticated members anyway. Nevertheless, this feature should be removed anyway in Contao 5, I agree :)

Toflar:
Okay. This PR should be RTM then. The only difference to the previous implementation is that currently the mobile page layout cannot be cached by the proxy. I'll work on this though. On the other hand we can now use real proxies such as Varnish and also benefit by ESI implementations in our own application code which is a huge advantage in my opinion.

aschempp:
If you cannot cache a mobile layout, how do you cache a desktop layout? Or do you mean caching is completly disabled if there is a mobile layout assigned to a page?

Toflar:
> Or do you mean caching is completly disabled if there is a mobile layout assigned to a page?

Exactly.

leofeyer:
Since we are not getting ESI in Contao 4.3, do we still want to merge this PR?

Toflar:
Of course! That's why it's a spin-off PR...

aschempp:
Absolutely, storing the output in the proxy cache would still be way better than our current listener.

aschempp:
I don't think this is RTM though. There are still a few things I'd do differently, from the controller name to the question of data upgrade. Not sure about deadlines though?

discordier:
@aschempp How about writing down what controller name you would like better?

src/Controller/EsiController.php
```diff
@@ -0,0 +1,60 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Handles the Contao ESI requests. This might be subject to change
+ * in the very near future which is why this class is declared final.
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+final class EsiController extends Controller
```
aschempp: This should rather be `InsertTagsController`

src/Controller/EsiController.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->framework = $framework;
+    }
+
+    /**
+     * Renders non cacheable insert tags (such as {{request_token}}).
+     *
+     * @param string $insertTag
+     */
+    public function renderNonCacheableInsertTag($insertTag)
```
aschempp: which would make this `indexAction`?

src/EventListener/AddToSearchIndexListener.php
```diff
@@ -48,6 +55,10 @@ public function onKernelTerminate(PostResponseEvent $event)
             return;
         }
 
+        if (preg_match('/' . preg_quote($this->fragmentPath, '/') .'/', $event->getRequest()->getPathInfo())) {
```
aschempp: Maybe that should be checked in `indexPageIfApplicable`, that's what the function's name says it does?

src/Resources/contao/classes/Frontend.php
```diff
@@ -654,224 +654,16 @@ public static function indexPageIfApplicable(Response $objResponse)
...
-			$response->headers->set('Last-Modified', gmdate('D, d M Y H:i:s', time()) . ' GMT');
-			$response->headers->set('Expires', gmdate('D, d M Y H:i:s', $expire) . ' GMT');
-		}
-		else
-		{
-			$response->headers->set('Cache-Control', 'no-store, no-cache, must-revalidate, post-check=0, pre-check=0');
-			$response->headers->set('Last-Modified', gmdate('D, d M Y H:i:s') . ' GMT');
-			$response->headers->set('Expires', 'Fri, 06 Jun 1975 15:10:00 GMT');
-		}
+		@trigger_error('Using Frontend::getResponseFromCache() has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
```
aschempp: I think the `trigger_error` message should also include the information that the method does not work anymore (like the docblock header).

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -241,98 +240,13 @@ public function hasAuthenticatedBackendUser()
 
 	/**
 	 * Add the template output to the cache and add the cache headers
+	 *
+	 * @deprecated Deprecated since Contao 4.3, to be removed in Contao 5.0.
```
aschempp: Again, it's not only deprecated, it no longer does anything at all.

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -414,4 +328,39 @@ public function getCustomSections($strKey=null)
...
+		global $objPage;
+
+		if (false === $objPage->cache && false === $objPage->clientCache)
+		{
+			return;
+		}
+
+		// If FE_USER_LOGGED_IN or BE_USER_LOGGED_IN every request is private
+		// Moreover, mobile layout and protected pages are not cached either
+		// TODO: Add support for proxies so they can vary on member context and page layout
```
aschempp: If I remember our discussion correctly, that TODO can be removed?

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -414,4 +328,39 @@ public function getCustomSections($strKey=null)
...
+		if (true === FE_USER_LOGGED_IN || true === BE_USER_LOGGED_IN || true === $objPage->isMobile || $objPage->protected)
+		{
+			$response->setPrivate();
```
aschempp: The method should set the response as private initially (e.g. around line 341) and then just `return` in the necessary cases. `setSharedMaxAge` will set it to *public* and remove the *private* header.

src/Controller/EsiController.php
Toflar: No, it still only renders the non cacheable ones.

src/Controller/EsiController.php
Toflar: Jup.

src/EventListener/AddToSearchIndexListener.php
Toflar: No. I think it's better suited here.

src/Resources/contao/classes/Frontend.php
Toflar: I can add that, yes.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Agreed.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: No, that TODO still remains.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Disagree. It's more explicit like that. The other way would require one to know that `setSharedMaxAge` does reset the private header.

src/Controller/EsiController.php
aschempp: Well, not true, it renders **all** insert tags **including** cacheable ones. Also, shouldn't an action have `Action` suffix?

src/Resources/contao/models/PageModel.php
```diff
@@ -927,12 +930,6 @@ public function loadDetails()
 
 		$this->trail = array_reverse($trail);
 
-		// Do not cache protected pages
-		if ($this->protected)
```
leofeyer: Why did you remove this?

src/Controller/EsiController.php
leofeyer: Declaring a class `final` is not the appropriate thing to do if it is subject to change. We should rather declare it `@internal` instead, shouldn't we?

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -124,15 +127,11 @@ protected function compile($blnCheckRequest=false)
...
-		$this->addToCache();
-
 		// Unset only after the output has been cached (see #7824)
 		unset($_SESSION['LOGIN_ERROR']);
 
-		// Replace insert tags and then re-replace the request_token tag in case a form element has been loaded via insert tag
-		$this->strBuffer = $this->replaceInsertTags($this->strBuffer, false);
-		$this->strBuffer = str_replace(array('{{request_token}}', '[{]', '[}]'), array(REQUEST_TOKEN, '{{', '}}'), $this->strBuffer);
+		// Replace insert tags
+		$this->strBuffer = $this->replaceInsertTags($this->strBuffer);
```
leofeyer: Why did you remove the second argument here?

src/Controller/EsiController.php
leofeyer: They should.

leofeyer:
I will merge this as soon as we have discussed @aschempp's comments.

src/Controller/EsiController.php
Toflar: Well it shouldn't be subject to change anymore anyway as it's only handling insert tags. I'll remove it.

src/Controller/EsiController.php
Toflar: I'll call it `renderAction`.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Because I want it to be `true` so they are replaced by the `<esi>` tags and not ignored.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Does it? Hm, okay, will look into it.

src/Resources/contao/models/PageModel.php
Toflar: Because it's wrong and not needed. Caching should not be disabled just because the page is protected. In this PR, whenever a user is logged in, the response is set to private, so it never caches.

src/Controller/EsiController.php
Toflar: Implemented in 071d434e163b9b8924cf90048b33c2838d02171a.

src/Controller/EsiController.php
Toflar: Implemented in 071d434e163b9b8924cf90048b33c2838d02171a.

src/Resources/contao/classes/Frontend.php
Toflar: Added in d5f3bf53d18b787c676c18fa4daa11811b8ac9e6.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Added in d5f3bf53d18b787c676c18fa4daa11811b8ac9e6.

src/Resources/contao/classes/FrontendTemplate.php
Toflar: Implemented explicit private header in 69cc0223ebd61fcd2a982cea6e82cc93e10af3c6. A response is private by default anyway.

Toflar:
All review comments addressed.

discordier:
All good from my side also. Big thanks to @Toflar for making this happen.

Issue #574, opened 2016-08-29
Future features and their depencencies


Toflar:
Starting a list here so we know what issues to tackle first. To be extended/modified.

| Task | Short description | Prerequisites | Logical priority | Pull Request |
| --- | --- | --- | --- | --- |
| 1 | Edge Side Includes | 2, 3, 8, 9 |  | @toflar |
| 2 | No more server related super globals (`$_GET`, `$_POST`, `$_COOKIE`, `$_SERVER`, `$_SESSION`) in the code base. Everything must be either using `Input`, `Environment`, `Session` or already the current request from the request stack. |  | 1 | https://github.com/contao/core-bundle/pull/575 (@discordier) |
| 3 | No more `global $objPage`. Must be an attribute on the current request. |  | 2 | @leofeyer |
| 4 | New search index (page crawler). Way better if done using SF's request stack because everything can be done in the same process (no real web requests = no routing via webserver = faster). | 2 |  | @discordier |
| 5 | New permission system using voters. | 7 |  |  |
| 6 | Better HTTPS support (automatically use cookie_secure etc.) | 3 |  |  |
| 7a | Full support for SF authentication in BE |  | 3 | @bytehead |
| 7b | Full support for SF authentication in FE | 10 | 3 | @bytehead |
| 8 | Make sure page types, fe modules and content elements work like regular controller (request -> response) |  |  |  |
| 9 | Provide a route and integrate with HttpCache to allow splitting up `Cookie` into other headers to allow proxies to `Vary` on them. | 2, 3, 7 |  | #577 (@toflar) |
| 10 | Use the CMF router and provide the page object as request document | | | |

discordier:
Currently working on the super globals usage (number 2 in your list). PR upcoming.

discordier:
My work on the superglobals is a prerequisite of my actual work on number 4.

Toflar:
> My work on the superglobals is a prerequisite of my actual work on number 4.

Already on the list :)

discordier:
You can add #575 for task 2

leofeyer:
@Toflar I guess we have to overwork the list, don't we?

Toflar:
Closing this because in the meantime

1) we have achieved full ESI support 🥳 
2) Server super global access has largely disappeared in v5 🥳 
3) is still there mainly for compatibility reasons but it's also part of the request object 🥳 
4) has landed 🥳 
5) has landed in v5 🥳 
6) has landed 🥳 
[...]

PR #577, opened 2016-09-02
[WIP] Feature/preflight http cache

Toflar:
This PR implements the preflight request functionality to allow proxies to vary on different headers. It already contains the support for Symfony's HttpCache. You only need to edit your `AppCache.php` like this:

``` diff
- class AppCache extends Symfony\Bundle\FrameworkBundle\HttpCache\HttpCache
+ class AppCache extends Contao\CoreBundle\HttpKernel\HttpCache\PreflightRequestHttpCache
{
}
```

This PR is WIP and here for reference only. As written in (https://github.com/contao/core-bundle/issues/574) it will require numbers 2, 3 and 7 before it can be finished (the listeners have to be written) so there's no point in commenting on it yet (as I'm off for September now anyway).

### DIFF
--- a/src/Controller/PreflightRequestController.php
+++ b/src/Controller/PreflightRequestController.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2016 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\Controller;
+
+use Contao\CoreBundle\Event\ContaoCoreEvents;
+use Contao\CoreBundle\Event\PreflightRequestEvent;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Handles the Contao preflight route.
+ *
+ * @author Yanick Witschi <https://github.com/toflar>
+ */
+class PreflightRequestController extends Controller
+{
+    /**
+     * EventDispatcher
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
+    /**
+     * PreflightController constructor.
+     *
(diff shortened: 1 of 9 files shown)


leofeyer:
I only know preflighted requests in the context of cross-origin resource loading, but that does not seem to be the subject in this PR. What exactly does your code do?

Toflar:
It does essentially the same as a CORS request which is why I named it preflight. If the original request matches certain conditions (e.g. `Cookie ` present in the headers), it forwards the request to the `contao/preflight` route where any listener can add headers to the response. The response of the preflight request is then analyzed and all `Contao-Vary-*` headers of the response are transferred to the original request which is then executed again. This concept is not limited to my proxy implementation (extended SF http cache) but can be adapted to Varnish and Co. as well. 
Basically, the preflight request allows the application to split up the `Cookie` or `Authorization` (or others, as it's configurable) in additional headers so the app itself can then `Vary` on those again.

E.g. for the page layout setting, you would like to send a response header like `Vary: Contao-Vary-Page-Layout` so the proxy adds individual cache entries based on the content of that header. However, that content is based on the session and thus part of the `Cookie` header. The preflight concept will allow for the following to happen:

1. Original request comes in. Contains a `Cookie: PHPSESSID=04hqrq97c36oqs43jgtpgbeke4;` header.
2. Preflight request to `contao/preflight` is executed
3. A page layout listener analyzes the PHP session and based on its information adds the header `Contao-Vary-Page-Layout: mobile` to the preflight response
4. Preflight response is analyzed, and the original request is decorated so that the request now contains both, `Cookie: PHPSESSID=04hqrq97c36oqs43jgtpgbeke4;` as well as `Contao-Vary-Page-Layout: mobile`.
5. Original request is processed again. The proxy does not have a cache entry yet, so it forwards it to the app. The app responds with the content and adds a `Vary: Contao-Vary-Page-Layout` header.

The next time the same request happens, at step 5 the proxy has a cache entry and varies based on the `Contao-Vary-Page-Layout` header.

Toflar:
Meanwhile this is part of Contao 4.4 together with our `terminal42/replay-headers-bundle` 😎

PR #580, opened 2016-09-26
[RTM] Subsequent elseif should not match

ausi:
Follow up to #579, see contao/core#8491

### DIFF
--- a/src/Resources/contao/library/Contao/StringUtil.php
+++ b/src/Resources/contao/library/Contao/StringUtil.php
@@ -632,8 +632,11 @@ class StringUtil
 		};
 
 		// Parsing stack used to keep track of the nesting level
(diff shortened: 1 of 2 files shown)


leofeyer:
Is this PR ready to merge? I figure we also need to back port the changes into the LTS version, don't we?

Toflar:
> I figure we also need to back port the changes into the LTS version, don't we?

Sure, it is a pretty important fix.

ausi:
> Is this PR ready to merge?

Yes, I think so.

leofeyer:
Merged in 7589374a4332bd9aa1cdf47a7ecdcf19a5afb6a4 and back ported in contao/core@c5b9e81173a6aee14bac3d096c5e97d70e6a8dfe.

PR #582, opened 2016-09-28
Use class set in config.php for form module to allow customization

mruettgers:
Overriding the class used for the form frontend module does not work when using getForm() of  Controller.php.

Let's use Module::findClass() to resolve the class.

### DIFF
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -531,7 +531,13 @@ abstract class Controller extends \System
 
 		$objRow->typePrefix = 'ce_';
 		$objRow->form = $objRow->id;
-		$objElement = new \Form($objRow, $strColumn);
+
+		$class = Module::findClass('form');
+		if (!class_exists($class)) {
+			throw new \RuntimeException('Class for module of type \'form\' could not be found.');
+		}
+		$objElement = new $class($objRow, $strColumn);
+
 		$strBuffer = $objElement->generate();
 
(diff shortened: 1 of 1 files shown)


leofeyer:
I like the change, however, `Form` is a hybrid that can be either a content element or a module, therefore `Module::findClass()` is incorrect. We would have to add a `Hybrid::findClass()` method, which checks both the `FE_MOD` and the `TL_CTE` array.

@contao/developers The `form` keyword exists in both arrays and might point to different classes. How do we handle this case in a `Hybrid::findClass()` method? Does `FE_MOD` have precedence over `TL_CTE`?

Toflar:
In my opinion `Hybrid::findClass()` is wrong. A hybrid can be both, a module or a content element. But the caller has to know, whether they want it to render as a module or as a content element. So imho it's the developers choice to either go for `Module::findClass()` or `ContentElement::findClass()`.

leofeyer:
Da stimme ich Dir grundsätzlich zu, aber in der `Controller::getForm()` ist das nicht so einfach. Wir fügen auch schon immer das Prefix `ce_` hinzu, obwohl es auch `mod_` sein könnte:

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/Controller.php#L532

Toflar:
I did a PR to illustrate my take on this: https://github.com/contao/core/pull/8527

leofeyer:
Closed in favor of contao/core#8527.

PR #596, opened 2016-10-20
[RTM] Inherit composer verbosity in script handler

discordier:
When running composer with `-vvv` (or `-vv` or `-v`), the scripts of Contao still run without knowing these verbosity setting.

This PR changes the behavior so that the process inherits the verbosity of composer and aids in debugging deployment issues then.


leofeyer:
Merged in a04678427a589e3a49611dbef7d81b201757ea92.

PR #597, opened 2016-10-20
Don't use protocol-relative URLs for assets

aschempp:


leofeyer:
This fix is irrelevant.

The main reason why you should not use protocol-relative URLs, is that resources that are available via HTTPS should be included via HTTPS, even if the website itself uses HTTP.

This change, however, only appends `http://` or `https://` depending on the `useSSL` setting or the current environment, which means resources are still loaded via HTTP instead of HTTPS if the page uses HTTP. You might as well use protocol-relative URLs in this case.

aschempp:
Apparently my response was not delivered by email :(
The change also uses SSL if you explicitly enabled SSL in the root page, no matter the current protocol.

leofeyer:
I wonder if we should adjust the protocol at all. Wouldn't it be better to simply use what the user has configured? Then the code would be:

```php
define($strConstant, $url . \Environment::get('path') . '/');
```

aschempp:
I would like that, but the protocol is currently not mandatory on the input.

PR #606, opened 2016-10-27
[WIP] Split page rendering into multiple methods

aschempp:


src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -69,8 +67,15 @@ public function run()
 			throw new PageNotFoundException('Page not found: ' . \Environment::get('uri'));
 		}
 
-		// Get the current page object(s)
-		$objPage = \PageModel::findPublishedByIdOrAlias($pageId);
+		return $this->renderPage(\PageModel::findPublishedByIdOrAlias($pageId));
+	}
+
+
+	public function renderPage($pageModel = null)
```
leofeyer: It seems you need to also pass `$pageId` and `$objRootPage`, which are both used in the `renderPage()` method.

aschempp:
Interesting. If I understand the code correctly, `$objRootPage` is always `null` when passed to the 403 page?

leofeyer:
Merged in 992469b17b4718b27d30d09bd293469a6444fe1a.

PR #608, opened 2016-10-28
Fix autoloading of Doctrine diff command 

aschempp:
Because the command extends a parent class that is not available.


leofeyer:
The change will not fix the PHP IDE problem, because `require-dev` packages are only loaded within the root package, which is `contao/standard-edition`.

aschempp:
I'm sorry, how is this relevant?

leofeyer:
It is relevant, because if the package is not installed, PhpStorm will throw an error. Even though we are not registering the command, we are still referencing classes from the migration bundle in our code (as @discordier has already explained to you).

aschempp:
I'm sorry but you're wrong. This does not apply to the standard-edition, but only to the bundle itself. That's what dev-dependencies are for. Standard edition does not use PhpUnit, so you would get the same errors for your unit tests in core bundle.

leofeyer:
Just see for yourself:

leofeyer:
Merged in 4176f0f00367245dde0b35e1a7d23ff2a8d91849.

Issue #607, opened 2016-10-28
TemplateLoader and custom composer path repositories


dmolineus:
I'm used to have following setup when developing:

``` bash
# Root directory of the project
project/

# Custom extensions being stored outside
project/dev/custom-extension1
project/dev/custom-extension2

# Different test installations
/var/www/web1
var/www/web2
```

In my composer.json i add the custom extensions using path repositories which will be symlinked into the `vendor` directory.

``` json
{
    "repositories": [
        {
            "type": "path",
            "url": "/path/to/project/custom-extension1"
        }
    ]
}
```

This worked pretty well until Contao 4.3. I talked with @Toflar and he told me it's an known limitation. But I hope we can find an solution for it.

At the moment there is the following behaviour:
- Template paths from symlinked files get registered as absolute path, 
  e.g. `/path/to/project/custom-extension1/src/Resources/contao/templates/mod_custom.html5`
- So the template loader tries to load it in `/kernel_root_dir//path/to/project/custom-extension1/src/Resources/contao/templates/mod_custom.html5` which does not work
- The paths can not be fixed by using the legacy template loader as the TemplateLoader is called after the class loader (which loads the autoload.php) files

That means that the TemplateLoader section in the autoload.php has no effect anymore. That's the purpose I guess, but it's also a BC break. 

As we already have a BC break here I suggest to do it right. This issue could easily fixed when the template loader always store the absolute paths. Checking if a path is relative and add the kernel root dir in `TemplateLoader::addFile` should be pretty simple. The BC break would be in `TemplateLoader::getFiles` which would return absolute paths instead of relative (or as implemented right now a mix of absolute paths and relative paths when using the setup described above).

If you agree to store absolute paths in the template loader I don't mind to contribute the PR.

leofeyer:
Storing absolute paths breaks the portability of the installation, that is why we are storing relative paths everywhere. I can think of two solutions:

1. Storing symlinked files as relative paths.

Then `/path/to/project/custom-extension/…` becomes `../../../path/to/project/custom-extension/…`, which will also work if the loader prepends the kernel root directory.

2. Use `Filesystem::isAbsolutePath()`

By using `Filesystem::isAbsolutePath()`, we can check if the path is absolute before prepending the kernel root directory.

@contao/developers Which solution would you prefer?

dmolineus:
Afaik the generated templates.php cache of twig also contains absolute paths. So I'm not sure if  portability of an installation would work when sharing the cache.

[...]

PR #612, opened 2016-11-03
Fix template paths creation in the template loader (See #607).

dmolineus:
This pull request fixes the issue #607 in the template loader which is used when the cache is not warmed up.

### DIFF
--- a/src/Resources/contao/library/Contao/TemplateLoader.php
+++ b/src/Resources/contao/library/Contao/TemplateLoader.php
@@ -135,21 +135,21 @@ class TemplateLoader
 			return TL_ROOT . '/' . self::$files[$template] . '/' . $file;
 		}
 
-		$strRelpath = null;
+		$strAbsPath = null;
 
 		try
 		{
 			// Search for the template if it is not in the lookup array (last match wins)
 			foreach (\System::getContainer()->get('contao.resource_finder')->findIn('templates')->name($file) as $file)
 			{
-				$strRelpath = str_replace(TL_ROOT . DIRECTORY_SEPARATOR, '', $file->getPathname());
+				$strAbsPath = $file->getPathname();
 			}
 		}
 		catch (\InvalidArgumentException $e) {}
 
-		if ($strRelpath !== null)
+		if ($strAbsPath !== null)
 		{
-			return TL_ROOT . '/' . $strRelpath;
+			return $strAbsPath;
 		}
 
 		throw new \Exception('Could not find template "' . $template . '"');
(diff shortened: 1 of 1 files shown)


leofeyer:
I don't get why we need this change. I have symlinked the calendar bundle, so the paths in `var/cache/dev/contao/config/templates.php` look like this:

```php
  'cal_default' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/calendar',
  'cal_mini' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/calendar',
  'event_full' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/events',
  'event_list' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/events',
  'event_teaser' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/events',
  'event_upcoming' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/events',
  'mod_calendar' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/modules',
  'mod_eventlist' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/modules',
  'mod_eventmenu' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/modules',
  'mod_eventreader' => '../../../tmp/calendar-bundle/src/Resources/contao/tmplates/modules',
```

Then I have added `dump(\TemplateLoader::getDefaultPath('event_upcoming', 'html5'));` in the main front end controller (`FrontendIndex`). The output is:

```
/Users/leofeyer/Sites/contao/standard-edition/../../../temp/calendar-bundle/src/Resources/contao/templates/events/event_upcoming.html5
```

Without the cache, the output is:

```
/Users/leofeyer/Sites/contao/standard-edition/vendor/contao/calendar-bundle/src/Resources/contao/templates/events/event_upcoming.html5
```

IMO, the output **without** the cache is correct and we should adjust the cache file instead of what you have changed in this PR.

@contao/developers /cc

leofeyer:
In fact, we should never work with the real path, because it might lead to `open_basedir()` problems.

leofeyer:
Ok, I agree that it makes more sense to store absolute paths in the cache. I have changed the implementation in 1b832755d4733f13db7a103f028850b02ba60103, which partly adopts changes from this PR.

leofeyer:
Apparently, the Symfony devs think that it makes more sense to store relative paths:

https://github.com/symfony/symfony/pull/19687

leofeyer:
Ok, we are now using relative paths again (see d3e11a9593c6dc135aa3b125d29f12fcf446c1ff).

PR #615, opened 2016-11-05
[RFC] Skip the authentication routines if there is a ContaoToken already

leofeyer:
@contao/developers @DanielSchwiperich This is a fix for the double authentication issue as discussed at the Contao Camp this morning.


leofeyer:
As discussed at the Contao Camp, this change implies a BC break, because it would no longer be possible to call `authenticate()` and trigger the authentication routines more than once.

PR #619, opened 2016-11-08
[RTM] Hotfix/scope aware trait

discordier:
This PR makes the `ScopeAwareTrait` use optional parameters for determining the request to match.

After implementing this, I noticed that many unit tests failed.
This was due to the fact, that these tests were pushing one request instance to the container (by using `TestCase::mockContainerWithContaoScopes`) and working on a completely unrelated request instance which got created in the test.
These tests therefore reflected a setup that could never happen in application workflow and therefore I straightened them out by always passing the correct request to the trait.

This then ultimately lead to the removal of all `setContainer()` calls in the listeners using the `ScopeAwareTrait`.

After doing so, I realized that we should prohibit `setContainer()` calls on these classes and eventually came up with `ScopeTrait`, this is used by `ScopeAwareTrait` internally in a 100% backwards compatible way.

Deprecations:
- `ScopeAwareTrait` has been deprecated in favor of `ScopeTrait` and a warning is issued.
- Not passing a `Request` object to any of the `ScopeAwareTrait::is*Scope()` methods has been deprecated separately and triggers another warning.

### DIFF
--- a/src/Framework/ContaoFramework.php
+++ b/src/Framework/ContaoFramework.php
@@ -189,7 +189,7 @@ class ContaoFramework implements ContaoFrameworkInterface
         }
 
         // Define the login status constants in the back end (see #4099, #5279)
-        if (!$this->request || !$this->isFrontendScope($this->request)) {
+        if (!($this->request instanceof Request) || !$this->isFrontendScope($this->request)) {
             define('BE_USER_LOGGED_IN', false);
             define('FE_USER_LOGGED_IN', false);
         }
@@ -205,11 +205,15 @@ class ContaoFramework implements ContaoFrameworkInterface
      */
     private function getMode()
     {
-        if ($this->request && $this->isBackendScope($this->request)) {
+        if (!($this->request instanceof Request)) {
+            return null;
+        }
+
+        if ($this->isBackendScope($this->request)) {
             return 'BE';
         }
 
-        if ($this->request && $this->isFrontendScope($this->request)) {
+        if ($this->isFrontendScope($this->request)) {
             return 'FE';
         }
 
--- a/src/Framework/ContaoFramework.php
+++ b/src/Framework/ContaoFramework.php
@@ -19,6 +19,7 @@ use Contao\Input;
 use Contao\RequestToken;
 use Contao\System;
 use Contao\TemplateLoader;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
@@ -37,7 +38,8 @@ use Symfony\Component\Routing\RouterInterface;
  */
 class ContaoFramework implements ContaoFrameworkInterface
 {
-    use ScopeAwareTrait;
+    use ContainerAwareTrait;
+    use ScopeCheckingTrait;
 
     /**
      * @var bool
--- a/src/Framework/ContaoFramework.php
+++ b/src/Framework/ContaoFramework.php
@@ -187,7 +187,7 @@ class ContaoFramework implements ContaoFrameworkInterface
         }
 
         // Define the login status constants in the back end (see #4099, #5279)
-        if (!$this->isFrontendScope()) {
+        if (!$this->request || !$this->isFrontendScope($this->request)) {
             define('BE_USER_LOGGED_IN', false);
             define('FE_USER_LOGGED_IN', false);
         }
@@ -203,11 +203,11 @@ class ContaoFramework implements ContaoFrameworkInterface
      */
     private function getMode()
     {
-        if ($this->isBackendScope()) {
+        if ($this->request && $this->isBackendScope($this->request)) {
             return 'BE';
         }
 
-        if ($this->isFrontendScope()) {
+        if ($this->request && $this->isFrontendScope($this->request)) {
             return 'FE';
         }
 
--- a/src/Framework/ContaoFramework.php
+++ b/src/Framework/ContaoFramework.php
@@ -39,7 +39,7 @@ use Symfony\Component\Routing\RouterInterface;
 class ContaoFramework implements ContaoFrameworkInterface
 {
     use ContainerAwareTrait;
-    use ScopeCheckingTrait;
+    use ScopeTrait;
 
     /**
      * @var bool
--- a/src/Framework/ScopeAwareTrait.php
+++ b/src/Framework/ScopeAwareTrait.php
@@ -14,7 +14,7 @@ use Contao\CoreBundle\ContaoCoreBundle;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\Request;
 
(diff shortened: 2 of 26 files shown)


leofeyer:
Please revert renaming the trait to `ScopeCheckingTrait`.

discordier:
This will then force us to keep implementing the `ContainerAwareTrait` in all using classes of which most likely none will ever need the container.
See 679ab3934782cecb96814073c86f00d207c397ba and b6762b8, If we do not do this, we will have no way to tell users that they should not use the fallback to the request stack in the container.
We also have no way to remove the container from derived classes.

discordier:
I have renamed the trait as discussed in slack

src/Framework/ContaoFramework.php
```diff
@@ -187,7 +189,7 @@ private function setConstants()
         }
 
         // Define the login status constants in the back end (see #4099, #5279)
-        if (!$this->isFrontendScope()) {
+        if (!$this->request || !$this->isFrontendScope($this->request)) {
```
leofeyer: `!$this->request` should not be necessary, because `isFrontendScope()` will handle `null`, right?

src/Framework/ContaoFramework.php
```diff
@@ -203,11 +205,11 @@ private function setConstants()
      */
     private function getMode()
     {
-        if ($this->isBackendScope()) {
+        if ($this->request && $this->isBackendScope($this->request)) {
```
leofeyer: `$this->request` should not be necessary, because `isFrontendScope()` will handle `null`, right?

src/Framework/ContaoFramework.php
```diff
@@ -203,11 +205,11 @@ private function setConstants()
...
             return 'BE';
         }
 
-        if ($this->isFrontendScope()) {
+        if ($this->request && $this->isFrontendScope($this->request)) {
```
leofeyer: `$this->request` should not be necessary, because `isFrontendScope()` will handle `null`, right?

src/Framework/ContaoFramework.php
discordier: Nope, you will get an error that instance of Request is expected.
You will only get null handled in the depreacted trait.

src/Framework/ContaoFramework.php
discordier: Nope, you will get an error that instance of Request is expected.
You will only get null handled in the depreacted trait.

src/Framework/ContaoFramework.php
leofeyer: Ok, then it should be `!($this->request instanceof Request)`. :smile:

src/Framework/ContaoFramework.php
leofeyer: And this should be `$this->request instanceof Request` accordingly.

src/Framework/ScopeAwareTrait.php
```diff
@@ -12,103 +12,83 @@
 
 use Contao\CoreBundle\ContaoCoreBundle;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
-use Symfony\Component\HttpKernel\Event\KernelEvent;
+use Symfony\Component\HttpFoundation\Request;
+
+@trigger_error('Trait Contao\CoreBundle\Framework\ScopeAwareTrait has been deprecated since Contao 4.3 and will get removed in Contao 5.0 - use Contao\CoreBundle\Framework\ScopeCheckingTrait instead', E_USER_DEPRECATED);
```
aschempp: needs to be adjusted for new trait name.

src/Framework/ContaoFramework.php
discordier: Fixed in ffb6171

src/Framework/ContaoFramework.php
discordier: Fixed in ffb6171

src/Framework/ScopeAwareTrait.php
discordier: Fixed in ef3d52e

discordier:
All issues addressed, I consider this RTM now.

leofeyer:
@aschempp and @bytehead have noted that the scope should be determined by the firewall and not by a request attribute, therefore we need to discuss the implementation once more I'm afraid.

aschempp:
This should not be merged, as it would introduce a new trait that will be deprecated in the next minor release if we introduce Symfony security.

discordier:
Then I need to copy & paste the trait code into my other PRs for checking the scope there and we can remove it then.
No problem for me, I simply wanted to make existing code usable as the current state is not usable outside the trait at all.

The initial need for this PR was to perform the check on a request instance instead of performing it on the request stack (which might differ and already did differ in the unit tests and therefore rendered some of them testing scenarios which can not happen in real world applications).
At least these tests should get fixed to not push one request to the stack and another one into the kernel event as this is simply plain wrong.

So the bottom line is:
> This should not be merged, as it would introduce a new trait that will be deprecated in the next minor release **if** we introduce Symfony security.

This emplaces now some questions:
* How certain is it that this makes it into 4.4?
* What is the replacement for detecting frontend or backend scope from within the application? (Thinking about the wildcard rendering)
* Do we drop the separation altogether?

Again, I have no hard feelings if this does not get merged, however, then I will be forced to copy&paste this new trait's code into multiple locations where I have to check the request and it feels wrong to do so (I bet @leofeyer won't like the resulting scrutinizer output).

leofeyer:
> I bet @leofeyer won't like the resulting scrutinizer output

He sure won't!

But maybe we can enhance the existing trait with an optional argument `$request = null` as discussed in an earlier approach?

discordier:
Which will then make the result unpredictable if you happen to pass in unchecked `null` (which is why I forbid doing so in the new trait) as it will fall back to the request stack and check if there is a request or and if so, use that one. We can do so but it bears new problems then. The using developer has to pay extra attention for things he simply should not have to.

On the other hand, I'd love to hear the opinion of @aschempp and @bytehead how the trait will get replaced in the future (see above questions) because until then this PR stays valid.

aschempp:
I've been working on this yesterday and last week. Here's the situation:

 1. The firewall must define if we are in Contao frontend or backend, because we must not allow to authenticate a `FrontendUser` if the request scope is set to backend. The information is duplicate.
 2. The firewall name (*context* or *providerKey*) is not available/readable from outside, at least not supposed to be (some tokens do have a `getProviderKey`, but at least the `AnonymousToken` does not)
 3. If someone defines a third firewall section (let's say `/api/`) that also authenticates in Contao (e.g. backend), we must also operate in Contao backend scope.

The scope identification could happen on the `$token->getUser()` object. Unfortunately, that is only available if there is a user actually authenticated, but not for the `AnonymousToken`. We can't use the firewall name because of 3. either.

I'm still digging the Symfony code for a clever way to get to that information. A custom firewall listener could handle that, but it would need to set state on a *scope services*. Not sure if that's a problem though, we do similar things with the language of the translator.

discordier:
Which in the end will turn out to be a checking trait again I suppose?
Keep in mind that sub requests might be located in another firewall but the firewall is only examined for the initial, externally originating request. So we need some way for context switching  (fe preview i.e.).

aschempp:
I don't think we should use a trait but probably a service. Also because the dependency injection of the service can change without affecting the consumer. Would we have done that in the first place we would not have the current trait issues.

Regarding sub-requests, context switching is not really supported. The user is authenticated against the browser cookie, therefore you can never authenticate a backend user in a frontend master request. Switching the context was never supported (due to the constant), and the framework cannot be initialized again in a different scope on a subrequest…

discordier:
Obsoleted by #677, closing therefore.

PR #620, opened 2016-11-08
[RTM] Fix the namespace declaration of various tests

discordier:
This PR corrects the namespaces for various test classes.

In addition, we also might want to change `composer.json`:
```diff
         "exclude-from-classmap": [
             "src/Resources/contao/config",
             "src/Resources/contao/dca",
-            "src/Resources/contao/helper",
             "src/Resources/contao/languages",
             "src/Resources/contao/templates",
             "src/Resources/contao/themes"
```

Otherwise the interfaces and deprecated exception classes can't get auto loaded.

### DIFF
--- a/tests/Analyzer/HtaccessAnalyzerTest.php
+++ b/tests/Analyzer/HtaccessAnalyzerTest.php
@@ -8,7 +8,7 @@
  * @license LGPL-3.0+
  */
 
-namespace Contao\CoreBundle\Test\HttpKernel\Bundle;
+namespace Contao\CoreBundle\Test\Analyzer;
 
 use Contao\CoreBundle\Analyzer\HtaccessAnalyzer;
 use Contao\CoreBundle\Test\TestCase;
--- a/tests/Cache/ContaoCacheClearerTest.php
+++ b/tests/Cache/ContaoCacheClearerTest.php
@@ -8,7 +8,7 @@
  * @license LGPL-3.0+
  */
 
-namespace Contao\CoreBundle\Test\Command;
+namespace Contao\CoreBundle\Test\Cache;
 
 use Contao\CoreBundle\Cache\ContaoCacheClearer;
 use Contao\CoreBundle\Test\TestCase;
--- a/tests/Cache/ContaoCacheWarmerTest.php
+++ b/tests/Cache/ContaoCacheWarmerTest.php
@@ -8,7 +8,7 @@
  * @license LGPL-3.0+
  */
 
-namespace Contao\CoreBundle\Test\Command;
+namespace Contao\CoreBundle\Test\Cache;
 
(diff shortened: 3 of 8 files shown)


leofeyer:
Merged in abd5ead4efe9005b3b9cb9fa946cc5aaec720be7.

> In addition, we also might want to change `composer.json`:

I don't think so. The resources are only loaded when the framework is booted, therefore they are excluded here. We can discuss this of course.

discordier:
For real resources, I agree.
However, in theses specific files there are only declarations (interfaces, classes and functions) which can not get loaded without framework.

We deprecated the class loader and now legacy classes can get loaded via composer (due to the class map entry) but the base interfaces for `\Contao\DataContainer` can not get loaded for example.

leofeyer:
As far as I understand, you want the helper resources to be loaded via Composer, right? Then would they not have to remain excluded from the class map and be loaded via

```json
    "autoload": {
        "files": ["helper/…"]
    }
```

instead?

discordier:
There are more than one solutions to this problem.
They are currently already added via
```
"classmap": [
            "src/Resources/contao"
        ],
```
but removed again by above mentioned line.
Adding them as "files" would of course also work, however they will then always get loaded, regardless if needed or not which I personally dislike and therefore suggested to remove above exemption.
This will still not autoload the `functions.php` (which has indeed to get added as "files"), but at least the interfaces and exception classes.

leofeyer:
Your argumentation is correct, still the change feels all wrong. We would end up with the following entries, all pointing to the same file:

```php
    'editable' => $vendorDir . '/contao/core-bundle/src/Resources/contao/helper/interface.php',
    'executable' => $vendorDir . '/contao/core-bundle/src/Resources/contao/helper/interface.php',
    'listable' => $vendorDir . '/contao/core-bundle/src/Resources/contao/helper/interface.php',
    'uploadable' => $vendorDir . '/contao/core-bundle/src/Resources/contao/helper/interface.php',
```

I don't think you can classmap files with multiple class definitions. @contao/developers /cc

aschempp:
Probably not, but you can include the file (https://getcomposer.org/doc/04-schema.md#files)

leofeyer:
Please read the ticket. 😄

discordier:
@leofeyer This will not impose an error, as all classes/interfaces will get loaded upon loading the first one from said file and there will not be any attempts to include the file again as the file is already loaded.

There is also no problem pointing several classes to the same file (at least none I am aware of and there wasn't recently any I was aware of).

The only downside is, it is loading more than one interface but heck, the whole stuff is so minimal, it probably won't harm in opposition to not having them loaded at all without booting the framework.

In the end, it only was a suggestion of how this can be further improved and I am fine with either change (including in class map and using files autoloader) but I prefer the class map way.

PR #621, opened 2016-11-09
[RFC] Deprecate the user aware trait

leofeyer:
This PR will deprecate the `UserAwareTrait` trait and adjust the two event listeners using it.


aschempp:
I remember someone once saying we should use setter injection 🙈 

Anyway, 👍 from me, but I'm not sure if the user check is correct. Not being an anonymous token does not necessarily qualify for a Contao user. But I don't know if that's in question at all places? If so, the check should rather be like

```php
if (null !== $token && $token->getUser() instanceof Contao\User)
```

But then again, this depends on #559 ;-)

leofeyer:
> Not being an anonymous token does not necessarily qualify for a Contao user.

This is your original implementation. And I'm pretty sure it is correct in our current scenario.

PR #626, opened 2016-11-20
[RFC] Use a cache for image dimensions

ausi:
Uses the `FilesystemAdapter` of the Symfony cache component to add an image dimensions cache.

I think an alternative would be to increase the requirement to Symfony 3.1 and configure the cache via `framework.cache.pools`.

### Todo:

* [ ] Release contao/image 0.4.0 once contao/image#25 ist merged and update composer.json
* [ ] Use `framework.cache.pools` instead of a service if #630 gets merged


src/Image/ImageFactory.php
```diff
@@ -73,9 +74,14 @@ class ImageFactory implements ImageFactoryInterface
...
     /**
+     * @var CacheItemPoolInterface
+     */
+    private $dimensionsCache;
+
+    /**
      * {@inheritdoc}
      */
-    public function __construct(ResizerInterface $resizer, ImagineInterface $imagine, ImagineInterface $imagineSvg, Filesystem $filesystem, ContaoFrameworkInterface $framework, $bypassCache, array $imagineOptions, array $validExtensions)
+    public function __construct(ResizerInterface $resizer, ImagineInterface $imagine, ImagineInterface $imagineSvg, Filesystem $filesystem, ContaoFrameworkInterface $framework, $bypassCache, array $imagineOptions, array $validExtensions, CacheItemPoolInterface $dimensionsCache = null)
```
Toflar: As it is optional, I'd use setter injection here :)

Toflar:
I'd even go for 3.2 as this introduces `PhpFilesAdapter` for opcode cache support. Symfony will then automatically prefer this over the filesystem one which is way faster :-)

src/Image/ImageFactory.php
ausi: Done.

ausi:
I don’t fully understand that. If Symfony 3.2 automatically switches to the new `PhpFilesAdapter` for `framework.cache`, requiring 3.1 would be enough AFAIK.

Toflar:
> If Symfony 3.2 automatically switches to the new PhpFilesAdapter for framework.cache, requiring 3.1 would be enough AFAIK.

That's true. I just meant that if we were about to raise the minimum requirement we should go for 3.2 directly. But yes, not really because of this PR.

composer.json
```diff
@@ -40,7 +42,7 @@
         "simplepie/simplepie": "~1.3",
         "tecnickcom/tcpdf": "~6.0",
         "true/punycode": "~1.0",
-        "contao/image": "^0.3",
+        "contao/image": "dev-feature/image-dimensions-cache as 0.4.0",
```
ausi: Agree. That’s why I added the todo in the first comment of this PR :)

src/Image/ImageFactory.php
```diff
@@ -88,6 +94,20 @@ public function __construct(ResizerInterface $resizer, ImagineInterface $imagine
     }
 
     /**
+     * Sets the cache pool used to cache image dimensions.
+     *
+     * @param CacheItemPoolInterface $dimensionsCache
+     *
+     * @return static
+     */
+    public function setDimensionsCache(CacheItemPoolInterface $dimensionsCache = null)
```
ausi: Done.

ausi:
We don’t want to merge contao/image#25 because caching of the dimensions should not belong to the `Image` class, so this PR becomes obsolete. The performance bug should get fixed by #632 for version 4.3. 

For Contao 4.4 I want to create a pull request that introduces a Symfony 3.2 tagged cache for `ImageFactory::create()` and `PictureFactory::create()` to also reduce the database queries.

PR #627, opened 2016-11-20
[RTM] Remove constructors from image and picture factory interfaces

ausi:
The constructors should be removed from the interfaces as we did in contao/image#10.


leofeyer:
Merged in e4c2bd559d3100cb5ced40a7802f31997781543c.

PR #630, opened 2016-11-21
[RTM] Make Symfony 3.2 the minimum requirement.

leofeyer:
This PR makes Symfony 3.1 the minimum requirement and removes the container scope usage.


leofeyer:

aschempp:
Can you write down here for the public why we are increasing the requirements?

aschempp:
yeah, that's what I don't understand…

 1. not a valid reason, because as you said *if it's not needed, don't do it*
 2. As you can see in the pull request, there is almost no BC code
 3. We definitely should benefit from these features, but we should not raise minimum requirements **before we actually use them** …

Toflar:
1. You don't read what I write. "if it makes your life easier (and it definitely does in this case) then go for it.".
2. You don't read what I write. "which was present in the SF code base and might have confused people".
3. We do use them. This PR here is the foundation for #626 which would be even better if it used cache pools introduced in 3.1.

BTW:
Raise minimum req: 4
Stay where we are: 1

aschempp:
Yeah, I just don't understand why we discussed this at the user meeting and agreed to not raise it until we actually need/use one of the new features…

aschempp:
This should be rebased on develop and reopened.

leofeyer:
I have raised the minimum Symfony version to 3.2 now.

PR #631, opened 2016-11-22
[RTM] Fix the JPEG interlacing

ausi:
Use the new `interlace` option from contao/image#24


leofeyer:
Merged in d1a9365941133bdd0ae30182d39db1aea080e066.

PR #632, opened 2016-11-22
[RFC] Fix performance of File::$imageSize

ausi:
On an article overview page with 50 pages/articles this change brings the response time from about 1800ms down to about 700ms.

This change should be OK to merge into 4.3 I think.

This is the alternative of #626

It may be useful to merge this back into 3.5...


leofeyer:
Merged in 0a1a9da158dd72a8020f3b27018d6cbb8124912b.

leofeyer:
Back ported to Contao 3.5 in contao/core@6723df9a7217b416cf1c95d4e553405c98772a09.

PR #644, opened 2016-12-05
Index options were added in Doctrine DBAL 2.5

aschempp:
In my current installation it got doctrine/dbal 2.4.5 installed after `composer update`. Probably because doctrine/dbal 2.5 seems to conflict with doctrine/common 2.7, and for whatever reason 2.7.1 was installed.

The options array key was not defined and raised a notice error, causing the doctrine diff command to fail.

### DIFF
--- a/src/EventListener/DoctrineSchemaListener.php
+++ b/src/EventListener/DoctrineSchemaListener.php
@@ -78,7 +78,8 @@ class DoctrineSchemaListener
                     $data['unique'],
                     $data['primary'],
                     $data['flags'],
-                    $data['options'])
+                    isset($data['options']) ? $data['options'] : null
+                )
             );
 
             $event->preventDefault();


leofeyer:
I've run into the same issue, but I fixed it by adding `"doctrine/dbal": "~2.5"` to my `composer.json` file. Version 2.5 contains quite a few bug fixes, therefore I'd prefer using it.

aschempp:
I thought about this too, but then decided not to change that because we're not using any feature of 2.5… as you know, we'd actually need 2.6 ;)

leofeyer:
> as you know, we'd actually need 2.6 ;)

I know. 😄 

Still version 2.5 contains a lot of bug fixes and those definitely matter to us, don't they?

aschempp:
Yeah, but we're not updating dependencies because of bug fixes. We would need to specify the latest version of everything in that case ;-)

leofeyer:
Only the ones that are downgraded for no obvious reason. ;)

aschempp:
Closed due to https://github.com/contao/core-bundle/issues/648

PR #647, opened 2016-12-12
[RFC] Cache build fails without autoload.php

aschempp:
Since we added template autoloading in Contao 4.3 the `autoload.php` is not used anymore. If no bundle contains it, the cache warmup will fail.

### DIFF
--- a/src/Cache/ContaoCacheWarmer.php
+++ b/src/Cache/ContaoCacheWarmer.php
@@ -116,8 +116,17 @@ class ContaoCacheWarmer implements CacheWarmerInterface
     {
         $dumper = new CombinedFileDumper($this->filesystem, new PhpFileLoader(), $cacheDir.'/contao', true);
 
-        $dumper->dump($this->locator->locate('config/autoload.php', null, false), 'config/autoload.php');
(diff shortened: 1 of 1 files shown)


src/Cache/ContaoCacheWarmer.php
```diff
@@ -116,8 +116,17 @@ private function generateConfigCache($cacheDir)
...
-        $dumper->dump($this->locator->locate('config/autoload.php', null, false), 'config/autoload.php');
-        $dumper->dump($this->locator->locate('config/config.php', null, false), 'config/config.php');
+        try {
+            $dumper->dump($this->locator->locate('config/autoload.php', null, false), 'config/autoload.php');
+        } catch (\InvalidArgumentException $e) {
+            // Ignore, no autoload.php existed in any bundle
+        }
+
+        try {
+            $dumper->dump($this->locator->locate('config/config.php', null, false), 'config/config.php');
```
leofeyer: This change is not necessary, is it? The core bundle has a `config.php` file, so the case can never occur.

aschempp: That's correct. We also did not add a try-catch because the core bundle supplied an `autoload.php`. I guess it's safe to add it as well, we never know about future changes ;-)

PR #652, opened 2016-12-14
Allow to run Contao without a localconfig.php

aschempp:
Since the localconfig parameters can be specified in the container, we don't need the file to be present.

### DIFF
--- a/src/Resources/contao/library/Contao/Config.php
+++ b/src/Resources/contao/library/Contao/Config.php
@@ -317,17 +317,7 @@ class Config
 	 */
 	public static function isComplete()
 	{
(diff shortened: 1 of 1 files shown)


leofeyer:
What happens if someone changes the configuration in the back end settings?

aschempp:
> What happens if someone changes the configuration in the back end settings?

A localconfig file will be created. Same as when running the install tool for the first time.
https://github.com/aschempp/contao-core-bundle/blob/3c860fd8c3e4319aa915f066e72ad7c20f08e8ac/src/Resources/contao/library/Contao/Config.php#L179

In my case, I disabled the backend settings module for installations where I set settings in the container.

Issue #638, opened 2016-11-30
Kernel::getContaoRootDir()


leofeyer:
I think we should provide a `Kernel::getContaoRootDir()` method instead of relying on `TL_ROOT` always being `dirname($container->getParameter('kernel.root_dir'))`. @contao/developers WDYT?

Toflar:
Makes no sense. It's the app root dir and not related to Contao at all.

ausi:
Also a parameter that can be used instead of `"%kernel.root_dir%/.."` would be great.

discordier:
IMHO such a method is not needed at all.
Instead of relying on the constant you will have to fetch the kernel always then which is almost as bad as the constant.

A parameter (which can be injected into services) makes more sense.

In the end, we should remove the references to `TL_ROOT` altogether and use the proper locations (cache dir etc.) directly.

aschempp:
We should first find out why we need that parameter, and why Symfony does not…

leofeyer:
We need the parameter, because our paths are all relative to `TL_ROOT` and Symfony's paths are relative to `TL_ROOT.'/app'` (or where ever the app dir is).

At the moment, we assume (hardcoded!) that `TL_ROOT === dirname('kernel.root_dir')`, which of course only works as long as the kernel is in `app/` and `app/` is in the root level. This implicit directory structure requirement needs to be removed IMO.

aschempp:
I'm very well aware of the current situation. I just wonder why anyone would need the root path at all. 

For example you cannot use the root to find your bundles files, because you would hardcode the vendor path which is wrong too.

leofeyer:
Here are some places where it would be useful:

* https://github.com/contao/core-bundle/blob/hotfix/4.3.1/src/Cache/ContaoCacheWarmer.php#L256
* https://github.com/contao/core-bundle/blob/hotfix/4.3.1/src/Resources/config/services.yml#L158 (@ausi's example)
* https://github.com/contao/core-bundle/blob/hotfix/4.3.1/src/HttpKernel/Bundle/ContaoModuleBundle.php#L33
* https://github.com/contao/core-bundle/blob/hotfix/4.3.1/src/DependencyInjection/Compiler/AddResourcesPathsPass.php#L41
* https://github.com/contao/core-bundle/blob/hotfix/4.3.1/src/Command/InstallCommand.php#L96
[...]

PR #653, opened 2016-12-14
[RFC] Change contao.image.target_path to an absolute value

aschempp:
Fixed one issue mentioned in https://github.com/contao/core-bundle/issues/638

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -90,7 +90,16 @@ class Configuration implements ConfigurationInterface
                             ->defaultValue($this->debug)
                         ->end()
                         ->scalarNode('target_path')
-                            ->defaultValue('assets/images')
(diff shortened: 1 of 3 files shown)


leofeyer:
This is not correct unfortunately, as the end result should be `'%contao.root_dir%/assets/images'`.

discordier:
Reopen and retarget? Was against contao:hotfix/4.3.1.

leofeyer:
We can actually implement this if we use a different parameter name:

```php
                        ->scalarNode('target_dir')
                            ->defaultValue('%contao.root_dir%/assets/images')
                        ->end()
```

Then we can deprecate `contao.image.target_path` in favor of `contao.image.target_dir`.

aschempp:
To be fully BC the default value would need to be `%kernel.root_dir%/%contao.image.target_path%`. But I don't like to create a new parameter just for that…

leofeyer:
See #684.

PR #654, opened 2016-12-14
[RTM] Respect the XLIFF source-language attribute

### LINKED ISSUE #642: XLIFF Original language not english
Hi,
It would be a good thing for the translation of the modules that the English language is not compulsory.

In the language directory `en` only `<source>` is read, in others language directory only `<target>` is read (if I'm not mistaken). So we can't make a translation de to es for example with the logic of the XLIFF format.
It will be necessary to take into account the attributes `original`, `source-language` and `target-language` of `<xliff version="1.1"><file>`

Best,
Antoine

ausi:
Fixes #642

The loader should behave the same way as before, but instead of the hardcoded assumption that the source language is `en` it respects the value of the XLIFF attribute `source-language`.

- [x] Check if the language format in Contao and XLIFF is the same

### DIFF
--- a/src/Config/Loader/XliffFileLoader.php
+++ b/src/Config/Loader/XliffFileLoader.php
@@ -127,7 +127,21 @@ class XliffFileLoader extends Loader
      */
     private function getNodeByLanguage(\DOMElement $unit, $language)
     {
-        return ('en' === $language) ? $unit->getElementsByTagName('source') : $unit->getElementsByTagName('target');
+        $tagName = 'target';
+        $node = $unit;
+
+        // Use the source tag if the source language matches
+        while ($node = $node->parentNode) {
+            if (
+                'file' === $node->nodeName
+                && strtolower($node->getAttribute('source-language')) === strtolower($language)
+            ) {
+                $tagName = 'source';
+                break;
+            }
+        }
+
+        return $unit->getElementsByTagName($tagName);
     }
 
     /**
--- a/src/Config/Loader/XliffFileLoader.php
+++ b/src/Config/Loader/XliffFileLoader.php
@@ -102,7 +102,7 @@ class XliffFileLoader extends Loader
      *
      * @return string
      */
-    private function getPhpFromFileNode($fileNode, $tagName)
+    private function getPhpFromFileNode(\DOMElement $fileNode, $tagName)
     {
         $return = '';
         $units = $fileNode->getElementsByTagName('trans-unit');
--- a/src/Config/Loader/XliffFileLoader.php
+++ b/src/Config/Loader/XliffFileLoader.php
@@ -76,11 +76,40 @@ class XliffFileLoader extends Loader
         $xml = $this->getDomDocumentFromFile($name);
 
         $return = "\n// ".str_replace(strtr(dirname($this->rootDir), '\\', '/').'/', '', strtr($name, '\\', '/'))."\n";
-        $units = $xml->getElementsByTagName('trans-unit');
+        $fileNodes = $xml->getElementsByTagName('file');
+        $language = strtolower($language);
+
+        /** @var \DOMElement[] $fileNodes */
+        foreach ($fileNodes as $fileNode) {
+            $tagName = 'target';
+
+            // Use the source tag if the source language matches
+            if (strtolower($fileNode->getAttribute('source-language')) === $language) {
+                $tagName = 'source';
+            }
+
+            $return .= $this->getPhpFromFileNode($fileNode, $tagName);
+        }
+
+        return $return;
+    }
+
+    /**
+     * Converts an XLIFF file node into PHP code.
+     *
+     * @param \DOMElement $fileNode
+     * @param string      $tagName
+     *
+     * @return string
+     */
+    private function getPhpFromFileNode($fileNode, $tagName)
+    {
+        $return = '';
+        $units = $fileNode->getElementsByTagName('trans-unit');
 
         /** @var \DOMElement[] $units */
         foreach ($units as $unit) {
-            $node = $this->getNodeByLanguage($unit, $language);
+            $node = $unit->getElementsByTagName($tagName);
 
             if (null === $node || null === $node->item(0)) {
                 continue;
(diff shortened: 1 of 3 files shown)


aschempp:
1. what if `target` does not exist? Shouldn't we read `source` in that case regardless of the attribute?
 2. your while-loop would search for a parent `file` node if the first found does not have the attribute. I don't think that's correct?
 3. are you sure the language format in XLIFF and Contao are the same? LanguageTag vs. LocalID (de-DE vs. de_DE)?

ausi:
> what if `target` does not exist? Shouldn't we read `source` in that case regardless of the attribute?

I don’t think so, we didn’t do that in the past and we shouldn’t change this behavior in a bugfix release IMO.

> your while-loop would search for a parent `file` node if the first found does not have the attribute. I don't think that's correct?

`<file>` nodes cannot be nested in XLIFF.

> are you sure the language format in XLIFF and Contao are the same? LanguageTag vs. LocalID (de-DE vs. de_DE)?

I’m not sure, i will check and update the PR if necessary.

discordier:
> what if target does not exist? Shouldn't we read source in that case regardless of the attribute?

Never ever! This will violate everything XLIFF is designed for and in addition will break current behavior of skipping missing language keys.
Imagine a translation file where the source is outdated but has not been translated yet.
That file would overwrite the new value of the original language then with the old value, which definitely is not what we want.

> > your while-loop would search for a parent file node if the first found does not have the attribute. I don't think that's correct?

> <file> nodes cannot be nested in XLIFF.

Does this also apply to `<xliff>` in general? I mean, is it required to be the root element of the `\DomDocument`?
If not you might cycle a big tree upwards where no hit will be possible anymore.
Imagine a document like this (namespace declarations and verbose boilerplate omitted):
```xml
<foo>
  <!-- many nested opening elements -->
  <xliff>
    <file>
      <unit>
      </unit>
    </file>
  <xliff>
  <!-- many nested closing elements -->
</foo>
```
I wonder if this is possible, however the spec does not clearify this.

>> are you sure the language format in XLIFF and Contao are the same? LanguageTag vs. LocalID (de-DE vs. de_DE)?

> I’m not sure, i will check and update the PR if necessary.

The specs are pretty clear on this by defining the [target-language](http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#target-language):
> A language code as described in the [RFC 4646](http://www.ietf.org/rfc/rfc4646.txt), the successor to [RFC 3066]. The values for this attribute follow the same rules as the values for xml:lang. Unlike the other XLIFF attributes, the values for xml:lang are not case-sensitive. For more information see [the section on xml:lang in the XML specification](https://www.w3.org/TR/REC-xml/#sec-lang-tag), and the [erratum E11](http://www.w3.org/XML/xml-V10-2e-errata#E11) (which replaces RFC 1766 by RFC 3066).
> The target language can be also specified by xml:lang in each <target> element. The values of target-language and xml:lang in <target> can be different only when in an <alt-trans> element.

So they have to use `-` delimiting.

aschempp:
> So they have to use - delimiting.

And what is Contao using?

leofeyer:
I guess it would make more sense to store the tag information in the `convertXlfToPhp()` method (one time) instead of while looping through the tree in the `getNodeByLanguage()` method (several times), wouldn't it?

discordier:
@leofeyer No, as the language is defined on a per `<file>`-element level and you may have multiple of these per document.
Each of these files may contain different combinations of source and target languages.
As we are scanning all `<trans-units>` and not looping over the files in an outer loop, the `<file>` element may change from one `<trans-unit>` to the next `<trans-unit>`.

leofeyer:
Then we should maybe loop over the files in an outer loop.

discordier:
Can safely do so, as all methods are private here. Don't know about C3.5 though.

leofeyer:
IMO, this is a feature anyway, so it should be targeted against version 4.4 and not be back ported to version 3.5.

discordier:
I have to formally protest. The xliff format was not parsed correctly. Therefore this is a a bugfix.
If you would be reading the wrong data from `localconfig.php` or database it would clearly qualify as bugfix. IMO reading the wrong stuff from the translation files is not different.
Implementing entire XLIFF 2.0 support would be a feature but simply fixing the reading of existing data instead of wild guessing and reading half of it definitely is a bugfix!

leofeyer:
Then I guess we have to fix it in Contao 3.5, too.

ausi:
>> So they have to use - delimiting.
> 
> And what is Contao using?

I think Contao uses the `-` too, [library/Contao/Validator.php:246](https://github.com/contao/core-bundle/blob/87ae8c6ff95fe6428ba96e8cdf0aae3dbf37de24/src/Resources/contao/library/Contao/Validator.php#L246).

> I guess it would make more sense to store the tag information in the `convertXlfToPhp()` method (one time) instead of while-looping through the tree in the `getNodeByLanguage()` method (several times), wouldn't it?

Then we would have to iterate over all `<file>` elements first. I don’t think it’s necessary because the language files get cached anyway.

> IMO, this is a feature anyway

IMO it’s a bug.

leofeyer:
> Then we would have to iterate over all `<file>` elements first. I don’t think it’s necessary because the language files get cached anyway.

Only if the cache is warmed up. Otherwise the file is loaded and parsed on the fly, therefore I would prefer to improve the implementation. If you are busy, I can create a PR?

discordier:
Add label "needs-backport" plz

leofeyer:
Hm, this cannot easily be back ported, as we have rewritten most of the routines for Contao 4.

discordier:
Shouldn't it be simply a matter of adding an outer loop around [here](https://github.com/contao/core/blob/e398ed783a28c840ed01644b7a070b57bf6baf49/system/modules/core/library/Contao/System.php#L732) for all `<file>` nodes and determining the language in there instead of [here](https://github.com/contao/core/blob/e398ed783a28c840ed01644b7a070b57bf6baf49/system/modules/core/library/Contao/System.php#L734)?

leofeyer:
Probably yes. You could test it and create a pull request if it works. 😄

discordier:
See PR.

leofeyer:
Thank you, @discordier.

Issue #636, opened 2016-11-30
Move all images (resized and unresized) to the assets folder


agoat:
When uploaded a file smaller then 50x50 Pixel there will be no preview shown in the filetree when the folder is not made public. This happens because no images will be upscaled (generally ok) or resized to the same size. In these cases the path to the files folder is used but normally is not accessible via http. 

![untitled-2](https://cloud.githubusercontent.com/assets/1402434/20746302/7ef0283e-b6e5-11e6-9976-8ea825378a30.png)

This happens in the backend as well as in the frontend. Particularly difficult when using images sizes and coincidental one of the sources have the same size or is bigger than the original.

**All (used) images should be moved to the assets folder!**

ausi:
Does that work in Contao 3.5?

agoat:
I didn't testet it right now but I think as long as the folder is not protected in contao 3.5 there is no problem. But when the protection is activated, the same will happen. 
In contao 4+ folders are protected by default, so the images would not be loaded from the /files/.. folder by default.

agoat:
I fixed this temporarily for me in my extension that deferres the image calculation:
https://github.com/agoat/contao-deferredimages-bundle/blob/contao4.3/develop/src/Image/DeferredResizer.php?ts=4#L110-L120

leofeyer:
@contao/developers A more general question is whether we should generate publicly accessible thumbnails for protected images at all? We might as well use inline images in this case.

aschempp:
I don't feel like piping is a good idea… but obviously thumbnails mean that protected images will be publicly accessible...

Toflar:
I think piping is an excellent idea and the only one that solves this issue for protected images.
Given that the image itself knows whether it's protected or not (we should be able to find out if that's the case or not), we could provide a second filesystem cache in the image factory that is protected and add a controller for those images (`_contao/protected-images/<cachekey>` for example). @ausi do you think that would be possible?

agoat:
The source image is protected, but the thumnails generated from are publicly accessible. Isn´t that the idea behind the protection of the files folder?? The source is protected. 

Why should I upload images to a cms and don´t give publicly access in some way.

Currently, if a images is required in a smaller size, a smaller version is rendered into the assets folder. We just have to handle images without resizing.

Another question is how to recognize images required through css files?? - Maybe that can be implemented into the css parser when the urls are rewritten.

Toflar:
> Why should I upload images to a cms and don´t give publicly access in some way.

[...]

PR #655, opened 2016-12-16
Version 4.3.1

leofeyer:


src/Command/SymlinksCommand.php
```diff
@@ -284,4 +281,16 @@ private function filterNestedPaths(Finder $finder, $prepend)
...
+    /**
+     * Returns the path relative to the root directory.
+     *
+     * @param string $path
+     *
+     * @return string
+     */
+    private function getRelativePath($path)
+    {
+        return str_replace(strtr($this->rootDir, '\\', '/').'/', '', strtr($path, '\\', '/'));
```
leofeyer: Please don't try to fix things just by looking at them. The whole symlinking is not trivial, which is why we have thoroughly unit tested everything. And this method is definitely doing what it is supposed to to – which also includes returning an absolute path if `$path ` is not within `$this->rootDir`. 😉

leofeyer:
@contao/developers The release of version 4.3.1 now basically depends on #657 and #636. Both are up for discussion and could use some feedback and/or work.

src/Cache/ContaoCacheWarmer.php
```diff
@@ -116,8 +116,17 @@ private function generateConfigCache($cacheDir)
     {
         $dumper = new CombinedFileDumper($this->filesystem, new PhpFileLoader(), $cacheDir.'/contao', true);
 
-        $dumper->dump($this->locator->locate('config/autoload.php', null, false), 'config/autoload.php');
-        $dumper->dump($this->locator->locate('config/config.php', null, false), 'config/config.php');
+        try {
+            $dumper->dump($this->locator->locate('config/autoload.php', null, false), 'config/autoload.php');
+        } catch (\InvalidArgumentException $e) {
+            // No autoload.php files found
+        }
```
discordier: IMO this should be handled by a method:
```php
private function findConfig($fileName)
{
    try {
        return $this->locator->locate('config/' . $fileName, null, false);
    } catch (\InvalidArgumentException $e) {
        return [];
    }
}
```
This not only lowers the complexity of this method but also will not mask away any `\InvalidArgumentException` which could be thrown within the `$dumper->dump()` call but will only catch the ones originating from the `FileLocator`.

Same below for the `autoload.php`

src/Cache/ContaoCacheWarmer.php
```diff
@@ -130,8 +139,12 @@ private function generateDcaCache($cacheDir)
         $dumper = new CombinedFileDumper($this->filesystem, new PhpFileLoader(), $cacheDir.'/contao', true);
         $processed = [];
 
-        /** @var SplFileInfo[] $files */
-        $files = $this->finder->findIn('dca')->files()->name('*.php');
+        try {
```
discordier: We now have this block of 6 lines in two locations, this IMO justifies creation of a private method to prevent future errors.

src/Config/Loader/XliffFileLoader.php
```diff
@@ -76,11 +76,40 @@ private function convertXlfToPhp($name, $language)
...
+        $language = strtolower($language);
+
+        /** @var \DOMElement[] $fileNodes */
+        foreach ($fileNodes as $fileNode) {
+            $tagName = 'target';
+
+            // Use the source tag if the source language matches
+            if (strtolower($fileNode->getAttribute('source-language')) === $language) {
+                $tagName = 'source';
+            }
```
discordier: IMO we should add a check here if either the `target-language` or the `source-language` match and if neither matches, return empty.
It is entirely possible to have something with source "de" and target "fr" and now the content from "fr" will be used on the "en" page which IMO should be prevented.
The above cited use case could happen when multiple entries have been combined into one big XLIFF file where each originates from a different translator.

ausi: The `target-language` attribute is optional, so I don’t think this check is necessary. Also, we didn’t want to change the behavior of the current XLIFF loader in a patch release, but only fix a bug.

ausi:
> The release of version 4.3.1 now basically depends on #657 and #636.

IMO #636 can be delayed for a later version if necessary.

src/Config/Loader/XliffFileLoader.php
discordier: Fair enough.

src/Cache/ContaoCacheWarmer.php
leofeyer: Changed in eca5f9299f76e3907896d890522df5aa2917f2e5. Much better readable now.

leofeyer:
> IMO #636 can be delayed for a later version if necessary.

I guess it could, but I would really like to have it fixed. @ausi If you tell me how you want to fix it, I'll happily create a PR.

aschempp:
I think #636 is not an easy task and should not be done in this patch release that should be released asap due to the security issue.

leofeyer:
I am not asking to change the whole logic in a patch release. I only want to create a cache file for images smaller than 50 pixel, so they are shown in the file manager.

The logic change should be done in a feature release of course.

leofeyer:
The release has been scheduled for December 22nd (if there are no further objections).

Issue #637, opened 2016-11-30
Uploads marked as failed but uploaded and resized correctly


agoat:
When uploading files(images) mostly all files marked as failed. I tested this with a clean contao 4.3 installation. 

![untitled-1](https://cloud.githubusercontent.com/assets/1402434/20749021/dae32c76-b6f0-11e6-988e-505940e4664b.png)

Afterwards it look like this in the filetree:
![untitled-3](https://cloud.githubusercontent.com/assets/1402434/20749100/2cae102a-b6f1-11e6-8618-7b2e41976574.png)

Could not find where it goes wrong.

aschempp:
ddce4dfebae4fe15105c6e929c92ce61392e64a8 is breaking the uploader because `AjaxRedirectResponseException` no longer returns a 2xx status code.

agoat:
An extra question: Is it possible that we can add support of html5 chunked upload?? I know dropzone.js doesn´t support that but there is a fork which implemented chunked upload with dropzone (from 2012): https://github.com/bootle/DropZone
Another solution could be this one:
https://github.com/23/resumable.js (also have a look into this discussion: https://github.com/enyo/dropzone/issues/339)

Maybe worth to have a look. 
I think this would be a great benefit for contao to supports large file uploads (e.g. videos).

aschempp:
Not related to this ticket…

agoat:
Ok, I opened an extra ticket for chunk support: https://github.com/contao/core-bundle/issues/650

leofeyer:
Fixed in 19f560b69eebece7d0278e9950c3fdedabcee2ed.

aschempp:
Not sure if that fix is really correct. A 302 response without a location header sounds invalid to me?

discordier:
@aschempp It is a 204 response now?

aschempp:
In this specific case, yes. But the AjaxRedirectResponseException is 302 without a location.

discordier:
According to RFC, a 302 SHOULD indicate an alternative location but does not have to.

PR #657, opened 2016-12-20
Correctly handle ajax uploads in DC_Folder

aschempp:
see https://github.com/contao/core-bundle/issues/637

### DIFF
--- a/src/Resources/contao/drivers/DC_Folder.php
+++ b/src/Resources/contao/drivers/DC_Folder.php
@@ -12,9 +12,11 @@ namespace Contao;
 
 use Contao\CoreBundle\Exception\AccessDeniedException;
 use Contao\CoreBundle\Exception\InternalServerErrorException;
+use Contao\CoreBundle\Exception\ResponseException;
 use Contao\CoreBundle\Util\SymlinkUtil;
 use Contao\Image\ResizeConfiguration;
 use Patchwork\Utf8;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 
@@ -1085,6 +1087,11 @@ class DC_Folder extends \DataContainer implements \listable, \editable
 
 				if (empty($arrUploaded) && !$objUploader->hasError())
 				{
+					if ($blnIsAjax)
+					{
+						throw new ResponseException(new Response($GLOBALS['TL_LANG']['ERR']['emptyUpload'], 400));
+					}
+
 					\Message::addError($GLOBALS['TL_LANG']['ERR']['emptyUpload']);
 					$this->reload();
 				}
@@ -1126,6 +1133,11 @@ class DC_Folder extends \DataContainer implements \listable, \editable
 			// Redirect or reload
 			if (!$objUploader->hasError())
 			{
+				if ($blnIsAjax)
+				{
+					throw new ResponseException(new Response(\Message::generateUnwrapped(), 201));
+				}
+
 				// Do not purge the html folder (see #2898)
 				if (isset($_POST['uploadNback']) && !$objUploader->hasResized())
 				{
--- a/src/Resources/contao/classes/Ajax.php
+++ b/src/Resources/contao/classes/Ajax.php
@@ -10,7 +10,7 @@
 
 namespace Contao;
 
-use Contao\CoreBundle\Exception\AjaxRedirectResponseException;
+use Contao\CoreBundle\Exception\InternalServerErrorHttpException;
 use Contao\CoreBundle\Exception\NoContentResponseException;
 use Contao\CoreBundle\Exception\ResponseException;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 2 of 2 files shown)


aschempp:
Still does not handle that I think 302 is wrong for `AjaxRedirectResponseException`

leofeyer:
Which status code would you send instead?

aschempp:
That's hard to say. The main issue is that the `X-Ajax-Location` header is currently something special for Contao. And for this special case, I would keep `204` for BC reasons and because "it does not matter".
However, a **real** 302 response **with** Location header would obviously be correct for a redirect, no matter if ajax or not. It can be handled by ajax as well so its not *wrong* by definition…

aschempp:
I also wonder if a *no content response* is correct in the `Ajax.php` because it should never reach that point and should actually result in an internal server error or thelike…

bytehead:
> I also wonder if a no content response is correct in the Ajax.php because it should never reach that point and should actually result in an internal server error or thelike…

That's totally weird to me - wasted hours on that `204 no content`.

leofeyer:
If an Ajax request is successful, the server should send a 2** status code back. How weird would it be to get an internal server error in response to a successful request???

bytehead:
AFAIR I got `204` as well during errors.

leofeyer:
That's a different issue though.

aschempp:
The issue is that the `204` is sent regardless of the action. As you can see in my PR, there's a case where no file uploads are posted. That's not a 204, that's an error. And the case where `move()` does not return a response is not supported in an ajax state, that should never render a `204`.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -1126,6 +1133,11 @@ public function move($blnIsAjax=false)
 			// Redirect or reload
 			if (!$objUploader->hasError())
 			{
+				if ($blnIsAjax)
+				{
+					throw new ResponseException(new Response(\Message::generateUnwrapped(), 201));
```
leofeyer: This is the wrong status code. If the response has content, it should be `200`, otherwise `204`.

leofeyer: Using a `204` status code would only work for single uploads; then you would have to return the URL to the uploaded file in the `Location` header. But `move()` processes multiple uploads at once, so …

leofeyer: So the question is, do we want to send the HTML messages with the reponse (status 200) or just a no content response (status 204)?

aschempp: I disagree. Please read carefully.

[RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616.html) (*Hypertext Transfer Protocol -- HTTP/1.1*)

> The request has been fulfilled and resulted in a new resource being created. The newly created resource can be referenced by the URI(s) returned in the entity of the response, with the most specific URI for the resource given by a Location header field. The response SHOULD include an entity containing a list of resource characteristics and location(s) from which the user or user agent can choose the one most appropriate. The entity format is specified by the media type given in the Content-Type header field. 

RFC 2119 (*Key words for use in RFCs to Indicate Requirement Levels*):
> **SHOULD**   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.


1. the valid reason to not include the list is that there can be multiple, so it can't be represented with a `Location` field. Except we could return the `Location` of the parent folder, but I think that's unneccessary.
2. There is actually returned a list of entities. The media type is `text/html`, and the returned list are HTML strings (the messages) of the uploads.

leofeyer: I am only going to merge this if all @contao/developers confirm that this is correct in their opinion, because I definitely think that you are wrong.

discordier: IMO 204 is out of question, as we are returning a response containing `\Message::generateUnwrapped()` and therefore have response content.

For the `Location` header, please see [rfc7231 section 6.3.2](https://tools.ietf.org/html/rfc7231#section-6.3.2) (which obsoletes your referenced RFC 2616):

> The 201 (Created) status code indicates that the request has been fulfilled and has resulted in one or more new resources being created.  The primary resource created by the request is identified by **either a Location header field in the response or, if no Location field is received, by the effective request URI.**
> The 201 response payload typically describes and links to the resource(s) created.

Therefore we **MUST** either send a `Location` header OR the URI it got posted at is effectively the correct one. The standard has been made more strict.

However, I see a solution by sending a `Location` header for the first uploaded file and `Link` headers (see RFC [rfc5988 section 5](https://tools.ietf.org/html/rfc5988#section-5)) for any subsequent file in the original payload.

leofeyer: > IMO 204 is out of question, as we are returning a response

This is not correct. Right now we are not returning a response; only the PR changes this.

aschempp: Sure we can overcomplicate things and send headers that noone cares about… Or we can just accept that it does not matter at all and the effective request URL is the folder where files are uploaded to and that should be enough…

PR #660, opened 2016-12-29
[RTM] Set Contao encryption key from kernel secret by default

aschempp:
Dunno why we didn't add that in the first place. This will make https://github.com/contao/standard-edition/blob/master/app/config/config.yml#L29 obsolete.


leofeyer:
Then we can also omit `encryption_key: "%kernel.secret%"` in the `config.yml` file, right?

aschempp:
Itsn't that what I linked above? 😂

leofeyer:
So while we are at this, why don't we do the same for the `contao.prepend_locale` parameter? Then we could remove the entire Contao configuration from the `config.yml` file.

leofeyer:
Actually, this can be done even more simple:

```php
        $rootNode
            ->children()
                ->booleanNode('prepend_locale')
                    ->defaultValue("%prepend_locale%")
                ->end()
                ->scalarNode('encryption_key')
                    ->cannotBeEmpty()
                    ->defaultValue("%kernel.secret%")
                ->end()
                // …
```

I don't know why we are using constructor injection here at all?

leofeyer:
Ah, probably because we use `!$this->debug` instead of `$this->debug?`.

leofeyer:
See my follow up PR: #664

aschempp:
I have updated the PR.

PR #661, opened 2016-12-29
[RFC] Fix missing TL_FILES_URL for pictures

### LINKED ISSUE #659: The new image service does not recognize the assets url (TL_ASSETS_URL)
Tested with a clean contao 4.3.1 installation.

When setting the assets url in the root page settings, the css and script asset paths are using the assets url, but not the images.

I couldn't find any use of the contant TL_ASSETS_URL in the contao/image repository. Not sure if the image service should use TL_ASSETS_URL in the image->getUrl() method or contao/core have to bring this in (with the second prefix parameter).

ausi:
Fixes #659

The legacy Picture class in Contao 4.2 used the `TL_FILES_URL` constant in [Picture.php:281](https://github.com/contao/core-bundle/blob/58cf7360ddebbcffb51020d7ddd2892d71bd5c79/src/Resources/contao/library/Contao/Picture.php#L281). This is missing now and so the `TL_FILES_URL` is currently ignored in pictures. To fix it I added the constant to all `Picture::getImg()` and `Picture::getSources()` calls.

### DIFF
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -1511,8 +1511,8 @@ abstract class Controller extends \System
 
 			$picture = array
 			(
-				'img' => $picture->getImg(TL_ROOT),
-				'sources' => $picture->getSources(TL_ROOT)
+				'img' => $picture->getImg(TL_ROOT, TL_FILES_URL),
+				'sources' => $picture->getSources(TL_ROOT, TL_FILES_URL)
 			);
 
 			if ($src !== $arrItem['singleSRC'])
--- a/src/Resources/contao/library/Contao/InsertTags.php
+++ b/src/Resources/contao/library/Contao/InsertTags.php
@@ -836,8 +836,8 @@ class InsertTags extends \Controller
 
 							$picture = array
 							(
-								'img' => $picture->getImg(TL_ROOT),
-								'sources' => $picture->getSources(TL_ROOT)
+								'img' => $picture->getImg(TL_ROOT, TL_FILES_URL),
+								'sources' => $picture->getSources(TL_ROOT, TL_FILES_URL)
 							);
 
 							$picture['alt'] = $alt;
--- a/src/Resources/contao/library/Contao/Picture.php
+++ b/src/Resources/contao/library/Contao/Picture.php
@@ -249,8 +249,8 @@ class Picture
 
 		return array
 		(
-			'img' => $picture->getImg(TL_ROOT),
-			'sources' => $picture->getSources(TL_ROOT),
+			'img' => $picture->getImg(TL_ROOT, TL_FILES_URL),
+			'sources' => $picture->getSources(TL_ROOT, TL_FILES_URL),
 		);
 	}
 
--- a/tests/Contao/PictureTest.php
+++ b/tests/Contao/PictureTest.php
@@ -74,7 +74,7 @@ class PictureTest extends TestCase
         $GLOBALS['TL_CONFIG']['validImageTypes'] = 'jpeg,jpg,svg,svgz';
 
         define('TL_ERROR', 'ERROR');
-        define('TL_FILES_URL', '');
+        define('TL_FILES_URL', 'http://example.com/');
         define('TL_ROOT', self::$rootDir);
 
         $container = $this->mockContainerWithContaoScopes();
@@ -142,8 +142,8 @@ class PictureTest extends TestCase
 
         $this->assertEquals(200, $pictureData['img']['width']);
         $this->assertEquals(200, $pictureData['img']['height']);
-        $this->assertEquals('dummy.jpg', $pictureData['img']['src']);
-        $this->assertEquals('dummy.jpg', $pictureData['img']['srcset']);
+        $this->assertEquals('http://example.com/dummy.jpg', $pictureData['img']['src']);
+        $this->assertEquals('http://example.com/dummy.jpg', $pictureData['img']['srcset']);
         $this->assertEquals([], $pictureData['sources']);
     }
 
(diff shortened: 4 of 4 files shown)


agoat:
Shouldn't it be TL_ASSETS_URL instead of TL_FILES_URL?

Maybe I don´t understand the logic. What is the different between those two?

agoat:
The contao/image class differentiate between assets and files:
```php
724.  $static = (strncmp($src, 'assets/', 7) === 0) ? TL_ASSETS_URL : TL_FILES_URL;
725.
726.  return '<img src="' . $static . \System::urlEncode($src) . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="' . \StringUtil::specialchars($alt) . '"' . (($attributes != '') ? ' ' . $attributes : '') . '>';
```
https://github.com/contao/core-bundle/blob/4.3.1/src/Resources/contao/library/Contao/Image.php#L724

ausi:
> Shouldn't it be TL_ASSETS_URL instead of TL_FILES_URL?

`TL_FILES_URL` was used for resized images since its introduction in Contao 2.10, see [d798baa](https://github.com/contao/core/commit/d798baaed17d56a85d8828bd3dd60d2f2726d1d0#diff-d8e920c55c8dee7f3d54963d023642b2R2798). The name of the two constants may be a bit misleading, I think the names of the database columns are more clear: `staticFiles` for `TL_FILES_URL` and `staticPlugins` for `TL_ASSETS_URL`.

> What is the different between those two?

`TL_FILES_URL` is used for files from the *files/* directory and resized images, `TL_ASSETS_URL` is used for all other assets AFAIK.

> The contao/image class differentiate between assets and files

This is only in the `getHtml` method where the check is for images from `assets/contao/images`, this method is mostly used for icons.

leofeyer:
But since the resized images are stored in `assets/img`, I think we should use `TL_ASSETS_URL` here.

ausi:
How about removing `TL_FILES_URL` altogether (in 4.4) and only leave the setting for `TL_ASSETS_URL`? For backwards compatibility we can set both constants to the same value.

For the bugfix in 4.3 we should stick with `TL_FILES_URL` IMO.

leofeyer:
But we are still using both constants, aren't we?

ausi:
Yes, we are. But if we want to change the behavior, I would suggest to replace `TL_FILES_URL` with `TL_ASSETS_URL` everywhere and remove the setting for `TL_FILES_URL` completely.

aschempp:
I would not change anything as the whole property is obsolete with HTTP/2 …

leofeyer:
Me neither.

agoat:
IMO we should replace TL_FILES_URL to TL_ASSETS_URL everywhere as @ausi recommended.

> I would not change anything as the whole property is obsolete with HTTP/2 …

Although it will be obsolete with HTTP/2, currently it is necessary since only a few provider supports HTTP/2 today and this will be changing slowly in 2017. 

And what about CDN?

leofeyer:
Also, please consider that replacing the `TL_FILES_URL` is a BC break an thus only eligible for Contao 5. However, the constants will be removed all together in Contao 5, so …

ausi:
> I would not change anything as the whole property is obsolete with HTTP/2

I don’t think so. Serving the assets of a webpage via a CDN is perfectly fine with HTTP/2. Why should that be obsolete?

> As discussed, we should use the `TL_ASSETS_URL` constant instead of `TL_FILES_URL`.

Why should we change the behavior in a bugfix release? It was so since version 2.10.

> Also, please consider that replacing the `TL_FILES_URL` is a BC break an thus only eligible for Contao 5.

Why? The constant would still be there, it would just point to the same URL as `TL_ASSETS_URL`.

leofeyer:
> Why? The constant would still be there, it would just point to the same URL as `TL_ASSETS_URL`.

Exactly. And right now the two URLs can be different, which would not be possible after the change.

> Why should we change the behavior in a bugfix release? It was so since version 2.10.

No it was not? `$static = (strncmp($src, 'assets/', 7) === 0) ? TL_ASSETS_URL : TL_FILES_URL;`

ausi:
> And right now the two URLs can be different, which would not be possible after the change.

IMO that would be OK for a minor release because the distinction between the two makes no sense to me. But I’m OK with it, if we don’t want to break that behavior.

> No it was not? `$static = (strncmp($src, 'assets/', 7) === 0) ? TL_ASSETS_URL : TL_FILES_URL;`

As already mentioned in https://github.com/contao/core-bundle/pull/661#issuecomment-269697134 *This is only in the `getHtml` method where the check is for images from `assets/contao/images`, this method is mostly used for icons.* Take a look at [2.10.0/Controller.php:3161](https://github.com/contao/core/blob/2.10.0/system/libraries/Controller.php#L3161) and [4.3.1/Controller.php:1597](https://github.com/contao/core-bundle/blob/4.3.1/src/Resources/contao/library/Contao/Controller.php#L1597).

agoat:
> How about removing TL_FILES_URL altogether (in 4.4) and only leave the setting for TL_ASSETS_URL? For backwards compatibility we can set both constants to the same value.
> 
> For the bugfix in 4.3 we should stick with TL_FILES_URL IMO.

I would do what @ausi already suggested.

aschempp:
This PR is still approved from me, I would not change it to `TL_ASSETS_URL`. 

Regarding the CDN question, the only time I used a CDN I would set it on the actual host name to deliver everything including HTML files from CDN. Depending on your CDN that would even be better due to ESI includes, and you can usually also exclude .html from CDN if you prefer not to do that. But sure, there might be limited cases where you want a different domain for assets and HTML.

leofeyer:
Merged in 910587d18f868669b6cff6c5fbb78c0e1962c3f4 actually.

PR #662, opened 2016-12-30
[RFC] Add the contao.root_dir parameter

leofeyer:
Add the contao.root_dir parameter and canonicalize the path (see #638).


src/DependencyInjection/Compiler/ContaoRootDirPass.php
```diff
@@ -0,0 +1,30 @@
...
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class ContaoRootDirPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $container->setParameter('contao.root_dir', realpath($container->getParameter('contao.root_dir')));
```
ausi: Is `realpath` the right function to use here? It does not only resolve `/../` but also expands symlinks, checks for existence of the directory, may normalize the character case and could change unix style paths to windows style.

leofeyer: > but also expands symlinks,

That seems ok to me in this case.

> checks for existence of the directory,

This is an extra benefit, because the directory should exist.

> may normalize the character case and could change unix style paths to windows style

I don't think so. This is what `PathUtil::canonicalize()` does, therefore I am not using the library here.

ausi: >> may normalize the character case and could change unix style paths to windows style
>
> I don't think so. This is what `PathUtil::canonicalize()` does, therefore I am not using the library here.

`PathUtil::canonicalize()` converts everything into unix-style separators `/` never to windows style and does not change the character case.

But `realpath()` does according to the [PHP documentation](http://php.net/manual/en/function.realpath.php):

> For case-insensitive filesystems realpath() may or may not normalize the character case.  
> On windows realpath() will change unix style paths to windows style.

src/Resources/config/services.yml
```diff
@@ -155,7 +155,7 @@ services:
         class: Contao\CoreBundle\Image\LegacyResizer
         public: false
         arguments:
-            - "%kernel.root_dir%/../%contao.image.target_path%"
+            - "@=parameter('contao.root_dir')~'/'~parameter('contao.image.target_path')"
```
leofeyer: This is just temporary until we implement #653.

src/DependencyInjection/Compiler/ContaoRootDirPass.php
leofeyer: Well, actually we do not want any path separator change at all. We need to keep back slashes on Windows and slashes on *nix systems, so `realpath()` seems to be ok.

leofeyer:
I had to change the implementation, because this will only work if the `contao.root_dir` parameter is defined in the app kernel. Otherwise the information is not available in the install tool. Also, the new implementation is much less cumbersome than the one from 1fab114042f5a3ab194017fe6eb9ae805d698f02, which had to use expression language and `realpath()`.

leofeyer:
> I'll see if I can up with an alternative solution.

You don't have to. I have an alternative solution here, which does a lot of configuration processing in the install tool to retrieve the `contao.root_dir` value. The approach is just more complex and cumbersome, but I'm also not too big a fan of the custom kernel.

> In addition to that requiring a concrete implementation of the kernel instead of an interface is wrong

It is an abstract class, therefore it is not wrong. But as I said above, …

leofeyer:
We have already discussed this in #638 and in Mumble and we have agreed that we want to add this feature to remove the hardcoded assumption that the Contao root directory is always `%kernel.root_dir%/..`. No need to discuss it again here IMO.

aschempp:
What @discordier is exactly my gut feeling as well. However, it's true that the hardcoded assumption is not good either.

leofeyer:
I'm not saying that this is the best solution, either. But the only alternative would be to move the Contao folders into the `app/` folder so `TL_ROOT === $container->getParameter('kernel.root_dir')`. However, we have ruled out this solution in our Mumble discussion.

PR #664, opened 2017-01-02
[RTM] Set all defaults in the Configuration class

leofeyer:
This is the follow up PR for #660.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -49,11 +49,11 @@ class Configuration implements ConfigurationInterface
         $rootNode
             ->children()
                 ->booleanNode('prepend_locale')
-                    ->defaultFalse()
+                    ->defaultValue('%prepend_locale%')
                 ->end()
                 ->scalarNode('encryption_key')
-                    ->isRequired()
                     ->cannotBeEmpty()
+                    ->defaultValue('%kernel.secret%')
                 ->end()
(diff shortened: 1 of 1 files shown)


aschempp:
This does not work, see http://symfony.com/doc/current/configuration/using_parameters_in_dic.html

leofeyer:
You are right, it does not work for anything but strings, so it does not work for `prepend_locale` but it will work for `encryption_key`.

leofeyer:
So what do we do now? Shall I change this PR accordingly or do you want to rework #660?

aschempp:
You're right, I misunderstood the manual that it would result in a string value instead of the parameters. I would simply just change the `kernel.secret` and keep `prepend_locale` as-is, also because it would be a potential config break.

leofeyer:
I would prefer to have all the default configuration in the `Configuration` class, so we do not need a `contao:` section in the `config.yml` file at all.

aschempp:
We cannot rely on a global parameter for our configuration, as it would imply it to be set in the `parameters.yml`. Same goes with `kernel.secret`, it does not default to `secret` but that's a setup of the application.

leofeyer:
Then just do `$container->hasParameter('prepend_locale') ? $container->getParameter('prepend_locale') : false`.

aschempp:
That would still be "assuming application configuration".

SGehle:
Why is the pid get-parameter ignored?

PR #674, opened 2017-01-17
[RTM] Allow to manually pass a value to any widget

Toflar:
I suppose it will probably still take some time until our widgets become obsolete so here's something we should have had ever since there was Contao :-P

### DIFF
--- a/src/Resources/contao/library/Contao/Widget.php
+++ b/src/Resources/contao/library/Contao/Widget.php
@@ -752,14 +752,13 @@ abstract class Widget extends \Controller
 
 
 	/**
-	 * By default, Widget will use its getPost() method when validate() is
-	 * called. This method allows you to fetch the value from elsewhere.
+	 * Set a callback to fetch the widget input instead of using getPost()
 	 *
-	 * @param callable $callable
+	 * @param callable $callable The callback
 	 *
-	 * @return $this
+	 * @return $this The widget object
 	 */
-	public function setInputCallback(callable $callback = null)
+	public function setInputCallback(callable $callback=null)
 	{
 		$this->inputCallback = $callback;
 
@@ -772,7 +771,7 @@ abstract class Widget extends \Controller
 	 */
 	public function validate()
 	{
-        $varValue = (null !== $this->inputCallback ? call_user_func($this->inputCallback) : $this->getPost($this->strName));
+        $varValue = (is_callable($this->inputCallback) ? call_user_func($this->inputCallback) : $this->getPost($this->strName));
 		$varValue = $this->validator($varValue);
 
 		if ($this->hasErrors())
--- a/src/Resources/contao/library/Contao/Widget.php
+++ b/src/Resources/contao/library/Contao/Widget.php
@@ -771,7 +771,7 @@ abstract class Widget extends \Controller
 	 */
 	public function validate()
 	{
-        $varValue = (is_callable($this->inputCallback) ? call_user_func($this->inputCallback) : $this->getPost($this->strName));
+		$varValue = (is_callable($this->inputCallback) ? call_user_func($this->inputCallback) : $this->getPost($this->strName));
 		$varValue = $this->validator($varValue);
 
 		if ($this->hasErrors())
--- a/src/Resources/contao/library/Contao/Widget.php
+++ b/src/Resources/contao/library/Contao/Widget.php
@@ -122,10 +122,16 @@ abstract class Widget extends \Controller
 	protected $strLabel;
 
 	/**
-	 * Value
+	 * Value (set after validate())
 	 * @var mixed
 	 */
 	protected $varValue;
+
+	/**
+	 * Input callback
+	 * @var callable
+	 */
+	protected $inputCallback;
 
 	/**
 	 * CSS class
@@ -381,11 +387,11 @@ abstract class Widget extends \Controller
 
 			case 'value':
 				// Encrypt the value
-				if ($this->arrConfiguration['encrypt'])
+				if (isset($this->arrConfiguration['encrypt']) && true === $this->arrConfiguration['encrypt'])
 				{
 					return \Encryption::encrypt($this->varValue);
 				}
-				elseif ($this->varValue == '')
+				elseif ($this->varValue === '')
 				{
 					return $this->getEmptyStringOrNull();
 				}
@@ -746,11 +752,28 @@ abstract class Widget extends \Controller
 
 
 	/**
+	 * By default, Widget will use its getPost() method when validate() is
+	 * called. This method allows you to fetch the value from elsewhere.
+	 *
+	 * @param callable $callable
+	 *
+	 * @return $this
+	 */
+	public function setInputCallback(callable $callback = null)
+	{
+		$this->inputCallback = $callback;
+
+		return $this;
+	}
+
+
+	/**
 	 * Validate the user input and set the value
 	 */
 	public function validate()
 	{
-		$varValue = $this->validator($this->getPost($this->strName));
+        $varValue = (null !== $this->inputCallback ? call_user_func($this->inputCallback) : $this->getPost($this->strName));
+		$varValue = $this->validator($varValue);
 
 		if ($this->hasErrors())
 		{
--- a/src/Resources/contao/library/Contao/Widget.php
+++ b/src/Resources/contao/library/Contao/Widget.php
@@ -122,7 +122,7 @@ abstract class Widget extends \Controller
 	protected $strLabel;
 
 	/**
-	 * Value (set after validate())
+	 * Value
 	 * @var mixed
 	 */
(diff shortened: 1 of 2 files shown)


src/Resources/contao/library/Contao/Widget.php
```diff
@@ -381,11 +381,11 @@ public function __get($strKey)
 
 			case 'value':
 				// Encrypt the value
-				if ($this->arrConfiguration['encrypt'])
+				if (isset($this->arrConfiguration['encrypt']) && true === $this->arrConfiguration['encrypt'])
```
leofeyer: This change – although it is not wrong – is not technically required, is it?

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -381,11 +381,11 @@ public function __get($strKey)
...
 				{
 					return \Encryption::encrypt($this->varValue);
 				}
-				elseif ($this->varValue == '')
+				elseif ($this->varValue === '')
```
leofeyer: This change – although it is not wrong – is not technically required, is it?

tests/Contao/WidgetTest.php
```diff
@@ -66,6 +67,27 @@ public function testGetPost($key, $input, $value, $expected)
...
+            ->willReturnArgument(0);
+
+        /** @var Widget $widget */
+        $widget->validate('foobar');
+        $this->assertSame('foobar', $widget->value);
+
+        /** @var Widget $widget */
+        $widget->validate(null, true);
+        $this->assertNull($widget->value);
+    }
```
leofeyer: There should be another test for `->validate(null, false)`.

src/Resources/contao/library/Contao/Widget.php
Toflar: It somehow is because if I left it as is, I would have needed to adjust my unit test and suppress "undefined index encrypt" notices and I figured instead of fixing the unit test I might as well just fix the source of the problem :D

src/Resources/contao/library/Contao/Widget.php
Toflar: It is. `$this->varValue == '')` would be `true` if `$this->varValue` is `null` ;-) A case I would have never thought of without unit tests ;-)

tests/Contao/WidgetTest.php
Toflar: Right, there was even a bug which I fixed with the unit test covering that case :-D See df901dfa58b38a67c82a4c39753342294183cf93.

Toflar:
Failing tests not related.

Toflar:
Thougt of this too but it doesn't change the signature to be clear. Both values are optional.

aschempp:
If my widget has currently implemented the `validate()` method, PHP will raise an error because my widget does no longer follow the parent. Same issue we're having in https://github.com/isotope/core/issues/1753

A method like `validateCustom($value)` would not need a second parameter because calling that method would always mean `null` should be treated as value.

Toflar:
Okay, proposal for a method name? Not sure if setValue() is too simple? :D

aschempp:
`setValue` does not imply that it's validated. How about `validateValue()`?

leofeyer:
How about `overrideValue()`? We could also make it `setPost()` if the change is supposed to be permanent.

Toflar:
> setValue does not imply that it's validated. How about validateValue()?

It does not necessarily need to be validated imho. So `setValue()` is actually correct.
Basically it should just take a custom value instead of searching in the post values when you call `validate()`. A different method like `validateValue()` would not be a good idea as you could not set the value in one component and let it validate in another.

> How about overrideValue()? We could also make it setPost() if the change is supposed to be permanent.

Excuse me? That would be completely wrong and would make this PR completely useless? I can already set the post value now but that's exactly what's wrong and I want to omit?

I think `$widget->setValue()->validate()`. Would be correct, I'm adjusting the PR so you can see what I mean.

Toflar:
Okay, next round. Just one more comment on this version now:

Yes I did think of adding only one method instead of two but I want to cover the `null` value case. The only other idea was to make `getPost()` `public` because then you could do `$widget->setCustomValue($widget->getPost($widget->name))` to set it back to the original behavior. But this would again change the method signature of a method so I guess we're safer the way I built it now.

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -746,11 +758,41 @@ public function getAttribute($strKey)
...
+	}
+
+
+	/**
 	 * Validate the user input and set the value
 	 */
 	public function validate()
 	{
-		$varValue = $this->validator($this->getPost($this->strName));
+		$varValue = ($this->blnCustomValueSet) ? $this->varCustomValue : $this->getPost($this->strName);
```
leofeyer: The parentheses around `$this->blnCustomValueSet` can be omitted. Also, I'd prefer `blnHasCustomValue`.

aschempp:
This does not make sense to me at all. If you want to set a value, simply use `$widget->value = …`. The `__set()` method will take care of assigning it to the variable. If you want validation, that's why I suggested `validateValue()`.

Toflar:
> If you want validation, that's why I suggested validateValue().

And again, this cannot work as I pointed out before. The method cannot be named differently because all the current code calls `validate()` now.

leofeyer:
The current implementation dose make sense, just the naming is not good yet. How about `setInput()` or `setInputValue()`?

aschempp:
So @Toflar explained the issue to me. I didn't realize he's using the regular form generator and therefore cannot modify the *flow*. I just came up with another idea that I like much more than those two ugly properties.

I suggest to add a `setInputCallback()` method that accepts a callable. If the callable is set, it will be called instead of `getPost`. This will allow for more flexibility as well, and it is easier to understand.

**Getting input from input value is fairly easy.**
```php
$widget->setInputCallback(function ($strName) {
    return \Input::get($strName);
});
```

**so is getting it from prepared data**
```php
$widget->setInputCallback([$myclass, getValue]);

// MyClass
function getValue($strName) {
    return $this->data[$strName];
}
```

The Widget::validate method then should look like this:
```php
$varValue = (null !== $this->inputCallback ? call_user_func($this->inputCallback, $strName) : $this->getPost($this->strName));
```

leofeyer:
Fine with me, too.

Toflar:
PR and tests updated accordingly. RTM.

src/Resources/contao/library/Contao/Widget.php
discordier: Isn't `isset($this->arrConfiguration['encrypt'])` also true for `$this->arrConfiguration['encrypt'] = false;`?
Therefore this seems to be required IMO.

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -746,11 +752,28 @@ public function getAttribute($strKey)
...
+	}
+
+
+	/**
 	 * Validate the user input and set the value
 	 */
 	public function validate()
 	{
-		$varValue = $this->validator($this->getPost($this->strName));
+        $varValue = (null !== $this->inputCallback ? call_user_func($this->inputCallback) : $this->getPost($this->strName));
```
discordier: Shouldn't this rather be:
```diff
-		$varValue = $this->validator($this->getPost($this->strName));
+        $varValue = (is_callable($this->inputCallback) ? call_user_func($this->inputCallback) : $this->getPost($this->strName));
```
IMO this will prevent errors when wrong values get passed or a class is not loadable or the like.
What do you think?

Toflar: There's a `callable` type hint, you cannot set anything else than `callable` or `null` ;)

discordier: Of course I can, I can inject `['SomeUnloadableClass', 'someMethod']` where `is_callable` will try to autoload and return `false` where your solution will trigger a "Fatal error: class not found".

leofeyer:
What about @discordier's unanswered [comment](https://github.com/contao/core-bundle/pull/674#discussion_r97854422)? I cannot merge the PR until it has been addressed.

Toflar:
It has been addressed. I wrote there is a type hint for `callable`. You cannot inject an array...dunno what @discordier was writing. The PR is RTM to me.

aschempp:
@Toflar `callable` does allow an array, maybe you were looking for `\Closure`?

Toflar:
Yeah we could do this but I don't understand why it would be relevant in any way. If you provide a wrong `callable` you did something wrong? An error is better than falling back to the `Input::post()` stuff?

aschempp:
I agree, if a callable is set we should try to call it and not simply ignore…

leofeyer:
I also think that we should support `[$obj, 'method']`, which can be passed to `call_user_func()`.

PR #677, opened 2017-01-22
[RTM] Match security firewall based on request scope

aschempp:
There is one fundamental problem in the current implementation:
 - There is a *frontend* and a *backend* firewall **context** based on the route (= path / pattern).
 - Each route can have a *_scope* that defines `TL_MODE`. 
 - These two are independet, so theroretically a route with a path that matches the *frontend firewall scope* can have a *_scope attribute backend*.

With this PR the firewall is no longer matching the path but only the request scope. This also means a route without request scope will no longer have a Contao firewall.

This also replaces #619 as a request matcher is the right way to detect Contao frontend or backend.

**TODOs**
 - [x] Update unit tests
 - [x] Update DEPRECATED.md


bytehead:
👍

aschempp:
PR is updated to use the new request matchers instead of the `ScopeAwareTrait`. One thing to discuss is the class name `RequestContext`. It's not that wrong because *context* is what it's called in the firewall, but there already is a `RequestContext` class in Symfony which could be confusing.

src/Routing/ScopeMatcher.php
```diff
@@ -0,0 +1,115 @@
...
+     * @param RequestMatcherInterface $backendMatcher
+     */
+    public function __construct(RequestMatcherInterface $frontendMatcher, RequestMatcherInterface $backendMatcher)
+    {
+        $this->frontendMatcher = $frontendMatcher;
+        $this->backendMatcher = $backendMatcher;
+    }
+
+    /**
+     * Checks whether the request is a Contao the master request.
```
xchs: … is a Contao master request.

PR #679, opened 2017-01-23
[RTM] Use the contao.web_dir parameter in the Combiner

ausi:
### ToDo:

- [x] Add tests

### DIFF
--- a/src/Resources/contao/library/Contao/Combiner.php
+++ b/src/Resources/contao/library/Contao/Combiner.php
@@ -75,7 +75,7 @@ class Combiner extends \System
 	protected $arrFiles = array();
 
 	/**
-	 * Webdir relative to TL_ROOT
+	 * Web dir relative to TL_ROOT
 	 * @var string
 	 */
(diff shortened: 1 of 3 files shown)


src/Resources/contao/library/Contao/Combiner.php
```diff
@@ -73,12 +73,29 @@ class Combiner extends \System
...
+
 
 	/**
 	 * Public constructor required
 	 */
 	public function __construct()
 	{
+		$strWebDir = \System::getContainer()->getParameter('contao.web_dir');
+
+		if (strncmp($strWebDir, TL_ROOT . '/', strlen(TL_ROOT . '/')) === 0)
```
leofeyer: I think this should be `DIRECTORY_SEPARATOR`, because `kernel.root_dir` will be `C:\…` on Windows.

ausi: Better in 696148b9704672b4a2f25e07f165e89b659112ad ?

PR #685, opened 2017-01-24
[RTM] Use Symfony Security for authentication purposes in Contao

bytehead:
This is a working draft to replace the current `simple_preauth` mechanism to login backend users and frontend members with the `form_login` from the Symfony Security Component.

### ToDos:
- [x] Keep redirect on login
- [x] Implement role switcher / impersonate any `BackendUser`
- [x] Session Life Time (`framework.session.cookie_lifetime`)
- [x] Rename firewalls (prefix with `contao_`)
- [x] Implement Security for `FrontendUser`
- [x] Replace `postAuthenticate`
- [x] Replace `postLogin`
- [x] Replace `checkCredentials` (needs a custom `AuthenticationProvider`)
- [x] Replace `importUser`
- [x] Replace `postLogout`
- [x] Replace current `accountLocked` stuff
- [x] Test Frontend Preview
- [x] Write tests
- [x] Check Code style
- [x] Deprecate all old hooks
- [x] Provide Event for legacy `postAuthenticate` hook
- [x] Provide Event for legacy `checkCredentials` hook
- [x] Provide Event for legacy `importUser` hook
- [x] Provide Event for legacy `postLogout` hook
- [x] Show backend login errors

### Fixes needed:
- [x] ~~Backend login redirect: Simple test: login, go somewhere, delete your session cookie.~~
- [x] UserProvider refreshUser without credentials
- [x] Provide exception messages
- [x] Replace usage of static legacy class methods (e.g. `Config`, `Environment`)
- [x] Call `importUser` only on `POST` request
- [x] Additional check with `UserChecker` in [`Frontend`](https://github.com/contao/core-bundle/pull/685#pullrequestreview-78677905) 
- [x] Implement the `EquatableInterface` on User to compare the token object and DB object
- [x] on `BackendUser::getInstance` check if there is a valid `_security_contao_backend` token in the actual session to restore the user from
- [x] on `FrontendUser::getInstance` check if there is a valid `_security_contao_frontend` token in the actual session to restore the user from

### Nice to have (possibly future PRs):
- use `DoctrineTokenProvider` for persistent remember me tokens

### Possible new features (future PRs):

- Allow login also with the email address (not only username)
- Allow some user action via command like `activate`, `change-password` or similar


src/Resources/contao/classes/BackendUser.php
```diff
@@ -30,7 +31,7 @@
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
-class BackendUser extends \User
+class BackendUser extends User
```
leofeyer: This is a BC break, isn't it?

src/Resources/contao/library/Contao/User.php
```diff
@@ -92,7 +96,7 @@
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
-abstract class User extends \System
+abstract class User extends System implements AdvancedUserInterface, EncoderAwareInterface, \Serializable
```
leofeyer: This is also a BC break.

src/Resources/contao/dca/tl_user.php
```diff
@@ -634,9 +634,14 @@ public function deleteUser($row, $href, $label, $title, $icon, $attributes)
...
 	 *
 	 * @throws Exception
+	 *
+	 * @deprecated Deprecated since Contao 4.x, to be removed in Contao 5.0.
+	 *             Use the switch_user_button_generator service instead.
 	 */
 	public function switchUser($row, $href, $label, $title, $icon)
 	{
+		@trigger_error('Using tl_user->switchUser() has been deprecated and will no longer work in Contao 5.0. Use the switch_user_button_generator service instead', E_USER_DEPRECATED);
+
```
aschempp: The rest of this method should be removed or replaces by the new switch user stuff.

src/EventListener/SwitchUserListener.php
```diff
@@ -0,0 +1,31 @@
...
+
+    public function onSwitchUser(SwitchUserEvent $event)
+    {
+        /** @var BackendUser $user */
+        $user = $this->tokenStorage->getToken()->getUser();
+
+        /** @var BackendUser $targetUser */
+        $targetUser = $event->getTargetUser();
+
+        $this->logger->info(sprintf("User %s has switched to user %s.", $user->username, $targetUser->username));
```
aschempp: The only purpose of this listener is to log the action? If that's a replacement of an existing log, we should log with a `ContaoContext` to make it appear in the backend log.

src/Resources/config/services.yml
```diff
@@ -222,6 +222,21 @@ services:
...
             - ["setContainer", ["@service_container"]]
 
+    contao.security.backend_user_provider:
+        class: Contao\CoreBundle\Security\User\ContaoBackendUserProvider
+        public: false
+        arguments:
+            - "@contao.framework"
+
+    contao.security.switch_user_button_generator:
+        class: Contao\CoreBundle\Security\User\SwitchUserButtonGenerator
```
aschempp: To me this is an event (callback) listener and should be in the according namespace.

src/Resources/contao/controllers/BackendMain.php
```diff
@@ -203,6 +205,20 @@ protected function output()
...
+		$authorizationChecker = \System::getContainer()->get('security.authorization_checker');
+
+		/** @var RouterInterface $router */
+		$router = \System::getContainer()->get('router');
+
+		$logoutLink = $router->generate('contao_backend_logout');
+
+		if ($authorizationChecker->isGranted('ROLE_PREVIOUS_ADMIN')) {
+			$logoutLink = $router->generate('contao_backend', [
+				'_switch_user' => '_exit',
```
aschempp: this is a new feature, right? Previously a login just logged out whatever user was currently in. Might be confusing if we don't change the logout label to "switch back to X".

src/Security/User/SwitchUserButtonGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+        if (!$this->authorizationChecker->isGranted('ROLE_ALLOWED_TO_SWITCH')) {
+            return '';
+        }
+
+        $stmt = $this->connection->prepare("SELECT id, username FROM tl_user WHERE id = :id");
+        $stmt->bindValue('id', $row['id']);
+        $stmt->execute();
+
+        if (0 === $stmt->rowCount()) {
+            throw new UserNotFoundException('Invalid user ID' . $row['id']);
```
aschempp: This should never happen, so I would simply return empty string (remove action) in this case and not add a new exception for a never-happening case :)

src/Security/User/SwitchUserButtonGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if ($tokenUser->username === $user->username) {
+            return '';
+        }
+
+        $url = $this->router->generate('contao_backend', [
+            '_switch_user' => $user->username
+        ]);
+
+        return $this->twig->render('@ContaoCore/Backend/switch_user.html.twig', [
```
aschempp: Do we need a twig template for this? Generally good, but there are so many places where we currently don't…

src/EventListener/SwitchUserListener.php
bytehead: That should be the replacement of the existing log, but I didn't get it how it's done to make it appear in the backend.

src/Resources/contao/dca/tl_user.php
bytehead: 👍

src/Resources/contao/controllers/BackendMain.php
bytehead: Good point :)

src/Security/User/SwitchUserButtonGenerator.php
bytehead: Yes IMHO we need exactly that. Otherwise we will end up like the DCA classes are right now with template code in the logic parts. And we have to start using (twig) templates at some point.

src/EventListener/SwitchUserListener.php
aschempp: https://github.com/contao/core-bundle/blob/7f746f41156f81867aefbb812f5e086e543d5bdd/src/Resources/contao/library/Contao/System.php#L246

src/Security/User/SwitchUserButtonGenerator.php
aschempp: ok, but then this should be a general button template, right?

src/EventListener/SwitchUserListener.php
bytehead: Ah, thank you! I'll update it later.

src/Security/User/SwitchUserButtonGenerator.php
bytehead: Good idea. Should be discussed with the others probably?

src/Resources/contao/library/Contao/User.php
```diff
@@ -357,17 +357,14 @@ public function authenticate()
 	 */
 	public function login()
 	{
+		dump(\Input::post('username'));
```
Toflar: @leofeyer ?

bytehead: It's not completed yet 😀

Toflar: Ah fuck, sorry. I thought this was a 4.4.5 commit, sorry about that!

bytehead:
As discussed on Mumble we will go for a custom `AuthenticationProvider` to support old hooks like `checkCredentials`.

src/Resources/contao/controllers/BackendMain.php
bytehead: The current label is `Close the current session` and thus not wrong in my opinion.

src/Security/User/ContaoUserProvider.php
```diff
@@ -58,9 +64,13 @@ public function __construct(ContaoFrameworkInterface $framework, ScopeMatcher $s
      * {@inheritdoc}
      *
      * @return BackendUser|FrontendUser
+     *
+     * @deprecated Using ContaoUserProvider::loadByUsername has been deprecated and will no longer work in Contao 5.0.
```
xchs: It should read `… ContaoUserProvider::loadUserByUsername …`, right?

src/Security/User/ContaoUserProvider.php
```diff
@@ -58,9 +64,13 @@ public function __construct(ContaoFrameworkInterface $framework, ScopeMatcher $s
...
      */
     public function loadUserByUsername($username): User
     {
+        @trigger_error('Using ContaoUserProvider::loadByUsername has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
```
xchs: It should read `… ContaoUserProvider::loadUserByUsername …`, right?

bytehead:
After nearly one year and 101 commits later Symfony Security in Contao is feature complete and ready for comments!

aschempp:
@bytehead I just tried to use this feature. When I go to `example.com/contao` I get a 403 exception and I'm not redirected to the login screen 😢

aschempp:
I'm also unable to login through `contao/login`, the page just reloads and does nothing… Btw, I just checked out the feature branch, I assume I don't need any other version of `manager-bundle` or `installation-bundle`?

bytehead:
Do you have imported the `security.yml` config?
```
imports:
    - { resource: "@ContaoCoreBundle/Resources/config/security.yml" }
```

bytehead:
@aschempp Works like a charm on a fresh installation of mine. Could you provide your `composer.json`?

aschempp:
Sorry, the security.yml was not correctly loaded. There's another issue though, the backend login redirect does no longer know the referrer, so you're not redirected to the previous screen after login.

src/Resources/contao/classes/Frontend.php
```diff
@@ -685,4 +690,53 @@ public static function getResponseFromCache()
...
+	{
+		/** @var SessionInterface $session */
+		$session = \System::getContainer()->get('session');
+
+		if ($session->has($sessionKey) && $token = unserialize($session->get($sessionKey)))
+		{
+			/** @var TokenInterface $token */
+
+			// Validate the session ID and timeout
+			if ($token->isAuthenticated())
```
bytehead: Additional check with UserChecker

src/Resources/contao/classes/BackendUser.php
```diff
@@ -30,7 +29,7 @@
...
```
backbone87: This is a BC break. `Contao\BackendUser::getInstance() instanceof User` would return false, because `instanceof` does not trigger autoloads.

bytehead: Using `\User` leads to a `ClassNotFoundException`:
```
Attempted to load class "User" from the global namespace.
Did you forget a "use" statement for e.g. "Symfony\Component\Security\Core\User\User" or "Symfony\Bridge\Doctrine\Tests\Fixtures\User"?
```

bytehead: We could use '\Contao\User` - no `ClassNotFoundException` then but that doesn't solve the problem as well.

backbone87: The class aliases are still present in C4, so why would `BackendUser externds \User` fail?

src/Resources/contao/library/Contao/User.php
```diff
@@ -95,7 +98,7 @@
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
-abstract class User extends \System
+abstract class User extends System implements AdvancedUserInterface, EncoderAwareInterface, \Serializable
```
backbone87: This causes problems. What Symfony Security means with their `UserInterface` is not the user entity, but the user identity. Making your user entity also the identity causes problems, when stored roles change (and other stuff). I would like to explain this to you via mumble.

aschempp: It would be better to publicly explain, because then everyone can follow and find about it

bytehead: What I actually read on the documentation doesn't sound so wrong:
> Regardless of how your user are loaded or where they come from (a database, configuration, web service, etc), you will have a class that implements this interface. Objects that implement this interface are created and loaded by different objects that implement UserProviderInterface

src/Resources/contao/classes/BackendUser.php
bytehead: I'm not sure, if the class alias stuff is still present. The profiler doesn't show any class as aliased. There all composerized.

src/Resources/contao/library/Contao/User.php
backbone87: The symfony docs do not go very deep on this and dont address issues like role change. You can google for these kind of issues and will find solutions like removing the login token after you updated the user roles, but all this can be avoided by using a separate object "Identity" that is derived from your actual user entity.

> It would be better to publicly explain, because then everyone can follow and find about it

I never said, i wont do it publicly, but just that writing paragraphs here isnt the correct medium, to explain a design flaw

bytehead: I've never experienced issues like that and it's for sure not my first project with Symfony security.  Even the everywhere used [`FosUserBundle`](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Model/User.php) also makes no distinction between entity and identity.
Correct me, if I'm wrong.

backbone87: https://github.com/symfony/symfony/issues/12025
https://github.com/symfony/symfony/pull/24331

These solutions dont handle all edge cases (and arent even merged yet)

bytehead: Thanks! I'll have a deeper look into the issues 👍

aschempp: I can't think of any way where the Symfony user is not our legacy Contao user class, or we will run into a ton of other problems. I quickly read up on the issues but couldn't find any relation to entites or models?

backbone87: > I can't think of any way where the Symfony user is not our legacy Contao user class, or we will run into a ton of other problems.

That is not the point of separating out the identity from the user entity. The identity is an adapter of the user entity to integrate with the symfony security. So it is strictly bound to contaos user entity. It also keeps your user entity clean (since you dont have possible empty implementations of unused interface methods there). But the main point is that you can tailor the identitys runtime lifecycle to play nice with symfony without hacking your user entities lifecycle (especially when you want to use doctrine for the user entity).

> I quickly read up on the issues but couldn't find any relation to entites or models?

The problem comes from serialization. You dont know if your user entity comes from the DB or the session. If you use an adapter/proxy identity you know that the Identity will be always "complete" (because it can be serialized and unserialized) and always comes from the session or runtime and never from the DB. Think of the identity as a value object referencing your user entity.

src/Resources/contao/library/Contao/User.php
```diff
@@ -258,478 +277,425 @@ public function getData()
...
-			$intUserid = $objSession->pid;
-		}
+	/**
+	 * {@inheritdoc}
+	 */
+	public function serialize()
+	{
+		return serialize(array(
+			$this->id,
+			$this->username,
```
aschempp: I think we need to save `tstamp` here as well so we can use the timestamp to compare if the `refreshUser` returned a new object than was in the security token.

bytehead: Yeah, thats something still missing (see open task in the list).

src/Resources/contao/library/Contao/User.php
bytehead: The [`EquatableInterface`](http://api.symfony.com/4.0/Symfony/Component/Security/Core/User/EquatableInterface.html) looks like to be determined to be implemented to catch up with stuff like role changes (because role changes are not detected by the [`AbstractToken::hasUserChanged`](https://github.com/symfony/symfony/blob/3.2/src/Symfony/Component/Security/Core/Authentication/Token/AbstractToken.php#L244-L287). 

But I understand your POV stated here:
> But the main point is that you can tailor the identitys runtime lifecycle to play nice with symfony without hacking your user entities lifecycle (especially when you want to use doctrine for the user entity).

aschempp: I think we resolved the entity issue in the Mumble call. The main issue would be that a Contao model restored from the session would not be in the registry, potentially leading to duplicate entities. Same applies in Doctrine and the UnitOfWork. However, Contao users are **not** entities, so that issue actually does not apply.

backbone87:
I would still deprecate the `Contao\User`, `Contao\FrontendUser` and `Contao\BackendUser`

If you want to check permissons -> use the `AuthorizationCheckerInterface`
if you want to access data -> user `Contao\UserModel` or `Contao\MemberModel`

src/Resources/contao/modules/ModuleLogin.php
```diff
@@ -195,11 +128,27 @@ protected function compile()
...
+			$this->Template->targetPath = $objTarget->getAbsoluteUrl();
+		}
+
 		$this->Template->username = $GLOBALS['TL_LANG']['MSC']['username'];
 		$this->Template->password = $GLOBALS['TL_LANG']['MSC']['password'][0];
-		$this->Template->action = ampersand(\Environment::get('indexFreeRequest'));
+		$this->Template->action = $router->generate('contao_frontend_login');
 		$this->Template->slabel = \StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['login']);
-		$this->Template->value = \StringUtil::specialchars(\Input::post('username'));
+		$this->Template->value = $session->get(Security::LAST_USERNAME);
```
backbone87: missing encoding?

src/Resources/contao/classes/BackendUser.php
aschempp: I'm pretty sure the issue is serialization, and we have no reasonable way to boot the framework before unserializing the token. Therefore I think we should extend `\Contao\User` to have the class loaded by the Composer autoloader.

Regarding the `instanceof` check, that might be a general issue. Because checking for `$foobar instanceof \FrontendModule` would fail as well, everywhere in Contao. @backbone87 please open a separate issue for that.

backbone87:
But this PR heavily updates and reimplements the contao user classes insteadof leaving them aside and introduce own classes.

Since this PR just deals with authentication, the correct way would be to deprecate all authentication related methods in the contao user classes and provide new classes to interact with the symfony authentication.

After that the authorization part can be addressed and then we hopefully have almost all methods in the contao user classes deprecated and therefore deprecate the whole classes.

bytehead:
> the correct way would be to deprecate all authentication related methods in the contao user classes

They are already deprecated.

backbone87:
But you still use the contao user classes in the symfony authentication. That means the authentication is still coupled to those classes.

src/Resources/contao/classes/Frontend.php
bytehead: It's now checked on request with the `TokenLifetimeListener` implemented in 4c78db7

src/Resources/contao/controllers/BackendMain.php
bytehead: Changed in e2bed632d.

src/Event/CheckCredentialsEvent.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Gives a vote for the event.
+     *
+     * @param bool $vote
+     */
+    public function vote(bool $vote): void
+    {
+        $this->vote = $this->vote || $vote;
```
leofeyer: This means that once a listener has voted `true` it will remain `true`. Is this correct and intended?

bytehead: Yes, I think so. [This is](https://github.com/contao/core-bundle/blob/develop/src/Resources/contao/library/Contao/User.php#L432-L445) what the actual behaviour of the hook looks like.

aschempp: Why did we add events in the first place?

src/Resources/contao/classes/BackendUser.php
```diff
@@ -111,6 +123,32 @@ protected function __construct()
 		$this->strHash = \Input::cookie($this->strCookie);
 	}
 
+	public static function getInstance()
+	{
+		/** @var TokenInterface $token */
```
leofeyer: Shouldn't there be a `if (static::$objInstance !== null)` check somewhere? This code will be run upon every call instead of just upon the first call. Intended?

src/Resources/contao/classes/BackendUser.php
```diff
@@ -111,6 +123,32 @@ protected function __construct()
...
+		if ($token !== null && is_a($token->getUser(), static::class))
+		{
+			return $token->getUser();
+		}
+
+		/** @var SessionInterface $session */
+		$session = \System::getContainer()->get('session');
+
+		// Try to load possibly authenticated BackendUser from session
+		if ($session->has(self::SECURITY_SESSION_KEY) && $token = unserialize($session->get(self::SECURITY_SESSION_KEY)))
```
leofeyer: Maybe we should add `array('allowed_classes' => TokenInterface::class)` as second argument for `unserialize()`?

src/Event/CheckCredentialsEvent.php
bytehead: To get rid of the old hooks?

src/Resources/contao/classes/Frontend.php
```diff
@@ -685,4 +690,53 @@ public static function getResponseFromCache()
...
+	 * @return bool
+	 *
+	 * @deprecated Deprecated since Contao 4.x, to be removed in Contao 5.0.
+	 */
+	protected function getAuthenticationStatus($sessionKey = '')
+	{
+		/** @var SessionInterface $session */
+		$session = \System::getContainer()->get('session');
+
+		if ($session->has($sessionKey) && $token = unserialize($session->get($sessionKey)))
```
leofeyer: Again `array('allowed_classes' => TokenInterface::class)` could be added.

src/Resources/contao/classes/BackendUser.php
bytehead: Good point, should be checked then.

src/Resources/contao/classes/BackendUser.php
bytehead: 👍

src/Resources/contao/classes/Frontend.php
bytehead: 👍

src/Event/CheckCredentialsEvent.php
aschempp: Well we need to continue the exact same hook signature or it won't be BC. If we're not keeping that, we don't really need the hooks at all, because people then should simply create their own firewall listener?

src/Resources/contao/classes/BackendUser.php
aschempp: Be aware that there can be other classes. The user in the token is serialized as well. And you never know what else the token contains (e.g. a `FacebookToken` could have other dependencies).

src/Event/CheckCredentialsEvent.php
bytehead: It's the same signature:
`public function __construct(string $username, string $credentials, User $user)`
> The checkCredentials hook is triggered when a login attempt fails due to a wrong password. It passes the username and password as well as the user object as arguments and expects a boolean as return value.

src/Resources/contao/classes/BackendUser.php
bytehead: But that as well should implement the `TokenInterface`, or do I miss something here?

src/Resources/contao/controllers/BackendMain.php
```diff
@@ -59,13 +61,7 @@ public function __construct()
...
 		if ($this->User->pwChange)
 		{
-			$objSession = $this->Database->prepare("SELECT su FROM tl_session WHERE hash=?")
-										 ->execute($this->getSessionHash('BE_USER_AUTH'));
-
-			if (!$objSession->su)
-			{
-				$this->redirect('contao/password.php');
-			}
+			$this->redirect('contao/password.php');
```
leofeyer: We have to check for the admin role here, so the redirect is not executed if an admin impersonates another user.

src/Resources/contao/library/Contao/User.php
```diff
@@ -258,481 +279,475 @@ public function getData()
...
-		if ($this->checkAccountStatus() == false)
-		{
-			return false;
-		}
 
-		$blnNeedsRehash = true;
+	/**
+	 * {@inheritdoc}
+	 */
+	abstract public function getRoles();
```
leofeyer: This is a BC break. We must not alter the interfaces or abstract classes. We should make the method non-abstract and return a default value (probably an empty array?) instead.

bytehead: 🙈  true! Good catch!

src/Resources/contao/classes/BackendUser.php
bytehead: You can't provide interfaces nor parent classes for `allowed_classes`.

src/Resources/contao/controllers/BackendMain.php
```diff
@@ -223,14 +223,13 @@ protected function output()
...
-		$logout = $GLOBALS['TL_LANG']['MSC']['logoutBT'];
-		$logoutLink = $router->generate('contao_backend_logout');
-
+		// Generate the logout button
 		if ($authorizationChecker->isGranted('ROLE_PREVIOUS_ADMIN')) {
-			$logout = $GLOBALS['TL_LANG']['MSC']['exitUserBT'];
-			$logoutLink = $router->generate('contao_backend', [
-				'_switch_user' => '_exit',
-			]);
+			$logout = $GLOBALS['TL_LANG']['MSC']['backBT'];
```
bytehead: CS ;-)

src/Resources/contao/classes/BackendUser.php
aschempp: > But that as well has to implement the TokenInterface, or do I miss something here?

A user can be a `UserInterface` or what do you mean?

src/Event/CheckCredentialsEvent.php
aschempp: what are the event object used for then? We have not added object for any other hook from Contao 3. Sorry if I ask dump questions, I just saw the note by @leofeyer and haven't read through all the code.

src/Resources/contao/classes/FrontendUser.php
```diff
@@ -167,57 +190,30 @@ public function authenticate()
...
-
-			$this->createdOn = $time;
-			$this->autologin = $strToken;
-			$this->save();
-
-			$this->setCookie('FE_AUTO_LOGIN', $strToken, ($time + \Config::get('autologin')), null, null, \Environment::get('ssl'), true);
-		}
-
-		return true;
+		return false;
```
leofeyer: We are returning `parent::logout()` in the `logout()` method but `false` here. Should we return `parent::login()` instead?

bytehead: Yes please!

src/Event/CheckCredentialsEvent.php
bytehead: You can add an EventListener for that event and vote on the event object.

src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -0,0 +1,121 @@
...
+
+        if (!$request->hasSession()) {
+            return;
+        }
+
+        try {
+            /** @var FrontendUser $user */
+            $user = $this->userProvider->loadUserByUsername($username);
+        } catch (UsernameNotFoundException $e) {
+            $this->logger->info(
```
leofeyer: Are we sure that we want to log this? Could lead to a large number of irrelevant entries, couldn't it?

bytehead: It's only on frontend preview, if you try to login a frontend user that not exists.

src/Event/CheckCredentialsEvent.php
aschempp: Which would be wrong, because we should use security voters if we want to have votes? Do you intend to add a new feature or keep BC only?

bytehead: Security voters are still in a far distance. And stuff like this is mostly for BC and to have easy adaptions in the next releases. We can deprecate this, as soon as we have a working solution to work with Security voters.

src/Security/UserChecker.php
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    /**
+     * Check wether login is allowed (front end only).
+     *
+     * @param User $user
+     */
+    protected function checkIfLoginIsAllowed(User $user): void
+    {
+        if ($user instanceof FrontendUser && false === $user->login) {
```
leofeyer: Are you sure that the strict type check works here? `$user->login` is a `char(1)` column which is either `''` or `'1'` unless casted to boolean.

bytehead: Good catch! 👍

src/Resources/config/listener.yml
```diff
@@ -121,6 +121,33 @@ services:
...
+            - { name: kernel.event_listener, event: security.interactive_login, method: onInteractiveLogin }
+
+    contao.listener.security.switch_user:
+        class: Contao\CoreBundle\EventListener\SwitchUserListener
+        arguments:
+            - "@security.token_storage"
+            - "@logger"
+        tags:
+            - { name: kernel.event_listener, event: security.switch_user, method: onSwitchUser }
+            - { name: monolog.logger, channel: contao }
```
leofeyer: Is this a left-over tag?

bytehead: Yes, it's removed now.

src/Resources/contao/library/Contao/User.php
```diff
@@ -258,481 +282,466 @@ public function getData()
...
+		$user = new static();
 
-		// HOOK: pass credentials to callback functions
-		if (!$blnAuthenticated && isset($GLOBALS['TL_HOOKS']['checkCredentials']) && \is_array($GLOBALS['TL_HOOKS']['checkCredentials']))
+		// Load the user object
+		if ($user->findBy('username', $username) === false)
 		{
-			foreach ($GLOBALS['TL_HOOKS']['checkCredentials'] as $callback)
+			// Return if its not a real login attempt
+			if ($password === null && !$request->isMethod(Request::METHOD_POST))
```
leofeyer: What's the purpose of this line? What do you mean by "not a real login attempt"?

bytehead: `loadUserByUsername` will be called even on `refreshUser`, but the following logic only needs to be executed on a real login and not on a refresh.

src/Resources/contao/dca/tl_member.php
```diff
@@ -617,42 +612,6 @@ public function storeDateAdded($dc)
 	}
 
 
-	/**
-	 * Check whether the user session should be removed
-	 *
-	 * @param DataContainer $dc
-	 */
-	public function checkRemoveSession($dc)
```
bytehead: No BC break if you remove a public method?

src/Resources/views/Backend/switch_user.html.twig
```diff
@@ -1 +0,0 @@
-<a href="{{ url }}" title="{{ title }}">{{ image | raw }}</a>
```
bytehead: ?

src/Security/User/SwitchUserButtonGenerator.php
```diff
@@ -95,17 +88,12 @@ public function generateSwitchUserButton(array $row, string $href, string $label
...
-        ]);
+        $url = $this->router->generate('contao_backend', ['_switch_user' => $user->username]);
+        $title = StringUtil::specialchars($title);
 
-        return $this->twig->render('@ContaoCore/Backend/switch_user.html.twig', [
-            'url' => $url,
-            'title' => $title,
-            'image' => Image::getHtml($icon, $label),
-        ]);
+        return sprintf('<a href="%s" title="%s">%s</a>', $url, $title, Image::getHtml($icon, $label));
```
bytehead: I'm totally against to remove the template. Now we have again html somewhere in the logic.

leofeyer:
The PR is almost ready to merge. There are two open issues:

1. Do we want to log impersonations?
2. Do we need to change any `private` variables or methods to `protected` so people can overwrite certain things?

Regarding 1.: We are logging "switch user" actions but not "log in as member" actions. Is this ok or should we log both or none?

bytehead:
1. I would like to have impersonations logged. I'm fine to log frontend impersonations as well.
2. I'm against to make just everything private. Makes it unnecessary hard to extend existing stuff.

src/Resources/contao/controllers/BackendSwitch.php
```diff
@@ -106,25 +108,30 @@ public function run()
...
-						// Insert the new session
-						$this->Database->prepare("INSERT INTO tl_session (pid, tstamp, name, sessionID, ip, hash) VALUES (?, ?, ?, ?, ?, ?)")
-									   ->execute($objUser->id, $time, 'FE_USER_AUTH', \System::getContainer()->get('session')->getId(), \Environment::get('ip'), $strHash);
-
-						// Set the cookie
-						$this->setCookie('FE_USER_AUTH', $strHash, ($time + \Config::get('sessionTimeout')), null, null, \Environment::get('ssl'), true);
-						$objTemplate->user = \Input::post('user');
+						/** @var User $user */
+						$user = $token->getUser();
+						$objUser = \MemberModel::findByUsername($user->getUsername());
```
leofeyer: This is also something that makes me wonder. Shouldn't `$token->getUser()` return the user object already? Why are we loading the member model?

src/Resources/contao/dca/tl_member.php
leofeyer: It is a BC break for sure. The question is whether out BC promise includes DCA files, which I think they don't.

src/Resources/contao/templates/modules/mod_login.html5
```diff
@@ -13,16 +13,19 @@
       <?php endif; ?>
       <input type="hidden" name="FORM_SUBMIT" value="<?= $this->formId ?>">
       <input type="hidden" name="REQUEST_TOKEN" value="{{request_token}}">
+      <?php if ($this->targetPath): ?>
+        <input type="hidden" name="<?= $this->targetName ?>" value="<?= $this->targetPath ?>">
```
leofeyer: I also wonder if this is prone to manipulation? @ausi /cc

src/Resources/contao/controllers/BackendSwitch.php
bytehead: The `getUser()` on the token can have a fully loaded user instance, but you cannot be sure.
Normally it holds only the unserialized data from the session storage.

src/Resources/contao/templates/modules/mod_login.html5
bytehead: Maybe we should check, if it's an existing page from the installation?

leofeyer: I don't know. It seems it is a hidden field in Symfony, too: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/SecurityBundle/Tests/Functional/Bundle/CsrfFormLoginBundle/Form/UserLoginType.php#L47

aschempp: well you can only manipulate it through the web inspector. Which means the redirect target after login is manipulated. Can't really see an issue in that, it does not send any data to that URL.

ausi: `htmlspecialchars()` is missing for `$this->targetPath`, either in the template or in `ModuleLogin`. As it is now it’s vulnerable to XSS.

leofeyer: Fixed in 975070ea3adeb6f202ef55bdf280654a6663a353.

leofeyer:
Thank you, big time @bytehead.

PR #689, opened 2017-01-24
[RTM] Fix the search rebuilding cors headers

Toflar:


src/Resources/config/cors.yml
```diff
@@ -0,0 +1,3 @@
+nelmio_cors:
+    paths:
+        '/': # Nonsense configuration as the nelmio compiler pass will abort if not present
```
leofeyer: Why don't we use `'^/'` as in the [example configuration](https://github.com/nelmio/NelmioCorsBundle#configuration)?

Toflar: It's completely irrelevant. I'd rather not use anything at all.

leofeyer:
I would prefer to have the plugin related changes in a separate PR (there is contao/manager-bundle#14 already).

Toflar:
Sorry but I don't find the energy to split it up again. The other PR is closed for a reason.

leofeyer:
You told me not to mix multiple features in one pull request. 😄

leofeyer:
But I can split the PR myself, of course.

Toflar:
Yes, which is not the case here. Absolutely all changes are related to the very same bug fix.

leofeyer:
Not really, because we are not yet storing the manager configuration in the core bundle. This is a feature which is yet to be merged and not to be introduces half-way in a PR.

Toflar:
I disagree. But I don't have the time to discuss this, feel free to split it into multiple PR's.

aschempp:
Not true, we already do for security config.

leofeyer:
@aschempp Which is wrong and has been changed in the `develop` branch.

src/Resources/config/cors.yml
leofeyer: I'd go for `'^/'` as we are applying the logic to all routes, aren't we?

leofeyer: I have created a PR to fix this: https://github.com/nelmio/NelmioCorsBundle/pull/73

aschempp:
why was this merged in develop instead of the hotfix branch?

leofeyer:
Originally as basis for the other PR, which then has been closed. It has a "needs backport" label though.

leofeyer:
Back ported in 002a7a302c6b97d877f8004a30f0c7dba6693df7.

PR #692, opened 2017-01-28
[RTM] Fixed call to count() method on page objects

aschempp:
Passing a `PageModel` object is valid for the method, but the `count()` method does not exist.


leofeyer:
Do we have the same issue in Contao 3?

aschempp:
Well not really, because the method is not useable from outside and therefore the database query will always return a collection. That's also how I worked around it for now, but that's rather stupid:

```php
$page = PageModel::findBy('id', 15, ['return' => 'Collection']);
```

😉

PR #695, opened 2017-01-31
[RTM] Fixed issues where foreach was called on strings

aschempp:


leofeyer:
Do we need to back port the changes to Contao 3.5?

aschempp:
I guess so, yes

Issue #696, opened 2017-01-31
Bug : Download in news reader


antoinedavid:
Hi,

We can't use content download or content downloads in a news because the file are corrupt when downloading.

A part of the template are injected in the header and we don't have the end part of the file.

I found this issue on the demo site too. I've not test on Contao 4.0 but i've the same issue since Contao 4.1

Header of a PDF file for exemple:

```pdf

<div class="layout_full block" itemscope itemtype="http://schema.org/Article">

  <h1 itemprop="name">Hello world</h1>

      <p class="info"><time datetime="2017-01-30T17:27:00+01:00" itemprop="datePublished">2017-01-30 17:27</time> par Admin  </p>
  
  
      %PDF-1.5
%âãÏÓ
10 0 obj
<</Linearized 1/L 134289/O 12/E 129957/N 1/T 133984/H [ 777 289]>>
endobj
               
90 0 obj
<</DecodeParms<</Columns 5/Predictor 12>>/Filter/FlateDecode/ID[<E182E5868FCF0E01A64FE1F579ECD28B><BC0DE23470124040AC4DE59087625AF3>]/Index[10 135]/Info 9 0 R/Length 209/Prev 133985/Root 11 0 R/Size 145/Type/XRef/W[1 3 1]>>stream
hÞbbd```b``‘‘~ ’qˆä¬‘<A$×S°H˜¼	"ÙÝ@$·ˆäð³Áê¹[@¤À°9µ`2DƒH¦n°-ý@’1ì'Xl;ã™‘ ‹$Ho0»Lšƒõž³ÁzçÉû`RL
!±1eÑU2ÌER#„ ™„ÐEzÀì{`R˜!âßÍ›€ä&°	í@øÓ^dóÙH$"ÿY˜€±qã(I2ùŸùÏ€  Øã)
endstream
endobj
startxref
0
%%EOF...............
```
Best

asaage:
I can confirm.
Fortunately news-attachment is not affected.

leofeyer:
Which Contao version does this issue occur in?

asaage:
Whatever version the official-demo is running. i did't test 3.5.

antoinedavid:
@leofeyer C 4.1.x 4.2.x 4.3.x
I don't know for C 4.0.x, 
I don't have the on C 3.5.x

antoinedavid:
i've the solution

We need too check if no output buffer is active:
In library/Contao/File.php

```php
/**
[...]

PR #698, opened 2017-02-01
Make sure no output buffer is active before download

antoinedavid:
#696

### DIFF
--- a/src/Resources/contao/library/Contao/File.php
+++ b/src/Resources/contao/library/Contao/File.php
@@ -775,7 +775,11 @@ class File extends \System
 	 */
 	public function sendToBrowser($filename='')
 	{
-		$response = new BinaryFileResponse(TL_ROOT . '/' . $this->strFile);
+        // Make sure no output buffer is active
+		// @see http://ch2.php.net/manual/en/function.fpassthru.php#74080
+		while (@ob_end_clean());
+
+	    $response = new BinaryFileResponse(TL_ROOT . '/' . $this->strFile);
 
 		$response->setContentDisposition
 		(


leofeyer:
@contao/developers Do we also need to add this?

```php
		// Prevent session locking (see #2804)
		session_write_close();

		// Disable zlib.output_compression (see #6717)
		ini_set('zlib.output_compression', 'Off');
```

See [Contao/File.php#L802-L806](https://github.com/contao/core/blob/master/system/modules/core/library/Contao/File.php#L802-L806)

ausi:
I don’t think so. I’m also not sure if this is the right place for `ob_end_clean()` because we are throwing a `ResponseException` and do not output anything. Shouldn’t this be handled in the code that actually sends the output to the browser?

leofeyer:
It probably should, but it is not. :confused:

ausi:
> It probably should, but it is not. 😕

Why don’t we fix it there then?

Toflar:
This fix is not correct in my opinion. The Symfony `Response` class closes buffers on `send()`.

Toflar:
@antoinedavid You have to debug if the kernel does correctly handle our `ExceptionListener` that transforms the `ResponseException` to a `Response` and if your entry point (`app.php`) actually does call `$response->send()` on the correct response instance.

antoinedavid:
> I don’t think so. I’m also not sure if this is the right place for ob_end_clean() because we are throwing a ResponseException and do not output anything. Shouldn’t this be handled in the code that actually sends the output to the browser?

In the detail of a news item the output is not empty when `sendToBrowser()` is executed. if you make a simple `dump() exit;` you will see you'r news detail before your dump().

I think `ob_end_clean()` is a temporary solution i think the issue is more complex. Unfortunately I do not have time at the moment to deepen my research.

Toflar:
I tracked this down. So effectively Symfony makes sure, output buffers are closed AFTER the file is sent. We need to flush before we send it because we already have output buffers active. So turns out @antoinedavid's solution was correct, we should just reuse Symfony methods for that.
Here you go:

```diff
diff --git a/src/Resources/contao/library/Contao/File.php b/src/Resources/contao/library/Contao/File.php
index 7d150bf..8a87b79 100644
--- a/src/Resources/contao/library/Contao/File.php
+++ b/src/Resources/contao/library/Contao/File.php
@@ -14,6 +14,7 @@ use Contao\CoreBundle\Exception\ResponseException;
 use Contao\Image\Image as ContaoImage;
 use Contao\Image\ImageDimensions;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 
 
@@ -775,6 +776,9 @@ class File extends \System
 	 */
 	public function sendToBrowser($filename='')
 	{
+		// Close previous output buffers but do not flush
+		Response::closeOutputBuffers(0, false);
+
 		$response = new BinaryFileResponse(TL_ROOT . '/' . $this->strFile);
 
 		$response->setContentDisposition
```

ausi:
I’m still not sure if closing the buffers belongs to this method. What if I want to catch the `ResponseException` in my application and do something else with it? The buffers shouldn’t get closed then.

Where does the output buffer get opened? Shouldn’t it be handled there? Like so:

```php
ob_start();
try {
	// some code that might throw exceptions
}
finally {
	ob_end_clean();
}
```

Toflar:
> Where does the output buffer get opened? Shouldn’t it be handled there? Like so:

In any template...

leofeyer:
> Where does the output buffer get opened?

Everywhere where templates are rendered:

 * https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/InsertTags.php#L914
 * https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/TemplateInheritance.php#L181

ausi:
> Everywhere where templates are rendered

I think we should add `try-finally` blocks in these places. This would fix the output buffer for all types exceptions.

leofeyer:
As briefly discussed, I have added a workaround in fd15a93be4da0425050bf8879f4f2d8bc0d321f8 and @ausi will work on a "real" solution for Contao 4.4.

leofeyer:
See #714. Thank you @ausi.

PR #699, opened 2017-02-02
[RTM] Allow to pass objects in callback arrays

aschempp:
This simple change should get us closer to actually callables.

You can now use `[$this, 'foobar']` for a callback. 

**Example for current:**
```php
$GLOBALS['TL_DCA']['tl_example']['config']['onload_callback'][] = function () {
    $this->onLoad();
};
```

**Example for new:**
```php
$GLOBALS['TL_DCA']['tl_example']['config']['onload_callback'][] = [$this, 'onLoad'];
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.4.0-beta1 (2017-XX-XX)
 
+ * Support using objects in callback arrays (see #699).
  * Support importing form field options from a CSV file (see #444).
  * Add a Doctrine DBAL field type for UUIDs (see #415).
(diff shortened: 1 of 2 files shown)


discordier:
This improves the current style, however I advise anyone against using even callables in dca and Contao 4.
The devs should migrate all HOOKs and callbacks already to services for allowing easy migration in the future when we can have a "built container" (sometime in the future) and therefore this will get removed for Contao 5 again.

Therefore I am undecided if we should rather add this feature and discourage usage in the same time.

aschempp:
This is to support in the Contao core, which is not being migrated to events…

PR #703, opened 2017-02-10
[RTM] Use the selected template for custom section

agoat:
Currently if a section is insert in the page template with `$this->section('name');` the custom sections template selected in the layout will not be used. 

This PR extends the **section** method of the **FrontentTemplate** class to use use the custom sections template instead of just setting the template to `block_section`.


discordier:
General note about pull requests, please always base them on vanilla master from contao organization.
This PR is now polluted with merge commits where you merged vanilla master into your fork.

agoat:
Oh, sorry.
I think I have not yet fully understood the system of forks and commits.

I can't fork the contao/core-bundle twice to make a new branche from vanilla master. How can I get the vanilla master into my fork??

agoat:
Ok, thank you.
Looks like a new branch can't be created from the origin fork master on the web page. But with the client or on the console.

Should I crate a new PR from a new (clean) branch or can this on still be used?

discordier:
You can create a new branch locally and force push over your branch then.
This will update the PR.

leofeyer:
If you insert a section via `$this->section('name')`, you can pass the template as second argument, so e.g. `$this->section('name', 'block_custom');`. I don't know if inheriting the parent template is a good idea.

agoat:
> I don't know if inheriting the parent template is a good idea.

I don't understand. 

As you may see, if a template is passed as second argument it will be used. But it not, the template, wich is selected in the custom layouts section, will be used. Otherwise the template setting in the custom layout sections will be useless if the position is set to manual output.

leofeyer:
Well, to me "manual output" implies that the "Template" drop-down is ignored. But I understand that you see it differently.

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -172,7 +172,13 @@ public function section($key, $template=null)
 
 		if ($template === null)
 		{
-			$template = 'block_section';
```
leofeyer: Removing this line is wrong. If the foreach loop does not find a template, we'd end up with `null`.

agoat: Usually every position should have at least `block_section`. But I agree, to be save we shouldn't remove it.

leofeyer:
@ausi Does this make sense?

ausi:
I think it does make sense. But I don’t see why the `sections()` method would need the code snippet too, it handles the user-selected templates already in the `block_sections` template.

leofeyer:
You are right.

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -173,6 +173,14 @@ public function section($key, $template=null)
 		if ($template === null)
 		{
 			$template = 'block_section';
+			
+			foreach ($this->positions as $position)
+			{
+				if (isset($position[$key]))
```
leofeyer: Shouldn't we check `isset($position[$key]['template'])` here?

agoat: Normally the [$template] value should exist for sure, because in the dataset from the section wizard always save title, id , template and position. 
But now we are safe.

leofeyer:
Thank you @agoat.

PR #708, opened 2017-02-15
[RTM] Allow to disable input encoding for a whole dca

Toflar:
Here's another feature for Contao 4.4.

For my own modules I want to be able to disable the input encoding all together and output stuff using twig or encode it myself. I have an application which even doesn't output anything at all but manages content that's fetched through an API. I have to set the `eval` attributes on every field which is a bit tedious. I think it can be very helpful to disable this completely and prepare for the future already in your own bundles/applications :-)

### DIFF
--- a/src/Resources/contao/classes/DataContainer.php
+++ b/src/Resources/contao/classes/DataContainer.php
@@ -273,6 +273,12 @@ abstract class DataContainer extends \Backend
 			$this->varValue = \StringUtil::insertTagToSrc($this->varValue);
 		}
 
+		// Use raw request if set globally
+		if (isset($GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData']) && $GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData'] === true)
+		{
+			$arrData['eval']['useRawRequestData'] = true;
+		}
+
 		/** @var Widget $objWidget */
 		$objWidget = new $strClass($strClass::getAttributesFromDca($arrData, $this->strInputName, $this->varValue, $this->strField, $this->strTable, $this));
 
--- a/src/Resources/contao/classes/DataContainer.php
+++ b/src/Resources/contao/classes/DataContainer.php
@@ -273,8 +273,11 @@ abstract class DataContainer extends \Backend
 			$this->varValue = \StringUtil::insertTagToSrc($this->varValue);
 		}
 
-		// Use raw request if set globally
-		if (isset($GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData']) && $GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData'] === true)
+		// Use raw request if set globally but allow opting out setting useRawRequestData to false explicitly
+		$useRawGlobally = isset($GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData']) && $GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData'] === true;
+		$notRawForField = isset($arrData['eval']['useRawRequestData']) && $arrData['eval']['useRawRequestData'] === false;
+
+		if ($useRawGlobally && !$notRawForField)
 		{
 			$arrData['eval']['useRawRequestData'] = true;
 		}
--- a/src/Resources/contao/library/Contao/User.php
+++ b/src/Resources/contao/library/Contao/User.php
@@ -10,6 +10,7 @@
 
 namespace Contao;
 
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Http\Session\SessionAuthenticationStrategy;
 
 
@@ -320,6 +321,8 @@ abstract class User extends \System
 	 */
 	public function login()
 	{
+		/** @var Request $request */
+		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
 		\System::loadLanguageFile('default');
 
 		// Do not continue if username or password are missing
@@ -339,7 +342,7 @@ abstract class User extends \System
 				foreach ($GLOBALS['TL_HOOKS']['importUser'] as $callback)
 				{
 					$this->import($callback[0], 'objImport', true);
-					$blnLoaded = $this->objImport->{$callback[1]}(\Input::post('username', true), \Input::postUnsafeRaw('password'), $this->strTable);
+					$blnLoaded = $this->objImport->{$callback[1]}(\Input::post('username', true), $request->request->get('password'), $this->strTable);
 
 					// Load successfull
 					if ($blnLoaded === true)
@@ -399,17 +402,17 @@ abstract class User extends \System
 		// The password has been generated with crypt()
 		if (\Encryption::test($this->password))
 		{
-			$blnAuthenticated = \Encryption::verify(\Input::postUnsafeRaw('password'), $this->password);
+			$blnAuthenticated = \Encryption::verify($request->request->get('password'), $this->password);
 		}
 		else
 		{
 			list($strPassword, $strSalt) = explode(':', $this->password);
-			$blnAuthenticated = ($strSalt == '') ? ($strPassword === sha1(\Input::postUnsafeRaw('password'))) : ($strPassword === sha1($strSalt . \Input::postUnsafeRaw('password')));
+			$blnAuthenticated = ($strSalt == '') ? ($strPassword === sha1($request->request->get('password'))) : ($strPassword === sha1($strSalt . $request->request->get('password')));
 
 			// Store a SHA-512 encrpyted version of the password
 			if ($blnAuthenticated)
 			{
-				$this->password = \Encryption::hash(\Input::postUnsafeRaw('password'));
+				$this->password = \Encryption::hash($request->request->get('password'));
 			}
 		}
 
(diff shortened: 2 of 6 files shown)


aschempp:
Input would still be encoded, right? I mean script tags and null bytes stripped and other "safety checks". We would need to use the request class instead of the input class if the flag is enabled, right?

I mean I'm totally fine with the feature, but the name is a bit confusing. It would better be something like `allowHtmlForAll` 😂

Toflar:
It will use `Input::postRaw()` with this config.

leofeyer:
`Input::postRaw()` applies the following filters:

* `Input::preserveBasicEntities()`
* `Input::xssClean()`
* `Input::encodeInsertTags()` (FE only)

I guess that's why @aschempp suggested to use the Symfony Request class in this case, which will not alter the input values at all.

Toflar:
Ah, I see. Yeah, so we already have our own use case for https://github.com/contao/core-bundle/pull/674/files 😄

aschempp:
You mean the core would apply that if the flag is set? 😂

Toflar:
I don't have another option I guess :) So we need both, a new `config` and a new `eval` flag. How would you like to name it? What about `disableInputEncoding` is not fine? It's pretty much stating what it does :-)

aschempp:
It depends on the implementation. If it means flag all fields with `decodeEntities` or thelike I'm fine with that. But if it means we use Symfony Request I would prefer something else, like `useSymfonyRequest`

Toflar:
What do the others think? /cc @contao/developers

discordier:
I'd do some flag on dca level and forward this to the Widget class.
In the widget class, we check the flag and it will then use the raw value from the request.

I think we could combine this with my changes from #575, sadly I still did not come around to shift the remaining methods from `Input` over to `StringUtils` so it would be RTM.

ausi:
Looking at this, I thought we must already have such a flag for the password fields in the backend, but it looks like we haven’t. If I change my password in the backend to `foobar\0baz` I have to login with `foobarbaz`. So this would be another use case for this feature.

I think the name `disableInputEncoding` is fine. But I’m not sure if we should use the Symfony request or `Input::postUnsafeRaw()` instead.

discordier:
Request, definately!

Toflar:
Yeah, request for sure but why should I name the `eval` option `useSymfonyRequest`? It's just the regular, unmodified value coming from the current request. Whether it is a Symfony subrequest or not does not matter, it's just representing the current request value. So I'm still all for `disableInputEncoding`.

bytehead:
I'm fine with `disableInputEncoding`.

discordier:
So am I.

Toflar:
Okay, but I don't want to rebase all the time. I'll only finish once @leofeyer merged https://github.com/contao/core-bundle/pull/674.

leofeyer:
As discussed in Mumble on March 23rd, the flag shall be called `bypassInputEncoding`.

Toflar:
Note to self: Check password fields (they should not strip null bytes etc.)

Toflar:
PR is updated and ready for review.

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -263,6 +265,14 @@ public function __set($strKey, $varValue)
 				$this->strPrefix = $varValue;
 				break;
 
+			case 'useRawRequestData':
```
leofeyer: Since you have declared the property as a `boolean`, we should handle the `false` case as well, shouldn't we?

Toflar: Yeah not all attributes do unfortunately. So it's inconsistent already anyway but I added it.

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -263,6 +265,22 @@ public function __set($strKey, $varValue)
...
+				if ($varValue === true)
+				{
+					/** @var Request $request */
+					$request = \System::getContainer()->get('request_stack')->getCurrentRequest();
+					$this->setInputCallback(function() use ($request) {
```
leofeyer: You could (but don't have to) simplify this:

```php
$this->setInputCallback(function () {
    return \System::getContainer()->get('request_stack')->getCurrentRequest()->request->get($this->name);
});
```

src/Resources/contao/classes/DataContainer.php
```diff
@@ -273,6 +273,12 @@ protected function row($strPalette=null)
 			$this->varValue = \StringUtil::insertTagToSrc($this->varValue);
 		}
 
+		// Use raw request if set globally
+		if (isset($GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData']) && $GLOBALS['TL_DCA'][$this->strTable]['config']['useRawRequestData'] === true)
```
ausi: Should we add a `!isset($arrData['eval']['useRawRequestData'])` check here too? This would make it possible to use the raw request for the whole DCA but opt-out of it for single fields.

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -263,6 +265,22 @@ public function __set($strKey, $varValue)
...
+						return $request->request->get($this->name);
```
ausi: It looks like `Widget::getPost()` supports array syntax like `field[foo][bar]`, do we need this functionality here too?

src/Resources/contao/classes/DataContainer.php
Toflar: Good idea, added in fd2ef3fdb60d6327cdf1b6741335ff7d990b6fac.

src/Resources/contao/library/Contao/Widget.php
Toflar: I don't think so. If you want the raw request data you need to handle the array yourself if you need to, no?

Toflar:
> he second line would overwrite the already set input callback.

Exactly. Which is the desired workflow. There can only ever be one responsible function fetching the input. If you handle `useRawRequestData` within `getPost()` the `setInputCallback()` method could result having no effect at all which is very confusing I find.

src/Resources/contao/library/Contao/Widget.php
ausi: I have no idea where this array syntax is needed, maybe @aschempp can help us here? See [Widget.php:781](https://github.com/contao/core-bundle/blob/4758be406700f7861bdddfb39e140de1e61a6b73/src/Resources/contao/library/Contao/Widget.php#L781)

ausi:
What about this example:

```php
$widget->useRawRequestData = true;
$widget->setInputCallback(null);
```

This would not use the raw request.

I think this is a confusing public API, if `useRawRequestData` would just be a simple flag, I could check for this flag in my own input callback too:

```php
$widget->setInputCallback(function() {
	if ($widget->useRawRequestData) {
		...
	}
});
```

Toflar:
> I think this is a confusing public API, if useRawRequestData would just be a simple flag, I could check for this flag in my own input callback too:

Which means that you can change the behavior someone else wanted. In my opinion both ways can be confusing. It's just the way it is.

ausi:
Isn’t this inconsistent? `allowHtml => Input::postHtml()` and `preserveTags => Input::postRaw()` are already two flags that `Widget::getPost()` handles, why should `useRawRequestData` behave differently than the existing flags?

Toflar:
No idea. Because it's cleaner? :) I changed it.

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -83,6 +84,7 @@
  * @property string                  $slabel            The submit button label
  * @property boolean                 $preserveTags      Preserve HTML tags
  * @property boolean                 $decodeEntities    Decode HTML entities
+ * @property boolean                 useRawRequestData  Use the raw request data from the Symfony request
```
ausi: Should we add `useRawRequestData` to the switch-case on [line 345](https://github.com/Toflar/core-bundle/blob/509805560c0c5753138fbd2cacad8cad21ed62a1/src/Resources/contao/library/Contao/Widget.php#L345) too?

Toflar: Jop! Added in 31732f9e599aa9d05291513390cbe73fa8438b5e.

src/Resources/contao/library/Contao/User.php
```diff
@@ -320,6 +321,8 @@ public function authenticate()
 	 */
 	public function login()
 	{
+		/** @var Request $request */
+		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
```
aschempp: This is a behavior change, shouldn't we use `getMasterRequest`?

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -793,6 +796,13 @@ public function validate()
 	 */
 	protected function getPost($strKey)
 	{
+		if ($this->useRawRequestData === true)
+		{
+			/** @var Request $request */
+			$request = \System::getContainer()->get('request_stack')->getCurrentRequest();
```
aschempp: Same here, I think we should use the master request?

src/Resources/contao/library/Contao/User.php
Toflar: I think it's almost always wrong to take the master request. The use cases are only very, very limited. So no: the current request is the way to go.

PR #712, opened 2017-02-24
Set up a default public folder on contao:install command

Toflar:
I think a lot of users forget that as of Contao 4, folders are protected by default.
And even if you know about it, I guess every project needs public files and I found myself creating the same file structure over and over again. I'm always using

* files
    * public
    * protected

While people might not like the "protected" folder, I think it is a good idea to create a "public" folder by default when setting up Contao. It makes setup faster and gives the users a hint about having to explicitly make folders public automatically.

### DIFF
--- a/src/Command/InstallCommand.php
+++ b/src/Command/InstallCommand.php
@@ -15,6 +15,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
 
 /**
  * Installs the required Contao directories.
@@ -97,6 +98,7 @@ class InstallCommand extends AbstractLockedCommand
         $this->webDir = rtrim($input->getArgument('target'), '/');
 
         $this->addEmptyDirs();
+        $this->setupUploadDir();
         $this->addIgnoredDirs();
 
         if (!empty($this->rows)) {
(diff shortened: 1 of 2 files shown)


aschempp:
I would delete this folder on every installation. Why would I create just one public and one protected folder? I create folders for their content and usually publish them…

If so, I would prefer a solution that folders are public by default like in Contao 3 (when added through the backend).

Toflar:
No protected foder is created? Public folders are public recursively. So it makes a lot of sense to just put everything that should be public in one parent folder named `public`. It's very explicit and I consider this to be best practice.

aschempp:
I absolutely don't. I use folders like `files/layout/…`, `files/header-images/`, `files/foobar`. Why would I want them to be grouped in a `public` folder?

Toflar:
The point is that you need to provide the **users** with some folders and I think it makes sense to add a `public` folder for them so they know exactly whatever is in this `public` folder, is going to be publicly available.

If you like to add a `layout` on the topmost level and make this public too, you can.

Anyway, I consider this to be best practice and I think it serves as a hint for administrators to see immediately that they have to make folders publicly available explicitly. That's why I want this to be part of the core. Let's see what others say.

asaage:
I am using a folder for each website-root as the first-level under files so i would probably delete **public** and **protected** after install.

Why not put a readme.txt file into files?
and may be change the harddisk-icon to something that indicates that protected is the default...
image

Sometimes the recursiveness bothers me but that was the case in contao 3 too where you could not place a public-folder inside a protected-folder.

ausi:
I found myself forgetting about the “protected by default” too, so +1 for the public folder. And for users that don’t want it, deleting a folder isn’t that much overhead I think.

fritzmg:
Personally I wouldn't want use it this way, since it creates an additional level in the URL for any public files (in additional subfolders).

Also just adding a `public` folder won't be enough I think. Users will just see that folder but not realize its significance, since there is no other folder to compare it to. If you had a `protected` folder as well, you would see a difference more clearly, since the `protected` folder will have an icon and the `public` folder will not.

Still, even then users just might assume it's the `protected` folder that has a special setting, not the `public` one.

Generally I never liked that approach in Contao 4 all that much, since you can't simply have public files in the `/files` directory now. They _must_ be in a subfolder to be accessible publicly. But that's besides the point ;)

aschempp:
Can't we solve most of the issues if folders are just public by default? That's like 90% of the use case. Can't remember if we show that checkbox when creating a folder?

fritzmg:
Yes, the checkbox is definitely there when creating a folder.

leofeyer:
> I would delete this folder on every installation.

Me too.

> Can't we solve most of the issues if folders are just public by default?

Yes, we could enable the checkbox by default.

leofeyer:
As discussed in Mumble on March 23rd, the "Public" checkbox by default. Also, the "Public" checkbox should be checked and disabled in the subfolders of a public folder.

asaage:
Instead of checking and disabling the "Public" checkbox in the subfolders of a public folder can't we introduce something without that strict recursiveness?

ausi:
@asaage I don’t think thats possible, because if a folder gets symlinked all subfolders are automatically public too.

leofeyer:
Implemented in b2b5a18e7011a7ff83f759726dd3b6dbb4ec1071.

PR #714, opened 2017-02-25
[RFC] Use try-finally blocks to close all opened output buffers

ausi:
As discussed in https://github.com/contao/core-bundle/pull/698#issuecomment-279708701

@antoinedavid can you verify that this solves your issue with `sendToBrowser()`?

### DIFF
--- a/src/Resources/contao/controllers/FrontendIndex.php
+++ b/src/Resources/contao/controllers/FrontendIndex.php
@@ -288,9 +288,17 @@ class FrontendIndex extends \Frontend
 					if (!method_exists($objHandler, 'getResponse'))
 					{
 						ob_start();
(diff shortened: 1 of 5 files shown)


Toflar:
Excellent work @ausi, thanks!

aschempp:
Why not simply use `ob_get_clean()`?

ausi:
> Why not simply use `ob_get_clean()`?

What do you mean by that? In the finally block we only want to close the output buffer, the contents of the buffer are not needed there.

aschempp:
Nevermind. Instead of `ob_get_content()` and `finally` we could have used `ob_get_clean()` and `catch`. But does not really make a difference I guess :-)

PR #719, opened 2017-03-07
Added symlink for TCPDF config file (backwards compatibility)

aschempp:
see https://github.com/isotope/core/issues/1763

### DIFF
--- a/src/Command/InstallCommand.php
+++ b/src/Command/InstallCommand.php
@@ -106,7 +106,7 @@ class InstallCommand extends AbstractLockedCommand
         }
 
         $this->addInitializePhp();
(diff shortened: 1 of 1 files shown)


leofeyer:
Why should we do this? TCPDF does not load the file from the old location anymore.

aschempp:
As the title says, it's for backwards compatibility reasons. Extensions for Contao 3 might load the configuration from this place, Isotope does for example. The file is not loaded by TCPDF, its loaded by Contao or an extension before using TCPDF.

leofeyer:
As discussed in Mumble on August 10th, 2017, we want to add this symlink and also add a redirect route for `cron/cron.php`.

PR #721, opened 2017-03-08
[RFC] Correctly parse Doctrine SQL array

aschempp:
Since Contao 4.3 we can set an array for the SQL definition in DCAs. However, the default value and other configs were not correctly detected/merged.

### DIFF
--- a/src/Resources/contao/library/Contao/Widget.php
+++ b/src/Resources/contao/library/Contao/Widget.php
@@ -1463,7 +1463,8 @@ abstract class Widget extends \Controller
 			{
 				$sql = $sql['columnDefinition'];
 			}
-			else {
+			else
+			{
 				if (isset($sql['default']))
 				{
 					return $sql['default'];
--- a/src/Resources/contao/library/Contao/Widget.php
+++ b/src/Resources/contao/library/Contao/Widget.php
@@ -1380,6 +1380,19 @@ abstract class Widget extends \Controller
 			}
 		}
 
+		if (is_array($arrAttributes['sql']) && !isset($arrAttributes['sql']['columnDefinition']))
+		{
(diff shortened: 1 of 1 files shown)


qzminski:
It should also support arrays here:

https://github.com/aschempp/contao-core-bundle/blob/a0c675b4741742100f112bfa714395289974ec8f/src/Resources/contao/library/Contao/Widget.php#L1525

aschempp:
Fixed in 5f9207f69536ee0750f7b7ca709726b244b5e02f

src/Resources/contao/library/Contao/Widget.php
```diff
@@ -1484,20 +1525,12 @@ public function getEmptyStringOrNull()
...
-		// Strip the field type definition
-		list(, $def) = explode(' ', $sql, 2);
-
-		if (strpos($def, 'NULL') !== false && strpos($def, 'NOT NULL') === false)
-		{
-			return null;
-		}
-
-		return '';
+		return static::getEmptyValueByFieldType($sql) === null ? null : '';
```
leofeyer: This change is wrong. The method comment says:

> Return either an empty string or null based on the SQL string

Therefore we cannot use `getEmptyValueByFieldType()` here!

aschempp:
getEmptyValueByFieldType return null, empty string or zero, and zero would be converted to empty string?

leofeyer:
Yes, there is an `getEmptyStringOrNull()` method using the `getEmptyStringOrNullByFieldType()` method internally. And even though I do not remember why we need this method, we must not change the API in a bugfix release.

aschempp:
Why do you think this is an API change? The method still exists, and it returns the same thing as before?

leofeyer:
:open_mouth: My bad.

PR #726, opened 2017-03-17
Contao 4.3.6

leofeyer:


CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # Contao core bundle change log
 
+### DEV
```
Toflar: Version missing

src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -1006,7 +1006,8 @@ public static function standardize($strString, $blnPreserveUppercase=false)
 		$strString = Utf8::toAscii($strString);
 		$strString = preg_replace($arrSearch, $arrReplace, $strString);
 
-		if (is_numeric(substr($strString, 0, 1)))
+		// Prefix numeric values (see #707)
+		if (is_numeric($strString))
```
Toflar: Imho this is a BC break because the `StringUtil` (as the name implies) is a general util class and belongs to the public API. This is a behavioural change which I'm not comfortable with. I would prefer adding a third argument to trigger this case explicitly and make use of it when generating page aliases and html id attributes.

CHANGELOG.md
leofeyer: This is correct and will be replaced by the Contao release helper.

src/Resources/contao/library/Contao/StringUtil.php
leofeyer: Changed in f14eaf261c32a5d2b09dfcd94be35fcaa38fbbdd.

leofeyer:
The release has been scheduled for March 22nd, 2017.

CHANGELOG.md
Toflar: Ah, good to know :)

src/Command/InstallCommand.php
```diff
@@ -188,8 +188,11 @@ private function addInitializePhp()
 /** @var Composer\Autoload\ClassLoader */
 $loader = require __DIR__ . '/../app/autoload.php';
 
+$request = Request::create('/_contao/initialize', 'GET', [], $_COOKIE, [], $_SERVER);
+$request->attributes->set('_scope', ('BE' === TL_MODE ? 'backend' : 'frontend'));
```
Toflar: How does this work? `TL_MODE` is never defined, is it?

leofeyer: Yes it is; in the custom entry point: https://github.com/isotope/core/blob/master/system/modules/isotope/group.php#L22

PR #729, opened 2017-03-20
[RTM] Improve the findByIdOrAlias() method

Toflar:


Toflar:
> tl_page::generateAlias
> tl_page::addAliasButton

How is this related to this pull request?




> Also, we should probably change the other bundles, as the same rules should apply to news alias etc. in my opinion.

They all share the `Model` class as parent so automatically make use of these changes?

aschempp:
> How is this related to this pull request?

You're right, I just quickly searched for `alias=?` 🙈 

> They all share the Model class as parent so automatically make use of these changes?

Unfortunately, that's only partially true: https://github.com/aschempp/contao-news-bundle/blob/825cc0992692ace816d213d5c8d66b51697b3c03/src/Resources/contao/models/NewsModel.php#L176-L176

Toflar:
Yeah sure, there are other methods that could be optimized.

leofeyer:
Improved the other methods in f45ced28a09db772fd245a80434381c8fb4d3482. @Toflar /cc

PR #730, opened 2017-03-22
[RTM] Fixed combined files broke when using declare() statements

Toflar:


src/Config/Loader/PhpFileLoader.php
```diff
@@ -52,7 +52,29 @@ public function load($file, $type = null)
...
         }
 
-        return rtrim($code)."\n";
+        // declare() statements
+        $lines = explode("\n", $code);
+        $codeNew = '';
+        $processed = false;
+
+        foreach ($lines as $line) {
+            if ('' === $line || $processed) {
```
discordier: IMO this is wrong as it does not take file doc comments and lines with trailing whitespace into account.
IMO it will break on files with:
```php
<?php

/**
 * This file is part of foobar/baz.
 *
 * (c) 2017 Acme.
 * [...]
 */

declare(strict_types=1);

// [...]
```

src/Config/Loader/PhpFileLoader.php
```diff
@@ -52,7 +52,29 @@ public function load($file, $type = null)
...
+                $codeNew .= $line . "\n";
+                continue;
+            }
+
+            // Until now, every line was empty which means this is the first
+            // one that's not so it must be the declare statement if it's there
+            // as this one has to be the first statement right after <?php
+            if (preg_match('/^declare\([^)]+\);$/', $line)) {
```
discordier: Won't this also clear a `declare(ticks=1);`, `declare(encoding='ISO-8859-1');` (and any other declare rule which might get specified in PHP 7.0+) if it should be the first statement (See [PHP.net](http://php.net/manual/de/control-structures.declare.php))?
IMO we should explicitely only capture strict_types and hence the regex should be:
```regex
/^declare\(\s*strict_types\s*=\s*[^)]+\);$/
```
Note that whitespace is allowed and therefore we need to ignore it here as well.

Toflar: The whole purpose about this PR is to remove all `declare` statements because they're always forbidden!

src/Config/Loader/PhpFileLoader.php
Toflar: True, I tried using `token_get_all()` first but this didn't work for me. Better idea?

src/Config/Loader/PhpFileLoader.php
discordier: Not true for `declare(ticks = 1);` which can occur anywhere in code (it simply overrides any previous declaration.

src/Config/Loader/PhpFileLoader.php
discordier: None really elegant.
The only bullet proof solution would be to ignore the fact that it has to be the first statement and compare all lines if they contain the needle...
We could also `preg_replace_all` the whole file to check if any `declare(strict_types=1);` is declared and replace that one out then.
something like:
```php
// Match any line, which only contains of leading whitespace, "declare(strict_types=1);" suffixed by whitespace or comment.
$code = preg_replace(
    '/\r?\n[\t ]*(declare\(\s*strict_types\s*=\s*[^)]+\);)(?:[\t ]+\/[\/\*][^\r\n]+)?[\t ]*\r?\n/',
    '',
    $code
);
```
Which matches any line where there is only(!) `declare(strict_types=1);` and whitespace on it.
This matches:
```php
<?php
/*
 * [....]
 */

   declare(strict_types=1);    
```
But sadly also matches:
```php
<?php
/*
 * [....]
 */

$GLOBALS['TL_DCA']['tl_test'] = [
];

   declare(strict_types=1);    
```

However this would indicate that the initial file was already broken and trying therefore we can safely remove that one as well.

The last culprit for the matches is, that it also matches entities that are on a own line within a comment, but I think of this as a non issue:
```php
<?php
/*
   declare(strict_types=1);    
 */
```

On a performance aspect, I think we can marginalize that we are regexing the whole file because the execution will only be performed on container build time where micro optimization is not needed.

Toflar: So to be correct it would need to replace `declare(strict_types=1,tick=1);` by `declare(tick=1);`, right?

discordier: Where did you get that from? I understand the [docs](http://php.net/manual/en/control-structures.declare.php) that only one directive is permitted per `declare`.

Toflar: Ah true, was just a regular thought 😄  Want to PR your changes against my PR? It's your code after all 😄

discordier: Just tried it out and it indeed works...
```php
declare(strict_types = 1, ticks = 1, encoding='UTF-8');
```

Funny is, that the following only raises a warning, maybe we need to handle all declares then?
```php
declare(foo='bar');
```

Another thing some people might want to do:
```php
<?php declare(strict_types = 1, ticks = 1, encoding='UTF-8');

/***
 * file doc goes here
 */
```

This gets more and more a bottomless pit... :grimacing:

Toflar: Yeah well, then that's just not a supported case. It's `strict_types=1` in 99.99999999999999% of all cases anyway.

Toflar:
There you go. I'm now using the tokenizer and I guess I covered everything with unit tests.

* `declare()` statements in comments and other strings are ignored
* `declare(whatever=1,strict_types=1)` becomes `declare(whatever=1)`
* `declare(strict_types=1,whatever=1)` becomes `declare(whatever=1)`
* Works with whitespace so `declare(      strict_types=1     )` just works fine
* Works with new lines too

😎  /cc @discordier

Toflar:
Test failures not related.

src/Config/Loader/PhpFileLoader.php
```diff
@@ -32,6 +32,46 @@ public function load($file, $type = null)
...
+
+                if ($collect) {
+                    $declarePart .= $token[1];
+                } else {
+                    $codeNew .= $token[1];
+                }
+
+            } else {
+                if ($collect && ';' === $token) {
+                    $declarePart = preg_replace('/([\s]*,)?strict_types[\s]*=[\s]*1([\s]*,)?/', '', $declarePart);
```
ausi: In the beginning of the regex it should be `(,[\s]*)?` I think, so that `declare(whatever=1 , strict_types=1)` works. And I think all `[\s]` can be replaced by just `\s`.

Toflar: You're right. Added another test case and fixed in 1e0e9263f5cc64416058792c8d5a1ea7c1042b8b.

discordier:
Good :)
Have you thought about whitespace around the `=`?
I am in Hamburg and can not check myself right now.

Toflar:
Yes it works but I added an explicit check in 488b07cd8fcb44bfce0a684b858b479953b1bd68.

src/Config/Loader/PhpFileLoader.php
```diff
@@ -32,6 +32,47 @@ public function load($file, $type = null)
     {
         $code = rtrim(file_get_contents($file));
 
+        // declare() statements
+        $codeNew = '';
+        $tokens = token_get_all($code);
+        $collect = false;
+        $declarePart = '';
+
+        foreach ($tokens as $k => $token) {
```
leofeyer: It seems that `$k` is never used, is it?

Toflar: Yeah, I wonder why phpstorm did not tell me. Fixed.

leofeyer:
I like the PR, except that we are now using the tokenizer to remove the `declare()` statements and `preg_replace` to remove other tokens (e.g. `<?php`) and handle namespaces (see the [changes from Contao 3.5.26](https://github.com/contao/core/pull/8701/commits/bf1197073816ad2d9fedd87eb8788f8e2b5479dd#diff-cd3d75697fd2d0576c59e58d1505c692)). Wouldn't it be better to handle them all in the tokenizer?

Toflar:
> Wouldn't it be better to handle them all in the tokenizer?

Sure, feel free to adjust them if you want to migrate over. That's definitely not on my priority list.

leofeyer:
I have created a follow-up PR: https://github.com/contao/core-bundle/pull/774

Toflar:
👍

PR #738, opened 2017-03-27
Fixed the boolean checkbox labels not supporting other false types...

qzminski:
Fixed the boolean checkbox labels not supporting other false types than empty strings.

I use the Doctrine's field type `boolean` to store the data for single checkboxes. Internally it stores them as 0/1 value in the `tinyint` column type. Everything works great except from the labels in the backend. The passed `0` results in true for the `$value !=''` statement and thus I have a `yes/yes` labels in the filter dropdown instead of `no/yes`.

Logic works as expected, the labels are just wrong.


leofeyer:
Is this a bugfix? Then we should probably merge it into `contao/core`.

leofeyer:
Actually merged in 4e7c1ee63fd084a5d94b0b03dd6fdec2f85ec386.

Issue #740, opened 2017-03-28
PHP files do not work in app/Resources/contao/languages


fritzmg:
You are supposed to be able to provide your own translation files within `app/Resources/contao/languages`. In theory, you should be able to provide either XLF or PHP files there, for example
```
app/Resources/contao/languages/de/default.xlf
```
with
```xml
<?xml version="1.0" encoding="UTF-8"?>
<xliff version="1.1">
  <file datatype="php" original="src/Resources/contao/languages/en/default.php" source-language="en">
    <body>
      <trans-unit id="MSC.more">
        <target>mehr »</target>
      </trans-unit>
    </body>
  </file>
</xliff>
```
or
```
app/Resources/contao/languages/de/default.php
```
with
```php
<?php

$GLOBALS['TL_LANG']['MSC']['more'] = 'mehr »';
```
However, the latter does not seem to work.

_Note:_ I deleted the `dev` and `prod` cache after each change of course.

See also: https://community.contao.org/de/showthread.php?66223-Weiterlesen-Link-unter-Contao-4-umbenennen

leofeyer:
@aschempp /cc

aschempp:
I cannot confirm this atm. I'm actively using PHP language files in `app/Resources` and it works just fine…

discordier:
So am I, I have not a single xlf in my extension code but only xlf.
Therefore we need someone to confirm and pinpoint the problem.

aschempp:
You know this is not about extensions, right?

mvscheidt:
The files are loaded indeed. I purposely caused a Synthax error in my file (default.php) and it threw an error there after clearing the cache. So this is most likely not an issue.

However the loading order is important here.
Both the PHP Files AND the Xliff Loader set the $GLOBALS Array currently. The latter because the constructors second param is true: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/System.php#L376

The "PHP Loader": https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/System.php#L390-L402

The xlfLoader Loader however happens to run after the PHP Files are loaded (you probably see where this is going).

Xliff Loader: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/System.php#L404-L416

[...]

PR #741, opened 2017-03-28
[RFC] Update the DEPRECATED.md file

fritzmg:
I simply took the exact same wording from the `trigger_error` method [here](https://github.com/contao/core-bundle/blob/4.3.7/src/Resources/contao/library/Contao/System.php#L438).

It would be useful to provide some examples - however PHP files do not seem to work there currently (#740).

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -310,3 +310,9 @@ $GLOBALS['TL_DCA']['tl_example'] = array
 );
 
 ```
+
+### langconfig.php
+
+Using the `langconfig.php` file has been deprecated and will no longer work 
+in Contao 5.0. Create one or more language files in `app/Resources/contao/languages` 
+instead.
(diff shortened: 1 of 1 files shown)


fritzmg:
Agreed.

Am I right in assuming that any DCA changes should be done via `app/Resources/contao/…`?

Not sure what to write about the `initconfig.php` though other than it being deprecated.

leofeyer:
Symfony provides a lot of useful events that can be used instead of the `initconfig.php` file.

fritzmg:
Do you have a suitable example? I cannot think of a proper use-case for the `initconfig.php` right now anymore. I sometimes put things in there that could be done in the `dcaconfig.php` as well - but I chose to do it in the `initconfig.php` since it it only executed once.

aschempp:
As far as I remember there is no Contao framework initialization event, but `kernel.request` might be used in most of the places for example.

fritzmg:
I have updated the PR.

leofeyer:
Merged in aae0a69798c6672dbcb9a6ef1192e9ed2f9a7c28.

PR #743, opened 2017-03-28
[RFC] Removed proxy cache killing varying on User-Agent

Toflar:
I guess this goes without any further comment 😄

### DIFF
--- a/src/Resources/contao/library/Contao/Template.php
+++ b/src/Resources/contao/library/Contao/Template.php
@@ -312,7 +312,6 @@ abstract class Template extends \Controller
 		$this->compile();
 
 		$response = new Response($this->strBuffer);
-
 		$response->headers->set('Content-Type', $this->strContentType . '; charset=' . Config::get('characterSet'));
 
 		return $response;
--- a/src/Resources/contao/library/Contao/Template.php
+++ b/src/Resources/contao/library/Contao/Template.php
@@ -293,7 +293,6 @@ abstract class Template extends \Controller
 
 		$this->compile();
 
-		header('Vary: User-Agent', false);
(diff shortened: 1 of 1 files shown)


leofeyer:
A quick test has shown that the `{{ua::class}}` insert tag is no longer correct without the `Vary` header. Is `$_SERVER['HTTP_USER_AGENT']` set upon ESI requests?

leofeyer:
I actually cannot reproduce the UA problem. I ran some more tests and everything worked as expected. The cache file has this line:

```html
<body id="top" class="<?php echo $this->surrogate->handle($this, '/_fragment?_path=insertTag%3D%257B%257Bua%253A%253Aclass%257D%257D%26_format%3Dhtml%26_locale%3Den%26_controller%3Dcontao.controller.insert_tags%253ArenderAction&_hash=…', '', true) ?>
 start" itemscope itemtype="http://schema.org/WebPage">
```

I double checked that the cached content is loaded and the user agent string was always correct. 🤔

aschempp:
If the request is done by the proxy, how come the user agent can be correct?

Toflar:
> If the request is done by the proxy, how come the user agent can be correct?

Because it's the same?

leofeyer:
But why did you guys get a wrong UA class then during our test this morning?

Toflar:
I'll check that.

leofeyer:
How do we proceed with this one?

Toflar:
Merge as soon as https://github.com/contao/core-bundle/pull/749 is merged.

leofeyer:
Shouldn't it work without #749? The `{{ua::class}}` insert tag is rendered via ESI, so we should not have to vary on the user agent string anymore.

Toflar:
It should, yes.

aschempp:
Except for the mobile layout…

PR #744, opened 2017-03-29
[RTM] Skip folder DCs that are not database assisted in DcaExtractor

ausi:
The install tool currently shows the deprecation message *“Using database.sql files has been deprecated and will no longer work in Contao 5.0. Use a DCA file instead.”* because it thinks `tl_templates` uses a *database.sql* file.

With this PR `DC_Folder`s that are not database assisted get skipped in the `DcaExtractor`.

### DIFF
--- a/src/Resources/contao/library/Contao/DcaExtractor.php
+++ b/src/Resources/contao/library/Contao/DcaExtractor.php
@@ -401,6 +401,12 @@ class DcaExtractor extends \Controller
 			return;
 		}
 
+		// Return if the DC type is "Folder" and is not database assisted
+		if ($GLOBALS['TL_DCA'][$this->strTable]['config']['dataContainer'] == 'Folder' && empty($GLOBALS['TL_DCA'][$this->strTable]['config']['databaseAssisted']))
+		{
+			return;
+		}
+
 		$blnFromFile = false;
 		$arrRelations = array();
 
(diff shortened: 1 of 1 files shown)


src/Resources/contao/library/Contao/DcaExtractor.php
```diff
@@ -401,6 +401,12 @@ protected function createExtract()
 			return;
 		}
 
+		// Return if the DC type is "Folder" and is not database assisted
+		if ($GLOBALS['TL_DCA'][$this->strTable]['config']['dataContainer'] == 'Folder' && empty($GLOBALS['TL_DCA'][$this->strTable]['config']['databaseAssisted']))
```
leofeyer: I don't like the `empty()` check here (although it might be technically correct).

ausi: Should I replace it with `(!isset($GLOBALS['TL_DCA'][$this->strTable]['config']['databaseAssisted']) || !$GLOBALS['TL_DCA'][$this->strTable]['config']['databaseAssisted'])`?

leofeyer: Probably just `$GLOBALS['TL_DCA'][$this->strTable]['config']['databaseAssisted']`? You have now added `isset()` to `databaseAssisted` but not to `dataContainer` …

ausi: I thought `dataContainer` is a required field because there is no isset check 6 lines above. Wouldn’t removing the isset check result in “undefined index” notices?

PR #745, opened 2017-03-29
Fixed missing initialization of the framework

Toflar:
Otherwise page types are not found.


leofeyer:
The you can remove the call in line 304 (now 306) I guess.

aschempp:
This looks like it will boot the framework in all cases, even unnecessarily, like when rendering a service unavailable exception?

Toflar:
Yeah basically it was just a quickfix instead of a issue-only so that it can be fixed properly including tests. I just can't handle so many branches atm so please just fix accordingly :D

leofeyer:
Ok, will do.

leofeyer:
Fixed in 682b5b8c1b2c470031097bdfb3d931aaf8f4b169.

PR #749, opened 2017-04-01
[RTM] Support for replaying headers

Toflar:
Implements our [terminal42/header-replay-bundle](https://github.com/terminal42/header-replay-bundle) to support "splitting up" user context provided by cookies (or authorization header or whatever you like) into whatever headers you like that are then replayed as master request onto the application so you can vary on them. Please read the whole README of the header-replay-bundle to understand the concept. 

* Brings back the support for the mobile page layout because we can now `Vary` on it
* Always show the "live" content when logged in

ToDo's

* [x] Unit tests for the two listeners
* [x] Wait for the [FOSHttpCache](https://github.com/FriendsOfSymfony/FOSHttpCache) and [FOSHttpCacheBundle](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle) to be released as 2.0.0 stable so I can release terminal42/header-replay-bundle 1.0.0 stable too

Adjust your `AppCache` according to the README. PR for the Managed Edition: https://github.com/contao/manager-bundle/pull/26


composer.json
```diff
@@ -122,5 +124,7 @@
             "dev-develop": "4.4.x-dev"
         },
         "contao-manager-plugin": "Contao\\CoreBundle\\ContaoManager\\Plugin"
-    }
+    },
+    "minimum-stability": "dev",
```
aschempp: This should not be necessary if you suffix the requirement with `@dev`. If that's an issue with an included package, simply require it in the root as well (with `@dev`).

src/EventListener/HeaderReplay/PageLayoutListener.php
```diff
@@ -0,0 +1,64 @@
...
+                case 'mobile':
+                    $mobile = true;
+                    break;
+                case 'desktop':
+                    $mobile = false;
+            }
+        }
+
+        $headers = $event->getHeaders();
+        $headers->set('Contao-Page-Layout', $mobile ? 'mobile' : 'desktop');
```
aschempp: I would suggest to call this `Contao-Mobile-Layout` with `true` only value or the header not being present. Also, creating a `$headers` variable is fairly unnecessary here, but I guess you're aware of that. Just a matter of choice :D

src/Resources/config/listener.yml
```diff
@@ -69,6 +69,20 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.exception, method: onKernelException, priority: 96 }
 
+    contao.listener.header_replay.backend_session:
+        class: Contao\CoreBundle\EventListener\HeaderReplay\BackendSessionListener
+        arguments:
+            - "%contao.security.disable_ip_check%"
+        tags:
+            - { name: kernel.event_listener, event: t42.header_replay, method: onReplay }
```
aschempp: as far as I remember Symfony events, the *correct* way to name the method would be according to the event name. So `t42.header_replay` would result in an `onHeaderReplay` method. Also, if you follow that convention, you do not need to name the method (it will be discovered automatically).

On a side note, I think the event should be called `terminal42.header_replay` to not introduce another "namespace".

Toflar:
No. It has to be present due to dependencies. I know about the @dev setting.

src/Resources/config/listener.yml
Toflar: This would look shit, which is why I decided not to do it. `onTerminal42HeaderReplay()` is not a nice method name at all.

src/EventListener/HeaderReplay/PageLayoutListener.php
Toflar: I like my version better. This applies to both, the header name and the variable.

src/Resources/config/listener.yml
aschempp: it would be `onHeaderReplay`. The key is not used. Think of `kernel.response` => `onResponse`

src/EventListener/HeaderReplay/PageLayoutListener.php
aschempp: The idea was that the header would not need to be present for 90% of pages (if not in mobile mode). Also, the Vary would only be needed if the current page actually has a mobile layout.

src/Resources/config/listener.yml
Toflar: No, it's `onKernelResponse`.

src/EventListener/HeaderReplay/PageLayoutListener.php
discordier: We could introduce new layout types using the same header with @Toflar's variant. Therefore I am also in favor of his approach.

src/ContaoManager/Plugin.php
```diff
@@ -43,7 +44,7 @@ class Plugin implements BundlePluginInterface, RoutingPluginInterface
     public function getBundles(ParserInterface $parser)
     {
         return [
-            BundleConfig::create(KnpMenuBundle::class),
```
aschempp: sure this is correct?

Toflar: Nope, this is a rebase-bug! Fixed in 9fdb93eeff406adec46fe2f3aad69fec83e93cb4.

src/EventListener/HeaderReplay/PageLayoutListener.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * @param HeaderReplayEvent $event
+     */
+    public function onReplay(HeaderReplayEvent $event)
+    {
+        $request = $event->getRequest();
+
+        $this->framework->initialize();
```
aschempp: pretty sure we should check for a Contao scope before booting the framework on EVERY request?

Toflar: On every request with a `Cookie` or `Authorization` header, yeah. We could do that? How do I do this now?

src/EventListener/HeaderReplay/BackendSessionListener.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        $this->disableIpCheck = $disableIpCheck;
+    }
+
+    /**
+     * @param HeaderReplayEvent $event
+     */
+    public function onReplay(HeaderReplayEvent $event)
+    {
+        $request = $event->getRequest();
```
aschempp: Same applies here. The event listener should only be triggered on `$request->attributes->get('_scope') === 'backend'` (actually use the constant!)

Toflar:
Scope checks implemented.

src/EventListener/HeaderReplay/PageLayoutListener.php
```diff
@@ -0,0 +1,76 @@
...
+
+        if ($request->cookies->has('TL_VIEW')) {
+            switch ($request->cookies->get('TL_VIEW')) {
+                case 'mobile':
+                    $mobile = true;
+                    break;
+                case 'desktop':
+                    $mobile = false;
+            }
+        }
```
leofeyer: We could avoid initializing the framework if there is a `TL_VIEW` cookie:

```php
if ($request->cookies->has('TL_VIEW')) {
    $mobile = 'mobile' === $request->cookies->get('TL_VIEW');
} else {
    $this->framework->initialize();
    $mobile = $this->framework->getAdapter(Environment::class)->get('agent')->mobile;
}
```

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -371,9 +371,13 @@ private function setCacheHeaders(Response $response)
 			return $response->setPrivate();
 		}
 
-		// Do not cache the response if a user is logged in or the page is protected or uses a mobile layout
-		// TODO: Add support for proxies so they can vary on member context and page layout
-		if (FE_USER_LOGGED_IN === true || BE_USER_LOGGED_IN === true || $objPage->isMobile || $objPage->protected || $this->hasAuthenticatedBackendUser())
+		// Vary on page layout
+		$response->setVary(['Contao-Page-Layout'], false);
```
leofeyer: This is legacy code, so we must not use the short array notation.

Toflar:
@leofeyer comments addressed.

Issue #751, opened 2017-04-06
Contao not working with Symfony 3.2.7


agoat:
This night the Symfony 3.2.7 is released. Contao will not work after Symfony is updated to 3.2.7!!

I think the reason have to do with some changes in Symfony 3.2.7 to the `rootDir` or `webDir` variables.

In the backend no css or js files could be loaded, because the symlink for the theme/flexible is wrong.
In the frontend no images or css and js files could be loaded, because they will not be generated.

aschempp:
I had exactly this issue yesterday evening! Do you know what changed in Symfony?

agoat:
After some quick investigation it seems I was wrong. It's not the `rootDir` or `webDir` variable (wich are contao related and not from symfony).

I think the problem has to do with the changes in the normalize method fo the filesystem made in https://github.com/symfony/symfony/pull/22133. But I have to do some more tests..

agoat:
Ok, it's definitive the `makePathRelative` method used in the [SymlinkUtil, TemplateLoader and ContaoCacheWarmer classes](https://github.com/contao/core-bundle/search?utf8=%E2%9C%93&q=makePathRelative).

The method returns false relative paths. E.g. for the backend theme the paths are always one level to short ('../vendor/contao/xx' instead of '../../vendor/contao/xx').

I think it's a bug in Symfony!!

aschempp:
Thanks for debugging!! Will you provide a fix for Symfony (you'll get core contributor 😉)?

agoat:
Some other tests show that for the `makePathRelative` method it is necessary to have the same path roots for the `endPath` and `startPath` values.

I tested the following:
```
$fs = new \Symfony\Component\Filesystem\Filesystem();

dump($fs->makePathRelative('vendor/contao/core-bundle/src/Resources/contao/themes/flexible','system/themes'));
dump($fs->makePathRelative('install/dir/of/contao4/vendor/contao/core-bundle/src/Resources/contao/themes/flexible','install/dir/of/contao4/system/themes'));
```

Result on Symfony 3.2.6:
```
"../../contao/core-bundle/src/Resources/contao/themes/flexible/"
"../../vendor/contao/core-bundle/src/Resources/contao/themes/flexible/"
```
Result on Symfony 3.2.7:
```
"../contao/core-bundle/src/Resources/contao/themes/flexible/"
"../../vendor/contao/core-bundle/src/Resources/contao/themes/flexible/"
```

[...]

PR #752, opened 2017-04-06
Use the same path roots when calling makePathRelative

agoat:
This PR is related to https://github.com/contao/core-bundle/issues/751 and fixes the (wrong) use of the `makePathRelative` method in the `SymlinkUtil` class to be compatible with Symfony 3.2.7.


agoat:
@ausi made a PR in symfony/symfony https://github.com/symfony/symfony/pull/22321. 

We decided that it's a bug in symfony. So this change will not be needed **after** symfony fixes the bug.

PR #755, opened 2017-04-07
[RTM] Extended picker

leofeyer:
This PR is the basis for an extended picker that allows to pick news, events etc. (see #7656).

It uses the KnpMenuBundle to generate the menu and dispatches the `contao.build_picker_menu` event so you can extend it (see http://symfony.com/doc/current/bundles/KnpMenuBundle/events.html).

### TODO

- [x] Unit tests
- [x] Make the parent and list views pickable
- [ ] Add a picker for news, events and maybe more (articles, FAQs)
- [x] Add a "check all" checkbox


leofeyer:
With the news bundle item (don't mind the icon):

jankout:
Thank you very much. It is great that we have the compatibility between the Contao's extensions.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -430,6 +450,26 @@ public function showAll()
...
+		$strClass = '';
+		$strPicker = '';
+
+		// Add the picker attributes
+		if ($this->strPickerField)
+		{
+			$strClass = ' picker unselectable';
+			$strPicker .= ' id="tl_select"';
+
+			if (isset($GLOBALS['TL_DCA'][$this->strTable]['config']['picker']['insertTag']))
```
leofeyer: It is not limited to insert tags; the picker will return any given value.

But the JavaScript additionally supports the `data-inserttag` and `data-callback` attributes; the first one will wrap the given value in an insert tag and the second one will trigger an Ajax callback. Both are optional.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -430,6 +450,26 @@ public function showAll()
...
+			{
+				$strPicker .= ' data-inserttag="' . $GLOBALS['TL_DCA'][$this->strTable]['config']['picker']['insertTag'] . '"';
+			}
+
+			if (isset($GLOBALS['TL_DCA'][$this->strTable]['config']['picker']['callback']))
```
leofeyer: That's the `data-callback` attribute we are using already (since Contao 4).

https://github.com/contao/core-bundle/blob/master/src/Resources/public/core.js#L952

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -3071,4 +3151,46 @@ protected function isProtectedPath($path)
...
+
+		if (empty($varValue))
+		{
+			return;
+		}
+
+		$this->arrPickerValue = $varValue;
+
+		// TinyMCE will pass the path instead of the ID
+		if (strpos($varValue[0], \Config::get('uploadPath') . '/') === 0)
```
leofeyer: Again, this is not new code. We are using this already:

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/widgets/FileSelector.php#L625

src/Menu/PickerMenuBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+     * @param Request       $request
+     */
+    private function addFilePickerItem(ItemInterface $menu, Request $request)
+    {
+        $item = $this->factory->createItem(
+            $this->filePickerLabel,
+            ['uri' => $this->route('contao_backend', 'files', $request)]
+        );
+
+        $item->setLinkAttribute('class', 'filemounts');
```
leofeyer: I suppose this comment relates to line 177, doesn't it? Because `setCurrent()` will not set the link attribute.

src/Resources/contao/drivers/DC_Folder.php
Toflar: Ah, didn't know that, cool stuff.

src/Resources/contao/drivers/DC_Folder.php
Toflar: Ok.

src/Resources/contao/drivers/DC_Folder.php
Toflar: Ok. Still ;)

src/Menu/PickerMenuBuilder.php
Toflar: Yeah, this comment must have been moved.

src/Menu/PickerMenuBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+        $this->factory = $factory;
+        $this->renderer = $renderer;
+    }
+
+    /**
+     * Sets the page picker label.
+     *
+     * @param string $pagePickerLabel
+     */
+    public function setPagePickerLabel($pagePickerLabel)
```
leofeyer: I fully agree that the service must not be stateful.

However, I think it is ok if the service adds the two core navigation items. Do you really think we need a separate event listener for that?

src/Menu/PickerMenuBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Creates the file menu.
+     *
+     * @param BackendUser $user
+     *
+     * @return string
+     *
+     * @throws \RuntimeException
+     */
+    public function createMenu(BackendUser $user)
```
leofeyer: ACK.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -178,6 +178,26 @@ public function __construct($strTable)
 			$this->arrValidFileTypes = \StringUtil::trimsplit(',', strtolower($GLOBALS['TL_DCA'][$this->strTable]['config']['validFileTypes']));
 		}
 
+		// Configure the picker
```
leofeyer: ACK.

src/Resources/contao/drivers/DC_Folder.php
leofeyer: I know what you mean. :)

src/Menu/PickerMenuBuilder.php
Toflar: Yes, I think the core should not behave differently than all the other listeners. It will also prevent from stuff like "the core can do this, you cannot in the listener unfortunately".
It's separation of concerns, easier to test, easier to remove etc. pp. I mean, create an `AbstractDcaListener` or something that gets the url, label and css class from its child class. That way it's a breeze implementing FAQ and other regular DCA related listeners.

leofeyer: Changed in ab0ebb3fd2de37a78eddda7429fc3e9f4020c2a7.

src/Menu/PickerMenuBuilder.php
leofeyer: Changed in ab0ebb3fd2de37a78eddda7429fc3e9f4020c2a7.

src/Resources/contao/drivers/DC_Folder.php
leofeyer: I have checked it and it does not really make sense to move any code into the parent class. Especially as we would have to add another abstract method `setPickerValue()`, which will break compatibility with all existing DCs.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2692,12 +2740,28 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 				{
 					$return .= ' <a href="'.$this->addToUrl('&amp;act=move&amp;mode=2&amp;pid='.$currentEncoded).'" title="'.\StringUtil::specialchars(sprintf($GLOBALS['TL_LANG']['tl_files']['uploadFF'], $currentEncoded)).'">'.\Image::getHtml('new.svg', $GLOBALS['TL_LANG'][$this->strTable]['move'][0]).'</a>';
 				}
+
+				if ($this->strPickerField)
```
leofeyer: I checked this, too, but there are too many small differences to extract the code into a parent method.

src/Resources/contao/drivers/DC_Folder.php
Toflar: Obviously I would not make it `abstract`. I'd just add the "normal" case to `DataContainer` and override the stuff that differs in the child tables :-)

src/Resources/contao/drivers/DC_Folder.php
leofeyer: Hm, let me check again once everything is implemented. There might be some optimization potential after all.

leofeyer:
Ok, I have completely reworked the implementation according to what @Toflar and I have discussed. It now uses tagged services that provide the picker information and a custom route that redirects depending on the picker value. Works great. 😄 

Here's how to add a provider in your custom bundle:

https://github.com/contao/news-bundle/commit/8a1960d0c75de1c1048087b14a671f88bdc03b9c

src/Resources/contao/dca/tl_content.php
```diff
@@ -1579,7 +1579,11 @@ public function tableImportWizard()
 	 */
 	public function pagePicker(DataContainer $dc)
 	{
-		return ' <a href="' . (($dc->value == '' || strpos($dc->value, '{{link_url::') !== false) ? 'contao/page' : 'contao/file') . '?do=' . Input::get('do') . '&amp;table=' . $dc->table . '&amp;field=' . $dc->field . '&amp;value=' . rawurlencode(str_replace(array('{{link_url::', '}}'), '', $dc->value)) . '&amp;switch=1' . '" title="' . StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['pagepicker']) . '" onclick="Backend.getScrollOffset();Backend.openModalSelector({\'width\':768,\'title\':\'' . StringUtil::specialchars(str_replace("'", "\\'", $GLOBALS['TL_DCA'][$dc->table]['fields'][$dc->field]['label'][0])) . '\',\'url\':this.href,\'id\':\'' . $dc->field . '\',\'tag\':\'ctrl_'. $dc->field . ((Input::get('act') == 'editAll') ? '_' . $dc->id : '') . '\',\'self\':this});return false">' . Image::getHtml('pickpage.svg', $GLOBALS['TL_LANG']['MSC']['pagepicker']) . '</a>';
```
Toflar: Still asking here. This seems pretty universal now to me. Why not add an `'eval' => ['picker' => true']` setting here to ease it for devs?

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -430,6 +445,25 @@ public function showAll()
 			$hrfNew = $GLOBALS['TL_DCA'][$this->strTable]['list']['new']['href'];
 		}
 
+		$strClass = '';
+		$strPicker = '';
+
+		// Add the picker attributes
+		if ($this->strPickerField)
```
Toflar: Still a lot of repetition here. Can't we put this in its own method? :)

src/Resources/contao/drivers/DC_Folder.php
leofeyer: I have moved most of the code to the parent class now.

leofeyer:
The PR is now ready to merge. I'd prefer to merge it before I set up the pickers in the other bundles, because otherwise I have to mess with the Composer version constraints too much. 😄

src/Controller/BackendController.php
```diff
@@ -210,4 +212,20 @@ public function alertsAction()
...
+     *
+     * @return RedirectResponse
+     *
+     * @Route("/_contao/picker", name="contao_backend_picker")
+     */
+    public function pickerAction(Request $request)
+    {
+        $pickerBuilder = $this->container->get('contao.menu.picker_menu_builder');
+
+        return new RedirectResponse($pickerBuilder->getPickerUrl($request->query->all()));
```
Toflar: I'd prefer to pass the whole `Request $request`. Who knows, maybe somebody want's to redirect based on some headers?

src/DependencyInjection/Compiler/PickerMenuProviderPass.php
```diff
@@ -0,0 +1,40 @@
...
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->has('contao.menu.picker_menu_builder')) {
+            return;
+        }
+
+        $definition = $container->findDefinition('contao.menu.picker_menu_builder');
+        $services = $container->findTaggedServiceIds('contao.picker_menu_provider');
```
Toflar: Please use the `PriorityTaggedServiceTrait` and `$this->findAndSortTaggedServices()`. It will easily allow us to set priorities on the providers if we need to ;)

src/Menu/PickerMenuBuilder.php
```diff
@@ -0,0 +1,140 @@
...
+use Knp\Menu\FactoryInterface;
+use Knp\Menu\Renderer\RendererInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * Creates the picker menu.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+class PickerMenuBuilder
```
Toflar: This is a service that will be either injected or retrieved from the container. It should thus specify a `PickerMenuBuilderInterface` so one could replace it completely. Should be as easy as using phpStorm's Refactor->Extract->Interface functionality ;-)

src/Menu/PickerMenuProviderInterface.php
```diff
@@ -0,0 +1,66 @@
...
+    public function canHandle($value);
+
+    /**
+     * Returns the picker URL.
+     *
+     * @param array $params
+     *
+     * @return string
+     */
+    public function getPickerUrl(array $params = []);
```
Toflar: This here should be `getPickerUrl(Request $request);`

src/Resources/contao/classes/Ajax.php
```diff
@@ -432,6 +432,10 @@ public function executePostActions(DataContainer $dc)
 				throw new NoContentResponseException();
 
 			// DropZone file upload
```
Toflar: Wrong comment here

tests/Controller/BackendControllerTest.php
```diff
@@ -61,4 +63,33 @@ public function testActions()
...
+        ;
+
+        $container = $this->mockKernel()->getContainer();
+        $container->set('contao.menu.picker_menu_builder', $pickerBuilder);
+
+        $controller = new BackendController();
+        $controller->setContainer($container);
+
+        $this->assertInstanceOf(
+            'Symfony\Component\HttpFoundation\RedirectResponse',
```
Toflar: `RedirectResponse::class`

src/Controller/BackendController.php
leofeyer: I don't know if this is sufficient:

```php
public function getPickerUrl(Request $request)
{
    if ($request->query->has('value')) {
        foreach ($this->providers as $provider) {
            if ($provider->canHandle($request->query->get('value'))) {
                return $provider->getPickerUrl($request);
            }
        }
    }

    return $this->router->generate('contao_backend', array_merge(['do' => 'page'], $request->query->all()));
    }
```

The code still checks `$request->query->get('value')`. Is this ok?

leofeyer:
@Toflar I have addressed all your comments.

src/Resources/config/services.yml
```diff
@@ -80,6 +80,56 @@ services:
         calls:
             - ["setContainer", ["@service_container"]]
 
+    contao.menu.matcher:
+        class: Knp\Menu\Matcher\Matcher
+        public: false
+
+    contao.menu.picker_menu_builder:
+        class: Contao\CoreBundle\Menu\PickerMenuBuilder
```
ausi: Is’t `menu` here duplicated? Would `contao.menu.picker_builder` and `Menu\PickerBuilder` be better?

src/Resources/contao/classes/DataContainer.php
```diff
@@ -447,6 +466,45 @@ protected function row($strPalette=null)
   </script>';
 		}
 
+		// Page picker
+		if ($arrData['eval']['dcaPicker'])
```
ausi: I think it would be nicer to combine the `eval.dcaPicker*` keys into a single `dcaPicker` array like `'eval' => ['dcaPicker' => ['do' => 'files, 'icon' => '']]`.

src/Resources/public/core.js
```diff
@@ -941,20 +941,24 @@ var Backend =
...
 			for (i=0; i<inp.length; i++) {
 				if (inp[i].checked && !inp[i].id.match(/^(check_all_|reset_)/)) {
 					val.push(inp[i].get('value'));
 				}
 			}
-			if (opt.tag && (field = $(opt.tag))) {
+			if (opt.callback) {
+				opt.callback(ul.get('data-table'), val);
+			} else if (opt.tag && (field = $(opt.tag))) {
+				console.warn('Using the modal selector without a callback function is deprecated. Please adjust your Contao DCA file.');
```
ausi: `console.warn()` is not available in all browsers I think (e.g. IE9). We should wrap a `if (window.console && console.warn) {` around it to prevent errors.

src/Resources/public/core.js
```diff
@@ -941,20 +941,24 @@ var Backend =
...
 				field.value = val.join(',');
 				if (it = ul.get('data-inserttag')) {
 					field.value = '{{' + it + '::' + field.value + '}}';
 				}
 				opt.self.set('href', opt.self.get('href').replace(/&value=[^&]*/, '&value=' + val.join(',')));
 			} else if (opt.id && (field = $('ctrl_' + opt.id)) && (act = ul.get('data-callback'))) {
+				console.warn('Using the modal selector without a callback function is deprecated. Please adjust your Contao DCA file.');
```
ausi: Same here.

src/Resources/config/services.yml
leofeyer: The alias is built from the class name, so `Menu\PickerMenuBuilder` becomes `menu.picker_menu_builder`. That's the regular Symfony behavior I think.

leofeyer:
> - In my test the `knp_menu.factory` service was missing, I think this is because of a missing `BundleConfig::create(KnpMenuBundle::class)` in the manager plugin.

Yes, the PR for the manager bundle has yet to be made. I'll add it to the TODO list.

> - The old file picker executed the `onload_callback`s of the DCA by instantiating it in [BackendFile.php:81](https://github.com/contao/core-bundle/blob/4758be406700f7861bdddfb39e140de1e61a6b73/src/Resources/contao/controllers/BackendFile.php#L81) and setting the active record from the database. I think this is important, because the `onload_callback` may be used to dynamically add a field based on the values of the database record (for example the MultiColumnWizard or my Custom Elements extension). This currently results in a `Target field "..." does not exist` error.

I will add this.

> - Should we change the file picker to save a `{{file::[UUID]}}` insert tag instead of a path?

That is not yet possible, because it would not be backwards compatible on the one hand and things like TinyMCE need the path in order to show the image on the other hand.

> - The backend style of Contao consists mostly of white and light-gray areas with gray borders. Isn’t the yellow background of the checkboxes/radios too much?

I don't think so. It just draws enough attention so you realize what you need to do. 😄 

> - It may be a good idea to hide the filter and search toolbars by default so that the picker interface looks cleaner.

I don't agree. The toolbars are required to search and filter the records.

src/Resources/contao/classes/DataContainer.php
leofeyer: 👍

src/Resources/public/core.js
leofeyer: Will do.

src/Resources/config/services.yml
ausi: Agreed

ausi:
>> - In my test the `knp_menu.factory` service was missing, I think this is because of a missing `BundleConfig::create(KnpMenuBundle::class)` in the manager plugin.
>
> Yes, the PR for the manager bundle has yet to be made. I'll add it to the TODO list.

But the missing pice is in [src/ContaoManager/Plugin.php:35](https://github.com/contao/core-bundle/blob/028672beec41e6f47492eb41c6bfd49b234f6e79/src/ContaoManager/Plugin.php#L35) in the core-bundle, why not add it in this PR?

>> - It may be a good idea to hide the filter and search toolbars by default so that the picker interface looks cleaner.
>
> I don't agree. The toolbars are required to search and filter the records.

IMO most of the time the filters and the search are not needed and moving them behind a search and filter icon would just make them one click away.

src/Resources/contao/classes/DataContainer.php
leofeyer: I have changed this in the latest PR, however we now end up with `'dcaPicker'=>array()` most of the time, which feels a little unfamiliar.

leofeyer:
@ausi I have changed everything but the `onload_callback`. Since we now use the regular Contao back end (list view, parent view and tree view) to generate the picker, the `onload_callback` is no longer triggered. I'm not sure how to solve this?

src/Resources/contao/dca/tl_content.php
```diff
@@ -270,7 +270,7 @@
 			'exclude'                 => true,
 			'search'                  => true,
 			'inputType'               => 'text',
-			'eval'                    => array('rgxp'=>'url', 'decodeEntities'=>true, 'maxlength'=>255, 'dcaPicker'=>true, 'fieldType'=>'radio', 'filesOnly'=>true, 'tl_class'=>'w50 wizard'),
+			'eval'                    => array('rgxp'=>'url', 'decodeEntities'=>true, 'maxlength'=>255, 'dcaPicker'=>array(), 'fieldType'=>'radio', 'filesOnly'=>true, 'tl_class'=>'w50 wizard'),
```
Toflar: I think it should remain `'dcaPicker'=>true` but optionally accepting an array.

leofeyer:
@ausi We could add the following code [here](https://github.com/contao/core-bundle/blob/feature/dca-picker/src/Resources/contao/classes/DataContainer.php#L903):

```php
$this->setPickerValue();

$strDriver = 'DC_' . $GLOBALS['TL_DCA'][$this->strPickerTable]['config']['dataContainer'];
$objDca = new $strDriver($this->strPickerTable);
$objDca->field = $this->strPickerField;

// Set the active record
if ($this->Database->tableExists($this->strPickerTable))
{
	/** @var Model $strModel */
	$strModel = \Model::getClassFromTable($this->strPickerTable);

	if (class_exists($strModel))
	{
		$objModel = $strModel::findByPk(\Input::get('id')); // <----- NOT AVAILABLE ANYMORE!

		if ($objModel !== null)
		{
			$objDca->activeRecord = $objModel;
		}
	}
}

// Call the load_callback
if (is_array($GLOBALS['TL_DCA'][$this->strPickerTable]['fields'][$this->strPickerField]['load_callback']))
{
	foreach ($GLOBALS['TL_DCA'][$this->strPickerTable]['fields'][$this->strPickerField]['load_callback'] as $callback)
	{
		if (is_array($callback))
		{
			$this->import($callback[0]);
			$this->arrPickerValue = $this->{$callback[0]}->{$callback[1]}($this->arrPickerValue, $objDca);
		}
		elseif (is_callable($callback))
		{
			$this->arrPickerValue = $callback($this->arrPickerValue, $objDca);
		}
	}
}
```

But `Input::get('id')` is no longer available, so we can no longer set the active record, can we?

src/Resources/contao/dca/tl_content.php
leofeyer: 👍

ausi:
@leofeyer I think this is not enough because the ID of the databse record is missing. I created the PR #763, this should be a working solution.

leofeyer:
Nope, the ID is not missing. It is added by the `FileTree` and `PageTree` classes, therefore I have added the `isset($_GET['id'])` check.

```php
    <p><a href="' . ampersand(\System::getContainer()->get('router')->generate('contao_backend_picker', array('do'=>'page', 'target'=>$this->strTable.'.'.$this->strField, 'value'=>implode(',', $arrSet), 'id'=>$this->activeRecord->id, 'popup'=>1))) . '" class="tl_submit" id="pt_' . $this->strField . '">'.$GLOBALS['TL_LANG']['MSC']['changeSelection'].'</a></p>
```

(The changes have not been pushed yet.)

ausi:
But wouldn’t the `$_GET['id']` conflict with the data container that shows the list/tree view (like in a news archive)? That’s why I added the ID to the `target` parameter in my PR.

leofeyer:
Yes, you are right – as always. 😄

leofeyer:
Well, you were only half way right. It `id` parameter was only used in the `PageTree` and `FileTree` widgets, which are both not switchable, which is why the problem did not occur.

However, your [changes in line 484](https://github.com/contao/core-bundle/pull/763/files#diff-b187da7f3bf59a72fc3327dbb0246dd4R484) now always add the ID, which is a different behavior than before. I don't think it will break anything, but we should review it carefully.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -813,6 +877,143 @@ protected function generateGlobalButtons()
...
+
+		list($this->strPickerTable, $this->strPickerField, $this->intPickerId) = explode('.', \Input::get('target'), 3);
+		$this->intPickerId = (int) $this->intPickerId;
+
+		\Controller::loadDataContainer($this->strPickerTable);
+
+		if (!isset($GLOBALS['TL_DCA'][$this->strPickerTable]['fields'][$this->strPickerField]))
+		{
+			throw new InternalServerErrorException('Target field "' . $this->strPickerTable . '.' . $this->strPickerField . '" does not exist.');
+		}
```
ausi: This `if` needs to be moved to the end of the method, because the DCA `onload_callback` may add the field dynamically.

leofeyer: Changed in d86a642a19d4d32258396ee2b277656a7ed9a4ec.

src/Resources/public/core.js
```diff
@@ -941,20 +941,24 @@ var Backend =
 				alert('Could not find the SimpleModal frame');
 				return;
 			}
-			ul = frm.document.getElementById(opt.id);
+			ul = frm.document.getElementById('tl_select');
```
ausi: This change broke an existing URL picker (similar to the old one from `tl_content`) in my test. Without this change the picker works as expected, but with it the `{{link_url::*}}` insert tag around the ID is missing.

leofeyer: Hm, the core is not passing an `id` attribute anymore, because you are supposed to pass a callback function instead. Therefore, `tl_select` is kind of a convention.

Is it crucial that we support `opt.id` here, too?

ausi: I think it’s very likely that somebody copied the `tl_content::pagePicker()` method to use it for an extension or in a project, this would include `\'id\':\'' . $dc->field . '\',` too. Not supporting `opt.id` is be a BC break I think.

How about using `tl_select` as a fallback? 
`frm.document.getElementById(opt.id || 'tl_select')`

leofeyer: That's possible of course, but it does not really help. Do we want to support `opt.id` in Contao 5 as well? Then we should encourage people to provide it. Or do we want to use `tl_select` in Contao 5? Then we should add a deprecation notice if someone uses `opt.id`.

ausi: > Do we want to support `opt.id` in Contao 5 as well?

I think it’s OK to deprecate it.

leofeyer: Changed in 2b43a70ba9defbfde8697bbf5ea3ecd1f22394cf.

PR #757, opened 2017-04-11
[RTM] Fix bug overwriting cache/clientCache with zero values

ausi:
If you set a cache time in the root page of a website and want to overwrite it with `0` in a subpage it doesn’t work currently because the check doesn’t differentiate between `0` and `false`.


ausi:
`false` means no cache setting was set, `0` means the cache was set to “0 (do not cache)”. I addressed this in 581562129b64a97dd1754e2ce597301da9f17aa8

PR #763, opened 2017-04-18
[RFC] Followup #755 Extended picker

ausi:
This adresses the issue discussed in https://github.com/contao/core-bundle/pull/755#issuecomment-294716212


leofeyer:
The changes have been added in 896a5f90cbee7051c9ee76e9add52820393acff2.

Toflar:
The `core.js` change is missing.

ausi:
The change in the `core.js` is not needed in leo’s version I think, because the ID and field name switched the order.

Toflar:
Ah, ok.

PR #774, opened 2017-04-19
Handle namespaces and declare() statements when combining files

leofeyer:
This is a follow-up PR on #730, which uses the tokenizer to handle everything (PHP tags, declare statements and namespaces).

### DIFF
--- a/tests/Config/Loader/PhpFileLoaderTest.php
+++ b/tests/Config/Loader/PhpFileLoaderTest.php
@@ -66,13 +66,18 @@ class PhpFileLoaderTest extends TestCase
      */
     public function testLoad()
     {
+        $expects = <<<'EOF'
+
+$GLOBALS['TL_TEST'] = true;
+
+EOF;
+
         $this->assertEquals(
-            "\n\n\$GLOBALS['TL_TEST'] = true;\n",
+            $expects,
             $this->loader->load($this->getRootDir().'/vendor/contao/test-bundle/Resources/contao/config/config.php')
         );
 
-        $content = <<<'EOF'
-
+        $expects = <<<'EOF'
 
 $GLOBALS['TL_DCA']['tl_test'] = [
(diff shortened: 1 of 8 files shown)


tests/Config/Loader/PhpFileLoaderTest.php
```diff
@@ -93,14 +98,83 @@ public function testLoad()
...
+EOF;
+
+        $this->assertEquals(
+            $expects,
+            $this->loader->load($this->getRootDir().'/vendor/contao/test-bundle/Resources/contao/dca/tl_declare.php')
+        );
+
+        $expects = <<<'EOF'
+
+//declare(strict_types  =   1,ticks = 1);
```
Toflar: This is not the correct behaviour. It needs to be `declare(ticks=1);`. Only strip `strict_types=1`.

leofeyer: You are right.

leofeyer:
> all my code is gone and all my commits/work disappeared (which somehow does bother me, yes)

You said "feel free to adjust them if you want to migrate over". It then got pretty complex and confusing with the PHP tokenizer, so I decided to use the `php-token-stream` library.

Now how do we solve this situation?

Toflar:
> Now how do we solve this situation?

Well you could've PR'ed to my PR. Even if you change the whole implementation this would've automatically alerted you that the tests fail because I had test cases for all the `declare()` situations ;)

leofeyer:
Done: https://github.com/Toflar/core-bundle/pull/6

tests/Config/Loader/PhpFileLoaderTest.php
leofeyer: I have fixed this in https://github.com/Toflar/core-bundle/pull/6, which uses your tests for the `declare()` stuff.

PR #776, opened 2017-04-21
Correct a symlink issue with Symfony 3.2.7

cyril-ponce:
Since Symfony 3.2.7, there is a change in makePathRelative method and symlink doesn't work on my host server.


leofeyer:
This one has been fixed already (see #752). Thank you anyway!

PR #794, opened 2017-04-27
[RTM] Replace the switch=1 option with a context parameter

leofeyer:
This PR removes the `switch=1` option and replaces it with a `context` parameter. Supported contexts are:

* file
* page
* link

The picker menu builder now asks the providers if they support the context, thus it is now e.g. possible to add another file picker tab to the file tree picker.


src/Resources/contao/widgets/PageTree.php
```diff
@@ -184,7 +184,7 @@ public function generate()
 		}
 
 		$return .= '</ul>
-    <p><a href="' . ampersand(\System::getContainer()->get('router')->generate('contao_backend_picker', array('do'=>'page', 'target'=>$this->strTable.'.'.$this->strField.'.'.$this->activeRecord->id, 'value'=>implode(',', $arrSet), 'popup'=>1))) . '" class="tl_submit" id="pt_' . $this->strField . '">'.$GLOBALS['TL_LANG']['MSC']['changeSelection'].'</a></p>
+    <p><a href="' . ampersand(\System::getContainer()->get('router')->generate('contao_backend_picker', array('do'=>'page', 'context'=>'page', 'target'=>$this->strTable.'.'.$this->strField.'.'.$this->activeRecord->id, 'value'=>implode(',', $arrSet), 'popup'=>1))) . '" class="tl_submit" id="pt_' . $this->strField . '">'.$GLOBALS['TL_LANG']['MSC']['changeSelection'].'</a></p>
```
Toflar: Not sure but can't we reuse the same logic as in `DataContainer` here now that we have contexts?

leofeyer: No, because the callback function is completely different.

Toflar: Then, tests and merge 👍

PR #800, opened 2017-04-28
install.php not working

xchs:
This adds the new/old route to the Contao install tool.

### DIFF
--- a/src/Resources/contao/languages/en/exception.xlf
+++ b/src/Resources/contao/languages/en/exception.xlf
@@ -48,7 +48,7 @@
         <source>The installation has not been completed, therefore Contao cannot work properly.</source>
       </trans-unit>
       <trans-unit id="XPT.incompleteFix">
-        <source>Please open the Contao install tool (&lt;code&gt;/install.php&lt;/code&gt;) in your browser.</source>
+        <source>Please open the Contao install tool (&lt;code&gt;/contao/install&lt;/code&gt;) in your browser.</source>
       </trans-unit>
       <trans-unit id="XPT.incompleteExplain">
         <source>For more information, please refer to the &lt;a href="https://contao.org/manual/installation.html" target="_blank"&gt;Contao manual&lt;/a&gt;.</source>


aschempp:
This should be merged into `develop` afaik, in 4.3 the `install.php` is still necessary.

leofeyer:
No, it is not (at least not in the standard edition). And using the `install.php` entry point is currently broken in Contao 4.3, so `/contao/install` is in fact the only way to install Contao 4.3.

aschempp:
As far as I know it is not necessary in the managed edition, but it is in the standard edition…

leofeyer:
Not if you install the standard edition via Composer, which I guess is what everyone using the standard edition does.

xchs:
> As far as I know it is not necessary in the managed edition

I'm using the ME and it IS actually necessary to open the Contao install tool via `/contao/install`. Otherwise you get an error.

aschempp:
install.php not working is a bug that should be fixed.

leofeyer:
Feel free to create a PR if you really thing that it is important – because I don't.

fritzmg:
This generates quite some confusion in the forum currently, since people are of course trying to open `/install.php` and do not know about the `/contao/install` route.

aschempp:
https://github.com/contao/manager-bundle/commit/4d2babb216341cf46c8146fc50730028af6e8275 is probably the best solution, maybe that should be merged in the next hotfix instead of develop @leofeyer ?

leofeyer:
@aschempp You can merge it in whatever branch you want. 😄

m-vo:
Btw.: the hint on the 'incomplete installation' error screen still points the user to ``/install.php``. Might be better to change it to ``/contao/install``.

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/languages/en/exception.xlf#L51

xchs:
> Btw.: the hint on the 'incomplete installation' error screen still points the user to /install.php. Might be better to change it to /contao/install.

This is exactly what this PR was made for.

aschempp:
Added in 4.3.10: https://github.com/contao/manager-bundle/commit/d8c9389ab5e6b0994634baddccc7bba519ff1857

leofeyer:
See https://github.com/contao/managed-edition/issues/7.

Issue #731, opened 2017-03-23
error page shown together with regular page on fatal error in insert tag


fritzmg:
If a fatal error occurs within an insert tag, the error page will be generated, but sent together with the regular page to the browser.

### Reproduction:

1. Create an HTML content element and insert `{{file::test.php}}`.
2. Create a file called `test.php` within `/templates` and generate a fatal error within, e.g. `<?php $foo = new Foo();` (assuming there is no class `Foo` loaded 😉).
3. Visit the page in the front-end (_not_ via `app_dev.php`).

### Result:

If you look at the source (or response), you will see the regular page, plus the error page:
```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title>Start - Website</title>
  <base href="http://c43.dev/">
  <meta name="robots" content="index,follow">
  <meta name="description" content="">
  <meta name="generator" content="Contao Open Source CMS">
  <!--[if lt IE 9]><script src="assets/html5shiv/js/html5shiv-printshiv.min.js"></script><![endif]-->
</head>
<body id="top" class="win firefox gecko fx52" itemscope itemtype="http://schema.org/WebPage">
  <div id="wrapper">
    <div id="container">
      <main id="main" itemscope itemtype="http://schema.org/WebPageElement" itemprop="mainContentOfPage">
        <div class="inside">
          <div class="mod_article first last block" id="article-1">
<!DOCTYPE html><html><head><meta charset="utf-8"><title>    An error occurred …
```
If you visit the front-end via `app_dev.php`, everything is fine (only the error page is generated and sent).

Toflar:
This is due to the ESI request. Not sure how to handle this properly yet but I'll have a look at it somewhen.

fritzmg:
Are all insert tags handled via ESI now, not just the "uncached" ones?

Toflar:
`{{file::` is an uncached one ;)

aschempp:
If I remember correctly, ESI has handling suggestions for http status codes.

fritzmg:
Ah, I though that `{{file::….html5/php}}` is not an uncached one in Contao 4 anymore, due to this problem: https://github.com/contao/core/issues/8658
@aschempp thought it was changed in Contao 4 (https://github.com/contao/core/issues/8658#issuecomment-283910850) but apparently that is not the case.

Toflar:
> If I remember correctly, ESI has handling suggestions for http status codes.

No.

[...]

PR #801, opened 2017-05-02
do not force {{file::PATH}} insert tags to be uncached

fritzmg:
See https://github.com/contao/core-bundle/issues/731#issuecomment-295773166

@Toflar and I discussed this in IRC and we came to the conclusion, that this condition
```php
($elements[0] == 'file' && !\Validator::isStringUuid($elements[1]))
```
could simply be dropped from [InsertTags.php#L119](https://github.com/contao/core-bundle/blob/4.3.9/src/Resources/contao/library/Contao/InsertTags.php#L119) altogether.

This way the user/developer can decide himself, whether or not a `{{file::PATH}}` insert tag should be uncached or not.

Previous behaviour in Contao 3 was, that `{{file::UUID}}` and `{{file::PATH}}` insert tags are always forced to be uncached (see [InsertTags.php#L94](https://github.com/contao/core/blob/3.5.27/system/modules/core/library/Contao/InsertTags.php#L94)). This creates a problem in Contao 3 where users are integrating a PHP script via `{{file::PATH}}` and this script wants to access information that is not available when the page is generated from the page cache (e.g. page layout information or other page object information, that is not saved within the page cache file). See https://github.com/contao/core/issues/8658 for instance.

In Contao 4 this restriction was removed from `{{file::UUID}}` insert tags (now you can decide for yourself whether or not that insert tag should be uncached), but still remained for `{{file::PATH}}` insert tags.

By removing this condition the user/developer can now decide whether to use

* `{{file::PATH}}`

or 

* `{{file::PATH|uncached}}`

### DIFF
--- a/src/Resources/contao/library/Contao/InsertTags.php
+++ b/src/Resources/contao/library/Contao/InsertTags.php
@@ -116,7 +116,7 @@ class InsertTags extends \Controller
 			// Skip certain elements if the output will be cached
 			if ($blnCache)
 			{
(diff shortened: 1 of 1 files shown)


fritzmg:
Well, may be add it as a "feature" in 4.4 instead of 4.3.

Issue #799, opened 2017-04-28
Use container autowiring


leofeyer:
Use container autowiring once there is a beta version of Symfony 3.3.

Toflar:
I am 👎  on this. Autowiring is a great addition for developing **applications**. Bundles should **never** rely on autowiring. If you meant the autoconfiguration feature, that's a different story.

leofeyer:
Actually I mean both. :)

http://symfony.com/blog/new-in-symfony-3-3-service-autoconfiguration

http://symfony.com/doc/master/service_container/autowiring.html

Where did you read that bundles should not rely on autowiring?

Toflar:
I did not read, it's obvious to me. Autowiring is always magic to some extent. I think they did a great job in making stuff as explicit as possible but it's still autowiring. Autowiring enables you to get to quicker results where you control **all** of the dependencies. That's only true when you are working on an **application**. A bundle, however, is imho not supposed to use any kind of autowire features.

leofeyer:
But where is the point in writing this

```yaml
    contao.image.image_sizes:
        class: Contao\CoreBundle\Image\ImageSizes
        arguments:
            - "@database_connection"
            - "@event_dispatcher"
            - "@contao.framework"
```

when you can write this

```yml
    contao.image.image_sizes:
        class: Contao\CoreBundle\Image\ImageSizes
        autowire: true
```

?

Toflar:
Autowiring is a RAD feature, it's not made for public bundles.

aschempp:
Totally agree with @Toflar

ausi:
> Where did you read that bundles should not rely on autowiring?

From the [Symfony documentation](http://symfony.com/doc/current/service_container/autowiring.html) (the last sentence):

> keep in mind to not use autowiring in public bundles nor in large projects with complex maintenance needs.

In your example, if I would register a second `@database_connection` service that implements the same interfaces, the autowiring would fail with `Unable to autowire...`.

leofeyer:
Ok. Then just auto-configuration.

leofeyer:
I have implemented auto configuration for commands in #804.

leofeyer:
Implemented in 491502964b99b685af20d12564f7ba95102dd4d6.

PR #804, opened 2017-05-02
[RTM] Auto configure the commands

leofeyer:
This PR implements #799 for the commands. (There are a few unrelated changes which are necessary to make the bundle work with Symfony 3.3.)

### DIFF
--- a/src/Resources/config/commands.yml
+++ b/src/Resources/config/commands.yml
@@ -1,34 +1,26 @@
 services:
+    _defaults:
+        autoconfigure: true
+
+    _instanceof:
+        Contao\CoreBundle\Framework\FrameworkAwareInterface:
+            calls:
+                - ["setFramework", ["@contao.framework"]]
+
     contao.command.automator:
         class: Contao\CoreBundle\Command\AutomatorCommand
-        calls:
-            - ["setFramework", ["@contao.framework"]]
-        tags:
-            -  { name: console.command }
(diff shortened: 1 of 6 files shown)


src/Resources/config/commands.yml
```diff
@@ -1,32 +1,26 @@
 services:
+    _defaults:
+        autoconfigure: true
```
Toflar: Can we add a `# Automatically tags commands as console.command` here and whenever we add something new that uses the `autoconfigure` extend this comment? Just for future reference and transparency :)

leofeyer:
> Can't we also use the autoconfiguration for `kernel.cache_clearer`, `kernel.cache_warmer` and others?

Yes, of course. But this PR only includes the commands. 😉

Toflar:
Ah shit, sorry I thought this was a complete SF 3.3 change. Reading titles would help. My bad!

src/Resources/config/commands.yml
leofeyer: I guess it is pretty obvious what the option does in the `commands.yml` file, isn't it?

Toflar: Yes, I just approved the PR. This comment was related to my initial thought that it should be for everything.

leofeyer:
For some reason it did not work with the trait. I tried this first. 😄

PR #807, opened 2017-05-03
[RTM] Use the file meta data by default when adding an image

leofeyer:
This PR implements contao/core#7774. It is still work in progress, but it already works for adding images to text elements.

### DIFF
--- a/src/Resources/contao/classes/Backend.php
+++ b/src/Resources/contao/classes/Backend.php
@@ -714,9 +714,13 @@ abstract class Backend extends \Controller
 	 * @param string  $strUuid
 	 * @param string  $strPtable
 	 * @param integer $intPid
+	 *
+	 * @deprecated Deprecated since Contao 4.4, to be removed in Contao 5.0.
 	 */
 	public static function addFileMetaInformationToRequest($strUuid, $strPtable, $intPid)
 	{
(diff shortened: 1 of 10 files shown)


leofeyer:
Default form:



Form with the "Overwrite meta data" option set:

leofeyer:
This PR is ready to merge now.

src/Resources/contao/classes/Backend.php
```diff
@@ -714,9 +714,13 @@ public static function findSearchablePages($pid=0, $domain='', $blnIsSitemap=fal
 	 * @param string  $strUuid
 	 * @param string  $strPtable
 	 * @param integer $intPid
+	 *
+	 * @deprecated Deprecated since Contao 4.4, to be removed in Contao 5.0.
 	 */
 	public static function addFileMetaInformationToRequest($strUuid, $strPtable, $intPid)
 	{
+		@trigger_error('Using Backend::addFileMetaInformationToRequest() has been deprecated and will no longer work in Contao 5.0.', E_USER_DEPRECATED);
```
Toflar: Use what instead?

leofeyer: Nothing. The elements now directly use the meta data, so we do not need this workaround anymore.

Toflar: Ok :)

PR #809, opened 2017-05-05
[RTM] Warn if another user has edited a record

leofeyer:
This PR implements contao/core#8412.


jankout:
Thank you very much.

src/Resources/contao/drivers/DC_Table.php
```diff
@@ -2172,6 +2181,16 @@ public function edit($intId=null, $ajaxId=null)
...
 			}
 
+			$arrRecords = $objSessionBag->get('edit_records', array());
+			$intLatestVersion = $objVersions->getLatestVersion();
+
+			// Show a warning if the record has been saved by another user (see #8412)
+			if ($intLatestVersion !== null && isset($arrRecords[$this->strTable][$this->intId]) && $intLatestVersion > ($arrRecords[$this->strTable][$this->intId] + 1))
+			{
+				\Message::addError(sprintf($GLOBALS['TL_LANG']['ERR']['versionWarning'], ($intLatestVersion - 1), $arrRecords[$this->strTable][$this->intId]));
+				$this->reload();
```
Toflar: When I read the code, the process is as follows: If somebody saved a new version until I hit the save button, it will reload and show the message but it will not store my data to the database. Is that correct?

ausi: I think saving is already done at this point.

Toflar: Ah, I see. So it does actually not prevent you from overriding the values but it just tells you that you did override it. That's fine for me I guess.

ausi: The overridden values are still available via the versions dropdown, so it should be fine I think.

ausi:
This implementation relies on the session, so if the same user (or another users logged in with the same account) edits the same element, the warning is not shown. Wouldn’t it be better to use a hidden field in the edit form instead of the `edit_records` array in the session?

Toflar:
> This implementation relies on the session, so if the same user (or another users logged in with the same account) edits the same element, the warning is not shown. Wouldn’t it be better to use a hidden field in the edit form instead of the edit_records array in the session?

I think that would be really cool. It also reduces session noise.

src/Resources/contao/drivers/DC_Table.php
leofeyer: No, that's wrong (of course). It just prevents any redirects so the user stays on the same page and actually sees the error message.

leofeyer:
I have reworked the implementation to use a hidden field instead of the session.

Issue #816, opened 2017-05-08
[4.4-beta1] File picker differences


ausi:
While testing the new file picker I noticed some backwards incompatible differences between the current and the old version:

- With `eval.files = false` only the directories were shown in the picker. Now everything is shown and can be selected but only for the directories a radio button is visible.
- With `eval.filesOnly = true` directories are still selectable, even though no radio button is shown for them.
- The `eval.extensions` setting has no effect anymore.
- The `eval.path` setting has no effect anymore.

leofeyer:
Of course, because now we do not have a separate picker anymore but the regular DC views with an additional checkbox or radio button. Those options are only supported in the old picker widget.

ausi:
And we do not want to add support for them? IMO `eval.extensions` and `eval.path` are very important features of the `fileTree` widget.

leofeyer:
They were features of the `FileSelector` class, not of the `FileTree` class. It is important to understand the difference. 

The file tree widget still works the same as in Contao 4.3, except it now uses the DCA picker instead of the file selector (which still exists in Contao 4.4 BTW). You want the `eval.extensions` and `eval.path` restrictions to apply to the DCA picker, too, however, as I said, the DCA picker is not a separate class but the regular DC views. It is like browsing the back end in the modal window.

I guess we could try to support `eval.extensions` and `eval.path` by modifying `$this->root`, but I don't like the idea to apply restrictions based on a GET parameter.

Toflar:
> I guess we could try to support eval.extensions and eval.path by modifying $this->root, but I don't like the idea to apply restrictions based on a GET parameter.

We already init the picker itself based on GET params. Why shouldn't we do more? Imho this needs to go to `DC_Folder` like this (parent `initPicker()` must return `false` on early return to get this working):

```php
protected function initPicker()
{
    if (false === parent::initPicker()) {
        return false;
    }

    // Apply root based on eval of $GLOBALS['TL_DCA'][$this->strPickerTable]['fields'][$this->strPickerField]
}
```

And I do consider this to be very important too 👍

leofeyer:
I have fixed the selection error in 02f206f1b6cd5521ced88dc48aa752e111fa854e.

leofeyer:
I have begun to create a PR addressing the issues in #819.

PR #819, opened 2017-05-09
[RTM] Re-add the file tree widget options

leofeyer:
This PR implements the missing features mentioned in #816.

**TODO**

* [x] Translate the error messages

### DIFF
--- a/src/DataContainer/DcaFilterInterface.php
+++ b/src/DataContainer/DcaFilterInterface.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2017 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\DataContainer;
+
+/**
+ * DCA filter interface.
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ */
+interface DcaFilterInterface
+{
+    /**
+     * Returns the filter array.
+     *
+     * @return array
+     */
+    public function getDcaFilter();
+}
--- a/src/DataContainer/DcaFilterInterface.php
+++ b/src/DataContainer/DcaFilterInterface.php
@@ -21,6 +21,9 @@ interface DcaFilterInterface
      * Returns the filter array.
      *
      * @return array
+     *
+     * @see DataContainer::setDcaFilter()
+     * @see DC_Folder::setDcaFilter()
      */
     public function getDcaFilter();
 }
--- a/src/Resources/contao/drivers/DC_Folder.php
+++ b/src/Resources/contao/drivers/DC_Folder.php
@@ -82,6 +82,12 @@ class DC_Folder extends \DataContainer implements \listable, \editable
 	 */
 	protected $blnIsDbAssisted = false;
 
+	/**
+	 * Hide files
+	 * @var boolean
+	 */
+	protected $blnHideFiles = false;
+
 
 	/**
 	 * Initialize the object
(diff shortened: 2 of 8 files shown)


src/DataContainer/DcaFilterInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Copyright (c) 2005-2017 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\DataContainer;
+
+/**
+ * DCA filter interface.
```
Toflar: I would like to have some more information here so that when you read it you actually know where it's used and why (see my PR maybe :-)).

leofeyer: Don't know what to write? The text from your PR does not apply anymore.

Toflar: What about this:

> Allows widgets that implement this interface to return a configuration array. In Contao
this is used for DataContainers that support the picker mode and thus might behave
differently depending on the settings of the widget. The widget implementing this interface must return an array of configuration settings. If the DataContainer supports these, they can work together.
See implementations of FileTree and DC_Folder for example.

leofeyer: We are currently using this for pickers only, but the `DataContainer::setDcaFilter()` method is meant to be a general method.

> @see DataContainer::setDcaFilter() or DC_Folder::setDcaFilter()

How about a short comment like above?

Toflar: Yeah that would be fine. It's just that if you see this interface and you never came across it you won't have no idea about why it's there and what it does.

leofeyer: Done.

Toflar: Go ahead an merge :-)

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -172,6 +178,8 @@ public function __construct($strTable)
 		$this->strTable = $strTable;
 		$this->blnIsDbAssisted = $GLOBALS['TL_DCA'][$strTable]['config']['databaseAssisted'];
 
+		$this->initPicker();
```
ausi: I think this needs a similar check as [`DC_Table` has on line 202](https://github.com/leofeyer/core-bundle/blob/21ff0a8e9cee252147ce5d934d17bf35f94c7e63/src/Resources/contao/drivers/DC_Table.php#L202). Currently this line leads to a `The table name must not be empty` exception if I open `/contao?do=files`.

src/Resources/contao/widgets/FileTree.php
```diff
@@ -126,6 +168,80 @@ protected function validator($varInput)
...
+			}
+
+			// Only certain file types can be selected
+			if ($this->extensions)
+			{
+				$extensions = \StringUtil::trimsplit(',', $this->extensions);
+
+				if (is_dir(TL_ROOT . '/' . $objFile->path))
+				{
+					$this->addError(sprintf($GLOBALS['TL_LANG']['ERR']['extensionsOnly'], $this->extensions));
```
ausi: I think this case should not be an error. The `extensions` setting should only apply for files, not for directories. In a gallery element for example, you can select folders and files with image extensions. This change would break this behavior.

IMO the parent if condition should be changed to `$this->extensions && !is_dir(...)`.

src/Resources/contao/drivers/DC_Folder.php
leofeyer: This line needs to be removed entirely, because the picker will be initialized 10 lines below. I must have missed this during rebase.

src/Resources/contao/widgets/FileTree.php
leofeyer: Changed in 524a80f3cc1ab9d1dbfdfa1588154178a1a43c25.

PR #832, opened 2017-05-18
[RFC] Hide captcha field for users and add honeypot field

ausi:
With these changes we would achieve the following:

1. Normal users won’t see the captcha field and can just submit the form.
2. If a user is detected as a spam bot, they can still send the form by answering the captcha question.
3. Bots that don’t support JS, or that can’t resist the honeypot cannot send the form.

### DIFF
--- a/src/Resources/contao/templates/forms/form_captcha.html5
+++ b/src/Resources/contao/templates/forms/form_captcha.html5
@@ -17,7 +17,7 @@
   <span class="captcha_text<?php if ($this->class) echo ' ' . $this->class; ?>"><?= $this->getQuestion() ?></span>
 
   <?php if (!$this->hasErrors()): ?>
-    <div style="display: none;">
+    <div style="display:none">
       <label for="ctrl_<?= $this->id ?>_hp">Do not fill out this field</label>
       <input type="text" name="<?= $this->name ?>_name" id="ctrl_<?= $this->id ?>_hp" value="">
     </div>
--- a/src/Resources/contao/templates/forms/form_captcha.html5
+++ b/src/Resources/contao/templates/forms/form_captcha.html5
@@ -15,4 +15,16 @@
 
   <input type="text" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="captcha mandatory<?php if ($this->class) echo ' ' . $this->class; ?>" value=""<?= $this->getAttributes() ?>>
   <span class="captcha_text<?php if ($this->class) echo ' ' . $this->class; ?>"><?= $this->getQuestion() ?></span>
+
+  <?php if (!$this->hasErrors()): ?>
+    <div style="display: none;">
+      <label for="ctrl_<?= $this->id ?>_hp">Do not fill out this field</label>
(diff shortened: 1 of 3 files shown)


leofeyer:
Awesome, thank you @ausi.

src/Resources/contao/templates/forms/form_captcha.html5
```diff
@@ -15,4 +15,16 @@
 
   <input type="text" name="<?= $this->name ?>" id="ctrl_<?= $this->id ?>" class="captcha mandatory<?php if ($this->class) echo ' ' . $this->class; ?>" value=""<?= $this->getAttributes() ?>>
   <span class="captcha_text<?php if ($this->class) echo ' ' . $this->class; ?>"><?= $this->getQuestion() ?></span>
+
+  <?php if (!$this->hasErrors()): ?>
+    <div style="display:none">
+      <label for="ctrl_<?= $this->id ?>_hp">Do not fill out this field</label>
```
Toflar: In English should be `fill in this field` and I wonder if we should translate it anyway for screenreaders? Also I wonder if there's some WAI-ARIA directive to improve usability for handicapped people.

ausi: Yes, the PR wasn’t completely finished, thats why I added [RFC], not [RTM] :)
Because there is a `display:none` div directly around it, no user and no screenreader user will see that in 99.9999% of cases.
But yes, we should change it to `Do not fill in this field` and translate it.

ausi:
@leofeyer should we restructure `getQuestion()` and `getSum()` so that the it works independend of the order and how often you call these methods? Calling `getQuestion()` twice, would overwrite the session, and calling `getSum()` before `getQuestion()` doesn’t work.

src/Resources/contao/templates/forms/form_captcha.html5
leofeyer: With `display:none`, the message will never be read out by screen readers. We do not need to translate it therefore.

leofeyer: Fixed in 820ea4235e36c590d763d0099985b394e4af8287.

ausi:
Followup: #835

fenepedia:
@ausi Does this new function replace your anti-spam-extension? In other words, is it the same?

ausi:
It works with the same principles: time check, JS check and a honeypot field. So I think it should offer the same protection.

The biggest difference to the anti-spam extension is the naming of the honeypot field I think. The extension tries to use very attractive names, but this can have conflict problems with other fields in the form.

leofeyer:
We could improve the new feature like "if there is no field named `email`, name the CAPTCHA field `email`; if there is no field named `name`, name the CAPTCHA field `name`; and use the CAPTCHA field name with a suffix in all other cases". (Then the form would have to set the CAPTCHA field name I guess, because the widget does not know about the other widgets.)

Does this make sense? And is it worth the effort?

Toflar:
The "if there is no email" does not work reliably. You could determine that in the loadFormField hook in case of the form generator but the captcha is used elsewhere too. Like comments form or whatever the developers did directly with that widget. I think the honeypot field is going to work anyway.

PR #835, opened 2017-05-19
[RTM] Only generate captcha values once

ausi:
This should improve the stability of `getQuestion()` and `getSum()`. Now they can be called in any order and multiple times. See #832

### DIFF
--- a/src/Resources/contao/forms/FormCaptcha.php
+++ b/src/Resources/contao/forms/FormCaptcha.php
@@ -44,7 +44,7 @@ class FormCaptcha extends \Widget
 	/**
 	 * Captcha values
 	 *
-	 * @var string
+	 * @var array
 	 */
 	protected $arrCaptcha;
 
--- a/src/Resources/contao/forms/FormCaptcha.php
+++ b/src/Resources/contao/forms/FormCaptcha.php
@@ -42,11 +42,11 @@ class FormCaptcha extends \Widget
 	protected $strCaptchaKey;
 
 	/**
-	 * Security questions
+	 * Captcha values
 	 *
 	 * @var string
 	 */
-	protected $strQuestion;
+	protected $arrCaptcha;
 
 	/**
 	 * The CSS class prefix
@@ -116,7 +116,7 @@ class FormCaptcha extends \Widget
 				break;
 
 			case 'question':
-				return $this->strQuestion;
+				return $this->getQuestion();
 				break;
 
 			default:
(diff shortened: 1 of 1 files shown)


src/Resources/contao/forms/FormCaptcha.php
```diff
@@ -42,11 +42,11 @@ class FormCaptcha extends \Widget
 	protected $strCaptchaKey;
 
 	/**
-	 * Security questions
+	 * Captcha values
 	 *
 	 * @var string
```
leofeyer: This should be `@var array` I guess?

src/Resources/contao/forms/FormCaptcha.php
```diff
@@ -147,27 +147,44 @@ public function validate()
...
+
+
+	/**
+	 * Generate the captcha question
+	 *
+	 * @return string The question string
+	 */
+	protected function getQuestion()
+	{
+		$this->generateCaptcha();
```
leofeyer: I would prefer a more explicit check here:

```php
if (!isset($this->arrCaptcha['int1']) || !isset($this->arrCaptcha['int2']))
{
    $this->generateCaptcha();
}
```

Technically, this is not necessary of course, however, it would regenerate `$this->arrCaptcha` in case the two required properties are not set.

src/Resources/contao/forms/FormCaptcha.php
```diff
@@ -188,12 +205,9 @@ protected function getQuestion()
 	 */
 	protected function getSum()
 	{
-		/** @var SessionInterface $objSession */
-		$objSession = \System::getContainer()->get('session');
-
-		$arrCaptcha = $objSession->get('captcha_' . $this->strId);
+		$this->generateCaptcha();
```
leofeyer: I would prefer a more explicit check here:

```php
if (!isset($this->arrCaptcha['sum']))
{
    $this->generateCaptcha();
}
```

Technically, this is not necessary of course, however, it would regenerate `$this->arrCaptcha` in case the required property is not set.

src/Resources/contao/forms/FormCaptcha.php
```diff
@@ -147,27 +147,44 @@ public function validate()
...
 	/**
-	 * Generate the captcha question
-	 *
-	 * @return string The question string
+	 * Generate the captcha values and store them in the session
 	 */
-	protected function getQuestion()
+	protected function generateCaptcha()
 	{
+		if ($this->arrCaptcha !== null) {
```
leofeyer: This could be omitted if we added the more explicit check I have suggested below.

src/Resources/contao/forms/FormCaptcha.php
```diff
@@ -42,11 +42,11 @@ class FormCaptcha extends \Widget
 	protected $strCaptchaKey;
 
 	/**
-	 * Security questions
+	 * Captcha values
 	 *
 	 * @var string
 	 */
-	protected $strQuestion;
+	protected $arrCaptcha;
```
leofeyer: And this could be `protected $arrCaptcha = array()` if we added the more explicit check I have suggested.

src/Resources/contao/forms/FormCaptcha.php
ausi: Fixed in 319561511cb67d80277ecaec801256e69df64b94

ausi:
I don’t see a reason why we should move the checks if the captcha was already generated to the `getQuestion()` and `getSum()` methods. It would be duplicated code and there is no case where some of the keys of `$arrCaptcha` are there but others are missing. Even if such a case would be introduced by a child class, regenerating the captcha would not help in this case, because the previously retrieved values would not match with the new ones.

leofeyer:
Then at least set a default value `protected $arrCaptcha = array()` and check `!empty($this->arrCaptcha)` instead of `null !== $this->arrCaptcha`.

ausi:
Done in d396e26311364927b7a21363732b259edab4cf3a

PR #862, opened 2017-06-04
change sections field to blob

fritzmg:
This fixes https://github.com/contao/core/issues/8722. Due to the new section wizard the limit of 1022 characters (on MyISAM) for the sections field will be reached much sooner. Thus, depending on the length of section names, you would be limited to a small number of sections. The limit is probably even smaller on InnoDB. If you go past that limit, invalid serialized data will be saved into the field.

This PR changes the field type to `blob`, which is used for other fields with similar wizards as well.


leofeyer:
Merged in 4cd899ed98ffd1a43344c3868d7d6394da786c40.

PR #863, opened 2017-06-06
set table and sorting parameter for eliminateNestedPages

fritzmg:
`\Controller::eliminateNestedPages` is used in two new places, but without the (optional) parameters `$strTable` and `$blnSorting`. Without the former, the default table will be `tl_page`. However, this makes no sense if the table of the current DCA is something else. This will artificially limit the number of table records to be shown, depending on the IDs of your current root pages.

And due to the missing third parameter it will also display the records in the wrong order, since the sorting is omitted.

See https://github.com/codefog/contao-news_categories/issues/106 for example.

### DIFF
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -3522,12 +3522,12 @@ class DC_Table extends \DataContainer implements \listable, \editable
 					}
 
 					$arrFound = $arrRoot;
-					$this->root = $this->eliminateNestedPages($arrFound);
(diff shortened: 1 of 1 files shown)


leofeyer:
I had to change the implementation in 4087ac6d074d6d914de2f06ec43a6c11dff2af1e for another reason. Does this also solve your issue?

leofeyer:
Never mind, my changes are wrong. 😢

leofeyer:
Merged in 1fdd16f25989a8cb10c5c78c347f8ee2c4dd9fce.

PR #868, opened 2017-06-08
[RTM] Surround breacrumb with nav element

### LINKED ISSUE #786: Backend-Breadcrumb für Screenreader-Nutzer nicht erkennbar
Wenn man z.B. die Seitenstruktur verkürzt sehen will, kann man auf einen Seitennamen klicken. Danach zeigt Contao nur noch diese Seite mit ihren Unterseiten an. Ergänzend erscheint über der eigentlichen Seitenliste eine weitere Liste, die als Breadcrumb dient (ein Klick auf den ersten Eintrag „Alle“ zeigt wieder die komplette Seitenstruktur an).

Das Problem für Screenreader-Nutzer: Sie können nicht verstehen, was da geschieht. Für sie ist das einfach nur eine weitere Liste, ohne jegliche Zuordnung.

Daher möchte ich vorschlagen, dass die grundlegenden Infos ergänzt werden. Um die Liste sollte folgendes HTML ergänzt werden:
` <nav aria-label="Aktueller Standort in der verkürzten Listenansicht"> ...</nav> `(über die Formulierung der Beschriftung könnte man natürlich noch reden)

cliffparnitzky:
Closes #786


src/Resources/contao/classes/Backend.php
```diff
@@ -886,9 +886,11 @@ public static function addPagesBreadcrumb($strKey='tl_page_node')
 		// Insert breadcrumb menu
 		$GLOBALS['TL_DCA']['tl_page']['list']['sorting']['breadcrumb'] .= '
 
-<ul id="tl_breadcrumb">
-  <li>' . implode(' &gt; </li><li>', $arrLinks) . '</li>
-</ul>';
+<nav aria-label="' . $GLOBALS['TL_LANG']['MSC']['breadcrumbAriaLabel'] . '" role="navigation">
```
leofeyer: Why did you add the `aria-label` tag? `<nav role="navigation">` should suffice for any screen reader.

src/Resources/contao/languages/en/default.xlf
```diff
@@ -1679,6 +1679,9 @@ This e-mail has been generated by Contao. You can not reply to it directly.
       <trans-unit id="MSC.metaNoLanguages">
         <source>You have to add a root page in the site structure to store file meta information.</source>
       </trans-unit>
+      <trans-unit id="MSC.breadcrumbAriaLabel">
+        <source>Current location in the shortened list view</source>
```
leofeyer: This label says nothing and should be removed.

src/Resources/contao/classes/Backend.php
NinaG: If there are several navigation elements on one page, it is very helpful for screen reader users if you label them differently so that the user is able to distinguish their usage. Something like "main navigation", "breadcrumb navigation" etc.

src/Resources/contao/languages/en/default.xlf
NinaG: It makes a lot of sense for screen reader users.

leofeyer: It does not. Following your explanation above, the label making sense would be "pagination menu" and not "Current location in the shortened list view".

NinaG: It is not the pagination menu here, but the breadcrumb of the current page. So I think it would be enough help to call it: "Your current location" or "Your current location in Contao backend".

leofeyer: Or simply "breadcrumb menu".

aschempp:
I wonder if generic labels make sense? Or if they should be specific to the module, like either with a new field or a template adjustment? Because I do have multiple navigation modules on a lot of sites (e.g. header + footer) and they would get the same label…

NinaG:
This template should be used ONLY for the breadcrumb. That way I see no problem with a generic label.

aschempp:
Who says I only have one breadcrumb? And the same applies to #869

NinaG:
Seriously? You've got several breadcrumbs and several paginations on the same page in the Contao Backend? We are talking about the backend here, aren't we?

If we are also talking about the frontend, it would make sense that the label correspondends to the module, i.e. "Pagination Module X" or "Breadcrumb Module Y".

aschempp:
ah, sorry, I did not realize it's about the backend 🙈

cliffparnitzky:
Ok, thanks. To finalize the discussion, here my summary of changes:

- remove german translation
- change the aria-label value into "Breadcrumb menu"

Correct?

leofeyer:
Why did you close the PR?

cliffparnitzky:
oh f*ck, this was an accident ... sorry. Hope I can restore it.

cliffparnitzky:
I had to create a new PR, because there where al lot of problemes in my local dev environment. Hope thats okay. Follow Up PR is #1036

PR #869, opened 2017-06-08
[RFC] Surround pagination with nav element

### LINKED ISSUE #775: Backend-Pagination barrierefreier auszeichnen
Ich schlage für die be_pagination.html5 folgende Änderungen vor, damit sie für assistive Technologien (AT wie z.B. Screenreader) besser zugänglich ist:

Die Pagination sollte von einem NAV-Element umhüllt, sowie mit einer entsprechenden role und einem ARIA-Label versehen werden.

Im Template sähe das dann so aus:
```
<div class="pagination <?= $this->class ?>">

<nav role="navigation" aria-label="Pagination">
  <p><?= $this->total ?></p>
  <ul>
    <?php if ($this->hasFirst): ?>
      <li class="first"><a href="<?= $this->first['href'] ?>" class="first" title="<?= $this->first['title'] ?>"><?= $this->first['link'] ?></a></li>
    <?php endif; ?>
    <?php if ($this->hasPrevious): ?>
      <li class="previous"><a href="<?= $this->previous['href'] ?>" class="previous" title="<?= $this->previous['title'] ?>"><?= $this->previous['link'] ?></a></li>
    <?php endif; ?>
    <?= $this->items ?>
    <?php if ($this->hasNext): ?>
      <li class="next"><a href="<?= $this->next['href'] ?>" class="next" title="<?= $this->next['title'] ?>"><?= $this->next['link'] ?></a></li>
    <?php endif; ?>
    <?php if ($this->hasLast): ?>
      <li class="last"><a href="<?= $this->last['href'] ?>" class="last" title="<?= $this->last['title'] ?>"><?= $this->last['link'] ?></a></li>
    <?php endif; ?>
  </ul>
</nav>

</div>
```

// Aktualisiert, siehe nachfolgendes Kommentar von mir

cliffparnitzky:
Closes #775


src/Resources/contao/templates/backend/be_pagination.html5
```diff
@@ -1,22 +1,24 @@
...
-    <?php endif; ?>
-    <?= $this->items ?>
-    <?php if ($this->hasNext): ?>
-      <li class="next"><a href="<?= $this->next['href'] ?>" class="next" title="<?= $this->next['title'] ?>"><?= $this->next['link'] ?></a></li>
-    <?php endif; ?>
-    <?php if ($this->hasLast): ?>
-      <li class="last"><a href="<?= $this->last['href'] ?>" class="last" title="<?= $this->last['title'] ?>"><?= $this->last['link'] ?></a></li>
-    <?php endif; ?>
-  </ul>
+  <nav role="navigation" aria-label="<?= $GLOBALS['TL_LANG']['MSC']['paginationAriaLabel'] ?>">
```
leofeyer: I would prefer to replace the surrounding `<div>` with the `<nav>` element instead of adding another nesting level.

src/Resources/contao/languages/en/default.xlf
```diff
@@ -1679,6 +1679,9 @@ This e-mail has been generated by Contao. You can not reply to it directly.
       <trans-unit id="MSC.metaNoLanguages">
         <source>You have to add a root page in the site structure to store file meta information.</source>
       </trans-unit>
+      <trans-unit id="MSC.paginationAriaLabel">
+        <source>Pagination</source>
```
leofeyer: Again, what is the added value here?

NinaG: If there are several navigation elements on one page, it is very helpful for screen reader users if you label them differently so that the user is able to distinguish their usage. Something like "main navigation", "breadcrumb navigation" etc.

cliffparnitzky:
Ok, thanks. To finalize the discussion, here my summary of changes:

- remove german translation
- replace the surrounding `<div>` with the `<nav>` element (add the `class` attribute of the `div`)

Correct?

leofeyer:
I guess we should also add a [skip link](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/templates/backend/be_main.html5#L57) while we are on it.

cliffparnitzky:
I have replaced the surrounding div with the nav element.

I also added the skip link. The probleme here is the generic label. I don't know if it is the best solution using the `$GLOBALS` in the template. If not, the part in `DC_Table` and `Versions` has to be modified.

src/Resources/contao/templates/backend/be_pagination.html5
```diff
@@ -1,8 +1,9 @@
 
-<div class="pagination <?= $this->class ?>">
+<nav class="pagination <?= $this->class ?>" role="navigation" aria-label="<?= $GLOBALS['TL_LANG']['MSC']['paginationAriaLabel'] ?>">
+  <a href="<?= $this->request ?>#skipNavigation" class="invisible"><?= \StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['skipNavigation']); ?></a>
 
   <p><?= $this->total ?></p>
-
+
```
leofeyer: I think you have to revert this change.

src/Resources/contao/languages/en/default.xlf
leofeyer: @cliffparnitzky Can you change the label to "Pagination menu" please?

src/Resources/contao/templates/backend/be_pagination.html5
cliffparnitzky: @leofeyer whats exactly? The empty row, the blanks or the complete template change?

leofeyer: The line indentation.

cliffparnitzky: Okay, removed it.

src/Resources/contao/templates/backend/be_pagination.html5
```diff
@@ -1,5 +1,6 @@
 
-<div class="pagination <?= $this->class ?>">
+<nav class="pagination <?= $this->class ?>" role="navigation" aria-label="<?= $GLOBALS['TL_LANG']['MSC']['paginationAriaLabel'] ?>">
+  <a href="<?= $this->request ?>#skipNavigation" class="invisible"><?= \StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['skipNavigation']); ?></a>
```
leofeyer: This is not correct. First of all, you must not use a hard coded skip ID, because there might be more than one pagination menu and it would lead to ambiguous IDs. Second of all, you also have to add the target element: 

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/templates/modules/mod_navigation.html5#L13

cliffparnitzky: At first sorry for my late reply.

To your statement above: I think thats the same like her: https://github.com/contao/core-bundle/pull/868#issuecomment-307322372

We are talking about the backend here, aren't we?

cliffparnitzky: @leofeyer Am i right, or is anything to do here?

leofeyer: > or is anything to do here?

Of course there is:

> First of all, you must not use a hard coded skip ID, because there might be more than one pagination menu and it would lead to ambiguous IDs. Second of all, you also have to add the target element:

None of this has been fixed, has it?

cliffparnitzky: @leofeyer What about my comment?

> We are talking about the backend here, aren't we?

leofeyer: There might be more than one pagination menu in the back end, too.

cliffparnitzky: @leofeyer Now, I've added the target element for the naviation skipping.

BUT: I'm still of the opinion, that multiple paginations in the back end are not possible. In the current implementation there is no unique identifier for a pagination. And scrolling will be done via URL parameter `lp` ... if there is more than one pagination no decision can be done, which list to scroll.

leofeyer:
Implemented in 432ca9d77cbf0b1d2dcc8035a765cec6df172331.

PR #871, opened 2017-06-13
BC compat for tl_content::pagePicker

aschempp:
The `tl_content::pagePicker` method is used by any extension that wanted to add a page picker before Contao 4.4. I know `tl_content` is not a BC layer, but we sure should try to prevent such issues if we can…

### DIFF
--- a/src/Resources/contao/classes/Backend.php
+++ b/src/Resources/contao/classes/Backend.php
@@ -1099,9 +1099,9 @@ abstract class Backend extends \Controller
 			$params['context'] = $config['context'];
 		}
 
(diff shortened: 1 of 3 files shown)


leofeyer:
You mean that people have added

```php
'wizard' => array
(
	array('tl_content', 'pagePicker')
),
```

in their DCAs?

aschempp:
correct, on their custom fields in `tl_content`.

aschempp:
I just fixed another bug in https://github.com/contao/core-bundle/pull/871/commits/ba30baaba92c7bf6b5450699293886a10d477ee4. On multi-edit the field name would result in duplicate ID attributes.

leofeyer:
Merged in 480e16df58e88d62cf155b834d901d572bbd6cb8.

PR #880, opened 2017-06-17
Show Symfony error message when in debug mode

bytehead:
Show the built-in Symfony error page (with stack trace etc) when the application is in `DEV` mode.
In `DEV` mode I don't want to have to look at the log files when I can have the Symfony error report right in the browser.

So it will basically display an error screen like this:
bildschirmfoto 2017-06-17 um 18 26 50

### DIFF
--- a/src/EventListener/PrettyErrorScreenListener.php
+++ b/src/EventListener/PrettyErrorScreenListener.php
@@ -87,7 +87,6 @@ class PrettyErrorScreenListener
      * @param ContaoFrameworkInterface $framework
      * @param TokenStorageInterface    $tokenStorage
      * @param LoggerInterface|null     $logger
-     * @param bool                     $debug
      */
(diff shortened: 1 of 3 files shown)


aschempp:
This should already be the case by the `$prettyErrorScreens` flag which is set according to `debug`?

bytehead:
Didn't check - but it doesn't work anyway.

bytehead:
I'll check.

aschempp:
Can you find out why this is not already the case in https://github.com/bytehead/core-bundle/blob/4d59c9a177c2817aaf02baf18a44823f3dc1a79f/src/EventListener/PrettyErrorScreenListener.php#L261 ?

bytehead:
It actually works in the COD (I'll have to check in my custom application).
But why returning late and not [here](https://github.com/bytehead/core-bundle/blob/4d59c9a177c2817aaf02baf18a44823f3dc1a79f/src/EventListener/PrettyErrorScreenListener.php#L107)?

aschempp:
not sure… I think exception logging is still done in the listener then?

bytehead:
Looks like that.
Thanks anyway - I'll close this.

PR #887, opened 2017-06-20
[WIP] Fix CORS detection

fritzmg:
Fix for https://github.com/contao/installation-bundle/issues/56, see [explanation](https://github.com/contao/installation-bundle/issues/56#issuecomment-309824158).

This PR changes the CORS detection to the same methodology as used in [nelmio/NelmioCorsBundle/EventListener/CorsListener.php#L68](https://github.com/nelmio/NelmioCorsBundle/blob/1.5.3/EventListener/CorsListener.php#L68).

This way the Install Tool works in browsers like Chrome and Safari as well (who set the `Origin` header on `POST` requests).

### DIFF
--- a/src/Cors/WebsiteRootsConfigProvider.php
+++ b/src/Cors/WebsiteRootsConfigProvider.php
@@ -69,7 +69,7 @@ class WebsiteRootsConfigProvider implements ProviderInterface
      */
     private function isCors(Request $request)
     {
-        return $request->headers->has('Origin') && $request->headers->get('Origin') != $request->getSchemeAndHttpHost();
+        return $request->headers->has('Origin') && $request->headers->get('Origin') !== $request->getSchemeAndHttpHost();
     }
 
     /**
--- a/src/Cors/WebsiteRootsConfigProvider.php
+++ b/src/Cors/WebsiteRootsConfigProvider.php
@@ -41,7 +41,7 @@ class WebsiteRootsConfigProvider implements ProviderInterface
      */
     public function getOptions(Request $request)
     {
-        if (!$this->isCors($request) || !$this->canRunDbQuery()) {
+        if (!$this->isCorsRequest($request) || !$this->canRunDbQuery()) {
             return [];
         }
 
@@ -67,9 +67,11 @@ class WebsiteRootsConfigProvider implements ProviderInterface
      *
      * @return bool
      */
-    private function isCors(Request $request)
+    private function isCorsRequest(Request $request)
     {
-        return $request->headers->has('Origin') && $request->headers->get('Origin') !== $request->getSchemeAndHttpHost();
+        return $request->headers->has('Origin')
+            && $request->headers->get('Origin') !== $request->getSchemeAndHttpHost()
+        ;
     }
 
     /**
--- a/src/Cors/WebsiteRootsConfigProvider.php
+++ b/src/Cors/WebsiteRootsConfigProvider.php
@@ -41,7 +41,7 @@ class WebsiteRootsConfigProvider implements ProviderInterface
      */
     public function getOptions(Request $request)
     {
-        if (!$this->hasOrigin($request) || !$this->canRunDbQuery()) {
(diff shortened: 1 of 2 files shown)


fritzmg:
However, this fix is incomplete, since the problem should have prevented by
```php
/**
 * Checks if the database connection and the table exist.
 *
 * @return bool
 */
private function canRunDbQuery()
{
    return $this->connection->isConnected() && $this->connection->getSchemaManager()->tablesExist(['tl_page']);
}
```
But `$this->connection->isConnected()` returns `true` in the Contao Managed Edition, since the `manager-bundle` already provides basic database information, that makes this function return `true`.

src/Cors/WebsiteRootsConfigProvider.php
```diff
@@ -67,9 +67,9 @@ public function getOptions(Request $request)
      *
      * @return bool
      */
-    private function hasOrigin(Request $request)
+    private function isCors(Request $request)
     {
-        return $request->headers->has('Origin') && '' !== $request->headers->get('Origin');
+        return $request->headers->has('Origin') && $request->headers->get('Origin') != $request->getSchemeAndHttpHost();
```
ausi: I think this should be `!==` instead of `!=`

fritzmg: True, done.

aschempp:
Just FYI, @Toflar is currently on holidays.

Toflar:
I think this check should be done by the CORS bundle (maybe we need to adjust the settings in our provider though). Will check in 2 weeks, thanks for taking care anyway :)

fritzmg:
If you mean `NelmioCorsBundle`, then that bundle already makes this check. However, it makes this check __after__ `$this->configurationResolver->getOptions($request)` (see [nelmio/NelmioCorsBundle/EventListener/CorsListener.php#L58](https://github.com/nelmio/NelmioCorsBundle/blob/1.5.3/EventListener/CorsListener.php#L58)), which is a call to `Contao\CoreBundle\Cors\WebsiteRootsConfigProvider->getOptions`, which in turn causes the problem.

leofeyer:
Apparently I forgot to adjust the CORS fix from Contao 3.5 for Contao 4.4.

https://github.com/contao/core-bundle/commit/2f5fff09b86ac55065a24a634b1a9733fdc84471

fritzmg:
> Apparently I forgot to adjust the CORS fix from Contao 3.5 for Contao 4.4: 2f5fff0

This is an incomplete fix imho. It still assumes that any request where the `Origin` header is set is a CORS request. It does not matter as much in Contao 3, because right after the request method is checked:
```php
$arrAllowedMethods = array('GET', 'HEAD', 'OPTIONS');
```
So `POST` requests from Webkit browsers (where the `Origin` header is set) are still ignored.

This is not the case in Contao 4.4.0 currently. Reproduction:

1. Generate a form with the form generator.
2. Add the form to a page.
3. Open the page both with FireFox and Chrome and with `app_dev.php`.
4. Submit the form in both browsers.
5. Look at the database queries.

In Chrome there will be two additional database queries:
```MySQL
SHOW FULL TABLES WHERE Table_type = 'BASE TABLE'
SELECT id FROM tl_page WHERE type='root' AND dns=:dns
```
since it sets the `Origin` header and thus `Contao\CoreBundle\Cors\WebsiteRootsConfigProvider->getOptions` will be executed where `hasOrigin` will be true.

leofeyer:
@Toflar Can you confirm that the fix is correct?

src/Cors/WebsiteRootsConfigProvider.php
```diff
@@ -67,9 +67,9 @@ public function getOptions(Request $request)
      *
      * @return bool
      */
-    private function hasOrigin(Request $request)
+    private function isCors(Request $request)
```
Toflar: I would rename this to `isCorsRequest`.

leofeyer:
Merged in 7acb61d9afc49bd6de1734435a11c77b43ddcc7c actually.

PR #889, opened 2017-06-22
Fix install tool displaying "incomplete installation" exception

aschempp:
Due to your MySQL server setup (e.g. allow anonymous connection), the existing check was successful but the call to `tableExists` results in an exception.


leofeyer:
I have also added the try-catch block to the `CommandSchedulerListeren` class in 9fe43824a71eac874f10ba09217dfdad2c7da0f8.

PR #900, opened 2017-06-26
Correctly handle active menu states in the new menu markup

aschempp:
Since Contao 4.4, the backend menu requires extensions to always render submenu items. I have fixed this in the core by always passing `$blnShowAll=true` to the hooks and then handle collapsing in the navigation method. This also correctly marks active items if they follow the default naming, which were previously ignored in extensions.

### DIFF
--- a/src/Resources/contao/classes/BackendUser.php
+++ b/src/Resources/contao/classes/BackendUser.php
@@ -484,7 +484,7 @@ class BackendUser extends \User
 				$arrModules[$strGroupName]['label'] = (($label = is_array($GLOBALS['TL_LANG']['MOD'][$strGroupName]) ? $GLOBALS['TL_LANG']['MOD'][$strGroupName][0] : $GLOBALS['TL_LANG']['MOD'][$strGroupName]) != false) ? $label : $strGroupName;
 				$arrModules[$strGroupName]['href'] = $router->generate('contao_backend', array('do'=>\Input::get('do'), 'mtg'=>$strGroupName, 'ref'=>TL_REFERER_ID));
 				$arrModules[$strGroupName]['ajaxUrl'] = $router->generate('contao_backend');
-				$arrModules[$strGroupName]['icon'] = 'modPlus.gif';
+				$arrModules[$strGroupName]['icon'] = 'modPlus.gif'; // backwards compatibility with e.g. EasyThemes
 
 				foreach ($arrGroupModules as $strModuleName=>$arrModuleConfig)
 				{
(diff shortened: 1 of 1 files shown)


src/Resources/contao/classes/BackendUser.php
```diff
@@ -484,15 +484,7 @@ public function navigation($blnShowAll=false)
...
-				$arrModules[$strGroupName]['isClosed'] = false;
-
-				// Do not show the modules if the group is closed
-				if (!$blnShowAll && isset($session['backend_modules'][$strGroupName]) && $session['backend_modules'][$strGroupName] < 1)
-				{
-					$arrModules[$strGroupName]['class'] = ' node-collapsed';
-					$arrModules[$strGroupName]['title'] = \StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['expandNode']);
-					$arrModules[$strGroupName]['isClosed'] = true;
-				}
+				$arrModules[$strGroupName]['icon'] = 'modPlus.gif';
```
leofeyer: Why did you add this? We are no longer using the GIF icons.

src/Resources/contao/classes/BackendUser.php
```diff
@@ -523,7 +508,33 @@ public function navigation($blnShowAll=false)
 			foreach ($GLOBALS['TL_HOOKS']['getUserNavigation'] as $callback)
 			{
 				$this->import($callback[0]);
-				$arrModules = $this->{$callback[0]}->{$callback[1]}($arrModules, $blnShowAll);
+				$arrModules = $this->{$callback[0]}->{$callback[1]}($arrModules, true);
+			}
+		}
+
+		foreach ($arrModules as $strGroupName => $arrGroupModules)
```
leofeyer: Why did you move the code?

src/Resources/contao/classes/BackendUser.php
aschempp: Because that was the/a way to detect if a menu is active in previous versions. [EasyThemes](https://github.com/terminal42/contao-easy_themes) relies on it to know if a menu is open.

src/Resources/contao/classes/BackendUser.php
aschempp: Because that code needs to be applied for new menus added by the hook as well?

src/Resources/contao/classes/BackendUser.php
leofeyer: But we have removed this in version 4.2 already (see f49802b0facc6f7e93411936789b025c7aa835cc) and I don't think that we should re-add it.

src/Resources/contao/classes/BackendUser.php
```diff
@@ -523,7 +508,33 @@ public function navigation($blnShowAll=false)
...
+		{
+			$arrModules[$strGroupName]['isClosed'] = false;
+
+			// Do not show the modules if the group is closed
+			if (!$blnShowAll && isset($session['backend_modules'][$strGroupName]) && $session['backend_modules'][$strGroupName] < 1)
+			{
+				$arrModules[$strGroupName]['class'] = str_replace('node-expanded', '', $arrModules[$strGroupName]['class']) . ' node-collapsed';
```
leofeyer: Why not `str_replace('node-expanded', 'node-collapsed', $arrModules[$strGroupName]['class'])`?

aschempp: Because `node-expanded` does not necessarily exist.

leofeyer:
Thank you @aschempp.

PR #906, opened 2017-06-28
Allow recursive dca picker in the same table

### LINKED ISSUE #883: Seitenstruktur: Interne Weiterleitung zeigt Modal ohne Radio-Buttons
Wenn ich eine interne Weiterleitung anlegen will, fehlen die Radio-Buttons, um eine Seite auszuwählen, siehe Screenshot. 
modal-interne-weiterleitung

aschempp:
fixes #883

Unfortunately, there's no cleaner solution…

Also tested to edit a page within the page picker and pick another page for the redirect 💪

### DIFF
--- a/src/Resources/contao/classes/DataContainer.php
+++ b/src/Resources/contao/classes/DataContainer.php
@@ -882,9 +882,14 @@ abstract class DataContainer extends \Backend
 		list($this->strPickerTable, $this->strPickerField, $this->intPickerId) = explode('.', \Input::get('target'), 3);
 		\Input::setGet('target', null);
 
+		if (empty($this->strPickerTable) || empty($this->strPickerField) || empty($this->intPickerId))
+		{
+			return false;
+		}
+
 		$menuBuilder = \System::getContainer()->get('contao.menu.picker_menu_builder');
 
-		if (empty($this->strPickerTable) || empty($this->strPickerField) || empty($this->intPickerId) || !$menuBuilder->supportsTable($this->strTable))
+		if (!$menuBuilder->supportsTable($this->strTable))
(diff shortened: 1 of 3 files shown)


src/Resources/contao/classes/DataContainer.php
```diff
@@ -874,14 +874,21 @@ protected function generateGlobalButtons()
...
+			return false;
+		}
+
+		list($this->strPickerTable, $this->strPickerField, $this->intPickerId) = explode('.', \Input::get('target'), 3);
+		\Input::setGet('target', null);
+
 		$menuBuilder = \System::getContainer()->get('contao.menu.picker_menu_builder');
 
-		if (!$menuBuilder->supportsTable($this->strTable))
+		if (!$menuBuilder->supportsTable($this->strTable) || empty($this->strPickerTable) || empty($this->strPickerField) || empty($this->intPickerId))
```
leofeyer: We should run those checks before we initialize the menu builder, shouldn't we?

aschempp: Agree, changed in 4665cd5cd7c82dc688726cfbcfba35288c29913a

leofeyer:
For some reason, this change breaks the link picker (TinyMCE).

leofeyer:
It is the `empty($this->intPickerId)` part. Why must the picker ID be set?

leofeyer:
Also, the picker is not applied correctly anymore now:



There should be no yellow background in the news archive view.

leofeyer:
Reverted and fixed in 887bac78b8cf12a02864a97bab8a724324356f70 now.

Issue #747, opened 2017-03-31
Adding a ArticleTree widget (like PageTree)


agoat:
I think to be able to select articles like pages (e.g. in the content element article and teaser) would be a big benefit!!

Currently there is only a simple select field to select articles. Very difficult to select an article when having a big page structure. And only the teaser element restricts the selection to the current root page tree.

I can start a PR for this, but I think I will need help from the core-team. After a quick try I realized, that this isn't a simple task..

Is this of interest (also for other extension developers)?

Toflar:
> Is this of interest (also for other extension developers)?
In 9 years I have never ever had the need to pick a single article. I have never used this teaser feature so I cannot really help here.
But @leofeyer is working on a rather general picker solution for events, news etc. so maybe implementing this gets easier. You should maybe wait for this 😄

agoat:
I don't use the content elements to insert article or teaser but I use to article select field in extensions (e.g. to select related articles or so).

> But @leofeyer is working on a rather general picker solution for events, news etc. so maybe implementing this gets easier.

That sounds interesting ..

leofeyer:
> But @leofeyer is working on a rather general picker solution for events, news etc.

I am working on a rather specific picker for new and events only. We have already discussed and discarded implementing a general picker by adjusting the `DC_Table` class.

But at least my implementation will make it easier to add a custom link to the picker menu.

asaage:
> I am working on a rather specific picker for new and events only.

👍 

FAQ & Articles should be added to cover all core-entities that provide a distinct url.

leofeyer:
Feel free to create a pull request. 👍

agoat:
> But at least my implementation will make it easier to add a custom link to the picker menu.

Will we have some general structure for tree/list picker functions?

I don't know the discussion about implementing a general picker and therefor not the reason to discard the implementation of a general picker. But I think a general method would be helpful. 
Currently the `TreePage` widget uses the `PageSelection` widget to generate the page tree. But the PageSelection is a customized duplication of what the DC_Table is generating for the page structure menu. For a ArticleTree picker we will have nearly the same, and also for news, events, a.s.o.

Is it possible to have some helper methods (render the tree view of different tables - page+article, news_archiv+news, calender+calendar_event)??

leofeyer:
Implemented in #755.

PR #912, opened 2017-06-30
[RFC] Feature/articletree widget

agoat:
This PR adds an article tree widget as suggested in #747. It allows selecting articles the same way as pages with the new dca picker (for the content elements `Article` and `Article teaser`).

It depends on the PR #906. Otherwise no articles can be selected.


agoat:
It looks like I have to rework this PR as there will be a lot of changes to the dcaPicker introduced in the next bugfix release.

leofeyer:
> as there will be a lot of changes to the dcaPicker introduced in the next bugfix release

Probably only in the next minor version (Contao 4.5).

agoat:
As the dca picker will be rewritten (#950), I will wait for the new implementation and adjust this PR accordingly.

agoat:
I close this in favour for #999.

PR #920, opened 2017-07-03
Wrong if condition in new record check in tl_page.php

agoat:
During development I got some odd results and found a wrong if condition in the `tl_page.php`.

The `generateArticle` method is called with the `onsubmit_callback`. It then checks if there is a new record and adds an article for the page.

The second if condition checking for a new record is wrong. The mistake comes only into account, if there is another new record from another table (e.g. tl_style_sheet) in the `new_record` bag.

As the if statements are processed sequentially, no extra brackets are necessary here.


leofeyer:
Thank you @agoat.

PR #923, opened 2017-07-04
Fixed support doctrine boolean types for subpalette checkboxes

Toflar:


src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3252,7 +3252,7 @@ public function getPalette()
 
 					if ($trigger != '')
```
qzminski: Maybe it would be just better to have ```if ($trigger) {``` here...

Toflar:
@qzminski I wasn't sure if it does have an effect on the `else` and I didn't want to break anything 😄  But I analyzed it more in depth now and you're right, it works exactly the same. Updated the PR, works like a charm 😄

aschempp:
This only works for subpalettes now, right? Not for Checkbox widgets etc…

Toflar:
> This only works for subpalettes now, right? Not for Checkbox widgets etc…

? A checkbox is a subpalette?

Issue #886, opened 2017-06-20
Contao 4.4: new DCA picker does not work for non-admins


fritzmg:
### Reproduction

1. Log-in as a non-admin user.
2. Create a new text element and write some text.
3. Select the text and choose _Edit/insert link_.
4. Click on the "browse" icon.

You will see the message "Forbidden - Invalid content element ID".

### Cause

The picker popup is opened with the following GET parameters:

```
array(5) {
  ["target"] => string(20) "tl_content.singleSRC"
  ["value"] => string(2) "10"
  ["context"] => string(4) "link"
  ["popup"] => string(1) "1"
  ["do"] => string(4) "page"
}
```
However, the function `tl_content::checkPermissions` requires an `id` parameter, containing the `id` of the article. Since this is missing, `tl_content::checkAccessToElement` is called with `NULL` for `$id` and thus the `AccessDeniedException` with `Invalid content element ID ` is thrown.

It still works for admins, since no such permissions are checked for admins of course.

ausi:
I can confirm this issue in Contao 4.4.0. It also happens with the page picker for the “Image link target” field where the `target` parameter is set to something like `tl_content.imageUrl.123` where `123` is the ID of the content element.

rabauss:
we have similiar problems with the new dca-picker in the isotope filetree popup.
All users (also admins) get permission errors ("File or folder "" is not within the files directory.") for actions in the filetree for product attributes.
we discovered, that without the get-parameter 'target' the actions are allowed.

for uploads via dropzone there is a backlink missing in the lightbox!

leofeyer:
@contao/developers I have no idea how to fix this except removing the permission check here:

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/dca/tl_content.php#L902

But that also seems like a bad idea. 😄

leofeyer:
Here's another solution. We could replace [this line](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/classes/DataContainer.php#L892) with:

```php
$ref = new \ReflectionClass($strDriver);
$objDca = $ref->newInstanceWithoutConstructor();
```

This would prevent the `onload_callback` from being executed. Side effects are possible though.

leofeyer:
Here's the PR as discussed: #926

PR #926, opened 2017-07-05
Picker without a DC object

leofeyer:
This PR is a possible fix for #886.

### DIFF
--- a/src/Resources/contao/classes/DataContainer.php
+++ b/src/Resources/contao/classes/DataContainer.php
@@ -1043,9 +1043,14 @@ class DataContainer extends \Backend
 	/**
 	 * Return the name of the current palette
 	 *
+	 * This method is not abstract, because you can use this class as a concrete
+	 * class if you need to pass a DataContainer object as callback argument.
+	 *
 	 * @return string
 	 *
 	 * @throws \Exception
+	 *
+	 * @see initPicker()
 	 */
 	public function getPalette()
 	{
@@ -1055,9 +1060,16 @@ class DataContainer extends \Backend
 	/**
 	 * Save the current value
 	 *
+	 * This method is not abstract, because you can use this class as a concrete
+	 * class if you need to pass a DataContainer object as callback argument.
+	 *
+	 * @param mixed $varValue
+	 *
 	 * @throws \Exception
+	 *
+	 * @see initPicker()
 	 */
-	protected function save()
+	protected function save($varValue)
 	{
 		throw new \LogicException('You must override the save() method in the concrete data container class.');
 	}
--- a/src/Resources/contao/classes/DataContainer.php
+++ b/src/Resources/contao/classes/DataContainer.php
@@ -31,7 +31,7 @@ use Imagine\Gd\Imagine;
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
-class DataContainer extends \Backend
+abstract class DataContainer extends \Backend
 {
 
 	/**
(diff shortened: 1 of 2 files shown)


aschempp:
I would prefer to not make the `DataContainer` abstract anymore, and add the method stubs with exceptions like Symfony commands do.

leofeyer:
Then we'd lose the method signature, wouldn't we? Otherwise fine with me.

Toflar:
Guess like so:

```diff
diff --git a/src/Resources/contao/classes/DataContainer.php b/src/Resources/contao/classes/DataContainer.php
index 67356fc..fd7e1ee 100644
--- a/src/Resources/contao/classes/DataContainer.php
+++ b/src/Resources/contao/classes/DataContainer.php
@@ -27,7 +27,7 @@ use Contao\CoreBundle\Exception\AccessDeniedException;
  *
  * @author Leo Feyer <https://github.com/leofeyer>
  */
-abstract class DataContainer extends \Backend
+class DataContainer extends \Backend
 {
 
 	/**
@@ -817,8 +817,13 @@ abstract class DataContainer extends \Backend
 	 * Return the name of the current palette
 	 *
 	 * @return string
+	 *
+	 * @throws \Exception
 	 */
-	abstract public function getPalette();
+	public function getPalette()
+	{
+ 		throw new \Exception('You must override the getPalette() method in the concrete DataContainer class.');
+	}
 
 	/**
 	 * Save the current value
@@ -827,5 +832,8 @@ abstract class DataContainer extends \Backend
 	 *
 	 * @throws \Exception
 	 */
-	abstract protected function save($varValue);
+	protected function save($varValue)
+	{
+		throw new \Exception('You must override the save() method in the concrete DataContainer class.');
+	}
 }
```

I'm happy with that too 👍

leofeyer:
I have updated the PR. There is a warning regarding the method signature now:

leofeyer:
Also, the class has public `help()` method now.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -1043,15 +1044,21 @@ protected function getPickerInputField($value, $attributes='')
...
 
 	/**
 	 * Save the current value
 	 *
-	 * @param mixed $varValue
-	 *
 	 * @throws \Exception
 	 */
-	abstract protected function save($varValue);
+	protected function save()
```
Toflar: Imho `$varValue` has to remain here. You can ignore the phpStorm inspection.

leofeyer: Why should it remain? It is not used.

Toflar: It is in subclasses though. The `Command` class in Symfony is exactly the same.

leofeyer: Re-added in 6a01448a2193791b130d70f5b953a82d715ba1c6.

aschempp: `$varValue` is part of the "interface", so it sure should remain.

aschempp: Also, the value is passed in the `DataContainer` class.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -902,8 +902,9 @@ protected function initPicker()
 
 		$this->setPickerValue();
 
-		$strDriver = 'DC_' . $GLOBALS['TL_DCA'][$this->strPickerTable]['config']['dataContainer'];
-		$objDca = new $strDriver($this->strPickerTable);
+		// Create a DataContainer object (see #886)
+		$objDca = new \DataContainer();
```
aschempp: This will be an issue because the `onload_callback` is no longer executed… And no, it's not executed in the real table, because most of the time the `strPickerTable` will be different for the popup.

leofeyer: You are wrong. The callbacks are executed for the real DCA object, making sure that all permissions are checked and that the user only sees their allowed modules. The helper object is only used in the callbacks.

> because most of the time the `strPickerTable` will be different for the popup

That's exactly why we **must not** trigger the callbacks. You cannot run the permission checks for `target=tl_content.singleSRC.2` in e.g. an `do=news` environment.

leofeyer: Also, please remember that you solved the **exact same case** in Haste with a `strClass` object (see https://github.com/codefog/contao-haste/pull/75). This should illustrate that we only need a helper object.

aschempp: Unfortunately, the `onload_callback` does not only check permissions, it might also alter the DCA, add new fields etc. that are important when listing or editing the records.

leofeyer: All of that has been done already by the "real" DCA.

leofeyer:
@ausi I have updated the PR as discussed. Please check if you can use the "loadDataContainer" hook to add your custom fields.

ausi:
@leofeyer Yes, creating the field via the `loadDataContainer` hook works. 🎉

agoat:
Why is the `load_callback` hook not called anymore ([removed in this PR](https://github.com/contao/core-bundle/pull/926/files#diff-b187da7f3bf59a72fc3327dbb0246dd4L927))? 
This may result in 'Target field "tl_content.virtualField" does not exist.'.

Is there a reason why the hook isn't called now? 
If the hook is used to add virtual fields (using the picker), the check for an existing picker field will currently fail, as the field only exists if the hook is called.

agoat:
Ah, sorry. I have just seen that there will be another solution from PR #934.

aschempp:
see #942

PR #930, opened 2017-07-05
Removed access to $strPickerTable instead of DcaFilterInterface all over DC files

aschempp:


src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2751,7 +2757,7 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 
 				if ($this->strPickerField)
 				{
-					$_buttons .= $this->getPickerInputField($currentEncoded, ($GLOBALS['TL_DCA'][$this->strPickerTable]['fields'][$this->strPickerField]['eval']['files'] || $GLOBALS['TL_DCA'][$this->strPickerTable]['fields'][$this->strPickerField]['eval']['filesOnly']) ? '' : ' disabled');
+					$_buttons .= $this->getPickerInputField($currentEncoded);
```
leofeyer: Why did you remove the `$this->blnFilesOnly ? '' : 'disabled'` part?

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2571,7 +2577,7 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 				{
 					--$countFiles;
 				}
-				elseif ($this->blnHideFiles && !is_dir(TL_ROOT . '/' . $currentFolder . '/' . $file))
+				elseif (!$this->blnShowFiles && !is_dir(TL_ROOT . '/' . $currentFolder . '/' . $file))
```
leofeyer: Why did you change the "hide files" logic?

src/Resources/contao/drivers/DC_Folder.php
aschempp: Because that line is only executed if there are files, and if there are files they are always selectable. `blnFilesOnly` only defines that folders are not to be selected.

src/Resources/contao/drivers/DC_Folder.php
aschempp: That's how it was always implemented. It's `eval => ['files' => true]`, and by default it's `false`.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2661,7 +2667,7 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 			}
 		}
 
-		if ($this->blnHideFiles)
+		if (!$this->blnFiles)
```
leofeyer: This is wrong and should be `if (!$this->blnFiles && ! $this->blnFilesOnly)`. See:

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/widgets/FileSelector.php#L535

PR #934, opened 2017-07-06
[RTM] Make the DCA picker work for non admins

### LINKED ISSUE #886: Contao 4.4: new DCA picker does not work for non-admins

leofeyer:
This should eventually fix #886.


src/Resources/contao/classes/DataContainer.php
```diff
@@ -1027,4 +1064,38 @@ protected function getPickerInputField($value, $attributes='')
...
+
+
+	/**
+	 * Create a data container object in the picker environment
+	 *
+	 * @return DataContainer
+	 */
+	private function getDataContainer()
+	{
+		$do = \Input::get('do');
```
aschempp: looks like `table` is missing to be set?

src/Resources/contao/dca/tl_article.php
```diff
@@ -258,7 +258,7 @@
 			'exclude'                 => true,
 			'inputType'               => 'textarea',
 			'search'                  => true,
-			'eval'                    => array('rte'=>'tinyMCE', 'tl_class'=>'clr'),
+			'eval'                    => array('rte'=>'tinyMCE', 'dcaPicker'=>true, 'fieldType'=>'radio', 'filesOnly'=>true, 'tl_class'=>'clr'),
```
aschempp: as discussed, I disagree with these options because they are specific to a DC table

src/Resources/contao/widgets/TextArea.php
```diff
@@ -101,6 +107,39 @@ public function __set($strKey, $varValue)
 
 
 	/**
+	 * {@inheritdoc}
+	 */
+	public function getDcaFilter()
```
aschempp: The filters depend on the table of the DCA. Why can I not set a root array for the `tl_page` here?

src/Resources/contao/classes/DataContainer.php
leofeyer: `table` is actually not required for the permission checks.

src/Resources/contao/dca/tl_article.php
leofeyer: No, they are specific to the field, just like the `rte` setting. And the field knows which RTE and which picker option it can handle.

src/Resources/contao/widgets/TextArea.php
leofeyer: What is the use case?

src/Resources/contao/classes/DataContainer.php
ausi: But custom checks might use `table` too.
Not setting the `table` parameter could lead to confusing bugs IMO. I think we should set it too.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -1027,4 +1064,38 @@ protected function getPickerInputField($value, $attributes='')
...
 	 */
 	abstract protected function save($varValue);
+
+
+	/**
+	 * Create a data container object in the picker environment
+	 *
+	 * @return DataContainer
+	 */
+	private function getDataContainer()
```
ausi: Isn’t `getDataContainer` a confusing name?
I’d like `getPickerDataContainer` better I think.

src/Resources/contao/classes/DataContainer.php
leofeyer: Added in 4d205066b3a97c30a23f0f0317e02873064ca131.

src/Resources/contao/classes/DataContainer.php
leofeyer: Well, it returns a `DataContainer` object, so the name seems suitable to me. Also, it is a private method (unless we want to change that?).

leofeyer:
> The filters depend on the table of the DCA. Why can I not set a root array for the `tl_page` here?
> 
> ```php
> // dca/tl_content.php
> 'foobar' => [
>     'eval' => [
>         'dcaPicker' => [
>             'context' => 'foobar',
>             'tl_page' => ['root' => [1,2,3,4]],
>             'tl_news' => ['root' => [5,6,7,8]],
>             'tl_files' => ['files' => true],
>         ]
>     ]
> ]
> ```

This is a new feature and has nothing to do with the bugfix actually. We should create a separate PR for that.

aschempp:
Is it? It's new because the picker now supports multiple tables, but it's a "bug" that I cannot set different root for news and pages.

leofeyer:
You are right. Added in 8d80f5a653f11e27c17245d30dce7c9be756d825.

leofeyer:
```php
// dca/tl_content.php
'foobar' => [
    'eval' => [
        'fieldType' => 'radio', // this is field specific
        'dcaPicker' => [
            'page' => ['root' => [1,2,3,4]], // this only applies to the page picker
            'files' => ['files' => true], // this only applies to the file picker
        ]
    ]
]
```

aschempp:
why not use the actual table name? So I can extend the options for new tables if necessary?

src/Resources/contao/classes/DataContainer.php
```diff
@@ -1027,4 +1071,90 @@ protected function getPickerInputField($value, $attributes='')
...
+			{
+				$arrModule =& $arrGroup[$this->strPickerDo];
+				break;
+			}
+		}
+
+		$this->import('BackendUser', 'User');
+
+		// Dynamically add the "personal data" module (see #4193)
+		if ($this->strPickerDo == 'login')
```
Toflar: This hardcoded stuff here smells to me. If the core needs this, somebody else might need it too?

leofeyer:
> why not use the actual table name? So I can extend the options for new tables if necessary?

Because the table name is not available at this point.

aschempp:
I don't understand how this is supposed to work… and why would the table not be available in the widget?

leofeyer:
Because you need the table of the picker target field and not the table of the widget.

Besides, using the table is wrong. We need to evaluate the `do` parameter. I have tested it extensively and it works like a charm.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -901,25 +908,62 @@ protected function initPicker()
 			return false;
 		}
 
-		list($this->strPickerTable, $this->strPickerField, $this->intPickerId) = explode('.', \Input::get('target'), 3);
+		$strTarget = \Input::get('target');
+
+		if (\Validator::isInsecurePath($strTarget))
```
ausi: I don’t think this check is necessary because `validatePickerParams()` will check the values a few lines later. 
And `$strTarget` is not really a path.

leofeyer: I see it as a kind of early return, because why continue if we already know that there is a problem?

leofeyer: Also, `validatePickerParams()` will not validate the field name.

aschempp:
see #942

PR #938, opened 2017-07-07
[WIP] Added options to configure YouTube embedding

aschempp:
Probably needs a database migration too to set the default options on update.

### DIFF
--- a/src/Resources/contao/dca/tl_content.php
+++ b/src/Resources/contao/dca/tl_content.php
@@ -627,6 +627,18 @@ $GLOBALS['TL_DCA']['tl_content'] = array
 			),
 			'sql'                     => "varchar(16) NOT NULL default ''"
 		),
+		'youtubeOptions' => array
+		(
+			'label'                   => &$GLOBALS['TL_LANG']['tl_content']['youtubeOptions'],
+			'default'                 => array('youtube_suggest', 'youtube_controls', 'youtube_intro'),
+			'exclude'                 => true,
+			'search'                  => true,
+			'inputType'               => 'checkbox',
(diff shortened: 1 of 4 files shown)


leofeyer:
Wouldn't it be much easier if you reversed the logic?

 * Hide suggested videos when the video finishes
 * Hide player controls
 * Hide video title and player actions

That would be backwards compatible and does not require any migration.

aschempp:
The options are taken straight from youtube…

fritzmg:
In my extension ([fritzmg/contao-youtube-iframe](https://github.com/fritzmg/contao-youtube-iframe/)) I also designed the checkboxes this way so that if you check none of them, the default values of YouTube are active. Currently I support:

* Show closed captions by default (`cc_load_policy=1`)
* White progress bar (`color=white`)
* Disallow fullscreen (`fs=0`)
* Disable annotations by default (`iv_load_policy=3`)
* Modest branding (`modestbranding=1`)
* Disable related videos (`rel=0`)
* Disable information (`showinfo=0`)

leofeyer:
If @fritzmg has already made an extension, there is no reason to merge this PR, is there?

fritzmg:
I think it would make sense to provide such functionality in the core.

leofeyer:
Ok, then @aschempp please update the PR and reverse the logic as discussed above.

aschempp:
I wonder if we should merge @fritzmg's features instead, apparently there are way more options than my PR provides?

leofeyer:
Implemented in c7b4f387ea111871e05e9d704b0626ceefc5d2e9.

PR #940, opened 2017-07-08
[RFC] show start/stop information

fritzmg:
If you have a lot of content elements with lots of different start/stop settings it would be convenient to see at a glance, which elements have such a setting defined and which are currently visible in the front-end and which aren't.

active_inactive

From top to bottom:

1. Not invisible, no start/stop setting.
2. Invisible, no start/stop setting.
3. Not invisible, start setting in the past.
4. Not invisible, start setting in the future.

This feature was originally requested [here](https://community.contao.org/de/showthread.php?65548-Anzeigen-von-und-Anzeigen-bis-Element-unsichtbar-sichtbar-setzen-möglich).

_Notes:_

* The icons are altered versions of the `clock.svg` icon from http://feathericons.com/ and I do not consider this one finished. I only used it for this proof of concept.
* Instead of showing this only for content elements to the left of the content element type, this could also be done as an additional button on the right instead (e.g. after the publish/unpublish button) which would show you the start/stop info on mouse over for example. The same concept could then be expanded to all record types.


m-vo:
I like this. Not sure if 'start setting in the past' should be any different to a regular entry if no 'stop setting in the future' is present.

asaage:
I think that is an improvement.
The clock-icon should probably not replace the in/visible-icon but be added to it to indicate weather a start/stop is set and if we are currently within that window. 

*(Personally i would prefer that start/stop-times automatically trigger a change of the un/publish attribute of the element but this would be a totally different approach)*

agoat:
+1

It should also be applied for articles and pages. Maybe with an additional clock-icon like the indication for pages hidden from the navigation.

Or with a clock indicator on the visibility button at the right. IMO thats the best and homogenous solution to give information if the ce, article or page is invisible by timer.

leofeyer:
I like the idea, however, I don't like the icon. Actually, I do like the clock icon, but it does not look good with the eye icon. Maybe we can find a replacement for either one?

leofeyer:
Anyone? @frontendschlampe /cc

frontendschlampe:
yeah ... I will check for icons, but we use the eye icon everywhere in Contao - so we have to replace everywhere?

I think, there's also a fifth variant:

> 1. Not invisible, no start/stop setting.
> 2. Invisible, no start/stop setting.
> 3. Not invisible, start setting in the past.
> 4. Not invisible, start setting in the future.

5. Not invisible, end setting in the past

And we're also not consistent to pages and articles, because there we use gray color for not visible elements. Maybe we should change this or we use the gray color for state 5?!

m-vo:
I think we should group the use cases. Imo there should only be a clock icon present as long the visibility of the element is *currently* affected by a clock setting: either beeing displayed *not yet* or beeing displayed *only until* or *not* beeing displayed *anymore*.

Not yet
* start setting in the future / stop setting in the future
* start setting in the future / stop setting in the future
* start setting in the future / stop setting in the future

Only until
* stop setting in the future

Not anymore
* 

 * Not invisible, no start/stop setting.
    Invisible, no start/stop setting.
    Not invisible, start setting in the past.
    Not invisible, start setting in the future.
Not invisible, end setting in the past

m-vo:
I think we should group the use cases. Imo there should only be a clock indication present as long as the visibility of the element is **currently affected** by a clock setting: either beeing forced to be displayed *not yet* or beeing displayed *only until* or *not* beeing displayed *anymore*.

**1) Cases**
* start setting in the future / stop setting in the future
* ~~start setting in the future / stop setting in the past~~ (illegal configuration)
* start setting in the past / stop setting in the future
* start setting in the past / stop setting in the past


**2) Assigning the cases**

Not yet (e.g. grayed out clock icon)
* start setting in the future / stop setting in the future

Only until (e.g. green clock icon)
* start setting in the past / stop setting in the future

Not anymore  (e.g. red clock icon)
* start setting in the past / stop setting in the past

frontendschlampe:
we also have to delete the eye icon in front of element name

agoat:
To avoid having to deal with too many cases, the easiest thing to do is to change the eye icon in a way (e. g. an additional icon such as a red clock) that indicates that the item is published but is not visible at the moment due to the start-stop times.

leofeyer:
@frontendschlampe What's the status of this ticket?

frontendschlampe:
I try to make a draft next week.

CMSworker:
Looking forward for this feature...

But a shot accessibility note: Please keep in mind that icons should be distinguishable by form also, not only by color change. Especially with the colors red and green some people might get problems because of color blindness...

clock-icons-deuteranopie

leofeyer:
Is there anyone who can help us with the icons?

frontendschlampe:
I'm sorry! :-(

Do we have a final idea? Just the clock and the eye?

frontendschlampe:
Here's an idea of icons:

1. Element visible, active start setting (element visible)
2. Element visible, start setting in future (element not visible)
3. Element visible, end setting in past (element not visible)
4. Element invisible, active start setting (element not visible)
5. Element invisible, start setting in future (element not visible)
6. Element invisible, end setting in past (element not visible)

If there's no start/stop setting, the clock is not visible.

What do you think?

v1_start-stop-icon

bezin:
How can the element be visible if 2. or 3. is the case 🤔

fritzmg:
Hm... I am confused by this 😬 

Imho the text should always be red, when the element is invisible, either through start/stop or visibility setting. The clock should always be red, when the element is invisible due to start/stop settings.

The arrows are not quite clear to me either. What about

clock_start_stop_visible
clock_stop_invisible
clock_stop_visible

etc. (with bigger arrows... was too lazy to make it better 😁)

frontendschlampe:
> How can the element be visible if 2. or 3. is the case 🤔

The color red or green is just for the visibility setting (eye):

red = invisible
green = visible

It has nothing to do with the setting of start and stop.

> Imho the text should always be red, when the element is invisible, either through start/stop or visibility setting.

I don't think so. The visibility setting has nothing to do with setting start/stop. Maybe the icons for future start and past end could be red, but I think, this could be confusing, too.

> The arrows are not quite clear to me either. What about
> etc. (with bigger arrows... was too lazy to make it better 😁)

Nearly this was my first try, but you can't see the arrows clear. When you make the arrows bigger, you can't see the clock clear. That's why I search for another solution with the arrow around the clock.

For start setting in future I make the arrow clockwise and for stop setting in past I make it anticlockwise.

frontendschlampe:
There's also the idea to show information about start/stop on mouse over the clock.

fritzmg:
> I don't think so. The visibility setting has nothing to do with setting start/stop. Maybe the icons for future start and past end could be red, but I think, this could be confusing, too.

Yes, but you should see at a glance whether an element is visible in general or not. And this must include the visibility setting, plus the start/stop setting.

> Nearly this was my first try, but you can't see the arrows clear. When you make the arrows bigger, you can't see the clock clear. That's why I search for another solution with the arrow around the clock.

Hm what about putting the arrows in the bottom left and right?

frontendschlampe:
> Yes, but you should see at a glance whether an element is visible in general or not. And this must include the visibility setting, plus the start/stop setting.

I don't know. I think not. But I can't decide! ;-)

> Hm what about putting the arrows in the bottom left and right?

The icons will be too big and the position is different with arrow before and behind.

fritzmg:
> I think not.

Why not? Being able to see which elements are currently visible in the front end surely would be an incredibly useful piece of information for the back end? ;)

frontendschlampe:
> Why not? Being able to see which elements are currently visible in the front end surely would be an incredibly useful piece of information for the back end? ;)

yes ... it's very useful, but I want to see, if the element is invisible for ever or having start/stop information. Red means not visible - never. Green means visible - and with clock icon it could be visible in future or was visible in past.

I see it from another side: if I see red (without checking the eye icon or any other icon), then I think, the element will never be visible - but with your coloring I have to check the icons, too.

fritzmg:
> yes ... it's very useful, but I want to see, if the element is invisible for ever or having start/stop information

Hence the icon.

> I see it from another side: if I see red (without checking the eye icon or any other icon), then I think, the element will never be visible - but with your coloring I have to check the icons, too.

Yes of course - but the advantage is, that you can do all of this just by _looking_. Any other solution would require you to either edit the content element or hover over the icon and then you would still have to actively check the time and think about whether this element is currently visible or not.

I think the whole point should be that _Contao_ tells the back-end user _somehow_ whether this element is currently visible or not.

fritzmg:
> Red means not visible - never. Green means visible - and with clock icon it could be visible in future or was visible in past.

Why is it different for green? If red means _never visible_ then green should als mean _always visible_. Neither of which is true ;)

frontendschlampe:
> Yes of course - but the advantage is, that you can do all of this just by looking. Any other solution would require you to either edit the content element or hover over the icon and then you would still have to actively check the time and think about whether this element is currently visible or not.
> 
> I think the whole point should be that Contao tells the back-end user somehow whether this element is currently visible or not.

you said it already:

> Hence the icon.

The icon also shows the status - not the color! ;-)

fritzmg:
> The icon also shows the status

Hm, then I did not get the icons. There is only 3 different icons - but you would need at least the following states, if you do not use the color:

* Start setting defined and start is in the future
* Start setting defined and start is in the past
* Stop setting defined and stop is in the future
* Stop setting defined and stop is in the past
* Start and stop setting defined and current time is in between start and stop
* Start and stop setting defined and current time is outside start and stop

frontendschlampe:
> Why is it different for green? If red means never visible then green should mean always visible. Neither of which is true ;)

Because green and visible could be a bigger problem, then red and unvisible. ;-)

frontendschlampe:
> Start setting defined and start is in the future

number 2 on my example

> Start setting defined and start is in the past
> Stop setting defined and stop is in the future
> Start and stop setting defined and current time is in between start and stop

both the same ... number 1 on my example

> Stop setting defined and stop is in the past

number 3 on my example

> Start and stop setting defined and current time is outside start and stop

number 2 or 3 in my example

But you can't display all these states just with icons. Nobody will get it.

fritzmg:
I still think users would get it (more), if you colour both the text and the icon depending on the current visibility state ;)

asaage:
my personal opinion on this:
- do not put any icons on the left
- put the clock icon next to the eye on the right
- let it be grey (or do not show it all) if no start/stop is set
- let it be red if start/stop-settings prevent the element from showing up
- let it be green if the element is visible due to start/stop-settings
- click the clock to take me to the start/stop palette
- use the colored cte-type to indicate the **current over-all-state**

I dont't think we need differently shaped clock-icons to determine past or futue.
Maybe we should replace all the grey-eye-icons with a red ones.

CMSworker:
I think integrating arrows within the clock icon is hard to recognize (too small). And a colored clock is also difficult, because of the many different setting options.

So here another trial (or suggestion): what about some kind of traffic light to indicate the frontend status and/or some arrow icons for the timing settings? (clock icon optional, also not really required here)

Contao-Icons

CMSworker:
Or even more simplified... (just a quick test):

Contao-Icons2

leofeyer:
@CMSworker Very nice 😎 I prefer the first example, but I'd be fine with either one.

aschempp:
Love the traffic light idea! ❤️

Toflar:
I also fancy the traffic light idea! Very nice! 🥇

fritzmg:
@CMSworker hm, but what about having both start & stop set? In your example you could not tell that an element with start & stop might not be visible anymore in the future.

ausi:
I think combining the small traffic lights from the second example with the arrows of the first example would be best. Would also solve @fritzmg s concern.

leofeyer:
However, the arrows are not yet consistent. For instance,  means "start in the past, stop in the future" as well as "start in the future, stop in the future". There are three states for both start and end date:

* not set
* in the past
* in the future

The arrows should represent all 9 combinations.

CMSworker:
> For instance,  means "start in the past, stop in the future" as well as "start in the future, stop in the future".

- When start and stop are in future: traffic lights are orange because article is "on hold".
- When start is in past and stop in future: traffic lights are green because article is visibible in FE.
- When start and stop are in past: traffic lights are orange again (as first case). Could also be made red in this case, but someone said before, it should only be red when visibility (eye icon) is disabled.

Arrows representing all 9 time setting cases (in fact only 8 because one is invalid) might get ugly / too complex: How can you visualize that start or stop date is in future or past? Adding more arrows?

Contao-Icons-20190618-Arrow-Test
Is it really required (or even possible) to show all possible setting within an icon in the element overview?

Maybe the arrows concept is not ideal for this and we need another metapher. Or another way of designing the arrows?

leofeyer:
> might get ugly / too complex

You are right. We should only use the arrows to indicate whether a start or stop date has been set then. This means that we do not need , because the traffic light will already indicate that the stop date is in the past. I have changed your image accordingly:

59516974-c451b700-8ec2-11e9-9198-80cdf07ec4c0

frontendschlampe:
I think, we go in a wrong way. So much information in elements view is very hard. And how we will do it with articles and pages? Everywhere will be icons ... more icons ... and nobody knows, what they mean. And there're much more information, that I want to see in overview: css-classes, access protection maybe more?! And the colors - the backend becomes motley. Please don't get me wrong - I think it's a good idea, but it's too much. Maybe we should think about a status overlay for any element, article, page, where I can see the detailed information?! Maybe only icons that there's a setting for start/stop, css-classes, access protection ... yes ... it's the info window, but there're all information displayed and I can't see the important and relevant information, that I want.

Toflar:
I know what you mean but this here is different imho. We‘re talking about the published state. The current green eye may be misleading because you think something is published but it‘s actually not. It‘s not comparable to CSS classes etc.

frontendschlampe:
> The current green eye may be misleading because you think something is published but it‘s actually not.

I don't want to say, that we should do nothing, but what we want to do now, is overkill - I think.

Toflar:
We might as well just show an orange eye if any of the start or stop fields is not empty. As simple as that you mean?

frontendschlampe:
> We might as well just show an orange eye if any of the start or stop fields is not empty. As simple as that you mean?

yes ... for example ...

fritzmg:
What would the orange eye represent though? Star or Stop is set _and_ the element is visible? Or invisible?

Toflar:
„Published but...“ - that‘s all. If you want to know more, use the info icon or edit view.
Just as an example. I agree that representing too many states is impossible to do in small icons so we have to simplify things.

fritzmg:
> „Published but...“ - that‘s all. If you want to know more, use the info icon or edit view.

But then you are still not able to determine at a glance, whether an element is visible in the front end or not.

> I agree that representing too many states is impossible to do in small icons so we have to simplify things.

I agree with that too. My original proposal just introduces one additional icon which - depending on the state, replaces the eye on the left and the current visibility status is determined by the color.

leofeyer:
> We might as well just show an orange eye

We must not use icons that only differ in color though. That's why the traffic light is a great idea.

CMSworker:
I'll try to sum up some points:
- The eye icon (metapher for visibility) is not ideal, because it does not indicate the real visibility status in frontend. So it should be replaced with a better metapher for the toggle function "published/unpublished" (or "approved/blocked" or "activated/disabled").
- Frontend visibility is depending on both, publish status (eye) and time settings. As a result it would probably be best to place both icons next to each other, I guess.
- Something like a traffic light could indicate the 3 important cases related to frontent visibility:
    - Red: unpublished or (permanent) invisible
    - Orange: published but invisible "on hold"
    - Green: visible
- Icons must not only differ by color to help colorblind users (accessibility).
- Keep it simple: Showing all possible time setting options with icons won't work (information overload), we should focus on the most important ones, related to the current frontend visibility status:
    - Element is currently and permanent visible
    - Element is currently visible, but will stop some time in future
    - Element is currently on hold (start planned but in future)
    - Element is currently and permanent invisible (end date is in the past)

So here another suggestion (new draft):

contao-icons-20190619

CMSworker:
Could be implemented like this for example (changed icon order/position optional):

ce-list

article-list

asaage:
I like this:
image
which is basically what i described [earlier](https://github.com/contao/core-bundle/pull/940#issuecomment-502124504)
I still think the amount of clock-icons can be reduced to 3 (in range / out of range / not set) if you provide a link-action wich shows me the start/stop palette.
KISS principle.

Did anyone ever think about the possibility to ditch the published state as manual input and make it a function of start/stop?
I think this would simplify things quiet a lot.
The default start-date could be the creation date - if no start-date is set it' "unpublished".....

m-vo:
Wow, there is a lot going on in this issue and some beautiful icons :smile:. Nevertheless here is another view on all these things:

I thought about how the start/stop functionality could be extended in the future to get an idea if we're going in the right direction. There might be things like recurring phases, visibility dependent on a day of the week, and so on. If we forget about how the implementation works for a second this can all be seen as automation of the visibility parameter instead of a complex state. A user may imagine there is some system in the background (:robot:) that flips the switch at a given time (rather than the state being evaluated every visit). 

This principle of parameter automation is used in a lot of audio software for example, where it is quite common to have both direct input (e.g. set a volume level) and automated input (e.g. fade out at 1m04s). And to keep a clutter free UI the parameter under automation control gets a little indication (e.g. red dot + tooltip: "automated by ..."). 

Maybe that could be something for Contao as well:
* clearly show the *effective* state right *now* (= currently visible in the FE)
* add icon/symbol beneath the automated parameter (here: visibility) that means "under automation control" + a tooltip that has more information like "will be deactivated at 09:15pm"
* user / automation parallel input:
  * either don't allow the user to change state if it's under automation control
  * or let the last input be the effective one, so the automation will overwrite the user when getting active
* allow the user to easily disable the automation if she wishes to do so

This principle is of course not limited to visibility but can be used for virtually all properties. For this to work out the implementation must effectively have the same result as if a scheduler/cron job would be executed which can be done with evaluation on access.

CMSworker:
And another one, in case you don't like the filigree "icon in circle" style, with an "infinite loop" clock for first case and an alternative to the stop square:

contao-icons-20190620

Would look like this in the icon row:

article-list

@asaage: I think "ditch the published state" is not a good idea. I have some project where the time setting is not used at all and it would be an annoying work to define start dates everywhere just to publish something. Also the publish/unpublish icon is a very usefull toggle link to make something very quickly invisible in frontend, especially for new or occational users. The publish icon can also be used when working with multiple people with different user rights: Someone can edit and prepare an article and also plan the timing for frontend visibility, but another person has to proof it first and can then publish it...

@m-vo: That's an interessting idea - but I think automatism should be discussed in a separate issue, because currently this is not a core function.

I think also the critic of @frontendschlampe about too many icons and missing notes about other things like css classes should be discussed in another issue. That would be more a general discussion about user interface optimation or backend redesign in general. There are already multiple other issues related to the icon user interface that should be taken in consideration then as well...

m-vo:
> @m-vo: That's an interessting idea - but I think automatism should be discussed in a separate issue, because currently this is not a core function.

All of this could be implemented in the model I guess. Accessing the records directly was a problem ever since as you have to manually evaluate the state yourself. But I know where you are going.

I just wonder if it's the right thing to add so much visual complexity for a single property and behavior. Although the icons are as reduced and clear as possible and the logic is well structured, the user has to create a quite complex mental model of the system to understand what is happening or will happen. IMHO this is not that intuitive + not extendable.

CMSworker:
Agree – but how would that look like? And how could that solve [the problem (see forum)](https://community.contao.org/de/showthread.php?65548-Anzeigen-von-und-Anzeigen-bis-Element-unsichtbar-sichtbar-setzen-m%C3%B6glich) that was the reason for this issue/pull request?

m-vo:
>[from the issue / forum] Hier wäre es super -wenn es im Backend ersichtlich wäre - welches Element momentan sichtbar ist und welches nicht.

That's what i meant with effective state. Maybe like the visible setting behaves right now: green/red title (or what I'd prefer for the color blind: normal/grey or dimmed element). 

>  but how would that look like?

🤷‍♂  I don't know. Something like that maybe:

contao_startstop

CMSworker:
I'm not sure, if I get this right: The red dot on the eye only indicates, that there is a time setting (or automatism) definied - so how do you know by that, if the article is visible in frontent or not?

When the eye is green (element published/approved) and there is a red dot (time defined), but the stop time is reached (so element is not visible in frontend), the text preview will get grey/faded out?

How can you know, that this article might go online soon (is planned) or will go online again (eg automatic every friday) or has stopped permanently?

What happens, when you click on the eye with a red dot? Will it deactivate the automatism or delete the time setting? If you link this to the time setting options, what will happen when you click on the eye and there is no dot? Also showing time setting options? So to publish or unpublish it, you have to open the article edit mode again?

m-vo:
> I'm not sure, if I get this right: The red dot on the eye only indicates, that there is a time setting (or automatism) definied - so how do you know by that, if the article is visible in frontent or not?

No, the color and eye (grey/green) always reflect the *current* state in the frontend. As soon as a start/stop (or other) automation is enabled, the red dot appears. This simply means: this state might change in the future / there is an automatism active that can change this state. If you want to know what exactly was setup, go and look into the element's settings (or get some quick information by a tooltip). 

It could work like this: A click on the eye without a dot behaves like normal. A click on the eye with a dot disables the automation and toggles the state (maybe with a warning dialog), so that a click does always have an effect.

(Right now: Imagine you've got a stop time set that's already in the past, then the eye symbol has no function: Whatever state you set the element won't be displayed. So in this case the symbol is superfluous anyway.)

fritzmg:
Just to weigh in here again:

* The original problem was that customers complained, that it is hard to identify which content elements are currently visible in the front end and which are not, if they use the start/stop feature.
* The original PR, as a PoC, tried to solve exactly that issue. However it would only work for content elements, due to the additional information available in the back end on the left side.
* Solving this issue for all elements (e.g. articles as well) surely is desirable - however, I think a flood of new icons would be counter productive. Making those icons easy to understand for everyone is even harder - and if you cannot understand them at a glance at all, then they do not really help with the original issue either.

mlaubacher:
Meiner Meinung nach muss folgendes bereinigt werden:
- Status indication und Toggle-funktion muss getrennt werden. (Ein korrektes oder explizites Verhalten ist nicht möglich im Falle wenn Daten gesetzt sind.)

Mögliche Lösungsvariante:
- Toggle nur vorhanden wenn kein Datum gesetzt ist
- Status indication hat 3 Stati: Veröffentlicht, Unveröffentlicht, Datum gesetzt

zonky2:
I'm a bit confused why so many icon variants have appeared here - but maybe I didn't understand the explanations either.
For me, the best overview would be to see the start and end time of the elements in the list - optionally also a colored icon, whether the element is currently visible or not
such as (* as not set)

\* - 05.06.2020
12.12.2019 - 11.12.2020
01.01.20201 - *

pagestyles:
ich schliesse mich @zonky2  an.
In diesem Fall finde ich auch dass die konkreten Daten (Datum) angezeigt werden sollten, auch wenn ich sehr auf grafische Lösungen stehe, weil ich sie intuitiver finde. Allerdings sind sehr gute Ideen dabei!

Als ich die Überschrift las kam mir als erstes die Lösung von @zonky2 in höhe der 3 Punkte zum erweitern der Ansicht in den Sinn.

zonky2:
> Als ich die Überschrift las kam mir als erstes die Lösung von @zonky2 in Höhe der 3 Punkte zum erweitern der Ansicht in den Sinn.

Bei den CEs ist ja genug Platz - da hätte ich die Datumsangabe eher oben in der Zeile der Icons links neben den Stift gepackt

Bei den Artikeln/Seiten ist etwas weniger Platz aber hier sollte es auch noch passen, die Datumsangaben rein zu packen - in C4.9 kann man auf "fullscreen" umschalten, wenns zu knirsch ist

m-vo:
Ich denke dieses Issue geht erst weiter, wenn jemand einen konkreten Vorschlag als PR macht.

PR #942, opened 2017-07-09
Use signed URLs instead of DcaFilterInterface to forward DCA configuration

aschempp:
As discussed

- removed use of DcaFilterInterface
- the Interface file still exists for BC reasons
- config for pageTree and fileTree widgets are supported, but for no other

I still want to discuss the `blnHideFiles` etc. props in DC_Folder. I think they're incorrectly implemented and the code was hard to adjust.

### DIFF
--- a/src/DataContainer/DcaFilterInterface.php
+++ b/src/DataContainer/DcaFilterInterface.php
@@ -14,6 +14,8 @@ namespace Contao\CoreBundle\DataContainer;
  * DCA filter interface.
  *
  * @author Leo Feyer <https://github.com/leofeyer>
+ *
+ * @deprecated Deprecated since Contao 4.4.1, to be removed in Contao 5.
  */
 interface DcaFilterInterface
 {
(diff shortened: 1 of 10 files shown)


aschempp:
Unfortunately, this breaks as soon as someone does anything with the URL, like toggle the tree 😢

src/DataContainer/DcaFilterInterface.php
```diff
@@ -14,6 +14,8 @@
  * DCA filter interface.
  *
  * @author Leo Feyer <https://github.com/leofeyer>
+ *
+ * @deprecated Deprecated since Contao 4.4.1, to be removed in Contao 5.
```
leofeyer: Since the interface has only been added in version 4.4.0 and the functionality is broken, we can remove it completely.

src/Menu/FilePickerProvider.php
```diff
@@ -103,6 +105,10 @@ public function processSelection($value)
      */
     public function canHandle(Request $request)
     {
+        if ($request->query->get('context') === 'file') {
```
leofeyer: Why did you add this?

leofeyer:
> Unfortunately, this breaks as soon as someone does anything with the URL, like toggle the tree 😢 

Does this mean that we cannot use this approach at all?

aschempp:
> Does this mean that we cannot use this approach at all?

Yeah, I guess so :(

Toflar:
We have a few ideas here, working on a solution :)

PR #943, opened 2017-07-10
Allow to override the mailer configuration in Email class

aschempp:
In Contao 3 it was possible to override the mailer configuration by replacing the global constants. We did such things in Notification Center (see #128).

In Contao 4 this no longer works because the container mailer is used. It should be possible to override this. Ideally, this could be merged as a bugfix because otherwise a few extensions will have issues to fix such bugs until the next minor 😶

### DIFF
--- a/src/Resources/contao/library/Contao/Email.php
+++ b/src/Resources/contao/library/Contao/Email.php
@@ -42,6 +42,12 @@ class Email
 {
 
 	/**
+	 * Mailer object
+	 * @var \Swift_Mailer
+	 */
+	protected $objMailer;
+
+	/**
 	 * Message object
 	 * @var \Swift_Message
 	 */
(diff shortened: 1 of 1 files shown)


discordier:
I wonder if we should not rather suggest to build the mail directly in extensions and not use the Email class anymore instead of adding more features.

I never thought of it to be as flexible as it should be.

Aside from that, the PR looks good to me, I only wonder if we want to keep the class.

Toflar:
> Aside from that, the PR looks good to me, I only wonder if we want to keep the class.

We need to, for BC reasons...

discordier:
> We need to, for BC reasons...

Of course we do, but if we agree to not use it in future, we could leave it as is. It is using the default mailer instance currently anyway and therefore not broken. I just wanted to point out to not enhance stuff that will be gone in the next major.

leofeyer:
Are you sure that you have linked the correct issue? #128 is about trusted proxies.

hermannbrauer:
Hey Leo, the 128 is a bug reference to the notification center....

https://github.com/terminal42/contao-notification_center/issues/128

leofeyer:
Are you referring to [this method](https://github.com/terminal42/contao-notification_center/blob/master/library/NotificationCenter/Gateway/Email.php#L139)?

PR #950, opened 2017-07-11
[RFC] Rewrite the picker implementation

aschempp:
As discussed in several issues, the current picker implementation does not really work. Here's my rewrite to support all current features and some more!

**TODO:**
- [x] Improve method names
- [x] Fix coding style / docblocks
- [x] Add necessary interfaces
- [x] Add unit tests


leofeyer:
Please revert the naming and namespace changes. Otherwise it is impossible to understand the changes.

src/Resources/contao/templates/backend/be_tinyNews.html5
```diff
@@ -24,6 +33,7 @@ setTimeout(function() {
       }
       editor.on('focus', function(){ Backend.getScrollOffset(); });
     },
+    file_browser_callback_types: <?= json_encode($pickerTypes) ?>,
```
leofeyer: The whole logic needs to be moved to the `DataContainer` class, so we can use `$this->pickerTypes` in the templates.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -510,7 +514,19 @@ protected function row($strPalette=null)
 			}
 		}
 
-		$objWidget->wizard = $wizard;
+		if ($wizard != '')
```
leofeyer: How is this change related to the picker?

src/Resources/public/core.js
```diff
@@ -898,7 +898,7 @@ var Backend =
 		var opt = options || {},
 			maxWidth = (window.getSize().x - 20).toInt(),
 			maxHeight = (window.getSize().y - 192).toInt();
-		if (!opt.id) opt.id = 'tl_select';
+		if (!opt.id) opt.id = 'tl_listing';
```
leofeyer: You must not change this, because it is not backwards compatible with the old `PageTree` and `FileTree` widgets.

src/Resources/contao/classes/DataContainer.php
aschempp: It hides the picker icon and spacing if the dcaPicker is not supported (due to permissions etc.)

src/Resources/contao/templates/backend/be_tinyNews.html5
aschempp: Is tinyMCE initialized by DataContainer ?

leofeyer: Yes: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/classes/DataContainer.php#L537-L551

aschempp: Changed in 1f24de033a5f8fc4a3739ff8e6e8c53b1507f5ed

src/Picker/PickerBuilder.php
```diff
@@ -0,0 +1,135 @@
...
+            $this->menuFactory,
+            $providers,
+            $config
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createFromJson($json)
```
Toflar: I would actually add `createFromUrlValue()` and `saveForUrlValue()` or something like this. You can keep the others if you like but this would allow us using `gzencode()` etc. in these methods (or anything else in the future) to transparently and backwards compatible improve the size of the content :)

src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6135,17 +6135,13 @@ public function initPicker(PickerInterface $picker)
...
-				if (!empty($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root']))
+				if (!empty($this->root))
 				{
-					$arrRoot = array_intersect($arrRoot, $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root']);
-
-					if (!empty($arrRoot))
-					{
-						$arrRoot = $this->eliminateNestedPages($arrRoot);
-					}
+					$arrRoot = array_intersect($arrRoot, $this->Database->getChildRecords($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root'], 'tl_page'));
```
Toflar: This should be `$this->strTable` not `tl_page` right?

aschempp: Fixed in 368b144dd3e88de3e803cb728a5cf13aa97332fe

src/Picker/PickerBuilder.php
aschempp: implemented in a25b13229627dfc085db24115786da8499ec97dc

aschempp: implemented in a25b13229627dfc085db24115786da8499ec97dc, including Gzip compression

src/Controller/BackendController.php
```diff
@@ -224,8 +226,23 @@ public function alertsAction()
      */
     public function pickerAction(Request $request)
```
Toflar: Can we have some extended description in the docblock here? :) `Handles the picker redirect.` does not really help a lot.

src/Picker/AbstractPickerProvider.php
```diff
@@ -0,0 +1,124 @@
...
+     */
+    private $tokenStorage;
+
+    /**
+     * Constructor.
+     *
+     * @param FactoryInterface      $menuFactory
+     * @param TokenStorageInterface $tokenStorage
+     */
+    public function __construct(FactoryInterface $menuFactory, TokenStorageInterface $tokenStorage)
```
Toflar: I prefer having setter injection for the `TokenStorageInterface` because `getUser()` is absolutely optional for an `AbstractPickerProvider`. I think we should have `setTokenStorage()` and `getTokenStorage()`. The `getUser()` already considers the `null` case anyway :)

src/Picker/DcaPickerProviderInterface.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Copyright (c) 2005-2017 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\Picker;
+
+/**
+ * Interface for DCA picker providers.
```
Toflar: I would like to have more description here. `Use this if .....` or something. This would help very much.

src/Picker/PickerConfig.php
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * Encodes the picker configuration for use in the URL.
+     *
+     * @param bool $gzip
+     *
+     * @return string
+     */
+    public function urlEncode($gzip = true)
```
Toflar: I would name this `$compress = true`. Who cares if this is gzip.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -3042,67 +3037,58 @@ protected function isProtectedPath($path)
...
 	{
-		parent::setDcaFilter($arrFilter);
+		$attributes = parent::initPicker($picker);
 
-		$blnHideFiles = !isset($arrFilter['files']) && !isset($arrFilter['filesOnly']);
+		$this->blnFiles = false;
+		$this->blnFilesOnly = false;
 
-		if (isset($arrFilter['files']) && $arrFilter['files'] === false)
+		if (is_array($attributes))
```
Toflar: I'd prefer a `null` check as there can only be array or null.

src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6123,4 +6119,32 @@ protected function formatGroupHeader($field, $value, $mode, $row)
...
 	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function initPicker(PickerInterface $picker)
+	{
+		$attributes = parent::initPicker($picker);
+
+		if (is_array($attributes))
```
Toflar: Check for `null` here and do an early return please. I hate these 5-level nestings 😄

src/Picker/AbstractPickerProvider.php
leofeyer: I think this makes sense.

src/Picker/ArticlePickerProvider.php
```diff
@@ -0,0 +1,92 @@
...
+            $attributes['value'] = str_replace(['{{article_url::', '}}'], '', $config->getValue());
+        }
+
+        return $attributes;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertDcaValue(PickerConfig $config, $value)
```
leofeyer: We are actually converting the picked value here, aren't we?

src/Picker/PickerConfig.php
```diff
@@ -0,0 +1,201 @@
...
+            $this->extras,
+            $this->value,
+            $current
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function jsonSerialize()
```
leofeyer: The method says JSON serialize but it returns an array?

src/Picker/PickerConfig.php
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * Encodes the picker configuration for use in the URL.
+     *
+     * @param bool $gzip
+     *
+     * @return string
+     */
+    public function urlEncode($gzip = true)
+    {
+        $data = json_encode($this);
```
leofeyer: Why are you encoding `$this` here? Should it be `$this->jsonSerialize()`?

src/Picker/PickerInterface.php
```diff
@@ -0,0 +1,49 @@
...
+     * @return ItemInterface
+     */
+    public function getMenu();
+
+    /**
+     * Gets the current provider.
+     *
+     * @return PickerProviderInterface|null
+     */
+    public function getCurrentProvider();
```
leofeyer: I would prefer `getProvider()` here.

src/Picker/PickerInterface.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Gets URL to the current picker tab.
+     *
+     * @return string
+     */
+    public function getCurrentUrl();
```
leofeyer: I would prefer `getUrl()` here.

src/Picker/PickerProviderInterface.php
```diff
@@ -0,0 +1,64 @@
...
+    public function supportsValue(PickerConfig $config);
+
+    /**
+     * Returns whether the picker is currently active.
+     *
+     * @param PickerConfig $config
+     *
+     * @return bool
+     */
+    public function isCurrent(PickerConfig $config);
```
leofeyer: I would prefer `isActive()` here.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -558,17 +564,26 @@ protected function row($strPalette=null)
...
 			$objTemplate->type = $type;
+			$objTemplate->fileBrowserTypes = $fileBrowserTypes;
 
 			// Deprecated since Contao 4.0, to be removed in Contao 5.0
 			$objTemplate->language = \Backend::getTinyMceLanguage();
 
 			$updateMode = $objTemplate->parse();
 
-			unset($file, $type);
+			unset($file, $type, $pickerBuilder);
```
leofeyer: Also unset `$fileBrowserTypes`?

src/Resources/contao/controllers/BackendMain.php
```diff
@@ -113,7 +113,22 @@ public function run()
 		// Open a module
 		elseif (\Input::get('do'))
 		{
-			$this->Template->main .= $this->getBackendModule(\Input::get('do'));
+			$picker = null;
+
+			if (isset($_GET['picker']))
```
leofeyer: Using `isset($_GET)` cannot be "overwritten" via `Input::setGet()`. I know that I have been doing it, too – maybe we should discuss it and be consistent?

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2667,7 +2662,7 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 			}
 		}
 
-		if ($this->blnHideFiles)
+		if (!$this->blnFiles && !$this->blnFilesOnly)
```
leofeyer: Does this really work without `$this->strPickerField &&`?

src/Resources/contao/templates/backend/be_main.html5
```diff
@@ -94,7 +94,7 @@
 
   <?php if ($this->isPopup): ?>
     <script>
-      if ($$('.tl_tree_checkbox,.tl_tree_radio').length > 0) {
+      if ($$('.tl_tree_checkbox,.tl_tree_radio').length > 0 && !document.location.search.test('act=select')) {
```
leofeyer: I'd prefer to have this as an outer if condition:

```js
if (!document.location.search.test('act=select')) {
    if ($$('.tl_tree_checkbox,.tl_tree_radio').length > 0)) {
        // …
    } else {
        // …
    }
}
```

Otherwise the else condition would still be evaluated in "edit multiple" mode.

leofeyer:
Already looking very good! Thank you very much @aschempp.

src/Controller/BackendController.php
aschempp: Improved in 9e53148bafbf41b76aef3e407678d25129765cb4

src/Picker/DcaPickerProviderInterface.php
aschempp: Improved in 9e53148bafbf41b76aef3e407678d25129765cb4

src/Picker/PickerConfig.php
aschempp: Improved in 9e53148bafbf41b76aef3e407678d25129765cb4

src/Resources/contao/drivers/DC_Folder.php
aschempp: Improved in 9e53148bafbf41b76aef3e407678d25129765cb4

src/Resources/contao/drivers/DC_Table.php
aschempp: Improved in 9e53148bafbf41b76aef3e407678d25129765cb4

src/Picker/ArticlePickerProvider.php
aschempp: We're converting the value from DC/database to the one we can pick.

src/Picker/PickerConfig.php
aschempp: That's how the `\JsonSerializable` interface works 😉

src/Picker/PickerConfig.php
aschempp: That's how the `\JsonSerializable` interface works 😉

src/Picker/PickerInterface.php
aschempp: It returns the currently active provider (based on the value).

src/Picker/PickerInterface.php
aschempp: It returns the URL for the current provider (based on the value).

src/Picker/PickerProviderInterface.php
aschempp: I've named this after the `current` variable of the `KnpMenu`…

src/Resources/contao/classes/DataContainer.php
aschempp: Added in 9e53148bafbf41b76aef3e407678d25129765cb4

src/Resources/contao/drivers/DC_Folder.php
aschempp: Yes it does, because `$blnFiles` is [initialized to `true`](https://github.com/aschempp/contao-core-bundle/blob/6a7d052a4b08047e4d42cd9f084d27d195aea0b1/src/Resources/contao/drivers/DC_Folder.php#L84) and only [set to `false` on `initPicker`](https://github.com/aschempp/contao-core-bundle/blob/6a7d052a4b08047e4d42cd9f084d27d195aea0b1/src/Resources/contao/drivers/DC_Folder.php#L3046).

src/Resources/contao/templates/backend/be_main.html5
aschempp: That would mean the buttons are not disabled when in select mode, which is not what we want, right?

src/Picker/AbstractPickerProvider.php
aschempp: Changed in 37b9d7caa43b196d8b234d59440c1b09f3842e06

src/Picker/ArticlePickerProvider.php
leofeyer: I still prefer `convertValue()` or `convertPickedValue()`. 😄

src/Picker/PickerInterface.php
leofeyer: Then I prefer `getActiveProvide()`.

src/Picker/PickerConfig.php
leofeyer: This is actually because `json_encode()` can handle objects and not because of the interface. 😉

src/Picker/PickerInterface.php
aschempp: It's named after `isCurrent`, which is based on `KnpMenu` (as described below). Still wanna change it?

src/Picker/PickerProviderInterface.php
leofeyer: Hm, ok. We should either say `current` or `active` everywhere for consistency. So `getCurrentProvider()`, `getCurrentUrl()`, `isCurrent()` etc. or `getActiveProvider()`, `getActiveUrl()`, `isActive()` etc.

I prefer the latter, but as long as it is consistent, I can live with it.

src/Picker/ArticlePickerProvider.php
aschempp: You know the method is (only) part of the `DcaPickerProviderInterface`, right? That's why I named the methods so they are specific to the DCA, if we ever later add similar features for something else…

src/Picker/PickerConfig.php
aschempp: Well `json_encode` will realize the object implements `\JsonSerializable` and uses that method…

src/Resources/contao/controllers/BackendMain.php
leofeyer: You are right. :thumbsup:

src/Resources/contao/drivers/DC_Folder.php
leofeyer: Ok.

src/Picker/PickerProviderInterface.php
aschempp: Yeah, sort of agree. We name it *active* most of the time in Contao. But we cannot change `KnpMenu`, and the `isCurrent()` method is also called in https://github.com/aschempp/contao-core-bundle/blob/37b9d7caa43b196d8b234d59440c1b09f3842e06/src/Picker/AbstractPickerProvider.php#L64 to determine the current menu item…

src/Resources/contao/templates/backend/be_main.html5
leofeyer: Of course. Don't know what I was thinking. 😄

src/Picker/ArticlePickerProvider.php
leofeyer: I just think that `convertPickedValue()` is better understandable for everyone implementing the interface. `convertDcaValue()` sounds like the `load_callback` of the field is executed to convert the DCA value. 😉

aschempp: How about `convertDcaToPickerValue()`?

src/Picker/PickerProviderInterface.php
leofeyer: Just use "current" everywhere.

aschempp: 👍  so nothing to change 😎

src/Picker/ArticlePickerProvider.php
leofeyer: What bothers you about `convertPickedValue()`?

aschempp: Every other method in the `DcaPickerProviderInterface` contains `Dca` 🙈

leofeyer: Ok, first, all the `Dca` parts are unnecessary in my opinion, because the class already has `Dca` in it.

But second, the first two methods actually read data from the DCA, so using `Dca` in the method name is kind of correct. The third method, however, reads user input, so …

aschempp: You mean the interface does?
The `convertDcaValue` does not really read user input, it's basically a save_callback on the DCA value before it is rendered as a picker radio/checkbox value.

leofeyer: Of course it reads user input. The user picks page ID 12, submits the value and gets back `{{link_url::12}}`.

Again, I think that `convertPickedValue()` is much better understandable.

aschempp: That is not correct. There is no Ajax request anymore, the values are converted **before** the user selects them in the picker. This also results in an instant update of the field value, contrary to the previous implementation which took seconds to update 😉

leofeyer: Ah, why don't you say this right away? Then `convertDcaValue` is actually fine. 😄

leofeyer:
Actually merged in 874076de55cbf962fd5e41c00e853e4dec8842a6.

PR #951, opened 2017-07-11
make maxImageWidth mandatory

fritzmg:
Fix for https://github.com/contao/core/issues/8742


leofeyer:
You have already created https://github.com/contao/core/pull/8743, so no need to create the same PR again. If it is merged into version 3.5, it will automatically be added to version 4.4 as well.

fritzmg:
Oh I see :)

PR #969, opened 2017-07-17
Fixed merge http headers listener not working with Symfony's request stack

Toflar:
After hours of debugging @aschempp and me finally found the issue for some redirect and cookie issues in Contao 4.4. They were introduced by using our header-replay-bundle but it was actually not the reason for the bug. The bug is that the listener instance is created once and we do not lazily fetch the headers using `headers_list()`. Now when you are using multiple request->response iterations within the same php process (which is one big advantage of Symfony and which was not noticed before because nobody did that apparently) the headers that are sent in between these iterations are ignored. We have to always fetch them lazily.

### DIFF
--- a/src/EventListener/MergeHttpHeadersListener.php
+++ b/src/EventListener/MergeHttpHeadersListener.php
@@ -29,7 +29,7 @@ class MergeHttpHeadersListener
     /**
      * @var array|null
      */
-    private $headers = null;
+    private $headers;
 
     /**
      * @var array
@@ -46,8 +46,7 @@ class MergeHttpHeadersListener
      * Constructor.
      *
      * @param ContaoFrameworkInterface $framework
-     * @param array                    $headers Do not pass this argument in your code;
-     *                                          It is meant for unit testing only.
+     * @param array                    $headers   Meant for unit testing only!
      */
     public function __construct(ContaoFrameworkInterface $framework, array $headers = null)
     {
@@ -151,7 +150,6 @@ class MergeHttpHeadersListener
     private function getHeaders()
     {
         if (null === $this->headers) {
-
             return headers_list();
         }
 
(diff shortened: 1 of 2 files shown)


src/EventListener/MergeHttpHeadersListener.php
```diff
@@ -60,13 +59,21 @@ public function __construct(ContaoFrameworkInterface $framework)
      */
     public function getHeaders()
     {
+        if ([] === $this->headers) {
+
```
leofeyer: This line needs to be removed. 😄

leofeyer:
I would prefer to remove the `setHeaders()` method and use constructor injection instead.

```php
public function __construct(ContaoFrameworkInterface $framework, $headers = [])
{
    $this->headers = $headers;
}
```

src/EventListener/MergeHttpHeadersListener.php
```diff
@@ -60,13 +59,19 @@ public function __construct(ContaoFrameworkInterface $framework)
      */
     public function getHeaders()
     {
+        if ([] === $this->headers) {
```
aschempp: I think `$this->headers` should be `null` if not "initialized"

leofeyer: That would be ok for me, too.

Toflar:
Updated PR according to your comments.

src/EventListener/MergeHttpHeadersListener.php
```diff
@@ -150,7 +150,7 @@ private function mergeHttpHeaders(Response $response)
     private function getHeaders()
     {
         if (null === $this->headers) {
-            return headers_list();
+            $this->headers = headers_list();
```
aschempp: This is wrong! Exactly the reason why we had to change the listener. The headers can change between (sub)requests, that's why we have to fetch them every time.

PR #970, opened 2017-07-18
[RTM] Added the insert tag flag for absolute URLs

qzminski:
Works like this: `{{link_url::123|absolute}}`.

### DIFF
--- a/src/Resources/contao/library/Contao/InsertTags.php
+++ b/src/Resources/contao/library/Contao/InsertTags.php
@@ -403,13 +403,13 @@ class InsertTags extends \Controller
 
 								if ($objNext instanceof PageModel)
 								{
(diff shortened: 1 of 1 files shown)


qzminski:
Will certainly do if this is merged.

leofeyer:
This does not really work, does it? The insert tag is cached, so if someone has both `{{link_url::123|absolute}}` and `{{link_url::123}}`, they will always get the absolute URL (second call returns the cached value).

qzminski:
Well, the `$strTag` is cached which is the insert tag and the flag, so they are both different tags in fact. It seems to work fine for me. Or am I missing something?

leofeyer:
No, you are right. My bad.

leofeyer:
Thank you @qzminski.

PR #973, opened 2017-07-18
add cron.php info

fritzmg:


leofeyer:
I'm against merging this. We did not add an explanation for any other changed entry point, either.

fritzmg:
It has come up a lot of times in the forum now though. May be it should just be added in the Documentation. Currently it's only mentioned in the cookbook: https://docs.contao.org/books/cookbook/de/Cronjobs-in-Contao.html which is more or less only for Contao 3.

rabauss:
is there somewhere a docu for the changed entry points?

PR #984, opened 2017-07-24
Change tab URL to not pre-select based on current value

aschempp:


src/Picker/ArticlePickerProvider.php
```diff
@@ -82,7 +82,7 @@ protected function getLinkClass()
     /**
      * {@inheritdoc}
      */
-    protected function getRouteParameters(PickerConfig $config)
+    protected function getRouteParameters(PickerConfig $config = null)
```
leofeyer: It seems that you are not using `$config` in any of the child classes?

aschempp: Because the core pickers are not picking from subtables. News, Event and FAQ pickers do.

leofeyer: Yes, but `null` is never passed as argument, is it? We could leave it as it was.

aschempp: If that was true, then the URL would always link to the selected value 😉 
see https://github.com/contao/core-bundle/pull/984/files#diff-4e7f624198eef84ce3f501384ca6dcc8R135

src/Picker/AbstractPickerProvider.php
```diff
@@ -122,6 +122,15 @@ protected function getUser()
     }
 
     /**
+     * Returns the routing parameters for the backend picker.
```
xchs: backend -> back end

PR #999, opened 2017-08-03
[RFC] Add article tree widget

agoat:
This PR adds an article tree widget to complete/extend the new dcaPicker. 

This can be used in the DCA like the `pageTree` (multiple and rootNodes are also supported) to select articles with the new picker.

I'm not sure if the content elements `Article` and `Article teaser` should use the articleTree. All other `Include elements` use select fields by now.

### DIFF
--- a/src/Picker/ArticlePickerProvider.php
+++ b/src/Picker/ArticlePickerProvider.php
@@ -30,7 +30,7 @@ class ArticlePickerProvider extends AbstractPickerProvider implements DcaPickerP
      */
     public function supportsContext($context)
     {
-        return 'link' === $context && $this->getUser()->hasAccess('article', 'modules');
+        return in_array($context, ['article', 'link'], true) && $this->getUser()->hasAccess('article', 'modules');
     }
 
     /**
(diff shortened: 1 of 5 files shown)


agoat:
There is a problem with the rootNodes handling (#see #1001). This have to be considered here, too.

agoat:
Note for me: Set the radio button as default (see #989).

Toflar:
@agoat I like the PR although I have currently no use case for it :) But you never know ;) It is currently not yet tagged for 4.5 because tests are failing. Maybe you need to rebase to latest develop?

agoat:
> I have currently no use case for it

We could use it in the `article` and `article teaser` content elements. The `article teaser` limit the selection to the articles from the same root. This can be done here, too, since the correct root node handling (see #1001) is merged.

And I use it already in an extension (special branch together with an adjusted core-bundle).

Toflar:
I never use that feature but then I guess you should include the changes to tl_content already :)

leofeyer:
Can you please post a screenshot of the article picker in action?

leofeyer:
See contao/contao#714

PR #1001, opened 2017-08-04
Correctly handle root nodes for the dcaPicker

agoat:
The handling of the rootNodes in the new dcaPicker is not working right for non admin backend users.
If the backend user is restricted to a page mounts and also a rootNode from a different level is defined as attribute for the picker, no pages will be shown at all.

Little example:

Let's say we have the following page structure
rootnodes

If the user is restricted to `ID 5` and a rootNode `ID 1` is set as an attribute for the picker no pages will be shown. That's because the `ID 1` is not a child of `ID 5`. 
The same happens when the attribute rootNode for the picker is also `ID 5`. Because the rootNode itself is not included for the array intersection.

So first the given rootNode(s) have to be added for the intersection. And secondly all the children have to be traversed down for the the rootNode attribute, too.

### DIFF
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -6145,6 +6145,7 @@ class DC_Table extends \DataContainer implements \listable, \editable
 			// Allow only those roots that are allowed in root nodes
 			if (!empty($this->root) && $arrRoot != $this->root)
 			{
+				$arrRoot = array_merge($arrRoot, $this->Database->getChildRecords($arrRoot, $this->strTable));
 				$arrRoot = array_intersect($arrRoot, array_merge($this->root, $this->Database->getChildRecords($this->root, $this->strTable)));
 				$arrRoot = $this->eliminateNestedPages($arrRoot);
 			}
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -6147,7 +6147,7 @@ class DC_Table extends \DataContainer implements \listable, \editable
 			{
 				$arrRoot = array_merge($arrRoot, $this->Database->getChildRecords($arrRoot, $this->strTable));
 				$arrRoot = array_intersect($arrRoot, array_merge($this->root, $this->Database->getChildRecords($this->root, $this->strTable)));
-				$arrRoot = $this->eliminateNestedPages($arrRoot);
+				$arrRoot = $this->eliminateNestedPages(array_unique($arrRoot, SORT_NUMERIC));
 			}
 
 			$this->root = $arrRoot;


aschempp:
Interesting.

Why do we use `$GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root']` instead of `$this->root` in the first place??

agoat:
> Why do we use $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root'] instead of $this->root in the first place??

Not sure what you want to know exactly.
But I found some strange behaviour.

When I put some dumps into the method like so.. 
```php
	public function initPicker(PickerInterface $picker)
	{
		$attributes = parent::initPicker($picker);

		if (null === $attributes)
		{
			return null;
		}

		// Predefined node set (see #3563)
		if (isset($attributes['rootNodes']))
		{
			$arrRoot = (array) $attributes['rootNodes'];
dump($this->root);		
			// Allow only those roots that are allowed in root nodes
			if (!empty($this->root))
			{
				$table = ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 6) ? $this->ptable : $this->strTable;

				$arrRoot = array_merge($arrRoot, $this->Database->getChildRecords($arrRoot, $table));
				$arrRoot = array_intersect($arrRoot, array_merge((array) $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root'], $this->Database->getChildRecords($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['root'], $table)));
				$arrRoot = $this->eliminateNestedPages($arrRoot);
			}

			$this->root = $arrRoot;
dump($this->root);	
		}

		return $attributes;
	}
```
.. I noticed the following:

When I push the `rootNodes` attribute to the picker the first dump will show the users page mount ids (admin have all root nodes here). The second will have the correct interesection with the child pages ids.

But when there is no `rootNodes` attribute set, the first dump is empty. 
Where is the `$this->root` is set before??

leofeyer:
How exactly can I reproduce this?

agoat:
> How exactly can I reproduce this?

Let's say we have the following page structure:
pages2

Create a user and set the pagemount to 'subfolder'. 
Now change (for testing purpose) the `url` field in the `tl_content.php` to the following:
```php
		'url' => array
		(
			'label'                   => &$GLOBALS['TL_LANG']['MSC']['url'],
			'eval'                    => array
                        (
				'mandatory'		=> true, 
				'rgxp'			=> 'url', 
				'decodeEntities'	=> true, 
				'maxlength'		=> 255, 
				'dcaPicker'		=> true,
				'fieldType'		=> 'radio', 
				'filesOnly'		=> true, 
				'rootNodes'		=> array(ID of 'test root' or 'subfolder')
				'tl_class'		=>'w50 wizard'
			),
			'sql'                     => "varchar(255) NOT NULL default ''"
		),

```
If you now try to select a page in the hyperlink ce, no pages will be shown.

agoat:
As an example: I have an extension which restrict the selection to the current page root (think about multi domain or multi language setups). 

So I will pass the **root nodes `ID1`** and the user can select what is the result of the insersection of all child pages. Otherwise I first need to get the page mount of the user to do it myself but afterwards the half check is done again.

IMO it makes totally sence to support root node IDs above or outside of the page mount children range.

leofeyer:
As discussed in Mumble on September 7th, we want to support your use case. However, we are not sure about the `array_merge()` solution. Did you test this with multiple root nodes (e.g. `$arrRoot = array(1, 5)`?

agoat:
I tested this with multiple root nodes from root pages by setting the ids of all multilingual root pages. No problems here. 

But when I set (why ever) a root and a child page id, like your example `$arrRoot = array(1, 5)`, it returned the resulting root page(s) multiple times. So we have to be sure to return only unique values.

Fixed in 409e08c3a54d666c8af783a761ed86587403b4a2.

src/Resources/contao/drivers/DC_Table.php
```diff
@@ -6145,8 +6145,9 @@ public function initPicker(PickerInterface $picker)
 			// Allow only those roots that are allowed in root nodes
 			if (!empty($this->root) && $arrRoot != $this->root)
 			{
+				$arrRoot = array_merge($arrRoot, $this->Database->getChildRecords($arrRoot, $this->strTable));
 				$arrRoot = array_intersect($arrRoot, array_merge($this->root, $this->Database->getChildRecords($this->root, $this->strTable)));
-				$arrRoot = $this->eliminateNestedPages($arrRoot);
+				$arrRoot = $this->eliminateNestedPages(array_unique($arrRoot, SORT_NUMERIC));
```
leofeyer: Why did you add the `SORT_NUMERIC` flag?

leofeyer:
Fixed in f3aafc07a0d830900e2fdccccd22048991471a23.

src/Resources/contao/drivers/DC_Table.php
agoat: > Why did you add the SORT_NUMERIC flag?

To compare the array values numerically.

But your solution in f3aafc07a0d830900e2fdccccd22048991471a23 works, too.

PR #1002, opened 2017-08-04
[RTM] increase imageWidth/Height default

fritzmg:
The old defaults for the `imageWidth` and `imageHeight` setting seem a bit low to me, especially in these modern times, with responsive images and high DPI display devices.

I recently forgot to change these values and now a customer needs to re-upload all pictures again, since they all got automatically downscaled to 800x600 -_-

Not sure what the new defaults should be though. 1280? 1680? 1920?

### DIFF
--- a/src/Resources/contao/config/default.php
+++ b/src/Resources/contao/config/default.php
@@ -67,8 +67,8 @@ $GLOBALS['TL_CONFIG']['uploadTypes']
 	. 'ttf,ttc,otf,eot,woff,woff2,'
 	. 'css,scss,less,js,html,htm,txt,zip,rar,7z,cto';
 $GLOBALS['TL_CONFIG']['maxFileSize']    = 2048000;
-$GLOBALS['TL_CONFIG']['imageWidth']     = 1280;
-$GLOBALS['TL_CONFIG']['imageHeight']    = 1280;
+$GLOBALS['TL_CONFIG']['imageWidth']     = 0;
+$GLOBALS['TL_CONFIG']['imageHeight']    = 0;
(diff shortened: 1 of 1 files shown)


leofeyer:
@ausi /cc

aschempp:
Well the default value is not based on "usefulness" but on memory limits, right?

fritzmg:
> Well the default value is not based on "usefulness" but on memory limits, right?

You are probably thinking of the `gdMaxImgWidth` and `gdMaxImgHeight` setting, which is currently set to `3000` for both by default.

Aybee:
True, I have not even noticed, as I always upload with FTP.

Maybe the default setting should have no restriction.

gdMax... should take care of memory limits.

ausi:
> Maybe the default setting should have no restriction.

I agree, the default for `imageWidth` and `imageHeight` should be changed to `0`.

fritzmg:
Setting `imageWidth` and `imageHeight` to zero by default was not the solution unfortunately. This causes any front end uploads to not work anymore. You will get the following error message in the front end:
```
Die Datei … übersteigt die maximale Bildbreite von 0 Pixel!
```

ausi:
I think this is another bug then in https://github.com/contao/core-bundle/blob/133cebc0a9716f935637dcb64cd46a5b37920d02/src/Resources/contao/forms/FormFileUpload.php#L177 and https://github.com/contao/core-bundle/blob/133cebc0a9716f935637dcb64cd46a5b37920d02/src/Resources/contao/forms/FormFileUpload.php#L186

PR #1004, opened 2017-08-04
Cachecheck for uncombined scss files and adding TL_ASSETS_URL

agoat:
This PR is a solution for the problem mentioned in #987.

Before a combined scss (and css) file is generated, we look for an already existing cache file. Let's do the same for every single (not combined) scss file.

I also found another issue (IMO it's one). The TL_ASSETS_URL is only used when combining the asset files (css and js) and not when returning the single files.

### DIFF
--- a/src/Resources/contao/library/Contao/Combiner.php
+++ b/src/Resources/contao/library/Contao/Combiner.php
@@ -196,10 +196,17 @@ class Combiner extends \System
 	/**
 	 * Generates the files and returns the URLs.
 	 *
+	 * @param string $strUrl An optional URL to prepend
+	 *
 	 * @return array The file URLs
 	 */
-	public function getFileUrls()
+	public function getFileUrls($strUrl=null)
 	{
+		if ($strUrl === null)
+		{
+			$strUrl = TL_ASSETS_URL;
+		}
+		
 		$return = array();
 		$strTarget = substr($this->strMode, 1);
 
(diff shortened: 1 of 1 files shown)


leofeyer:
This does not fix the issue as @eki89 explained in the original ticket:

>An dieser Stelle würde eine Überprüfung, ob sich seit der letzten Kompilierung die SCSS-Quelldatei verändert hat, nicht ausreichen, da innerhalb der SCSS Datei befindliche Imports nicht berücksichtigt werden können.
> Vor der Kompilierung müsste also geprüft werden, ob sich eine eingebundene SCSS-Quelldatei, seit der letzten Kompilierung verändert hat, inklusive der darin importierten SCSS-Dateien, um daraufhin sämtliche SCSS Dateien neu zu generieren.

agoat:
> This does not fix the issue as @eki89 explained in the original ticket:

Mhh, I think I have to agree. I missed the point. 

Anyway, the cache check before compiling scss should also be done when not combining them! IMO it makes no sence why not, especially regarding http/2.
And the problem, to recognize file changes outside of the contao system, have to be solved elsewhere.

agoat:
Why is this closed?
The PR #1010 do not fix the problem addressed here.

Although I missed the main intension, this PR fixes some bugs.

A check for already compiled scss files and the use of `TL_ASSETS_URL` in the `getFileUrls` method should be added to get the same behaviour like the `getCombinedFileUrl` method.

leofeyer:
> A check for already compiled scss files and the use of `TL_ASSETS_URL` in the `getFileUrls` method should be added to get the same behaviour like the `getCombinedFileUrl` method.

Please open a new issue for this. It is a feature and not a bug fix.

PR #1009, opened 2017-08-08
Do not merge HTTP headers upon header replay

leofeyer:
When replaying HTTP headers, the `kernel.response` event is triggered twice, first with the `application/vnd.t42.header-replay` content type and then with the `text/html` content type.

The problem is that the `MergeHttpListeners` class adds the HTTP headers to the first response and then unsets them, so the second response (the actual response) ends up with no merged headers. This prevents tools like Blackfire from working correctly, because the `X-Blackfire-Query` header is not present.

### DIFF
--- a/src/EventListener/MergeHttpHeadersListener.php
+++ b/src/EventListener/MergeHttpHeadersListener.php
@@ -13,6 +13,7 @@ namespace Contao\CoreBundle\EventListener;
 use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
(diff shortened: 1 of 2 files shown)


Toflar:
I think this is not the correct approach. I think we should make sure that no response listener is triggered at all because it makes no sense to modify the response of the preflight request. If you want to do that you should register to the replay headers event. I'll see if I can fix that in the bundle.

leofeyer:
Thanks.

Toflar:
Fixed with https://github.com/terminal42/header-replay-bundle/commit/7da33c316b3d8055cff97d0dfc9969655a7be669 and released as 1.2.0.

leofeyer:
Is it possible that we have this issue in Contao 4.5 again?

leofeyer:
Never mind, it seems to be something else.

PR #1010, opened 2017-08-08
[RFC] Check if the SCSS/LESS files need to be recompiled

### LINKED ISSUE #987: Cache für SCSS-Dateien wird bei jedem Seitenaufruf neu erstellt.
Bei der Verwendung von SCSS-Dateien, werden diese bei jedem Seitenaufruf im Frontend neu kompiliert. Dadurch kommt es bei der Verwendung vieler SCSS-Dateien zu einer Verzögerung bei der Auslieferung der Seite.

Verantwortlich für die Kompilierung sind folgende Zeilen in der Combiner.php: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/Combiner.php#L215-L217

An dieser Stelle würde eine Überprüfung, ob sich seit der letzten Kompilierung die SCSS-Quelldatei verändert hat, nicht ausreichen, da innerhalb der SCSS Datei befindliche Imports nicht berücksichtigt werden können.
Vor der Kompilierung müsste also geprüft werden, ob sich eine eingebundene SCSS-Quelldatei, seit der letzten Kompilierung verändert hat, inklusive der darin importierten SCSS-Dateien, um daraufhin sämtliche SCSS Dateien neu zu generieren.

Kombiniert man die Dateien mit Hilfe des im Layout verankerten Flags "skripte zusammenfassen", ist Contao in der Lage die zusammengefasste Dateien gecached abzurufen. Änderungen an den SCSS-Quelldateien führen allerdings nicht zur Neukompilierung der zusammengefassten Cachedatei.

leofeyer:
This PR attempts to fix #987. It only checks the modification time of the main SCSS file and the files which are imported in it (so no nested imports).

### DIFF
--- a/src/Resources/contao/library/Contao/Combiner.php
+++ b/src/Resources/contao/library/Contao/Combiner.php
@@ -527,9 +527,35 @@ class Combiner extends \System
 			return true;
 		}
 
+		// Check the modification time of the main file
 		if (filemtime(TL_ROOT . '/' . $strFile) > filemtime(TL_ROOT . '/' . $strCacheFile))
 		{
 			return true;
+		}
+
+		$arrCheck = array();
+		$strPath = dirname($strFile);
+
+		$fh = fopen(TL_ROOT . '/' . $strFile, 'rb');
+
+		// Store the paths of the imported files
+		while (($line = fgets($fh)) !== false)
+		{
+			if (strpos($line, '@import') !== false)
+			{
+				$arrCheck[] = $strPath . '/' . str_replace(array('@import ', '"', "'", ';'), '', trim($line)) . '.scss';
+			}
+		}
+
+		fclose($fh);
+
+		// Recompile if the cache file is older than any of the imported files
+		foreach ($arrCheck as $file)
+		{
+			if (filemtime(TL_ROOT . '/' . $file) > filemtime(TL_ROOT . '/' . $strCacheFile))
+			{
+				return true;
+			}
 		}
 
 		return false;


leofeyer:
@contao/developers Do we need to support nested imports?

bytehead:
IMHO no.

Toflar:
I never use any of the built in CSS stuff. Local pre and post processors are way more powerful. But if we do provide such a feature at all, I think it should not be too hard to make a recursive method from what you have already (and I suppose there are libraries for that 😄 ).

fritzmg:
I agree with @Toflar. If it's done at all, it should be done 100% ;) (not that I care about it much anyway either).

agoat:
@leofeyer Have you done a performance test on this? Because this will run on ever(!) page call and if we parse all the scss files, a cache will be useless.

Even though I work in the opposite way and use only the buildin scss stuff (for me one of the benefits of contao), I have to agree to better using a pre/post processor when editing scss files outside of contao.

IMO a better solution would be to have a switch in the settings to deactivate the (s)css caching.

leofeyer:
> Have you done a performance test on this? Because this will run on every(!) page call and if we parse all the scss files, a cache will be useless.

I second this.

agoat:
I still prefer a solution to be able to deactivate the caching (force recompilation) at all. Then there is no need to check for newer scss files.

leofeyer:
As discussed in Mumble on August 10th, 2017, here's what we want to do:

 * If we are in debug mode, always recompile.
 * If in production and the cache file exists, use it.
 * If an SCSS/LESS file is edited in the back end, purge the script cache.
 * If the file is edited via FTP, the script cache needs to be purged manually.

The `filemtime()` check shall be removed.

leofeyer:
Fixed in 3158b45ce5d8a06070473af60fb4bcd271ddc373. Purging the cache has been implemented already: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/drivers/DC_Folder.php#L1914-L1919

codiso:
This behavior is not the best decision! If i work on server from my custoner, i have no choise to run contao in debug_mode ??? I use SCSS files every time i develop under contao. Now i have to clear the cache manualy every time i make changes …
Why ist there no switch in "Systemeinstellungen" to activate/bypass the caching?

Toflar:
Production mode: www.domain.de/ueber-uns.html
Debug mode: www.domain.de/app_dev.php/ueber-uns.html

The debug entrypoint is always there and is protected. The Contao Manager will make it very easy for you to set your own password in the future.
That's the way to go.

codiso:
OK, if Contao Manager will do that in the future, everything is fine. The new behavior was very irritating if you made it several years on the other way …

biwerr:
The current Caching Issue is realy annoying on dev environment which is not running with app_dev.php.
I also don't like the usage of app_dev.php because of the poor performance on some hosting packages due to the memory consumption for logging/debuging.

Is it realy a big deal to provide some backend/config functionality to control the caching system?

Issue #997, opened 2017-08-03
Wizard class is removed


dmolineus:
Before Contao 4.4.2 it was possible to set the `wizard` class manually using `tl_class` in the dca config. Useful if a custom backend widget provides always uses a widget.

The current implementation introduced with https://github.com/contao/core-bundle/commit/874076de55cbf962fd5e41c00e853e4dec8842a#diff-b187da7f3bf59a72fc3327dbb0246dd4R536 removes the wizard class if no custom wizard is added.

Furthermore it's would custom classes containing `wizard`, e.g. `wizard-2x`

leofeyer:
@aschempp /cc

aschempp:
> Useful if a custom backend widget provides always uses a widget.

What do you mean by that? Can you give an example?

aschempp:
🤔 if your widget always adds a wizard, it should also always add the wizard class? Not in the template, but in the widget attributes?

Btw, you should probably update your composer.json. [Support section](https://github.com/netzmacht/contao-leaflet-geocode-widget/blob/master/composer.json#L20) looks wrong and a `contao-bundle` does not require the composer plugin 😉

dmolineus:
> 🤔 if your widget always adds a wizard, it should also always add the wizard class? Not in the template, but in the widget attributes?

It depends on user configuration which class should be added. Often it's wizard, sometimes `wizard-2x` or `wizard-3x` if some custom wizards are added. 

In general there shouldn't be any magic changes to the classes which are applied manually in the dca settings. Maybe you could introduce a new css class instead which wasn't available in the previous releases to not break this beheaviour in an bugfix release.

> Btw, you should probably update your composer.json. Support section looks wrong and a contao-bundle does not require the composer plugin 😉

😮 Thanks.

leofeyer:
@aschempp Can't we simply remove the `str_replace()` call? It will not hurt if the `wizard` class is added twice, will it?

Toflar:
Completely untested but you get the point: https://github.com/contao/core-bundle/pull/1011

leofeyer:
See #1011 then.

PR #1011, opened 2017-08-09
Fix wizard css class

Toflar:
See #997

### DIFF
--- a/src/Resources/contao/classes/DataContainer.php
+++ b/src/Resources/contao/classes/DataContainer.php
@@ -566,7 +566,7 @@ abstract class DataContainer extends \Backend
 		
 		if (!emtpy($cssClasses))
 		{
-			arrData['eval']['tl_class'] = implode(' ', array_unique($cssClasses));
+			$arrData['eval']['tl_class'] = implode(' ', array_unique($cssClasses));
 		}
 
 		$updateMode = '';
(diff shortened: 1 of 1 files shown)


src/Resources/contao/classes/DataContainer.php
```diff
@@ -526,19 +526,21 @@ protected function row($strPalette=null)
...
 			{
-				$arrData['eval']['tl_class'] .= ' wizard';
+				$cssClasses[] = 'wizard';
 			}
 		}
-		elseif (isset($arrData['eval']['tl_class']) && false !== strpos($arrData['eval']['tl_class'], 'wizard'))
+		elseif (in_array('wizard', $cssClasses))
 		{
-			$arrData['eval']['tl_class'] = str_replace('wizard', '', $arrData['eval']['tl_class']);
+			unset($cssClasses[array_search('wizard', $cssClasses)]);
```
leofeyer: We should handle the case that `array_search` returns `false` here.

leofeyer:
I think this is the way to go. Thank you @Toflar.

src/Resources/contao/classes/DataContainer.php
Toflar: Cannot happen because we check if it's present :-D

leofeyer: True.

dmolineus:
This PR would fix the issue with css classes containing `wizard` but would still remove manually defined wizard classes.

leofeyer:
I don't think so. It would only remove a class called `wizard` but not `wizard-2x` or `wizard-3x`.

PR #1012, opened 2017-08-09
[RFC] Use TL_ASSETS_URL in the getFileUrls method

agoat:
This PR adds the same behaviour to the `getFileUrls` method like the `getCombinedFileUrl` method.

To be specific:
1. It adds a check for an already compiled scss file in the assets folder before a scss file will be compiled.
2. It uses/adds the `TL_ASSETS_URL` constant for all the single css urls.

One little question have to be answered: We have currently two constants: `TL_ASSETS_URL` and `TL_FILES_URL` (until Contao 5, see #661). The `TL_FILES_URL` is used for files from the `files` directory. Should we use [here](https://github.com/agoat/contao-core-bundle/blob/6647cd076cc86faa90b09ea8ec6a7e541fa65b13/src/Resources/contao/library/Contao/Combiner.php#L251) the `TL_FILES_URL` constant instead of the `TL_ASSETS_URL` for css and js files loaded from the `files` directory?


src/Resources/contao/library/Contao/Combiner.php
```diff
@@ -212,11 +219,18 @@ public function getFileUrls()
 			{
 				$strPath = 'assets/' . $strTarget . '/' . str_replace('/', '_', $arrFile['name']) . $this->strMode;
 
+				// Load the existing file
+				if (file_exists(TL_ROOT . '/' . $strPath))
```
leofeyer: This should be

```php
if (!file_exists(TL_ROOT . '/' . $strPath))
{
    $objFile = new \File($strPath);
    $objFile->write($this->handleScssLess($content, $arrFile));
    $objFile->close();
}
```

But the PR will conflict with the `hotfix/4.4.3` changes anyway once they are merged.

agoat: > But the PR will conflict with the hotfix/4.4.3 changes anyway once they are merged.

OK, you just added the same functionality like this. But what about the `TL_ASSETS_URL`?

agoat: > This should be ...

Changed in c93e5b9ed894e6f221340eb4f7be80fc3fe8ed9c

agoat:
I updated this PR, as @leofeyer added a check for already compiled scss files in the contao 4.4.3 version. Now the PR adds what is left over: The use of the TL_ASSETS_URL in the `getFileUrls` method.

leofeyer:
Thank you @agoat. I am a little concerned about the

```php
if ($strUrl === null)
{
	$strUrl = TL_ASSETS_URL;
}
```

part though, because it is not backwards compatible. If an extension relies on the method returning relative paths and prepends the asserts URL itself, this change would break it.

Toflar:
It is a BC break and cannot be accepted, yes.

agoat:
I agree with @Toflar. This will be breaking BC like @leofeyer described above. 

But IMO the both methods `getFileUrls` and `getCombinedFileUrl` should have the same behaviour when a `TL_ASSETS_URL` ist set. Currently css and js files are not called from the assets domain if the page is called in debug mode (app_dev.php).

What about adding a depreciated hint and add this to the 4.5 milestone?

aschempp:
Why would a developer rely on an empty url?

agoat:
I will close this PR.

As already mentioned above, it's a BC break.
The intention was to use the same arguments for the methods of `getFileUrls` and `getCombinedFile` (and to add automatically the `TL_ASSETS_URL` in both).

The handling of url prefixes when using the `getFileUrls` method have to be done in the calling function.

PR #1016, opened 2017-08-10
[RTM] Use ausi/slug-generator for the creation of aliases

ausi:
As discussed in #954

@contao/developers Is an event class the right thing for this use case?

ToDo:

- [x] Wait until CLDR 32.0 is released (19th October) and tag version 1.0.0 of ausi/slug-generator
- [x] Create pull requests for news-bundle and events-bundle: contao/news-bundle#20, contao/calendar-bundle#14


agoat:
Have only looked quickly over it. Great.

Just as a note: Your package require `php: ^7.1`. If this will be merged, contao should also raise the minimum php version to 7.1.

Toflar:
It already is ;) See the develop branch :)

ausi:
Yes, the PHP requirement is the same as in Contao 4.5. But it also requires `ext-intl`, `ext-mbstring` and `ext-pcre`, maybe we should discuss if it is OK to require these extensions?

leofeyer:
I guess so. Is there an alternative?

Toflar:
Of course! All of them are required for a decent multilingual (and thus multibyte/utf-8) setup.

ausi:
`ext-pcre` is already required by Contao and `patchwork/utf8`.

`ext-intl` and `ext-mbstring` are already required by `league/uri` which itself is required since Contao 4.1.

Therefore I think I can keep the requirements of the slug library as they are now.

ausi:
The `de-ASCII` rules are included in the latest CLDR version 32-beta3, so I released `ausi/slug-generator` version 1.0 and updated this PR.

It should be RTM now 🎉

tests/Slug/ValidCharactersTest.php
```diff
@@ -50,23 +54,43 @@ function (SlugValidCharactersEvent $event) use (&$path): bool {
...
-        ];
+        $translator
+            ->expects($this->at(0))
+            ->method('trans')
+            ->with('MSC.validCharacters.unicodeLowercase')
+            ->willReturn('Unicode numbers and lowercase letters')
+        ;
 
+        $translator
+            ->expects($this->at(1))
```
ausi: Is it necessary to test how the tranlator is used?

src/Slug/ValidCharacters.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @var TranslatorInterface
+     */
+    private $translator;
+
+    /**
+     * @var string[]
+     */
+    private $defaultOptions = [
```
leofeyer: Could this also be a constant or static property?

src/Event/ContaoCoreEvents.php
```diff
@@ -60,7 +60,7 @@
     public const BACKEND_MENU_BUILD = 'contao.backend_menu_build';
 
     /**
-     * The contao.slug_valid_characters event is triggered when the options for the valid characters setting for aliases are generated.
+     * The contao.slug_valid_characters event is triggered when the valid slug characters option is generated.
```
ausi: I think this should be plural “options are generated”

src/Resources/contao/languages/en/tl_page.xlf
```diff
@@ -99,10 +99,10 @@
         <source>The date and time format string will be parsed with the PHP date() function.</source>
       </trans-unit>
       <trans-unit id="tl_page.validAliasCharacters.0">
-        <source>Valid characters for aliases</source>
+        <source>Valid alias characters</source>
       </trans-unit>
       <trans-unit id="tl_page.validAliasCharacters.1">
-        <source>When aliases get created automatically, this character set gets used for the conversion.</source>
+        <source>Please choose the character for auto-generating page aliases.</source>
```
ausi: Also plural “characters”

src/Slug/ValidCharacters.php
ausi: Yes, a private constant should be fine.

src/Event/ContaoCoreEvents.php
leofeyer: Are there several valid slug characters options?

src/Resources/contao/languages/en/tl_page.xlf
leofeyer: True.

tests/Slug/ValidCharactersTest.php
leofeyer: Probably not. 😄

leofeyer: Although the test assures that the class iterates over all four default options.

ausi: But the order should not be relevant. I think we should remove the translator tests here.

src/Event/ContaoCoreEvents.php
ausi: Yes, by default there are four options: unicode lowercase, unicode, ASCII lowercase and ASCII.

leofeyer:
Very good job @ausi. Thanks a bunch.

backbone87:
So 4.4 LTS does not and will not support this?

ausi:
@backbone87 No, but there is an extension which backports the new behaviour to 4.4: [ausi/contao-slug-backport](https://github.com/ausi/contao-slug-backport)

backbone87:
@ausi is it worth to add a contao specific impl of Ausi\SlugGenerator\SlugGenerator that provides a method with a PageModel argument, that automatically extracts the slug configuration and builds it? that would remove the duplicated code that gathers this configuration in various listeners and eases the integration for 3rd party bundles

ausi:
I’m not sure if this is necessary. It’s not that much code that transforms the page model to the options array:

```php
$slugOptions['locale'] = $objPage->language;
if ($objPage->validAliasCharacters) {
    $slugOptions['validChars'] = $objPage->validAliasCharacters;
}
```

backbone87:
well, you also must make sure, that $page->loadDetails() is executed and possibly strip insert tags.

further StringUtil::generateAlias should be deprecated and changed to use the slug generator

ausi:
> further StringUtil::generateAlias should be deprecated and changed to use the slug generator

sounds reasonable, I think you should open a new ticket for that.

ausi:
Related: #1334

PR #1020, opened 2017-08-11
[RFC] Fetch file attributes only if file exists

agoat:
When creating a new file instance with 
```php
$file = new File('image.jpg')
``` 
and then calling 
```php
$file->get('mtime')
```
it will result in a **'ContextErrorException'** (HTTP 500 Internal Server Error in the FE) when trying to get `filemtime()` from a file that doesn't exist.

This PR adds a check to see if a file exist before it tries to fetch a file attribute.

### DIFF
--- a/src/Resources/contao/library/Contao/File.php
+++ b/src/Resources/contao/library/Contao/File.php
@@ -173,7 +173,7 @@ class File extends \System
 		{
 			case 'size':
 			case 'filesize':
-				return filesize(TL_ROOT . '/' . $this->strFile);
+				return ($this->exist) ? filesize(TL_ROOT . '/' . $this->strFile) : null;
 				break;
 
 			case 'name':
@@ -227,15 +227,15 @@ class File extends \System
 				break;
 
 			case 'ctime':
-				return filectime(TL_ROOT . '/' . $this->strFile);
+				return ($this->exist) ? filectime(TL_ROOT . '/' . $this->strFile) : null;
 				break;
 
 			case 'mtime':
-				return filemtime(TL_ROOT . '/' . $this->strFile);
+				return ($this->exist) ? filemtime(TL_ROOT . '/' . $this->strFile) : null;
 				break;
 
 			case 'atime':
-				return fileatime(TL_ROOT . '/' . $this->strFile);
+				return ($this->exist) ? fileatime(TL_ROOT . '/' . $this->strFile) : null;
 				break;
 
 			case 'icon':
--- a/src/Resources/contao/library/Contao/File.php
+++ b/src/Resources/contao/library/Contao/File.php
@@ -223,7 +223,7 @@ class File extends \System
 				break;
 
 			case 'hash':
-				return $this->getHash();
+				return ($this->exist) ? $this->getHash() : null;
 				break;
 
 			case 'ctime':
--- a/src/Resources/contao/library/Contao/File.php
+++ b/src/Resources/contao/library/Contao/File.php
@@ -173,7 +173,7 @@ class File extends \System
 		{
 			case 'size':
 			case 'filesize':
-				return ($this->exist) ? filesize(TL_ROOT . '/' . $this->strFile) : null;
+				return ($this->exists()) ? filesize(TL_ROOT . '/' . $this->strFile) : null;
 				break;
 
 			case 'name':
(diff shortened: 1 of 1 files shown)


leofeyer:
Wrong approach IMO. You should call `$file->exists` in your code before calling `$file->mtime` if the file might not exist.

agoat:
> You should call $file->exists in your code before calling $file->mtime

You should, but...
Sometimes it's not quite possible. At least the contao core don't check before trying to get file attributes (see the [File class](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/File.php#L259)).

In my particular case I defer the image resizing and the images don't exist. But the core try to get file attributes which result in an exception. 

IMO it's better to have a check in the file class to avoid exceptions before trying the fetch file attributes. 
It don't hurt and I can reduce the number of overwritten classes.

agoat:
Why is this closed?
Do you see a problem including a check if a file exists before starting io operations?

leofeyer:
The ticket has been closed automatically, because you have targeted it against the `hotfix/4.4.3` branch instead of the `master` branch.

agoat:
I tried to avoid to get file attributes from not existing files. But the problem is in the `addImageToTemplate` method in the `Controller` class. 
After the image is resized the width and height is taken from the file directly ($objFile->imageSize). This could also be taken from the result of the `picture_factory`. 

I was thinking about to open a PR for this, but not sure to make the `File` class io operations safe or to avoid getting attributes from files which are already calculated. 


```diff
try
 		{
 			$src = \System::getContainer()->get('contao.image.image_factory')->create(TL_ROOT . '/' . $arrItem['singleSRC'], $size)->getUrl(TL_ROOT);
 			$picture = \System::getContainer()->get('contao.image.picture_factory')->create(TL_ROOT . '/' . $arrItem['singleSRC'], $size);
 
 			$picture = array
 			(
 				'img' => $picture->getImg(TL_ROOT, TL_FILES_URL),
 				'sources' => $picture->getSources(TL_ROOT, TL_FILES_URL)
 			);
-
-			if ($src !== $arrItem['singleSRC'])
-			{
-				$objFile = new \File(rawurldecode($src));
-			}
 		}
 		catch (\Exception $e)
 		{
@@ -1550,10 +1545,10 @@ public static function addImageToTemplate($objTemplate, $arrItem, $intMaxWidth=n
 		}
 
 		// Image dimensions
-		if (($imgSize = $objFile->imageSize) !== false)
-		{
-			$objTemplate->arrSize = $imgSize;
-			$objTemplate->imgSize = ' width="' . $imgSize[0] . '" height="' . $imgSize[1] . '"';
+		if ($picture['img']['width'] !== null && $picture['img']['height'] !== null)
+		{	
+			$objTemplate->arrSize = array($picture['img']['width'], $picture['img']['height']);
+			$objTemplate->imgSize = ' width="' . $picture['img']['width'] . '" height="' . $picture['img']['height']  . '"';
 		}
```
(https://github.com/contao/core-bundle/compare/master...agoat:fix/readImagesSizeFromPictureFactory)


Performance wise it looks like there is no measureable difference to read the the width and height from the file or the array with arround 50 images.

leofeyer:
@ausi /cc

ausi:
`File->imageSize` is not just an array of width and height, it includes much more information as you can see in https://github.com/contao/core-bundle/blob/2ec59c9ebb6579d4fb63ff992808ccfb64eb238b/src/Resources/contao/library/Contao/File.php#L277-L286

So we have to use `File->imageSize` to get all these values for `$objTemplate->arrSize`.

I don’t think we should add an `exists()` check to all operations in the `File` class, this should be done in the calling code that uses the `File` class.

The problem in `Controller::addImageToTemplate()` should be fixed by changing:

```diff
-		if (($imgSize = $objFile->imageSize) !== false)
+		if ($objFile->exists() && ($imgSize = $objFile->imageSize) !== false)
```

agoat:
> The problem in Controller::addImageToTemplate() should be fixed by changing: ...
```php
if ($objFile->exists() && ($imgSize = $objFile->imageSize) !== false)
```

So easy. Why didn't I came up with this myself ;-)

leofeyer:
Fixed in 0ec64e76a6e0689e19a79e735a70466dc2a07978. Thank you @ausi.

PR #1023, opened 2017-08-14
[RFC] Clear script cache when editing js files

agoat:
Can we also clear the script cache when editing a js file with the source editor?


leofeyer:
Thank you @agoat.

PR #1025, opened 2017-08-16
Fixed disabling the button in modal

Toflar:


leofeyer:
As briefly discussed in Slack:

* The logic should only be applied to the "Apply" button. The "Close" button must never be disabled.
* If `act=select` is in the URL, the "Apply" button should to be disabled.
* In all other cases, the button must remain enabled.

leofeyer:
Nope, this is not correct. There is also the case that I have opened the picker but am now editing a record. The button needs to be disabled in this case.

leofeyer:
IMO the current code is correct. It disables the buttons except if there are picker checkboxes or radio buttons. Only the "Close" button should never be disabled.

leofeyer:
We should only apply the `disabled` attribute to the `btn.primary` button.

Toflar:
I agree 👍

aschempp:
I agree as well. But why is the button called "close" instead of "cancel"??

leofeyer:
Because it closes the modal window I guess. 😄

leofeyer:
Fixed in ddf2360b1b3e3fa9fabe7f52eec63cf54a719c5f.

PR #1046, opened 2017-08-29
[RFC] Add the possibility to save hidden version data

agoat:
With this PR I want to add the possibility to save versions in the `tl_version` table, which are not directly editable and therefore should not be  shown on the welcome page.

This makes the version table useable for extensions to save versions for extra data.

As far as I understand, the editUrl is always saved when a new version is created. So I simply added an extra WHERE statement. Another solution would be to have an extra `hide` column.

**To give an example for a use case:**
I have an extension which saves extra data related (and edited) with content elements in an extra table. When the content element is saved, a version is created and I use the `oncreate_version_callback` to save a version for the extra data, too. But these are not directly editable and makes the 'Last changes' overview unreadable.

### DIFF
--- a/src/Resources/contao/classes/Versions.php
+++ b/src/Resources/contao/classes/Versions.php
@@ -632,7 +632,7 @@ class Versions extends \Controller
 		$objDatabase = \Database::getInstance();
 
 		// Get the total number of versions
-		$objTotal = $objDatabase->prepare("SELECT COUNT(*) AS count FROM tl_version WHERE version>1" . (!$objUser->isAdmin ? " AND userid=?" : ""))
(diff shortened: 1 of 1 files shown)


leofeyer:
How is the extra data created/edited?

aschempp:
Using onversion and onrestore callbacks I guess… I tried that with Isotope as well, but it does show "unwanted" entries...

agoat:
> Using onversion and onrestore callbacks I guess… I tried that with Isotope as well, but it does show "unwanted" entries...

Exactly what @aschempp presumed..
The extra data is saved while editing a content element with the `onsubmit_callback` into an extra table (using virtual fields). And when a new version is created for the content element I have to save also a version for the extra data using the `oncreate_version_callback`. I don't have an editUrl because the data can only be edited through the content element. 
But the extra saved version will be displayed in the list with the last changes on the welcome page,  but that doesn't make much sense.

leofeyer:
@contao/developers Should we keep the explicit `IS NOT NULL` check or should we also ignore entries with an empty edit URL (empty string)?

PR #1055, opened 2017-09-07
Catch general exceptions during route generation

Toflar:
When there is an exception (such as if a route does not exist at all) during route generation, our pretty error screen listener fails because the framework does not boot properly. Is there any reason why we only catch the - global - InvalidArgumentException here?


leofeyer:
I guess we do because all the possible exceptions in the `UrlGeneratorInterface::generate()` method inherit from the `\InvalidArgumentException` exception. Which exception did you have?

Toflar:
A `FileLoaderLoadException`. If the route exists but the controller to it does not.

leofeyer:
I see. So if we caught all exceptions here, the `getRoute()` method would return `null`. Is this ok?

@aschempp /cc

aschempp:
Jup. Already confirmed with @Toflar

PR #1057, opened 2017-09-07
fixed GLOB_BRACE warning on alpine linux systems

iCodr8:
"Note: The GLOB_BRACE flag is not available on some non GNU systems, like Solaris." <= php.net

Docker based alpine systems do not support "GLOB_BRACE". The simple fix is done with this merge request, so that apline linux systems can work nevertheless.

See
- http://php.net/manual/de/function.glob.php#refsect1-function.glob-notes
- https://symfony.com/blog/new-in-symfony-3-3-psr-4-based-service-discovery#comment-21379
- https://github.com/symfony/symfony/pull/21289/files#diff-ad1ed76aba6a80df5a48dfa4585adcf3R115

### DIFF
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -101,7 +101,7 @@ abstract class Controller extends \System
 		}
 
 		$strBrace = '{' . implode(',', \StringUtil::trimsplit(',', strtolower(\Config::get('templateFiles')))) . '}';
-		$arrCustomized = glob(TL_ROOT . '/templates/' . $strPrefix . '*.' . $strBrace, GLOB_BRACE);
+		$arrCustomized = glob(TL_ROOT . '/templates/' . $strPrefix . '*.' . $strBrace, defined('GLOB_BRACE') ? GLOB_BRACE : 0);
 
 		// Add the customized templates
 		if (is_array($arrCustomized))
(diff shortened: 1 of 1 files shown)


leofeyer:
How can omitting the flag be a solution? Does the function return the correct result without it?

iCodr8:
Im my case (docker alpine linux) "GLOB_BRACE" returns an empty string, but an integer is required. So I searched for a solution and found that, how symfony fixed it. The problem is, that the templates from the templates folder in root couldn't be found anyway :(

leofeyer:
> The problem is, that the templates from the templates folder in root couldn't be found anyway :(

Of course not.

iCodr8:
To avoid the alpine problems, I changed from `php:7.1-fpm-alpine` to `php:7.1-fpm` now.

Toflar:
@iCodr8 a PR fixing this would be very welcome. It should look something like this:

```diff
- $arrCustomized = glob(TL_ROOT . '/templates/' . $strPrefix . '*.' . $strBrace, GLOB_BRACE);
+ $arrCustomized = $this->braceGlob(TL_ROOT . '/templates/' . $strPrefix . '*.' . $strBrace);
```

```php
private function braceGlob($resource)
{
    if (false === strpos($resource, '/**/') && (defined('GLOB_BRACE') || false === strpos($resource, '{'))) {
        return glob(resource, GLOB_BRACE);
    }

    $finder = new Finder();
    $regex = Symfony\Component\Finder\Glob::toRegex($resource);

    // Use finder to filter for the regular expression
}
```

So basically it checks if it even needs `GLOB_BRACE` and if it is defined and then falls back to the finder using a regular expression filtering which works too but is slower. See the Symfony code for reference: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Config/Resource/GlobResource.php#L98-L140

Toflar:
Maybe you can also use the `GlobResource` directly ;)

iCodr8:
I will test and solve it :)

leofeyer:
Siehe https://github.com/contao/core-bundle/pull/1071

PR #1063, opened 2017-09-11
[RFC] Improve the UrlGeneratorTest

aschempp:


src/Routing/UrlGenerator.php
```diff
@@ -141,6 +141,8 @@ private function prepareAlias($alias, array &$parameters)
         $hasAutoItem = false;
         $autoItems = $this->getAutoItems($parameters);
 
+        unset($parameters['auto_item']);
```
leofeyer: Why did you change this? It has worked without it, hasn't it?

tests/Routing/UrlGeneratorTest.php
```diff
@@ -364,4 +356,32 @@ private function getGenerator($prependLocale = false, $returnArgument = 1, $useA
...
     }
+
+    /**
+     * @param array $expectedParameters
+     * @param       $expectedRoute
+     * @param int   $referenceType
+     *
+     * @return \PHPUnit_Framework_MockObject_MockObject|UrlGeneratorInterface
+     */
+    private function expectRouter(
```
leofeyer: CS: no line breaks in method definitions. 😄

src/Routing/UrlGenerator.php
aschempp: No, the test failed after correctly implementing it 😉

tests/Routing/UrlGeneratorTest.php
aschempp: Fixed in 88d9d28703ef82958c464bb4909698b0eb206ced

leofeyer:
Merged in 1a1e035ba0b44d167092e30131a3a33f75720579 (`hotfix/4.4.5`).

PR #1065, opened 2017-09-11
[RTM] Store the request token as cookie instead of the session

ausi:
Use a [double submit cookie](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Double_Submit_Cookie) instead of storing the request token in the session of the user.


src/Csrf/MemoryTokenStorage.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        $this->tokens = $tokens;
+    }
+
+    /**
+     * Get all active tokens.
+     *
+     * @return array
+     */
+    public function getSaveTokens()
```
aschempp: I don't like this method name, shouldn't it be `getActiveTokens` if the internal variable is called `$activeTokens`? Also, I would prefer to name that `usedTokens` 😁 

The method can be simplified like this:

```php
return array_intersect_key($this->tokens, $this->activeTokens);
```

src/EventListener/CsrfTokenCookieListener.php
```diff
@@ -0,0 +1,119 @@
...
+    private $cookiePrefix;
+
+    /**
+     * Constructor.
+     *
+     * @param MemoryTokenStorage $tokenStorage
+     * @param int                $cookieLifetime
+     * @param string             $cookiePrefix
+     */
+    public function __construct(MemoryTokenStorage $tokenStorage, int $cookieLifetime = 86400, string $cookiePrefix = 'csrf_')
```
aschempp: Does this have to be a `MemoryTokenStorage`? Shouldn't we expect the interface?

src/EventListener/CsrfTokenCookieListener.php
```diff
@@ -0,0 +1,119 @@
...
+     * @param ParameterBag $cookies
+     *
+     * @return array
+     */
+    private function getTokensFromCookies(ParameterBag $cookies)
+    {
+        $tokens = [];
+
+        foreach ($cookies as $key => $value) {
+            if (strncmp($key, $this->cookiePrefix, strlen($this->cookiePrefix)) === 0) {
```
aschempp: `strpos($this->cookiePrefix, $key) === 0` ?

src/Resources/config/listener.yml
```diff
@@ -161,3 +161,11 @@ services:
...
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
             - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse }
+
+    contao.listener.csrf_token_cookie:
+        class: Contao\CoreBundle\EventListener\CsrfTokenCookieListener
+        arguments:
+            - "@contao.csrf.token_storage"
+        tags:
+            - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 20 }
+            - { name: kernel.event_listener, event: kernel.response, method: onKernelResponse, priority: 0 }
```
aschempp: you don't need to set a priority if it's `0`.

src/EventListener/CsrfTokenCookieListener.php
ausi: We call the `initialize()` method on it, so we cannot use the interface.

src/EventListener/CsrfTokenCookieListener.php
ausi: `strncmp()` should be way faster IMO.

src/Csrf/MemoryTokenStorage.php
ausi: Done in c3f6afb94f567a587c763f1a32e84d8ecd29bdcb

src/Resources/config/listener.yml
ausi: Done in d874349ecc242ae71e3fc096da15369a9c281aa3

src/EventListener/CsrfTokenCookieListener.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $cookieLifetime = $this->cookieLifetime ? $this->cookieLifetime + time() : 0;
+        $isSecure = $event->getRequest()->isSecure();
+
+        foreach ($this->tokenStorage->getUsedTokens() as $key => $value) {
+            $event->getResponse()->headers->setCookie(
```
aschempp: I think you should use `deleteCookie` if the value is `null`. This way Symfony handles the unset case.

ausi: deleteCookie needs the same arguments as setCookie, except value and lifetime. Using deleteCookie would make a block of duplicate code.

aschempp: but it would move the delete logic to the header bag instead of our class, right?

ausi: If you refer to `$value === null ? 1 : $cookieLifetime` as the delete logic, then yes.

PR #1066, opened 2017-09-11
[RFC] Use KnpMenu for the main backend menu

sheeep:
This PR refactors the way the backend menu is created.

* [x] Check `hasAccess` on backend modules
* [x] Remove / Deprecate old methods in `Contao\BackendUser`
* [x] Implement `getUserNavigation` hook
* [x] A lot of CS stuff


src/EventListener/BackendMenu/UserMenuListener.php
```diff
@@ -0,0 +1,65 @@
...
+                        'title' => $categoryOptions['title'],
+                        'href' => $categoryOptions['href']
+                    ]
+                ]);
+
+                $node->setDisplayChildren(strpos($categoryOptions['class'], 'node-expanded') !== false);
+                $tree->addChild($node);
+            }
+
+            // Create the child nodes
```
aschempp: I would prefer to refactor this to a method that generates the child.

src/Resources/config/listener.yml
```diff
@@ -40,6 +40,14 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.terminate, method: onKernelTerminate }
 
+    contao.listener.backend_menu.legacy_menu_listener:
```
aschempp: same here with the namespace

src/Resources/contao/templates/backend/be_main.html5
```diff
@@ -54,23 +54,7 @@
   <div id="container" class="cf">
     <?php if (!$this->isPopup): ?>
       <div id="left">
-        <nav id="tl_navigation" role="navigation">
```
aschempp: If we would keep this element here and just render the `ul` part, we could skip the skipNavigation part from our menu renderer (including the language labels in Twig 😉 )

src/Event/BackendMenuEvent.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Contao\CoreBundle\Event;
+
+use Knp\Menu\ItemInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class BackendMenuEvent extends Event
```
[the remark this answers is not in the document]
sheeep: > Or does such an event class maybe already exist in KnpMenu?

I don't find any such events in either KnpMenu and KnpMenuBundle.

> can or should we have a universal menu event?

We probably could. I can't think of any reasons or possibilities for having more than the tree in this event.

src/EventListener/BackendMenu/UserMenuListener.php
```diff
@@ -0,0 +1,65 @@
...
+
+    public function __construct(FactoryInterface $factory, TokenStorageInterface $tokenStorage)
+    {
+        $this->factory = $factory;
+        $this->tokenStorage = $tokenStorage;
+    }
+
+    public function onBuild(BackendMenuEvent $event)
+    {
+        $tree = $event->getTree();
```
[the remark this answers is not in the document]
sheeep: Correct, we don't need to return the tree.

src/Event/MenuEvent.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Contao\CoreBundle\Event;
+
+use Knp\Menu\ItemInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+class MenuEvent extends Event
+{
+    private $tree;
+
+    public function __construct(ItemInterface $tree)
```
aschempp: The event should also contain the menu factory, so it does not need to be injected into the listener.

src/Menu/BackendMenuRenderer.php
```diff
@@ -0,0 +1,29 @@
...
+
+    public function __construct(Environment $templating)
+    {
+        $this->templating = $templating;
+    }
+
+    public function render(ItemInterface $item, array $options = [])
+    {
+        $lang = [
+            'skipNavigation' => \StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['skipNavigation'])
```
aschempp: This can later be solved with https://github.com/contao/core-bundle/pull/1072

src/Resources/config/services.yml
```diff
@@ -138,6 +138,24 @@ services:
...
+        arguments:
+            - "@knp_menu.factory"
+            - "@event_dispatcher"
+
+    contao.menu.backend_menu:
+        class: Knp\Menu\MenuItem
+        factory: ["@contao.menu.backend_menu_builder", create]
+        arguments: ["@request_stack"]
+        tags:
+            - { name: knp_menu.menu, alias: backend }
```
aschempp: Not sure if we need this tag, but if then the alias is not useful.

src/Resources/config/services.yml
```diff
@@ -138,6 +138,24 @@ services:
...
+        arguments:
+            - "@twig"
+
+    contao.menu.backend_menu_builder:
+        class: Contao\CoreBundle\Menu\BackendMenuBuilder
+        arguments:
+            - "@knp_menu.factory"
+            - "@event_dispatcher"
+
+    contao.menu.backend_menu:
```
aschempp: According to discussion with @sheeep we don't need the menu as a service because of a custom renderer.

leofeyer:
I'm done fixing the coding style and switching to the translator. There are, however, two issue that need to be addressed in the `be_menu.html.twig` template:

**The skip navigation URL**

OLD

```php
<a href="<?= $this->request ?>#skipNavigation" class="invisible"><?= $this->skipNavigation ?></a>
```

NEW

```twig
<a href="#skipNavigation" class="invisible">{{ 'MSC.skipNavigation'|trans }}</a>
```

What happened to the `$this->request` variable?

**The ajaxUrl parameter**

OLD

```php
<li class="tl_level_1_group […] AjaxRequest.toggleNavigation( […] '<?= $arrModules['ajaxUrl']; ?> […]
```

NEW

```twig
<li class="tl_level_1_group […] AjaxRequest.toggleNavigation( […] '{{ path('contao_backend') }} […]
```

Do we no longer need the `ajaxUrl` parameter?

leofeyer:
Never mind, I just realized that the `contao_backend` route **is** the Ajax URL.

leofeyer:
Thanks a lot @sheeep.

Issue #1061, opened 2017-09-11
Windows builds on AppVeyor are failing


ausi:
`FilePickerProviderTest::testGetDcaAttributes` fails since 874076de55cbf962fd5e41c00e853e4dec8842a6

`InstallCommandTest::testInstallation` fails since c38e9fc28952c88cee324a3ee5a9d3a208cd0e1f

patrickjDE:
`InstallCommandTest::testInstallation` probably fails because of the known(?) limitation that creating symlinks via php on Windows requires administrator privileges.

fritzmg:
That's the result of FilePickerProviderTest on my system:
```
$ vendor/bin/phpunit --filter=FilePickerProviderTest
PHPUnit 5.7.21 by Sebastian Bergmann and contributors.

Testing Contao Core Bundle
..........F.                                                      12 / 12 (100%)

Time: 374 ms, Memory: 12.00MB

There was 1 failure:

1) Contao\CoreBundle\Tests\Picker\FilePickerProviderTest::testReturnsTheDcaAttributes
Failed asserting that Array &0 (
    'fieldType' => 'radio'
    'filesOnly' => true
    'value' => 'C%3A%5CUsers%5CSpooky%5CDocuments%5CVCS%5Ccontao-core-bundle%5Ctests%5CPicker/foobar'
) is identical to Array &0 (
    'fieldType' => 'radio'
    'filesOnly' => true
    'value' => 'C:\Users\Spooky\Documents\VCS\contao-core-bundle\tests\Picker/foobar'
).

C:\Users\Spooky\Documents\VCS\contao-core-bundle\tests\Picker\FilePickerProviderTest.php:329

FAILURES!
Tests: 12, Assertions: 27, Failures: 1.
```
But this is a general problem with the test I think. `FilePickerProvider::getDcaAttributes` returns the value url encoded, while the test simply asserts `__DIR__.'/foobar'`, which is not url encoded.

fritzmg:
> InstallCommandTest::testInstallation probably fails because of the known(?) limitation that creating symlinks via php on Windows requires administrator privileges.

No, the error happens regardless of that. Haven found the reason yet though.

fritzmg:
Analysis of `InstallCommandTest::testInstallation`:
```php
$this->getRootDir()
```
within the test cases will return
```
C:\Users\User\Documents\VCS\contao-core-bundle\tests\Fixtures
```
for example. Then, when
```php
[...]

PR #1069, opened 2017-09-12
[RFC] Fix failing tests on Windows

ausi:
See #1061 

@aschempp Is my change in *FilePickerProviderTest.php* correct? I think the `PickerConfig` only handles relative paths, not absolute ones.


ausi:
@leofeyer can we configure Coveralls that a coverage decrease of 0.5% is allowed and not marked as a failure? This way we could prevent failures like this:

coveralls-failed

PR #1071, opened 2017-09-12
[RFC] Created a new method braceGlob to polyfill GLOB_BRACE

sheeep:
Discussion in #1057.

### DIFF
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -2362,11 +2362,7 @@ abstract class Controller extends \System
 		{
 			$path = $info->getPathname();
 
-			if (preg_match($regex, $path) && $info->isFile()) {
-				return true;
-			}
-
-			return false;
+			return preg_match($regex, $path) && $info->isFile();
 		});
 
 		$files = iterator_to_array($filesIterator);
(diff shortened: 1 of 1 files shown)


src/Resources/contao/library/Contao/Controller.php
```diff
@@ -2341,4 +2343,30 @@ protected function createNewVersion($strTable, $intId)
 		$objVersions = new \Versions($strTable, $intId);
 		$objVersions->create();
 	}
+
+	protected static function braceGlob($resource)
+	{
+		if (false === strpos($resource, '/**/') && (defined('GLOB_BRACE') || false === strpos($resource, '{')))
+		{
+			return glob($resource, defined('GLOB_BRACE') ? GLOB_BRACE : 0);
```
Toflar: The `defined()` check here is redundant :)

sheeep: It's not.

Given that `false === strpos($resource, '/**/')` and `false === strpos($resource, '{'))` we still need that check. Even though the flag could be omitted IMO. This comes directly from [Symfonys `GlobalResource`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Config/Resource/GlobResource.php#L98-L99).

Toflar: Ah, true!

leofeyer:
Merged into the correct branch in e98544b739d013a5019a8a5cad64ca96ae3087ba.

Thank you very much @sheeep. I did some profiling and `glob()` is about twice as fast, so good that we have both solutions now. 👍 

with-glob

without-glob

PR #1072, opened 2017-09-12
[RTM] Translator

ausi:
Can be used as following:

```php
->trans('MSC.view', [], 'contao_default');
->trans('ERR.minlength', ['My Field', 1234], 'contao_default');
->trans('websiteTitle.0', [], 'contao_tl_settings');
->trans('MSC.key\.mit\.punkt\.und\\\\backslash', 'contao_default');
```

Twig:

```twig
{% trans_default_domain 'contao_default' %}
{{ 'MSC.view'|trans }}
{{ 'ERR.minlength'|trans(['My Field', 1234]) }}
{{ 'websiteTitle.0'|trans({}, 'contao_tl_settings') }}
{{ 'MSC.key\\.mit\\.punkt\\.und\\\\backslash'|trans }}
```

### DIFF
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -49,12 +49,9 @@ class Translator implements TranslatorInterface
             return $this->translator->trans($id, $parameters, $domain, $locale);
         }
 
-        if (!$this->framework->isInitialized()) {
-            return $id;
-        }
-
         $domain = substr($domain, 7);
 
+        $this->framework->initialize();
         $this->loadLanguageFile($domain);
 
         $translated = $this->getFromGlobals($id, $domain);
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -43,6 +43,10 @@ class Translator implements TranslatorInterface
 
     /**
      * {@inheritdoc}
+     *
+     * Gets the translation from Contao’s $GLOBALS['TL_LANG'] array
+     * if the message domain starts with "contao_".
+     * The locale parameter is ignored in this case.
      */
     public function trans($id, array $parameters = [], $domain = null, $locale = null)
     {
@@ -70,6 +74,8 @@ class Translator implements TranslatorInterface
 
     /**
      * {@inheritdoc}
+     *
+     * Forwards to the default translator, doesn’t handle $GLOBALS['TL_LANG'].
      */
     public function transChoice($id, $number, array $parameters = [], $domain = null, $locale = null)
     {
@@ -78,6 +84,8 @@ class Translator implements TranslatorInterface
 
     /**
      * {@inheritdoc}
+     *
+     * Forwards to the default translator, doesn’t handle $GLOBALS['TL_LANG'].
      */
     public function setLocale($locale)
     {
@@ -86,6 +94,8 @@ class Translator implements TranslatorInterface
 
     /**
      * {@inheritdoc}
+     *
+     * Forwards to the default translator, doesn’t handle $GLOBALS['TL_LANG'].
      */
     public function getLocale()
     {
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -106,8 +106,10 @@ class Translator implements TranslatorInterface
             $id = $domain.'.'.$id;
         }
 
-        preg_match_all('/(?:\\\\.|[^.])+/s', $id, $matches);
-        $parts = preg_replace('/\\\\(.)/s', '$1', $matches[0]);
+        // Split the ID into chunks separated by dots,
+        // escaped dots (\.) and backslashes (\\) are allowed.
+        preg_match_all('/(?:\\\\[.\\\\]|[^.])++/s', $id, $matches);
+        $parts = preg_replace('/\\\\([.\\\\])/s', '$1', $matches[0]);
 
         $item = &$GLOBALS['TL_LANG'];
 
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -11,6 +11,7 @@
 namespace Contao\CoreBundle\Translation;
 
 use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Contao\System;
 use Symfony\Component\Translation\TranslatorInterface;
 
 /**
@@ -127,8 +128,8 @@ class Translator implements TranslatorInterface
      */
     private function loadLanguageFile(string $name)
     {
-        /** @var \Contao\System */
-        $system = $this->framework->getAdapter('System');
+        /** @var System */
+        $system = $this->framework->getAdapter(System::class);
 
         $system->loadLanguageFile($name);
     }
--- a/src/Translation/Translator.php
+++ b/src/Translation/Translator.php
@@ -1,0 +1,132 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2017 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\Translation;
+
+use Contao\CoreBundle\Framework\FrameworkAwareInterface;
+use Contao\CoreBundle\Framework\FrameworkAwareTrait;
+use Symfony\Component\Translation\TranslatorInterface;
+
+/**
+ * Translator.
+ *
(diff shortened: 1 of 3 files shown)


leofeyer:
I like a lot!

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function trans($id, array $parameters = [], $domain = null, $locale = null)
+    {
+        if (null === $domain || strncmp($domain, 'contao_', 7) !== 0) {
+            return $this->translator->trans($id, $parameters, $domain, $locale);
+        }
+
+        if (!$this->framework->isInitialized()) {
```
aschempp: Should we initialize the framework here?

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+     * @return string|null
+     */
+    private function getFromGlobals(string $id, string $domain)
+    {
+        if ('default' !== $domain) {
+            $id = $domain.'.'.$id;
+        }
+
+        preg_match_all('/(?:\\\\.|[^.])+/s', $id, $matches);
+        $parts = preg_replace('/\\\\(.)/s', '$1', $matches[0]);
```
aschempp: According to my Regex tool, this should be like the following. Not sure what `|[^.]` is for?

```
preg_match_all('/(?:\\\.|[^.])+/s', $id, $matches);
$parts = preg_replace('/\\(\.)/s', '$1', $matches[0]);
```

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+
+        $item = &$GLOBALS['TL_LANG'];
+
+        foreach ($parts as $part) {
+            if (!isset($item[$part])) {
+                return null;
+            }
+            $item = &$item[$part];
```
aschempp: If I'm right, this will overwrite the referenced array? So labels would be lost?

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Load a Contao framework language file.
+     *
+     * @param string $name
+     */
+    private function loadLanguageFile(string $name)
+    {
+        /** @var \Contao\System */
+        $system = $this->framework->getAdapter('System');
```
aschempp: You must always use the FQCN, or just use `System::class` and import the class from `Contao\System`.

src/Translation/Translator.php
ausi: Yes. Done.

src/Translation/Translator.php
ausi: Done.

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+
+        // Split the ID into chunks separated by dots,
+        // escaped dots (\.) and backslashes (\\) are allowed.
+        preg_match_all('/(?:\\\\[.\\\\]|[^.])++/s', $id, $matches);
+        $parts = preg_replace('/\\\\([.\\\\])/s', '$1', $matches[0]);
+
+        $item = &$GLOBALS['TL_LANG'];
+
+        foreach ($parts as $part) {
+            if (!isset($item[$part])) {
```
aschempp: what if `$item` is not an array here? Will we get an error message trying to access a key of a scalar variable?

ausi: No, `isset()` doesn’t generate a warning for such cases. Tested in aea72e343519394ea70116b36fd9f39bba3cfbc1.

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
+<?php
```
DanielSchwiperich: since this PR is targeted for 4.5.0 (same as #1077) and is already using php 7 features like the `string` parameter type hints it should fully make use of it. 

right here `declare strict` declaration is missing

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+    public function __construct(TranslatorInterface $translator, ContaoFrameworkInterface $framework)
+    {
+        $this->translator = $translator;
+        $this->framework = $framework;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function trans($id, array $parameters = [], $domain = null, $locale = null)
```
DanielSchwiperich: the symfony interface is php5, it is ok to use parameter hints in the implementing method, also return type declaration is missing

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Get the translation from the $GLOBALS['TL_LANG'] array.
+     *
+     * @param string $id     Message id, e.g. "MSC.view"
+     * @param string $domain Message domain, e.g. "messages" or "tl_content"
+     *
+     * @return string|null
+     */
+    private function getFromGlobals(string $id, string $domain)
```
DanielSchwiperich: I saw leo using nullable return type, so I assume we have php 7.1 features available and we can set a return type `: ?string`

src/Translation/Translator.php
DanielSchwiperich: same for following methods

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Contao\System;
+use Symfony\Component\Translation\TranslatorInterface;
+
+/**
+ * Translator.
+ *
+ * @author Martin Auswöger <martin@auswoeger.com>
+ */
+class Translator implements TranslatorInterface
```
DanielSchwiperich: Since this Translator is only for usage with  `$Globals['TL_LANG']` and hopefully replaced someday by a solution which doesn't need to use a superglobal I would recommend a name like `GlobalsTranslator` or so

src/Translation/Translator.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    public function transChoice($id, $number, array $parameters = [], $domain = null, $locale = null)
+    {
+        return $this->translator->transChoice($id, $number, $parameters, $domain, $locale);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setLocale($locale)
```
DanielSchwiperich: I'd recommend to give info in the doc block that this method and the following do nothing in regard of contao.

src/Translation/Translator.php
ausi: No, parameter hints are not allowed for implemented methods in PHP 7.1, only in 7.2

ausi:
@DanielSchwiperich Adding PHP 7 features is not part of this pull request and should be done in #1077 IMO

src/Translation/Translator.php
ausi: Where this translator gets the messages from may change in the future. In the optimal case this class gets removed in the future and the default translater service can be used.

DanielSchwiperich:
ah ok in that case you would need to remove the `string` parameter type hints because they are only allowed from php 7 onwards. And then ignore most of my other comments 😄

src/Translation/Translator.php
DanielSchwiperich: you are absolutely right on this. Thought I had done this successfully in a project but was wrong

src/Translation/Translator.php
ausi: Done in bbf0b949dfb753d7507e1a4c4720d71a62ba057f.

aschempp:
Just found the `LoggingTranslatorPass` class. Does that affect us?

ausi:
The logging still works for the Symfony translator, Contao translations don’t get logged.

leofeyer:
Thank you @ausi.

DanielSchwiperich:
@leofeyer you have now merged code which requires php7, you know that?

Toflar:
The develop branch is php 7.1 only.

DanielSchwiperich:
then I don't understand why this PR was not made using all the php 7.1 features (see my comments). 
Either do it php 7(.1) way or not at all. imho

ausi:
@DanielSchwiperich the PHP 7.1 features for this PR got added in https://github.com/contao/core-bundle/pull/1077/files#diff-144b2e2beae5529dc5a75f36c717cbb7

PR #1079, opened 2017-09-13
[RFC] Extend \Image::getPath($src) method to return public vendor images

agoat:
This PR adds the ability to easily insert images from your vendor public folder, just like the contao theme images. This might be interesting for developers.

You will be able to insert a image from your vendors public folder into your html like so:
```php
$strOutputHTML .= \Image::getHtml('myimage.svg', 'my image', '');
```
or just fetch the path
```php
$imgUrl = \Image::getPath('myimage.svg');
```
It will return the relative path to the image (e.g. `bundles/agoatspecialbundle/myimage.svg`).

That's maybe much simpler as:
```php
\Image::getHTML('bundles/whattheheckismybundlessymlinkedname/myimage.svg');
```

The only drawback with this is, that only the first image path found is returned if the same image name is used in several bundles.

What do you think?

### DIFF
--- a/src/Resources/contao/library/Contao/Image.php
+++ b/src/Resources/contao/library/Contao/Image.php
@@ -677,7 +677,7 @@ class Image
 	
 			if (count($vendorAssets) > 0)
 			{
-				return StringUtil::stripRootDir($vendorAssets[0]);
+				return \StringUtil::stripRootDir($vendorAssets[0]);
 			}
 		}
 	}
--- a/src/Resources/contao/library/Contao/Image.php
+++ b/src/Resources/contao/library/Contao/Image.php
@@ -660,20 +660,25 @@ class Image
 		{
 			$theme = \Backend::getTheme();
 
-			if (pathinfo($src, PATHINFO_EXTENSION) == 'svg')
-			{
-				return 'system/themes/' . $theme . '/icons/' . $src;
-			}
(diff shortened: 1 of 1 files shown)


ausi:
I don’t think it’s a good idea to extend the features of the old classes. We should use a resource finder/locator service instead I think.

leofeyer:
It would probably be ok to modify the old class, but why would be want an additional and expensive `glob()` call that is made upon every method call?

agoat:
> I don’t think it’s a good idea to extend the features of the old classes. We should use a resource finder/locator service instead I think.

That sounds good. However, I must confess that I don't yet understand the whole possibilities of the symfony framework. That's probably a task for you guys. I look and learn how contao is doing it. So you are my teachers ;-)

> It would probably be ok to modify the old class, but why would be want an additional and expensive glob() call that is made upon every method call?

I already thought about the performance of the glob task. Maybe it's better to catch the public folders and go through them, but I don't know if this would be faster as a pure search on the disk system. And the glob() command is only called when nothing else could be found. 
I couldn't find a way the get informations of the bundles public assets folders from symfony directly.


This is more a quick approuch for a solution to show the idea.

agoat:
Ok. I'm not sad about this. Was just an idea. 
There are many things that are more important.

PR #1087, opened 2017-09-19
[RTM] Minimum password length enhancement

daniel-nemeth:
Check the minimum password length first from DCA 'eval' value, if it does not exists use the value from Config as fallback.

### DIFF
--- a/src/Resources/contao/forms/FormPassword.php
+++ b/src/Resources/contao/forms/FormPassword.php
@@ -125,7 +125,7 @@ class FormPassword extends \Widget
 		}
 
 		// Check password length either from DCA or use Config as fallback (#1087)
-		$intLength = ($this->minlength > 0) ? $this->minlength : \Config::get('minPasswordLength');
+		$intLength = $this->minlength ?: \Config::get('minPasswordLength');
 
 		if (Utf8::strlen($varInput) < $intLength)
 		{
--- a/src/Resources/contao/forms/FormPassword.php
+++ b/src/Resources/contao/forms/FormPassword.php
@@ -124,9 +124,12 @@ class FormPassword extends \Widget
 			return '';
 		}
 
-		if (Utf8::strlen($varInput) < \Config::get('minPasswordLength'))
+		// Check password length either from DCA or use Config as fallback (#1087)
+		$intLength = ($this->minlength > 0) ? $this->minlength : \Config::get('minPasswordLength');
+
+		if (Utf8::strlen($varInput) < $intLength)
 		{
-			$this->addError(sprintf($GLOBALS['TL_LANG']['ERR']['passwordLength'], \Config::get('minPasswordLength')));
+			$this->addError(sprintf($GLOBALS['TL_LANG']['ERR']['passwordLength'], $intLength));
 		}
 
 		if ($varInput != $this->getPost($this->strName . '_confirm'))
--- a/src/Resources/contao/widgets/Password.php
+++ b/src/Resources/contao/widgets/Password.php
@@ -114,9 +114,12 @@ class Password extends \Widget
 			return '*****';
 		}
 
-		if (Utf8::strlen($varInput) < \Config::get('minPasswordLength'))
+		// Check password length either from DCA or use Config as fallback (#1086)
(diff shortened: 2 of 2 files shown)


src/Resources/contao/widgets/Password.php
```diff
@@ -114,9 +114,12 @@ protected function validator($varInput)
 			return '*****';
 		}
 
-		if (Utf8::strlen($varInput) < \Config::get('minPasswordLength'))
+		// Check password length either from DCA or use Config as fallback (#1086)
+		$intLength = ($this->arrConfiguration['minlength'] > 0) ? $this->arrConfiguration['minlength'] : \Config::get('minPasswordLength');
```
Toflar: You can skip the `$this->arrConfiguration` array and just use `$this->minlength`, it will use the magic `__get()` method. Not saying this is better but we have to make sure it's in line with all the other widgets :) 
Otherwise this looks good to me, it's nice to see more developers trying to contribute! Excellent!

daniel-nemeth: You are right, thanks!

Toflar:
Any reason why `BackendPassword` should not get the same feature? :)

daniel-nemeth:
You mean FormPassword.php ?

Toflar:
Yes, of course. Sorry.

bytehead:
> Any reason why BackendPassword should not get the same feature? :)

Yes, IMHO it's not very clever to have the ability that allows you backend passwords shorter than 8 characters.

Toflar:
This PR is about using the widget independently from the system configuration (which makes sense in many ways). We're not the once to decide upon the minimum length of passwords (and it was configurable anyway so...).

bytehead:
I don't know, why `BackendPassword` should be independent from system configuration.

Toflar:
Example: I build my extension that allows to set a new password for some third party service which requires 16 characters?

bytehead:
It's only about the minimum length.

daniel-nemeth:
In the frontend and backend widget we have a DCA eval value to use for the check, but where do we define the minlength for `BackendPassword`?

aschempp:
This is a new feature so it should be Contao 4.5 (develop branch). Otherwise I'm fine with the change, but it should be in all widgets. `BackendPassword` is a widget as well, so it works exactly the same.

aschempp:
oh well… `BackendPassword` is probably the Controller? That should not be changed, both widgets are already included in this PR.

daniel-nemeth:
`BackendPassword` has currently no other configuration as the `Config`. So it's is enough to adjust both widgets I guess.

> Yes, IMHO it's not very clever to have the ability that allows you having backend passwords shorter than 8 characters.

So this discussion is then clear too.

src/Resources/contao/forms/FormPassword.php
```diff
@@ -124,9 +124,12 @@ protected function validator($varInput)
 			return '';
 		}
 
-		if (Utf8::strlen($varInput) < \Config::get('minPasswordLength'))
+		// Check password length either from DCA or use Config as fallback (#1087)
+		$intLength = ($this->minlength > 0) ? $this->minlength : \Config::get('minPasswordLength');
```
leofeyer: This can be simplified to `$intLength = $this->minlength ?: \Config::get('minPasswordLength');`

src/Resources/contao/widgets/Password.php
```diff
@@ -114,9 +114,12 @@ protected function validator($varInput)
 			return '*****';
 		}
 
-		if (Utf8::strlen($varInput) < \Config::get('minPasswordLength'))
+		// Check password length either from DCA or use Config as fallback (#1086)
+		$intLength = ($this->minlength > 0) ? $this->minlength : \Config::get('minPasswordLength');
```
leofeyer: This can be simplified to `$intLength = $this->minlength ?: \Config::get('minPasswordLength');`

leofeyer:
Thank you @daniel-nemeth.

daniel-nemeth:
You‘re welcome, and sorry, I didn‘t noticed the request for simplifing the password check since our new mail client blocked github mails.

Thanks for merging!

PR #1089, opened 2017-09-19
[RFC] Optimize and extend the treeView in DC_Table

agoat:
During development on an extended table driver for an extension to have a treeview that is sortable and has pagination (btw. both are technically possible, but in the end do not fit usablility, at least with pagination users lost orientation in the tree) i found some optimitation for the treeview.

The official demo page generates with a fully expanded treeview a significant amount of SQL Queries:

View | SQL count
-- | --
Page tree | 172
Article tree | 415

With the optimized generateTree method the SQL Query count will drop down to:

View | SQL count
-- | --
Page tree | 136
Article tree | 339

While debuging my SQL Queries I found a lot of the same query to the tl_page table called from the button callbacks in the tl_article dca. When using the PageModel instead of direct SQL, the count drops down to:

View | SQL count
-- | --
Page tree | 126
Article tree | 134

So there is general dropp from 415 to 134 queries in total (nearly 60% unnecessary SQL queries). This may be important to pages with many articles (I had one which cause over 1200 SQL Queries opening the article tree)

I opened an extra PR (#1090) for the button callbacks problem!!

This PR handles only the logic of the generateTree method and add some features (arguable!!):

- Don't count the child records and when called recursively request the complete child records again
- Have an expand buttons (plus/minus) down to the last page (better overview with finer display options)
- Adds a feature to group by an attribute (e.g. in the article view group articles by 'inColumn')
- Adds a filter for the parent table (e.g. show only regular pages)
- Switched the order of child pages and articles (pages first - like folders and files)

### Grouping
Grouping can be enabled by adding the `group` attribute to the dca with the following lines:
```php
@@ dca/tl_article.php
// List
	'list' => array
	(
		'sorting' => array
		(
			'mode'                    => 6,
			'fields'                  => array('published DESC', 'title', 'author'),
			'paste_button_callback'   => array('tl_article', 'pasteArticle'),
			'panelLayout'             => 'filter;search',
			'group'			  => 'inColumn'
		),
		'label' => array
		(
			'fields'                  => array('title'),
			'label_callback'          => array('tl_article', 'addIcon')
		),
```
This will look like this:
article1

Maybe there should be a better group icon.

### Parent table filter
The parent table filter will be configured by the `ptable` attribute in the list sorting array:
```php
@@ dca/tl_article.php
	// List
	'list' => array
	(
		'sorting' => array
		(
			'mode'                    => 6,
			'fields'                  => array('published DESC', 'title', 'author'),
			'paste_button_callback'   => array('tl_article', 'pasteArticle'),
			'panelLayout'             => 'filter;search',
			'group'			  => 'inColumn',
			'pfilter'		  => array('type=?', 'regular')
		),
```
And will result like this:
article2


src/Resources/contao/drivers/DC_Table.php
```diff
@@ -3433,11 +3433,11 @@ protected function treeView()
 			if (!is_array($session[$node]) || empty($session[$node]) || current($session[$node]) != 1)
 			{
 				$session[$node] = array();
-				$objNodes = $this->Database->execute("SELECT DISTINCT pid FROM " . $table . " WHERE pid>0");
+				$objNodes = $this->Database->execute("SELECT DISTINCT id FROM " . $table);
```
leofeyer: This cannot be right. The original statement says "select all PIDs greater 0" which is **not the same** as "select all IDs". Consider this:

```
------------
| id | pid |
------------
|  1 |   0 |
|  2 |   1 |
------------
```

The original query would return `1` and your query would return `1, 2`.

agoat: I changed the node toggle method, too. So in this context it's right.

In the old way, only pages with subpages could be toggled. The child table rows (e.g. articles) of a page are always visible. 
With this PR I changed it and all pages with subpages or child table rows can be toggled. 
That's why I need all page ids here.

leofeyer:
@contao/developers What about this PR?

Toflar:
Guess it's gonna die if the work isn't picked up. @agoat, wanna give it another shot on our monorepo at https://github.com/contao/contao/ for maybe 4.10?

agoat:
Ok, I will try making a new PR in the monorepo. But have to check this first on the latest contao version.

Last two years I was very busy with a big cloud project and all my extensions were hardly developed any further. First I guess I have to catch up with the latest contao version ;-) So many improvements during the last years..

aschempp:
Honestly, I think it will be impossible to review the changes. Is there any use case for the feature in the Contao core?

PR #1090, opened 2017-09-19
[RFC] Use the PageModel instead of direct SQL queries in button callbacks

agoat:
This PR handles the optimisation of button callbacks mentioned in #1089.

Is there a reason why all the button callbacks don't use Models??
Because if we will do, the amount of SQL Queries will drop from 415 to around 210 in the article view of the official contao demo page. 

Currently every rendered article row will call the buttons callbacks which itself will make the same SQL Query to the tl_page table for every page (around 6 queries per article). 

When the `\PageModel` is used, the Result is cached.

### DIFF
--- a/src/Resources/contao/dca/tl_article.php
+++ b/src/Resources/contao/dca/tl_article.php
@@ -720,9 +720,7 @@ class tl_article extends Backend
 	 */
 	public function editArticle($row, $href, $label, $title, $icon, $attributes)
 	{
-		$objPage = $this->Database->prepare("SELECT * FROM tl_page WHERE id=?")
-								  ->limit(1)
-								  ->execute($row['pid']);
+		$objPage = \PageModel::findById($row['pid']);
 
 		return $this->User->isAllowed(BackendUser::CAN_EDIT_ARTICLES, $objPage->row()) ? '<a href="'.$this->addToUrl($href.'&amp;id='.$row['id']).'" title="'.StringUtil::specialchars($title).'"'.$attributes.'>'.Image::getHtml($icon, $label).'</a> ' : Image::getHtml(preg_replace('/\.svg$/i', '_.svg', $icon)).' ';
 	}
@@ -747,9 +745,7 @@ class tl_article extends Backend
 			return Image::getHtml(preg_replace('/\.svg$/i', '_.svg', $icon)).' ';
 		}
 
-		$objPage = $this->Database->prepare("SELECT * FROM tl_page WHERE id=?")
-								  ->limit(1)
-								  ->execute($row['pid']);
+		$objPage = \PageModel::findById($row['pid']);
 
 		return $this->User->isAllowed(BackendUser::CAN_EDIT_ARTICLES, $objPage->row()) ? '<a href="'.$this->addToUrl($href.'&amp;id='.$row['id']).'" title="'.StringUtil::specialchars($title).'"'.$attributes.'>'.Image::getHtml($icon, $label).'</a> ' : Image::getHtml(preg_replace('/\.svg$/i', '_.svg', $icon)).' ';
 	}
@@ -775,9 +771,7 @@ class tl_article extends Backend
 			return '';
 		}
 
-		$objPage = $this->Database->prepare("SELECT * FROM tl_page WHERE id=?")
-								  ->limit(1)
-								  ->execute($row['pid']);
+		$objPage = \PageModel::findById($row['pid']);
 
 		return $this->User->isAllowed(BackendUser::CAN_EDIT_ARTICLE_HIERARCHY, $objPage->row()) ? '<a href="'.$this->addToUrl($href.'&amp;id='.$row['id']).'" title="'.StringUtil::specialchars($title).'"'.$attributes.'>'.Image::getHtml($icon, $label).'</a> ' : Image::getHtml(preg_replace('/\.svg$/i', '_.svg', $icon)).' ';
 	}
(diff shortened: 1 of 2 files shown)


Toflar:
There is no reason and this is an excellent PR, thank you very much!

src/Resources/contao/dca/tl_page.php
```diff
@@ -1433,11 +1433,9 @@ public function copyPageWithSubpages($row, $href, $label, $title, $icon, $attrib
 			return '';
 		}
 
-		$objSubpages = $this->Database->prepare("SELECT * FROM tl_page WHERE pid=?")
-									  ->limit(1)
-									  ->execute($row['id']);
+		$objSubpages = \PageModel::findByPid($row['id'], array('limit'=>1));
```
aschempp: You could actually use `countBy` here, because we only need to know if there are subpages.

agoat: But then it's not read from the model registry and an extra sql query is executed (see [Model.php](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/library/Contao/Model.php#L1158)).

aschempp: Well it does not read from the registry this way either... `findByPid` is not cacheable, and nothing is cacheable if you use options.

agoat: You're totally right. It's only read from registry if a model is requested by it's id.

But one question is left: Is there a (measurable) difference in performance, when setting a limit of one or query all child records and count them? 

I don't think so. The amount of data from the sql server is negligible unless someone have a thousand sub pages. So I'm fine with using `countBy`. Looks more logic.

agoat: Changed to the `countByPid` method in b89e33070b3e137bdbfcc1929050b2bd27f5a3b5

src/Resources/contao/dca/tl_article.php
```diff
@@ -720,9 +720,7 @@ public function getArticleTemplates()
 	 */
 	public function editArticle($row, $href, $label, $title, $icon, $attributes)
 	{
-		$objPage = $this->Database->prepare("SELECT * FROM tl_page WHERE id=?")
-								  ->limit(1)
-								  ->execute($row['pid']);
+		$objPage = \PageModel::findById($row['pid']);
```
leofeyer: The namespace prefix is not necessary here.

src/Resources/contao/dca/tl_page.php
```diff
@@ -1433,11 +1433,9 @@ public function copyPageWithSubpages($row, $href, $label, $title, $icon, $attrib
 			return '';
 		}
 
-		$objSubpages = $this->Database->prepare("SELECT * FROM tl_page WHERE pid=?")
-									  ->limit(1)
-									  ->execute($row['id']);
+		$subPages = \PageModel::countByPid($row['id']);
 
-		return ($objSubpages->numRows && $this->User->hasAccess($row['type'], 'alpty') && $this->User->isAllowed(BackendUser::CAN_EDIT_PAGE_HIERARCHY, $row)) ? '<a href="'.$this->addToUrl($href.'&amp;id='.$row['id']).'" title="'.StringUtil::specialchars($title).'"'.$attributes.'>'.Image::getHtml($icon, $label).'</a> ' : Image::getHtml(preg_replace('/\.svg$/i', '_.svg', $icon)).' ';
+		return ($subPages && $this->User->hasAccess($row['type'], 'alpty') && $this->User->isAllowed(BackendUser::CAN_EDIT_PAGE_HIERARCHY, $row)) ? '<a href="'.$this->addToUrl($href.'&amp;id='.$row['id']).'" title="'.StringUtil::specialchars($title).'"'.$attributes.'>'.Image::getHtml($icon, $label).'</a> ' : Image::getHtml(preg_replace('/\.svg$/i', '_.svg', $icon)).' ';
```
leofeyer: This should probably be `$subPages !== null`.

src/Resources/contao/dca/tl_page.php
```diff
@@ -1485,9 +1483,7 @@ public function pastePage(DataContainer $dc, $row, $table, $cr, $arrClipboard=nu
 		// Prevent adding non-root pages on top-level
 		if (Input::get('mode') != 'create' && $row['pid'] == 0)
 		{
-			$objPage = $this->Database->prepare("SELECT * FROM " . $table . " WHERE id=?")
-									  ->limit(1)
-									  ->execute(Input::get('id'));
+			$objPage = \PageModel::findById(Input::get('id'));
```
leofeyer: This is wrong. You have to use the model that corresponds to `$table`.

src/Resources/contao/dca/tl_page.php
```diff
@@ -1512,12 +1508,10 @@ public function pastePage(DataContainer $dc, $row, $table, $cr, $arrClipboard=nu
 				}
 			}
 
-			$objPage = $this->Database->prepare("SELECT * FROM " . $table . " WHERE id=?")
-									  ->limit(1)
-									  ->execute($row['pid']);
+			$objPage = \PageModel::findById($row['pid']);
```
leofeyer: Same here.

src/Resources/contao/dca/tl_page.php
```diff
@@ -1433,11 +1433,9 @@ public function copyPageWithSubpages($row, $href, $label, $title, $icon, $attrib
 			return '';
 		}
 
-		$objSubpages = $this->Database->prepare("SELECT * FROM tl_page WHERE pid=?")
-									  ->limit(1)
-									  ->execute($row['id']);
+		$subPages = \PageModel::countByPid($row['id']);
```
leofeyer: This has no effect. If you want to reuse the existing model, you have to use `findByPid()` here.

leofeyer:
Implemented in 57e2b6f3cde10cc04bb78d6ebf3923b60ee8f0dd. Thank you @agoat for analyzing this and pointing me in the right direction.

agoat:
You're welcome. 
Sorry, I didn't have time to look at this PR again. However, you know the core code better anyway and have found even more optimizations. Perfect.

PR #1094, opened 2017-09-20
[RTM] Allow to register hooks listeners as tagged services

dmolineus:
Since https://github.com/contao/core-bundle/pull/376 it's possible to use services for hooks. This is a great step to using services within Contao.

This PR completes this step by allowing to register hooks as services without using the `config.php`. You don't have to manage your hook registration at two places anymore.

```yaml

services:
  my.custom.hook_listener:
    class: 'My\Custom\Hook\Listener'
    tags:
      - { name: 'contao.hook', hook: 'parseTemplate', method: 'onParseTemplate'}
      - { name: 'contao.hook', hook: 'initializeSystem', method: 'onInitializeSystem', before: true}
```

I'm aware of the above mentioned PR discussed it before and this feature wasn't added because of lack of control the position of the hook.

To gain more control of the position it's possible to set the `before` attribute so that the hook is registered before the legacy hooks. If you really need more control, well, just use the globals.

To keep backward compatibility the registered hooks are added to the `$GLOBALS['TL_HOOKS']` array.

### DIFF
--- a/src/DependencyInjection/Compiler/RegisterHooksPass.php
+++ b/src/DependencyInjection/Compiler/RegisterHooksPass.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2017 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Find hook services and store them in an parameter.
+ *
+ * @author David Molineus <https://github.com/dmolineus>
+ */
+class RegisterHooksPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $serviceIds = $container->findTaggedServiceIds('contao.hook');
+        $before     = [];
+        $after      = [];
+
+        foreach ($serviceIds as $serviceId => $tags) {
+            foreach ($tags as $attributes) {
+                $this->guardRequiredAttributesExist($serviceId, $attributes);
+
+                if (!empty($attributes['before'])) {
+                    if (!isset($before[$attributes['hook']])) {
+                        $before[$attributes['hook']] = [];
+                    }
+
+                    $before[$attributes['hook']][] = [$serviceId, $attributes['method']];
+                } else {
+                    if (!isset($after[$attributes['hook']])) {
+                        $after[$attributes['hook']] = [];
+                    }
+
+                    $after[$attributes['hook']][] = [$serviceId, $attributes['method']];
+                }
+            }
+        }
+
+        if ($before) {
+            $container->setParameter('contao.hook_listeners.before', $before);
+        }
+
+        if ($after) {
+            $container->setParameter('contao.hook_listeners.after', $after);
+        }
+    }
+
+    /**
+     * Guard that required attributes (hook and method) are defined.
+     *
+     * @param string $serviceId  Service id.
+     * @param array  $attributes Tag attributes.
+     *
+     * @throws InvalidConfigurationException When an attribute is missing.
+     */
+    private function guardRequiredAttributesExist(string $serviceId, array $attributes): void
+    {
+        if (!isset($attributes['hook'])) {
+            throw new InvalidConfigurationException(
+                sprintf('Missing hook attribute in tagged hook service with service id "%s"', $serviceId)
+            );
+        }
+
+        if (!isset($attributes['method'])) {
+            throw new InvalidConfigurationException(
+                sprintf('Missing method attribute in tagged hook service with service id "%s"', $serviceId)
+            );
+        }
+    }
+}
--- a/src/DependencyInjection/Compiler/RegisterHooksPass.php
+++ b/src/DependencyInjection/Compiler/RegisterHooksPass.php
@@ -28,6 +28,10 @@ class RegisterHooksPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
+        if (!$container->hasDefinition('contao.framework')) {
+            return;
+        }
+
         $serviceIds = $container->findTaggedServiceIds('contao.hook');
         $hooks      = [];
 
@@ -48,7 +52,8 @@ class RegisterHooksPass implements CompilerPassInterface
                 krsort($hooks[$hook]);
             }
 
-            $container->setParameter('contao.hook_listeners', $hooks);
+            $definition = $container->getDefinition('contao.framework');
+            $definition->setArgument(6, $hooks);
         }
     }
 
--- a/src/DependencyInjection/Compiler/RegisterHooksPass.php
+++ b/src/DependencyInjection/Compiler/RegisterHooksPass.php
@@ -16,40 +16,34 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
-/**
- * Find hook services and store them in an parameter.
- *
- * @author David Molineus <https://github.com/dmolineus>
- */
 class RegisterHooksPass implements CompilerPassInterface
 {
     /**
      * {@inheritdoc}
      */
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         if (!$container->hasDefinition('contao.framework')) {
             return;
         }
 
         $serviceIds = $container->findTaggedServiceIds('contao.hook');
-        $hooks      = [];
+        $hooks = [];
 
         foreach ($serviceIds as $serviceId => $tags) {
             foreach ($tags as $attributes) {
-                $this->guardRequiredAttributesExist($serviceId, $attributes);
+                $this->checkRequiredAttributes($serviceId, $attributes);
 
                 $priority = (int) ($attributes['priority'] ?? 0);
-                $hook     = $attributes['hook'];
+                $hook = $attributes['hook'];
 
                 $hooks[$hook][$priority][] = [$serviceId, $attributes['method']];
             }
         }
 
-        if (count($hooks) > 0) {
-            // Apply priority sorting.
+        if (\count($hooks) > 0) {
             foreach (array_keys($hooks) as $hook) {
-                krsort($hooks[$hook]);
+                krsort($hooks[$hook]); // order by priority
             }
 
             $definition = $container->getDefinition('contao.framework');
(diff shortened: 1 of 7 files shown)


aschempp:
I very much like this idea! Even thought we agreed not to "improve" the "old" stuff, this will be helpful a lot. However, I would suggest two changes:

 1. Use priorities for the hooks. We can simply take the existing hook array keys as their internal priority (reversed). This way we will follow Symfony standards.
 2. The setup handling should not be done in the framework class, but using an event listener on the `initializeSystem` hook (which obviously needs to be registered using the old style ^^). Unfortunately, that would mean `initializeSystem` hooks are not taggable, but I can't see any other good solution.

leofeyer:
> Use priorities for the hooks.

+1

> The setup handling should not be done in the framework class

Since it is part of initializing the Contao 3 framework, it would be ok for me (not ideal of course).

dmolineus:
> Use priorities for the hooks. We can simply take the existing hook array keys as their internal priority (reversed). This way we will follow Symfony standards.

I thought about it, but wasn't sure how to handle the legacy hooks. To confirm if I got it right:

Given are registered hooks in `$GLOBALS['TL_HOOKS']` and I want to add my hook at a specific position:

```php
<?php
$GLOBALS['TL_HOOKS']['parseTemplate'] = [
  ['Foo', 'onParseTemplate'],
  ['Bar', 'onParseTemplate'],
  ['Baz', 'onParseTemplate'],
];
```
So your suggest is, that the `Foo` listener is handled as priority `2`, `Bar` as priority `1` and `Baz` ans priority `0`. To if I want to be at first position I have to use priority `4`. Did I understand you correctly?

What I dislike here that you never know how many hooks are registered. Symfony standard doesn't depend on the number of listeners as well. I suggest handle all legacy hooks as priority `0` instead.

Toflar:
Cool, I like the idea, well done @dmolineus!

aschempp:
I would agree with the priority 0, but I see issues here. The legacy order is important, because extensions are loaded after each other and might depend on it. So we need to make sure the legacy order stays the same. The most simple solution will be to `array_reverse` the old hooks and then use the keys as priorities.

aschempp:
Regarding initialization in the framework class, let's just make sure we agree on this once and for all, because @Toflar used the hook in the new *module as controller* pull request to merge legacy with new stuff. Everything should go into the framework class then, and that'll make it huge…

dmolineus:
> I would agree with the priority 0, but I see issues here. The legacy order is important, because extensions are loaded after each other and might depend on it. So we need to make sure the legacy order stays the same. The most simple solution will be to array_reverse the old hooks and then use the keys as priorities.

I would solve it this way:
 - Priority > 0 will be added before the legacy hooks depending on the priority
 - Now the legacy hooks are added in the ordered they are added
 - Priority = 0 comes after the legacy hooks
 - Priority < 0 comes after that

Another solution could be to to load the legacy hooks and hooks with priority null for each bundle at the same time. So the order would be:
 - Priority > 0 will be added before the legacy hooks depending on the priority
 - Bundle A legacy hooks
 - Bundle A hook listeners with priority 0
 - Bundle B legacy hooks
 - Bundle B hook listeners with priority 0
 - ...
 - Priority < 0 comes after that

dmolineus:
> Regarding initialization in the framework class, let's just make sure we agree on this once and for all, because @Toflar used the hook in the new module as controller pull request to merge legacy with new stuff. Everything should go into the framework class then, and that'll make it huge…

Maybe adding an event which is triggered before the initialization hook would allow to handle such tasks as the module as controller feature or the hook listener registration.

Toflar:
> let's just make sure we agree on this once and for all

Agree on what?

leofeyer:
> Regarding initialization in the framework class, let's just make sure we agree on this once and for all, because @Toflar used the hook in the new *module as controller* pull request to merge legacy with new stuff. Everything should go into the framework class then, and that'll make it huge…

He means "agree on not putting everything into the framework class and use listeners/hooks instead". Although this is a special case IMO, because of

> Unfortunately, that would mean `initializeSystem` hooks are not taggable, but I can't see any other good solution.

aschempp:
I would still prefer to call an explicit event listener manually instead of having everything in the framework class…

Regarding the priority 0, I'm not sure if the order of hooks inside one priority is kept when sorting them by priority?

dmolineus:
I reworked the hook registration using the priority attribute. I've chosen this strategy.

Priority > 0 will be added before the legacy hooks depending on the priority
Now the legacy hooks are added in the ordered they are added
Priority = 0 comes after the legacy hooks
Priority < 0 comes after that

Hooks with same priority will keep their internal order.

src/DependencyInjection/Compiler/RegisterHooksPass.php
```diff
@@ -0,0 +1,75 @@
...
+    public function process(ContainerBuilder $container)
+    {
+        $serviceIds = $container->findTaggedServiceIds('contao.hook');
+        $hooks      = [];
+
+        foreach ($serviceIds as $serviceId => $tags) {
+            foreach ($tags as $attributes) {
+                $this->guardRequiredAttributesExist($serviceId, $attributes);
+
+                $priority = $attributes['priority'] ?? 0;
```
aschempp: You should probably int-cast this for safety reasons.

src/DependencyInjection/Compiler/RegisterHooksPass.php
```diff
@@ -0,0 +1,75 @@
...
+                $hook     = $attributes['hook'];
+
+                $hooks[$hook][$priority][] = [$serviceId, $attributes['method']];
+            }
+        }
+
+        if (count($hooks) > 0) {
+            // Apply priority sorting.
+            krsort($hooks);
```
aschempp: are you sure this works? Because it sorts on the key of the hook name, not the priority? Also, you might want to use the `\SplPriorityQueue` which is exactly meant for this.

src/DependencyInjection/Compiler/RegisterHooksPass.php
```diff
@@ -0,0 +1,75 @@
...
+
+            $container->setParameter('contao.hook_listeners', $hooks);
```
aschempp: The hooks should not be registered as a parameter that is *always available*, but we should use dependency injection for it. That means something like `$container->getDefinition('contao.framework')->setArgument(6, $hooks)`.

src/Framework/ContaoFramework.php
```diff
@@ -476,4 +477,34 @@ private function canSkipTokenCheck(): bool
...
+     */
+    private function buildHookGlobals(): void
+    {
+        if ($this->container->hasParameter('contao.hook_listeners')) {
+            $config = (array) $this->container->getParameter('contao.hook_listeners');
+
+            foreach ($config as $hookName => $priorities) {
+                $hooks = [];
+
+                foreach ($priorities as $priority => $listeners) {
```
aschempp: I think this can be simplified to something like this:

```php
if (isset($GLOBALS['TL_HOOKS'][$hookName])) {
    if (isset($priorities[0]) {
        $priorities[0] = array_merge($GLOBALS['TL_HOOKS'][$hookName], $priorities[0]);
    } else {
        $priorities[0] = $GLOBALS['TL_HOOKS'][$hookName];
    }
}

$GLOBALS['TL_HOOKS'][$hookName] = array_merge_recursive($priorities);
```

src/DependencyInjection/Compiler/RegisterHooksPass.php
dmolineus: Done in ec49f63.

src/DependencyInjection/Compiler/RegisterHooksPass.php
dmolineus: Actually I was because I craeted unit tests for it. But I had a closer look. It didn't work. `assertEquals` does not detect order of keys which I thought it would (and which I was used to do by PHPSpec which I actually prefer).

Fixed it in 802350f. Just used simple arrays as Symfony does the same in the event dispatcher.

src/DependencyInjection/Compiler/RegisterHooksPass.php
dmolineus: Implemented in 728799e.

src/Framework/ContaoFramework.php
dmolineus: No, because array_merge_recursive would merge like this:

```php
$arrayA = ['foo','bar'];
$arrayB = ['baz'];

// would create [['foo', 'baz'], 'bar]
$merged = array_merge_recursive($arrayA, $arrayB);
```

Toflar:
I would have a use case just now again :) Looking forward to this!

aschempp:
I just looked into this again. Came up with the idea whether it might be clever to actually use a Symfony EventDispatcher for hooks? Meaning that for every hook service in the container, we register a closure that calls the eventDispatcher with a [GenericEvent](http://api.symfony.com/3.3/Symfony/Component/EventDispatcher/GenericEvent.html) ?

This way we could actually migrate from our hooks to real events. The only disadvantage would be that we can't have priorities mixed with the existing ones. Legacy hooks always come before or after the event dispatcher.

dmolineus:
> I just looked into this again. Came up with the idea whether it might be clever to actually use a Symfony EventDispatcher for hooks? Meaning that for every hook service in the container, we register a closure that calls the eventDispatcher with a GenericEvent ?

What would be the benefit of using a GenericEvent instead of named events for it as prepared in https://github.com/contao/core-bundle/pull/204? (Ingoring the initial work which has to be done to create it).

Toflar:
> The only disadvantage would be that we can't have priorities mixed with the existing ones. Legacy hooks always come before or after the event dispatcher.

That's a no go because that's exactly what I would use this PR for.

aschempp:
> What would be the benefit of using a GenericEvent instead of named events for it as prepared in #204? (Ingoring the initial work which has to be done to create it).

The difference to me is that you don't need to add event dispatcher code everywhere that hooks are already executed.

But I agree with @Toflar that priority settings is somewhat more important. PR looks RTM then to me.

src/Framework/ContaoFramework.php
```diff
@@ -472,4 +480,30 @@ private function canSkipTokenCheck(): bool
             || false === $this->request->attributes->get('_token_check')
         ;
     }
+
+    /**
+     * Build hooks globals.
+     */
+    private function buildHookGlobals(): void
+    {
+        foreach ($this->hookListeners as $hookName => $priorities) {
```
aschempp: I thought I already mentioned that somewhere, but can't find it. These lines can be much simplified like this:

```php
foreach ($this->hookListeners as $hookName => $priorities) {
    if (isset($GLOBALS['TL_HOOKS'][$hookName]) && is_array($GLOBALS['TL_HOOKS'][$hookName])) {
        if (isset($priorities[0]) {
            $priorities[0] = array_merge($GLOBALS['TL_HOOKS'][$hookName], $priorities[0]);
        } else {
            $priorities[0] = $GLOBALS['TL_HOOKS'][$hookName];
        }
    }

    $GLOBALS['TL_HOOKS'][$hookName] = array_merge_recursive($priorities);
}
```

dmolineus: You did, but it's still not true. `array_merge_recursive` with one argument will do nothing. Or do you mean `array_merge_recursive(...$priorities)`?

aschempp: Sorry, the correct call would be `call_user_func_array('array_merge', $priorities)`

dmolineus: Okay I added your suggest change in https://github.com/contao/core-bundle/pull/1094/commits/863f8137daf27824b665faa29a0ed1840b9160d9

src/DependencyInjection/Compiler/RegisterHooksPass.php
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+namespace Contao\CoreBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Find hook services and store them in an parameter.
```
aschempp: typo

Toflar:
So cool, thank you @dmolineus !

aschempp:
❤️

PR #1095, opened 2017-09-20
Fix URL encoding in XML sitemaps

ausi:
See #1085

### DIFF
--- a/src/Resources/contao/library/Contao/Automator.php
+++ b/src/Resources/contao/library/Contao/Automator.php
@@ -351,7 +351,7 @@ class Automator extends \System
 			foreach ($arrPages as $strUrl)
 			{
 				$strUrl = rawurlencode($strUrl);
-				$strUrl = str_replace(array('%2F', '%3F', '%3D', '%26', '%3A//'), array('/', '?', '=', '&', '://'), $strUrl);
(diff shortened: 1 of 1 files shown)


jankout:
I think that this concerns all special characters - e.g. with Czech special characters it does not work too.

ausi:
This pull request should fix all special characters, not only german ones. Can you give an example wich Czech special characters still cause problems?

jankout:
For example, one of the common characters: **_í_**. You can use following characters to see if it already works. 

**č, ř, š, í, á, é, ú, ů, ó, ň, ť, ď.** 

More characters for anther languages you can find [here](https://de.wikipedia.org/wiki/Liste_lateinischer_Alphabete#Von_A_bis_I_abgeleitete_Buchstaben).

ausi:
I just tested this pull request with the alias `čřšíáéúůóňťď` and it did work fine.

leofeyer:
@ausi Do we have to encode the sitemap URLs at all? If the browsers can handle `veröffentlichungen.html`, Google should be able to handle it in a `sitemap.xml` file as well, shouldn't it?

ausi:
@leofeyer the [sitemap protocol](https://www.sitemaps.org/protocol.html#escaping) defines *“all URLs ... must be URL-escaped”*.

leofeyer:
Merged in 2dbf3a3dd947ec2d655f4bc9f5c212493d9e415f.

PR #1096, opened 2017-09-20
Use the decoded URL in the search results

ausi:
See #1085

### DIFF
--- a/src/Resources/contao/templates/search/search_default.html5
+++ b/src/Resources/contao/templates/search/search_default.html5
@@ -7,6 +7,6 @@
     <p class="context"><?= $this->context ?></p>
   <?php endif; ?>
 
-  <p class="url"><?= $this->url ?><span class="filesize"> - <?= $this->filesize ?> kB</span></p>
+  <p class="url"><?= StringUtil::specialchars(urldecode($this->url), true, true) ?><span class="filesize"> - <?= $this->filesize ?> kB</span></p>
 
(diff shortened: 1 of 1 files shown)


leofeyer:
We should probably change this in https://github.com/contao/core-bundle/blob/master/src/Resources/contao/modules/ModuleSearch.php#L258-L260, don't you think?

ausi:
I don’t think so. If someone currently uses `$this->url` for the `href` attribute, it could lead to problems if we change the values in the module.

leofeyer:
I only want to change `$this->url` not `$this->href`. If someone used `$this->url` for the `href=""` attribute, they made a mistake.

ausi:
Makes sense. Should be OK then to change `$this->url` instead of modifying the template...

leofeyer:
Fixed in 89116b26de2b6dfc95629035aa2aa62050ca8d59.

ausi:
Shouldn’t we use 

```php
$objTemplate->url = \StringUtil::specialchars(urldecode($arrResult[$i]['url']), true, true)
```

?

leofeyer:
Why?

ausi:
Because the URL may include special characters (after urldecoding it) that have to be escaped for HTML.

leofeyer:
Fixed in 6e157e160a1f2c9ab2cae165c538c9fb73b0269d.

PR #1102, opened 2017-09-25
[RTM] Display form fieldsets as wrappers in the backend view.

dmolineus:
Wrapper elements provides a cleaner view of complex elements in the parent view. As the feature is not limited to CEs it would be nice to use it for form fieldsets as well.

- [x] Split fieldset into start and stop element
- [x] Update language files
- [x] Provide migration script

If this PR is accepted I'll provide the missing tasks as well.

### DIFF
--- a/src/Resources/contao/forms/FormFieldsetStop.php
+++ b/src/Resources/contao/forms/FormFieldsetStop.php
@@ -28,15 +28,6 @@ class FormFieldsetStop extends \Widget
 
 
 	/**
-	 * Do not validate
-	 */
-	public function validate()
-	{
-		return;
-	}
-
-
-	/**
 	 * Parse the template file and return it as string
 	 *
 	 * @param array $arrAttributes An optional attributes array
@@ -65,6 +56,6 @@ class FormFieldsetStop extends \Widget
 	 */
 	public function generate()
 	{
-		return "  </fieldset>\n";
+		return '</fieldset>';
 	}
 }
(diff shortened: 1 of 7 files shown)


fritzmg:
Oh yes finally 😁

dmolineus:
Migration script provided with https://github.com/contao/installation-bundle/pull/69/files

fritzmg:
This will break some extensions though, e.g. [terminal42/contao-conditionalformfields](https://github.com/terminal42/contao-conditionalformfields)

Toflar:
It's not API and I think it won't be really a problem to adjust the extension to it ;)

fritzmg:
I know, just saying ;)

src/Resources/contao/forms/FormFieldsetStop.php
```diff
@@ -0,0 +1,70 @@
...
+
+
+	/**
+	 * Generate the widget and return it as string
+	 *
+	 * @return string The widget markup
+	 */
+	public function generate()
+	{
+		return "  </fieldset>\n";
```
leofeyer: I wonder if we should throw an exception here? Ever since we've added the `Widget::parse()` method in Contao 3.3, using `Widget::generate()` was no longer recommended because it bypasses the template system. The existing widget classes still support the method to remain backwards compatibility, but this is a new class, so …

leofeyer:
Thank you @dmolineus.

PR #1104, opened 2017-09-25
[WIP] Twig for DC_Table / Refactoring

m-vo:
``DC_Table`` scares me and I'd really like to tame that 6kLOC monster somehow - otherwise I find it very hard to understand what's going on in there. However, refactoring the logic is hard as well :-) because the code is permeated by HTML everywhere. 

So here is an early attempt/draft to first refactor the HTML parts out of the class into individual Twig templates. I think this is managable and would allow further refactoring (as well as an easier way to improve the markup side). In this PR I currently only did this in a rough way for the ``show()``, ``showAll()`` and ``edit()`` methods (the latter one even not fully) as I'd like to get some feedback if this would be a valuable thing to do before putting energy into it.

Problems discovered so far: 
- button callbacks that contain HTML

### DIFF
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -291,7 +291,7 @@ class DC_Table extends \DataContainer implements \listable, \editable
 	 */
 	public function showAll()
 	{
-		$return = '';
+		$template = [];
 		$this->limit = '';
 		$this->bid = 'tl_buttons';
 
@@ -340,8 +340,8 @@ class DC_Table extends \DataContainer implements \listable, \editable
 		// Render view
 		if ($this->treeView)
 		{
-			$return .= $this->panel();
-			$return .= $this->treeView();
+            $template['panel'] = $this->panel();
+            $template['view'] = $this->treeView();
 		}
 		else
 		{
@@ -351,13 +351,13 @@ class DC_Table extends \DataContainer implements \listable, \editable
 				$this->values[] = CURRENT_ID;
 			}
 
-			$return .= $this->panel();
-			$return .= ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 4) ? $this->parentView() : $this->listView();
+            $template['panel'] = $this->panel();
+            $template['view'] = ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 4) ? $this->parentView() : $this->listView();
 
 			// Add another panel at the end of the page
 			if (strpos($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['panelLayout'], 'limit') !== false)
 			{
-				$return .= $this->paginationMenu();
+                $template['pagination'] = $this->paginationMenu();
 			}
 		}
 
@@ -366,7 +366,9 @@ class DC_Table extends \DataContainer implements \listable, \editable
 		$session['CURRENT']['IDS'] = $this->current;
 		$objSession->replace($session);
 
-		return $return;
+        return System::getContainer()
+            ->get('twig')
+            ->render('@ContaoCore/Backend/DC_Table/showAll.twig', $template);
 	}
 
 
@@ -391,8 +393,6 @@ class DC_Table extends \DataContainer implements \listable, \editable
 			return '';
 		}
 
-		$count = 1;
-		$return = '';
 		$row = $objRow->row();
 
 		// Get the order fields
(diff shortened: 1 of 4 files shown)


aschempp:
I really like the idea of refactoring, though we still are not sure if we want to keep the monster this way or just start with something from scratch… but I think we can never agree on something like that, because we can't determine the outcome. So I'm all in on refactoring.

Anyway, I don't think we should start with new "overall" templates for `edit` and `showAll` but just refactor small parts. The easiest is what you already did with the `show` template. I really like that part, also because I duplicated that template code in several other extensions (e.g. Isotope). I would suggest to create minimal PRs so they are manageable and easy to review.

Also, I suggest to use `DataContainer` instead of `DC_Table` as namespace, because most things (e.g. the show view) are the same for all drivers, not specific to DC_Table.

Let's first hear other feedbacks though 😉

m-vo:
> Also, I suggest to use DataContainer instead of DC_Table as namespace, because most things (e.g. the show view) are the same for all drivers, not specific to DC_Table.

You're right. I was already hoping that refactoring DC_Table lets the other drivers benefit automatically - that namespace was makeshift naming :-)


>Anyway, I don't think we should start with new "overall" templates for edit and showAll but just refactor small parts. 

I agree that having smaller and reusable templates should be the way to go. On the other hand this could be done very easily in a second step once you see patterns in the templates that can be merged into more general ones (including css changes). Imo, the logic can also be simplified a fair bit as soon there is no inline markup around anymore (which would probably lead to more code reusability and that again likely to template reusability). Right now the templates just pull the markup out of the respective methods.

It could get hard to do this effectivly in smaller chunks than on a per method basis. There are basically two steps necessary: 
* linear markup generation > logical markup generation 
* inline HTML > template

If the first step could be done on its own (e.g. refactoring logic into new methods) this would probably work out. But I have a hard time doing this or even seeing what the output is like before I step for step moved the parts over - a process of  having ``{{unrefactored-thing-at-the-beginning|raw}} ... {{unrefactored-thing-at-the-end|raw}}``. 😄  Have a look at the ``edit()`` method for example, maybe the part where the [rows are built](https://github.com/contao/core-bundle/blob/master/src/Resources/contao/drivers/DC_Table.php#L1927-L1995): independent of the problematic linear markup generation suddenly there is another return path mixed in that bubbles through the whole method...

**In general:** Would it make things easier to review if I'd create logical commits (refactoring a->b, refactoring b->c, refactoring c->d instead of refactoring a->d)? This would then require a review of every commit. I don't want to see b or c in the codebase as this is probably no production code - for example in case of ``DC_Table`` I first extracted the individual methods into traits because PHPStorm got laggy with realtime-refactoring...

> really like the idea of refactoring, though we still are not sure if we want to keep the monster this way or just start with something from scratch… 

Totally understand that. That's the primary goal of this PR - pushing a decision on how to go on. 😁
Ok, long text. Appreciate your comments!

aschempp:
> That's the primary goal of this PR - pushing a decision on how to go on.

😂 😂 👏 👏  perfect approach

Toflar:
Yeah, work on that DC_ monster is very welcome! 
So regarding the "how to go on"; I was thinking about refactoring too but then again this is such a monster that I wonder if reorganizing really makes sense. Or I mean if investing the time there makes sense.
Maybe a complete new approach that can be used side by side is a better idea? We talked about a `DC_API` once but I'd go further than that and just ignore the whole `DC_` idea because to be honest, we don't need different drivers. We need different views of the "same" data. The data should be retrievable using a general api that represents relations. Whether that's a tree relation or a list relation doesn't really matter from the API point of view. It's relevant for the view only.
I mean, I don't want to ask you to invent a whole new concept and work on it on your own but I just want to throw that into the discussion because it would be a pity having you working on refactoring a component that should actually be dropped anyway 😄

aschempp:
That's what I meant with the general discussion. Refactoring would be possible in 4.x though, and a new implementation is not. So even if we start a new implementation, refactoring is still a good thing because DC_Table will continue to live on for at least 5 years… It might also help us to learn something about these 6000 LOC before we start a new implementation ^^

m-vo:
Building a new thing from scratch (allthough I like that thought) would probably also postpone things like a BE rework (imo not really doable without templates everywhere) or larger modifications (think of dca 'quirks' like premature saving of invalid datasets or on ajax calls) quite a bit...

But I normally really like the idea of a concept target when refactoring. From my past experience with refactoring it either got me to a point where I got into a dead end but understood the problematic things enough to create a better aproach from scratch or at some point the code magically dissolved into logical essence. <-- But only if I made the right decisions when refactoring. I don't know if that holds true for the drivers (I still only have a limited understanding of the legacy code).

So maybe a good approach could be: 1) Discuss/Outline what we'd like to see in the future independent of what we have now. 2) Refactor in steps and see if (parts of) the concept can be achieved. 3) Reuse those things when redesigning for Contao 5.

asaage:
Just a note:
I was hoping that [cca/dc-general](https://github.com/contao-community-alliance/dc-general) would replace DC_Table one day and they hope that too.

bytehead:
> I was hoping that cca/dc-general would replace DC_Table one day

Please don't do this.

leofeyer:
>> I was hoping that cca/dc-general would replace DC_Table one day

> Please don't do this.

We won't. We have discussed it several times at various developer meetings and we always came to the same conclusion.

m-vo:
As discussed on the mumble call on 01/18 reworking the DC_* concept is too far in the future. I'm therefore closing this PR.

PR #1105, opened 2017-09-25
Validate array keys before saving to tl_user

ausi:
This should prevent `1054 Unknown column '...' in 'field list'` errors.

### DIFF
--- a/src/Resources/contao/library/Contao/User.php
+++ b/src/Resources/contao/library/Contao/User.php
@@ -566,8 +566,11 @@ abstract class User extends \System
 	 */
 	public function save()
 	{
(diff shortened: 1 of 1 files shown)


leofeyer:
This cannot be right. The error only occurs if the database is not up to date, doesn't it?

ausi:
We had the problem that if a custom permission gets added to `tl_user_group` but not to `tl_user`, this error occurs after the login.

leofeyer:
But that's a defect in the extension, isn't it?

ausi:
I’m not sure. Shouldn’t it be supported to have different settings in tl_user_group?

Issue #918, opened 2017-07-03
Problem mit MariaDB 10.2.4 oder höher


BugBuster1701:
Ab MariaDB 10.2.4 ist das Wort "rows" nun reserviert
Siehe: https://mariadb.com/kb/en/mariadb/reserved-words/

Das führt bei der Installlation von Contao 4.4. zum Error:
`SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'rows VARCHAR(8) DEFAULT '' NOT NULL`

```
[...]

PR #1106, opened 2017-09-25
[WIP] Fix MariaDB reserved keywords issue

ausi:
See contao/core-bundle#918

### DIFF
--- a/src/Resources/contao/library/Contao/Database.php
+++ b/src/Resources/contao/library/Contao/Database.php
@@ -707,6 +707,36 @@ class Database
 
 
 	/**
+	 * Escape the column name if it a reserved word
+	 *
+	 * @param string $name
+	 *
+	 * @return string
+	 */
+	public static function escapeColumnName($name)
+	{
+		if (in_array($name, ['rows'], true)) {
+			$name = '`'.$name.'`';
+		}
+
+		return $name;
+	}
+
+
+	/**
+	 * Escape a list of column names
+	 *
+	 * @param string[] $names
+	 *
+	 * @return string[]
+	 */
+	public static function escapeColumnNames(array $names)
+	{
+		return array_map([static::class, 'escapeColumnName'], $names);
+	}
+
+
+	/**
 	 * Execute a query and do not cache the result
 	 *
 	 * @param string $strQuery The query string
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -3143,12 +3143,12 @@ class DC_Table extends \DataContainer implements \listable, \editable
 			{
 				if ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 4)
 				{
-					$this->Database->prepare("UPDATE " . $this->strTable . " SET " . $this->strField . "='' WHERE pid=?")
+					$this->Database->prepare("UPDATE " . $this->strTable . " SET " . \Database::escapeColumnName($this->strField) . "='' WHERE pid=?")
 								   ->execute($this->activeRecord->pid);
 				}
 				else
 				{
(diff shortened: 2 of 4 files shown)


aschempp:
LGTM! we need to check/fix the install tool / Database Updater too though.

ausi:
@aschempp contao/installation-bundle#70

src/Resources/contao/library/Contao/Database.php
```diff
@@ -707,6 +707,36 @@ public function getUuid()
...
 	/**
+	 * Escape the column name if it a reserved word
+	 *
+	 * @param string $name
+	 *
+	 * @return string
+	 */
+	public static function escapeColumnName($name)
+	{
+		if (in_array($name, ['rows'], true)) {
```
Toflar: Can't we use the doctrine platform method here to quote everything we need (plus force `rows`)?

leofeyer: Isn't `in_array($name, ['rows'], true)` the same as `'rows' === $name`? Or did you just want to avoid the Yoda condition. 😄

Toflar: He just prepared for more in the future.

Toflar: @ausi, what about this? (should work in 3.5 too):



```php
// I'd name it "quote", not "escape"...
public static function quoteColumnName($name)
{
    if (method_exists('Contao\System', 'getContainer')) {
        $connection = Contao\System::getContainer()->get('connection');
        if ($connection instanceof Doctrine\DBAL\Connection) {
            $name = $connection->getDatabasePlatform()->quoteSingleIdentifier($name);
        }
    }

    // Force quoting "rows" for older Contao versions
    // (which will be ignored if already done by the platform in newer Contao versions)
    if (in_array($name, ['rows'], true)) {
        $name = '`'.$name.'`';
    }

    return $name;
}
```

ausi:
We already discussed parts of this issue in contao/installation-bundle#70 and I think there are open questions we have to agree on first:

1. Do we want to fix this for PHP 5.6 / dbal 2.5? If so, we should replace the `MySqlPlatform` and add `rows` to the keyword list.
2. Do we want to allow reserved keywords as column names in the future? If not we have to rename `rows` to something different in Contao 4.5

I would say YES to 1. and NO to 2.  
@contao/developers ?

Toflar:
To me, renaming the column name is not an option.

And then, there are 2 problems. One is the database schema and the other is querying the database. We need to fix querying anyway so this PR here is needed but imho can be improved as shown in my comment.
Speaking about the schema, I would fix it in 3.5 hardcoded on `rows` and for Contao 4 (installation bundle) I wouldn't do anything and just tell people to update to PHP 7.1 so they get dbal 2.6. I think it's reasonable to say that if you use the latest and greatest MariaDB you can also use the latest and greatest PHP version.

ausi:
> We need to fix querying anyway so this PR here is needed

Yes, if we want to support keywords as column names. But then we have to fix it everywhere in Contao (also in DC_Table for example). Do you think we should support keywords as table names too? Or only for columns and indexes?

Toflar:
I do think reserved keywords have to be properly escaped. To me it's best practice to do so in all the places you are not in control of the name of the column (= where you don't write the query yourself).  And it applies to all three of them, tables, columns and indexes although I tend to kind of ignore the tables.

ausi:
OK, then I will update this PR, so that we quote/escape all columns and indexes via doctrine.

leofeyer:
> I would say YES to 1. and NO to 2.

I agree, although it seems you have agreed on something else meanwhile. If so, never mind. 😄

aschempp:
I'm honestly not sure. Fixing this one issue is more important to me than adding a new feature to escape all fields in the deprecated database class... I would not invest too much time into this.

ausi:
Closed in favor of #1262

PR #1142, opened 2017-10-12
Fix XLIFF loading into $GLOBALS

discordier:
Without these checks we get:
`Undefined index: first` or `Illegal string offset 'with'` errors resulting in a 500 error when language files are not "compatible" to the expected array structure.


src/Config/Loader/XliffFileLoader.php
```diff
@@ -236,6 +236,9 @@ private function addGlobal(array $chunks, $value)
         $data = &$GLOBALS['TL_LANG'];
 
         foreach ($chunks as $key) {
+            if (!isset($data[$key]) || !is_array($data[$key])) {
+                $data[$key] = [];
+            }
```
ausi: Wouldn’t it be better to use something like the following?

```php
if (!isset($data) || !is_array($data)) {
    $data = [];
}
```

Otherwise every language string would be set to an array first and then be replaced by the message string.

leofeyer:
> language files are not "compatible" to the expected array structure

What does that mean?

src/Config/Loader/XliffFileLoader.php
discordier: That will not prevent us from the  `Undefined index:` on the line below as the array does not hold the key.
We could check if it is not the last element and only then define it as array but this would not bring much benefit as we would perform the test for each iteration and overwriting one array with a string should IMO be faster than testing all elements.
Have not benchmarked though.

discordier:
>> language files are not "compatible" to the expected array structure

> What does that mean?

See unit test.
file 1 defines:
```
$GLOBALS['TL_LANG']['MSC']['third'] = 'is-a-string';
```
file 2 defines:
```
$GLOBALS['TL_LANG']['MSC']['third']['with'][1]
```
The current approach breaks the whole system with a `ThrowableFatalError: Illegal string offset 'with'`

src/Config/Loader/XliffFileLoader.php
ausi: The array doesn’t have to hold the key if you accessing the reference of it.  
E.g. `$foo = []; $bar = &$foo['baz'];` doesn’t generate a warning.

discordier: You are of course right, I wonder why it didn't work on my system before.

PR #1153, opened 2017-10-20
[RFC] Allow prefilling of scripts and stylesheets

discordier:
IMO the backend template should only append variables and not kill values set by the developer in the parse method but rather append to them.

This way one may define custom stylesheets and scripts from within an custom backend controller that will not get overridden by the template.

I am using this in a custom backend controller template based upon `be_main.html.twig`:
```twig
{{ render_contao_backend_template({
    main: block('main'),
    error: block('error'),
    headline: block('headline'),
    stylesheets: block('stylesheets'),
    javascripts: block('javascripts'),
    mootools: block('mootools')
}) | raw }}

{# 
  Here we generate the stylesheets to be added to be_main instead of adding 
  stuff to $GLOBALS['TL_CSS'] within a Symfony controller.
#}
{% block stylesheets %}
    {% for stylesheet in stylesheets %}
        <link rel="stylesheet" href="{{ stylesheet }}">
    {% endfor %}
{% endblock %}
{# 
  Here we generate the javascripts to be added to be_main instead of adding 
  stuff to $GLOBALS['TL_JAVASCRIPT'] within a Symfony controller.
  TODO: add support for "async" here.
#}
{% block javascripts %}
    {% for javascript in javascripts %}
        <script src="{{ javascript }}"></script>
    {% endfor %}
{% endblock %}
{# 
  Here we have the mootools scripts to be added to be_main instead of adding 
  stuff to $GLOBALS['TL_MOOTOOLS'] within a Symfony controller.
#}
{% block mootools %}
    {% for mootool in mootools %}
        {{ mootool }}
    {% endfor %}
{% endblock %}
....
```

So we have now:
- three added dots in this PR
- a custom twig template

And I am able to do anything in a custom controller without any reference to `$GLOBALS`

@contao/developers what do you think? Shall we also expand the shipped template then?


discordier:
AppVeyor failed due to network outage apparently, this is unrelated to my changes.

src/Resources/contao/classes/BackendTemplate.php
```diff
@@ -73,7 +73,7 @@ protected function compile()
 				$strStyleSheets .= \Template::generateStyleTag($this->addStaticUrlTo($stylesheet), $options->media);
 			}
 
-			$this->stylesheets = $strStyleSheets;
+			$this->stylesheets .= $strStyleSheets;
```
bytehead: Shouldn't `$this->stylesheets` be of type `array` instead of `string`?
```twig
% block stylesheets %}
    {% for stylesheet in stylesheets %}
        <link rel="stylesheet" href="{{ stylesheet }}">
    {% endfor %}
{% endblock %}
```

Same for `$this->javascripts` and `$this->mootols`.

discordier: The block get's rendered to a string in the section you quoted?

I guess I don't see your point... :disappointed:
Could you clearify?

bytehead: Don't you have to pass the data to the twig template? 🤔

discordier: I still do not know what you mean, I simply added a template, expanded the contao-core template and off I go. Pretty close to the original code by @sheeep in his custom backend controller example.
To be honest, it is almost the same, I simply wanted to also be able to add stylesheets and javascripts in addition to headline, content and error.
I'll try to clearify again.

Controller (route annotation omitted as it does not relate):
```php
class DemoController
{
    /**
     * @return array
     *
     * @Template("AcmeBundle::dummy.html.twig")
     */
    public function __invoke()
    {
        // Without this PR we are forced to do this:
        // $GLOBALS['TL_CSS'][] = 'bundles/acmebundle/css/custom';
        // $GLOBALS['TL_JAVASCRIPT'][] = 'bundles/acmebundle/js/some-script';
        // $GLOBALS['TL_MOOTOOLS'][] = '<!-- MOOO comment! -->';

        // With this PR and the enhanced twig template we can do THIS:
        return [
                'stylesheets' => ['bundles/acmebundle/css/custom'],
                'javascripts' => ['bundles/acmebundle/js/some-script'],
                'mootools' => ['<!-- MOOO comment! -->'],
        ];
    }
}
```
AcmeBundle::dummy.html.twig:
```twig
{{ render_contao_backend_template({
    main: block('main'),
    error: block('error'),
    headline: block('headline'),
    stylesheets: block('stylesheets'),
    javascripts: block('javascripts'),
    mootools: block('mootools')
}) | raw }}

{# 
  Here we generate the stylesheets to be added to be_main instead of adding 
  stuff to $GLOBALS['TL_CSS'] within a Symfony controller.
#}
{% block stylesheets %}
    {% for stylesheet in stylesheets %}
        <link rel="stylesheet" href="{{ stylesheet }}">
    {% endfor %}
{% endblock %}
{# 
  Here we generate the javascripts to be added to be_main instead of adding 
  stuff to $GLOBALS['TL_JAVASCRIPT'] within a Symfony controller.
  TODO: add support for "async" here.
#}
{% block javascripts %}
    {% for javascript in javascripts %}
        <script src="{{ javascript }}"></script>
    {% endfor %}
{% endblock %}
{# 
  Here we have the mootools scripts to be added to be_main instead of adding 
  stuff to $GLOBALS['TL_MOOTOOLS'] within a Symfony controller.
#}
{% block mootools %}
    {% for mootool in mootools %}
        {{ mootool }}
    {% endfor %}
{% endblock %}

{% block headline %}
    This is headline as in example by @sheep.
{% endblock %}

{% block main %}
Hello there - I am the main content.
{% endblock %}
```

Rendering is as follows:
1. Annotation passes returned array data to template.
2. Template prepares blocks
3. Template passes blocks to the wrapper ContaoTemplateExtension from #512
4. The ContaoTemplateExtension adds ALL blocks to template (see [here](https://github.com/contao/core-bundle/pull/512/files#diff-5fb1652178784ba302a159c90bb5c2c4R75)
5. Contao BackendTemplate class renders result.

My PR solves the issue, that the in 4. added contents of `stylesheets`, `javascripts` and `mootools` are preserved and values from `$GLOBALS` get appended instead of overwriting.

I hope now you understand what I mean.

bytehead: nevermind, I'm in the wrong context - all good 😎

discordier:
@leofeyer I wonder if this should rather be tagged as bugfix as the whole purpose of the PR by @sheeep was to override values in `BackendTemplate` from a custom controller.
While, he added support for passing arbitrary keys to the template, he seems to have missed checking if these are actually overwritable in there.

If we really keep this as feature, custom backend controllers with custom assets will not be possible during the whole LTS phase without polluting `$GLOBALS`.

In the end, this is even a Bug in Contao 3.5, as in there the values passed to the template are silently overwritten as well.

src/Resources/contao/classes/BackendTemplate.php
discordier: It should not be needed because `BackendTemplate` extends `Template` which extends `Controller` which extends `System` which has a `__get()` returning `null` on unknown keys.
Therefore the `getter` will bubble up to `System` and the setter will set the value in the array.

On my System appending a string to `null` did not raise any notice but maybe I forgot some flag:
```
xtra@sigyn:~$ php -ddisplay_errors=stderr -derror_reporting=-1 \
    -r '$foo = null; $foo .= "bar"; var_export($foo);'
'bar'
```
So IMO we should be good to go.

@sheeep Did you accidentally forget these values or intentionally? So, is it a bugfix or feature?

sheeep: That was not intentional. No. So this might be a bugfix instead of a feature?

leofeyer:
And this does not require our `be_page.html.twig` template to be adjusted, too?

discordier:
@leofeyer: No, we do not have to, however we may do so - I tried to keep this PR as minimal as possible to eliminate possible BC breaks and the result were the "three dots".

In my application I have a custom base template which calls `render_contao_backend_template` (code see [above](https://github.com/contao/core-bundle/pull/1153#discussion_r145938507)).
So far every custom module must define an own template calling the extension to be able to add the stylesheets etc.
We could of course add the blocks to the base template - when doing so, however, we must check for variable existence as the fields are not necessarily present.

I'll post an example template later which we can discuss then.

PR #1161, opened 2017-10-26
Support checking the registry for table prefixed queries

Toflar:


src/Resources/contao/library/Contao/Model.php
```diff
@@ -1054,14 +1054,20 @@ protected static function find(array $arrOptions)
 		{
 			$arrColumn = (array) $arrOptions['column'];
 
-			if (count($arrColumn) == 1 && ($arrColumn[0] == static::$strPk || in_array($arrColumn[0], static::getUniqueFields())))
+			if (count($arrColumn) == 1)
 			{
-				$varKey = is_array($arrOptions['value']) ? $arrOptions['value'][0] : $arrOptions['value'];
-				$objModel = \Model\Registry::getInstance()->fetch(static::$strTable, $varKey, $arrColumn[0]);
+				// Support table prefixes
+				$arrColumn[0] = $result = preg_replace('/^' . preg_quote(static::getTable(), '/') . '\./', '', $arrColumn[0]);
```
leofeyer: Is the `$result = ` part supposed to be there?

src/Resources/contao/models/FilesModel.php
```diff
@@ -175,6 +175,14 @@ public static function findByUuid($strUuid, array $arrOptions=array())
 			$strUuid = \StringUtil::uuidToBin($strUuid);
 		}
 
+		// Check in model registry (does not work by default due to UNHEX())
+		$objModel = \Model\Registry::getInstance()->fetch(static::$strTable, $strUuid, 'uuid');
```
leofeyer: Similar to https://github.com/contao/core-bundle/blob/b894c22a85d68d2dea8fdfe3dbb7f5be1c2df38b/src/Resources/contao/models/PageModel.php#L660-L683 we should add a `empty($arrOptions)` check here, shouldn't we?

Toflar:
Updated :)

discordier:
I wonder if we should backport this to 3.5 as the non usage of the registry could be considered a bug.

leofeyer:
@Toflar What do you think?

Toflar:
Guess there are arguments for both, pro and contra :-) I‘d be okay with backporting it.

leofeyer:
Backported in 36bb1a78bbecc9ed5f561c7d9f607b62fbd68b62.

PR #1165, opened 2017-10-31
[RTM] Use Symfony asset component for Contao assets

aschempp:
I started using Symfony Encore for our application asset management and stumbled over the Symfony Asset service. I think there are a ton of useful feature for Contao.

 - It *knows* the path to an asset by asking for the composer package.
 - It automatically adds the `TL_ASSETS_URL` prefix if configured.
 - It automatically adds the package version to the URL, which results in clever cache busting on updates.

The provided code is totally working, but there are a lot more features we can implement. Missing **TODOs**:
 - [ ] Use asset URLs for `TL_JAVASCRIPT` and `TL_CSS`
 - [ ] Add support for asset URLs (absolute) to the `Contao\Combiner`
 - [x] Re-implement `Controller::setStaticUrls` instead of using constants in the context service.
 - [x] *Maybe* add `Resources/public` of all bundles as packages.
 - [ ] *Maybe* add a package for the backend theme
 - [x] Unit Tests

### DIFF
--- a/src/Asset/ContaoContext.php
+++ b/src/Asset/ContaoContext.php
@@ -2,46 +2,125 @@
 
 namespace Contao\CoreBundle\Asset;
 
+use Contao\Config;
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Contao\PageModel;
 use Symfony\Component\Asset\Context\ContextInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
 
 class ContaoContext implements ContextInterface
 {
     /**
+     * @var ContaoFrameworkInterface
+     */
+    private $framework;
+
+    /**
+     * @var RequestStack
+     */
+    private $requestStack;
+
+    /**
      * @var string
      */
-    private $constant;
+    private $field;
+
+    /**
+     * @var bool
+     */
+    private $debug;
+
+    /**
+     * @var PageModel|null
+     */
+    private $page;
 
     /**
      * Constructor.
      *
-     * @param string $constant
+     * @param ContaoFrameworkInterface $framework
+     * @param RequestStack             $requestStack
+     * @param string                   $field
+     * @param bool                     $debug
      */
-    public function __construct(string $constant)
+    public function __construct(ContaoFrameworkInterface $framework, RequestStack $requestStack, string $field, bool $debug = false)
     {
-        $this->constant = $constant;
+        $this->framework = $framework;
+        $this->requestStack = $requestStack;
+        $this->field = $field;
+        $this->debug = $debug;
     }
 
     /**
-     * Gets the base path.
+     * Sets the current page model.
      *
-     * @return string The base path
+     * @param PageModel|null $page
+     */
+    public function setPage(?PageModel $page)
+    {
+        $this->page = $page;
+    }
+
+    /**
+     * Gets the current page model.
+     *
+     * @return PageModel|null
+     */
+    public function getPage(): ?PageModel
+    {
+        return $this->page;
+    }
+
+    /**
+     * {@inheritdoc}
      */
     public function getBasePath()
     {
-        if (!defined($this->constant)) {
+        $request = $this->requestStack->getCurrentRequest();
+        $host = $this->getFieldValue();
+
+        if ($this->debug || '' === $host || null === $request) {
             return '';
         }
 
-        return rtrim(constant($this->constant), '/');
+        return sprintf(
+            '%s://%s%s',
+            $this->isSecure() ? 'https://' : 'http://',
+            preg_replace('@https?://@', '', $host),
+            $request->getBasePath()
+        );
     }
 
     /**
-     * Checks whether the request is secure or not.
-     *
-     * @return bool true if the request is secure, false otherwise
+     * {@inheritdoc}
      */
     public function isSecure()
     {
-        return defined($this->constant) && 0 === strpos(constant($this->constant), 'https://');
+        if (null === $this->page) {
+            $request = $this->requestStack->getCurrentRequest();
+
+            if (null === $request) {
+                return false;
+            }
+
+            return $request->isSecure();
+        }
+
+        return (bool) $this->page->loadDetails()->rootUseSSL;
+    }
+
+    /**
+     * @return string
+     */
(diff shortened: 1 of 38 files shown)


src/Asset/ContaoContext.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace Contao\CoreBundle\Asset;
+
+use Symfony\Component\Asset\Context\ContextInterface;
+
+class ContaoContext implements ContextInterface
```
Toflar: Nothing should rely on the constants anymore. If you need to access `TL_*` it's better to find a different way because the constants will not be available in 5 anymore.

leofeyer: The only other way is to read from `$GLOBALS` in the back end and from `global $objPage` in the front end, which does not seem less wrong than using the constants to me.

src/Asset/ContaoContext.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Gets the base path.
+     *
+     * @return string The base path
+     */
+    public function getBasePath()
+    {
+        if (!defined($this->constant)) {
+            return '';
+        }
```
DanielSchwiperich: is this check needed? `constant` cannot be undefined due to constructor string requirement and no method which unsets the property

src/Asset/ContaoContext.php
Toflar: Yeah, or we find a new way because apparently we will still need that feature in the future :-)

aschempp: Added ToDo no. 3 for this. We should probably inject the PageModel and "duplicate" the code from `Controller::setStaticUrls`.

aschempp:
I've added a dependency to https://github.com/terminal42/asset-bundle. This will automatically add all public bundle files as asset packages, completing the new Contao setup.


**in Twig templates:**
```twig
<img src="{{ asset('img/icon.png', 'rock_solid_slider') }}">
<img src="{{ asset('folder.gif', 'terminal42_folderpage') }}">
```

**in Contao templates:**
```html
<img src="<?= $this->asset('img/icon.png', 'rock_solid_slider') ?>">
<img src="<?= $this->asset('folder.gif', 'terminal42_folderpage') ?>">
```

**in Contao HTML module/element:**
```html
<img src="{{asset::img/icon.png::rock_solid_slider}}">
<img src="{{asset::folder.gif::terminal42_folderpage}}">
```

leofeyer:
I think we should finish this PR with the current feature set. The only item from the todo list that needs to be done is "Unit Tests". The rest can be added later.

aschempp:
@Toflar I have rewritten the contexts to use the page model, is that how you intended it?

src/Asset/ContaoContext.php
```diff
@@ -0,0 +1,125 @@
...
+
+        return (bool) $this->page->loadDetails()->rootUseSSL;
+    }
+
+    /**
+     * @return string
+     */
+    private function getFieldValue(): string
+    {
+        if (null === $this->page) {
```
leofeyer:
```php
        if (null !== $this->page) {
            return (string) $this->page->{$this->field};
        }

        /** @var Config $config */
        $config = $this->framework->createInstance(Config::class);

        return (string) $config->get($this->field);
```

aschempp:
> Why not have a ContaoAssetsContext and ContaoFilesContext?

I don't think that would make anything better. It means we need three classes instead of one, including an abstract. Which will be exactly the same, a class that will be configured, just by extending instead of a constructor argument.

src/Asset/ContaoContext.php
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @param PageModel|null $page
+     *
+     * @return string
+     */
+    private function getFieldValue(?PageModel $page): string
+    {
+        if (null === $page) {
+            /** @var Config $config */
+            $config = $this->framework->createInstance(Config::class);
```
leofeyer: Why are you using `createInstance()` instead of `getAdapter()` here? And shouldn't we initialize the framework first?

src/Resources/contao/library/Contao/Model.php
```diff
@@ -394,7 +394,7 @@ public function mergeRow(array $arrData)
 	 */
 	public function markModified($strKey)
 	{
-		if (!isset($this->arrModified[$strKey]))
+		if (!isset($this->arrModified[$strKey]) && isset($this->arrData[$strKey]))
```
leofeyer: Is this change related to the feature?

src/Asset/ContaoContext.php
Toflar: Both must be fixed indeed.

src/Resources/contao/library/Contao/Controller.php
```diff
@@ -1792,45 +1793,31 @@ public static function addEnclosuresToTemplate($objTemplate, $arrItem, $strKey='
 	/**
 	 * Set the static URL constants
 	 *
-	 * @param PageModel $objPage An optional page object
+	 * @deprecated Deprecated in Contao 4.5, to be removed in Contao 5. Use the asset contexts instead.
```
leofeyer: Since we have deprecated the method, we should no longer use it. Otherwise, if we still need to use it, we should not deprecate it. 😉

leofeyer:
Thank you @aschempp.

fritzmg:
In your examples you are using
```php
<img src="<?= $this->asset('folder.gif', 'terminal42_folderpage') ?>">
```
however, shouldn't it be
```php
<img src="<?= $this->asset('folder.gif', 'terminal42/contao-folderpage') ?>">
```
Or what exactly is `$packageName`?

Toflar:
Yes it is, the example comment was before some more commits ;)

aschempp:
@fritzmg `terminal42_folder` is correct. The bundle assets are based on the bundle classname, which in most cases and for correct setups should match the bundle alias for container configuration. We can't use the package name because we don't know which package a bundle belongs to (and a package theoretically can contain multiple bundles).

The diffeference is Contao componentes (assets), where the package name is the composer package (e.g. `contao-components/jquery`), because these do not have a bundle name.

PR #1166, opened 2017-11-02
[RTM] Generate the method name from hook

aschempp:
This will remove the need to set a method name for hook event listeners. The name is automatically generated from hook name, if it is not set in the tag.

```
// config.yml
services:
    App\EventListener\MyListener:
        tags:
            - { name: contao.hook, hook: generatePage }
```

```php
// MyListener.php

class MyListener
{
    public function onGeneratePage(…$foo)
    {
    }
}
```

### DIFF
--- a/src/DependencyInjection/Compiler/RegisterHookListenersPass.php
+++ b/src/DependencyInjection/Compiler/RegisterHookListenersPass.php
@@ -97,6 +97,6 @@ class RegisterHookListenersPass implements CompilerPassInterface
             return (string) $attributes['method'];
         }
 
-        return 'on' . ucfirst($attributes['hook']);
+        return 'on'.ucfirst($attributes['hook']);
     }
 }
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/Compiler/RegisterHookListenersPass.php
```diff
@@ -56,38 +56,47 @@ private function getHooks(ContainerBuilder $container): array
...
+     * Gets the method name from config or hook name.
+     *
+     * @param array $attributes
+     *
+     * @return string
+     */
+    private function getMethod(array $attributes): string
+    {
+        if (isset($attributes['method'])) {
+            return (string) $attributes['method'];
```
leofeyer: Why did you cast to string here?

aschempp: why? It could be an integer, can it not? Just making sure the return type of the function is met…

PR #1168, opened 2017-11-02
Use section templates for all fe_page sections

aschempp:


leofeyer:
As discussed in Mumble on November 2nd, 2017, the majority of the participants did not like this approach.

PR #1171, opened 2017-11-02
[WIP] Universal picker widget

qzminski:
As requested by @aschempp this pull request introduces the universal picker widget. The field configuration can look like this:

```php
$GLOBALS['TL_DCA']['tl_module']['fields']['article'] = [
    'label' => &$GLOBALS['TL_LANG']['tl_module']['article'],
    'exclude' => true,
    'inputType' => 'picker',
    'eval' => [
        'foreignTable' => 'tl_article', // Foreign table for widget record fetching
        'context' => 'article', // Context for picker
        'labelField' => 'title', // Field for the label
        'fieldType' => 'radio', // Field type (optional, defaults to radio)

        // Custom record label callback (icon is added anyway)
        'label_callback' => function (array $row) {
            return $row['title'] . ' (ID: ' . $row['id'] . ')';
        },

        // Multiple sortable selection
        //'fieldType'    => 'checkbox',
        //'sortable'     => true,
    ],
    'sql' => ['type' => 'integer', 'unsigned' => true],
];
```

Concerns:

1. Perhaps `foreignTable` can be fetched from the picker.
2. `label_callback` could make its way one level up in the array just where other callbacks sit.

The `ArticlePickerProvider` adjustments taken from https://github.com/contao/core-bundle/pull/999.


src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+		elseif (strpos($varInput, ',') === false)
+		{
+			return $this->multiple ? array(intval($varInput)) : intval($varInput);
+		}
+		else
+		{
+			$arrValue = array_map('intval', array_filter(explode(',', $varInput)));
+
+			return $this->multiple ? $arrValue : $arrValue[0];
+		}
```
rabauss: we had problems on saving multiple selected ids and afterwards search them via like in DB - so do we have to cast to string at this point?

```        
        if ($this->multiple)
        {
            // Cast to string to search in serialized array via db
            if (!empty($this->varValue)) {
                foreach ($this->varValue as &$value) {
                    $value = (string)$value;
                }
            }
        }
```

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+
+
+	/**
+	 * Load the database object
+	 *
+	 * @param array $arrAttributes
+	 */
+	public function __construct($arrAttributes=null)
+	{
+		$this->import('Database');
```
Toflar: Do not import it here, use `getInstance()` instead.

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+			if ($this->mandatory)
+			{
+				$this->addError(sprintf($GLOBALS['TL_LANG']['ERR']['mandatory'], $this->strLabel));
+			}
+
+			return '';
+		}
+		elseif (strpos($varInput, ',') === false)
+		{
+			return $this->multiple ? array(intval($varInput)) : intval($varInput);
```
Toflar: Sure we need `inval` here? What if I want to pick multiple string values like string uuids?

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+			// Label field
+			$label = $data[$this->labelField];
+		}
+		else
+		{
+			$label = $data['id'];
+		}
+
+		// Add the icon
+		if ($this->icon)
```
Toflar: Don't you think it would make sense to put the icon in the callback as well? E.g. the page picker has different icons per page type and this sounds like a typical requirement to me :)

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+	 */
+	public function generate()
+	{
+		$arrValues = array();
+
+		// Generate the records
+		if (!empty($this->varValue)) // Can be an array
+		{
+			$ids = array_map('intval', (array) $this->varValue);
+			$records = $this->Database->execute("SELECT * FROM {$this->foreignTable} WHERE id IN (" . implode(',', $ids) . ")" . ($this->sortable ? (" ORDER BY " . $this->Database->findInSet('id', $ids)) : ""));
```
Toflar: Generating the query should go to its own protected method so it can be easily overridden in a custom widget.

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+
+			while ($records->next())
+			{
+				$arrValues[$records->id] = $this->generateRecordLabel($records->row());
+			}
+		}
+
+		$return = '<input type="hidden" name="'.$this->strName.'" id="ctrl_'.$this->strId.'" value="'.implode(',', array_keys($arrValues)).'">
```
Toflar: Can we use a template here please? I know it's old code but still :)

src/Resources/contao/widgets/PickerWidget.php
qzminski: I have asked @aschempp  about the style of this *legacy* class with and he said I should just take the existing code (based on page tree widget) as is for now. Ideally we should not use `Database` but the db connection service.

src/Resources/contao/widgets/PickerWidget.php
qzminski: The `fileTree` widget will not extend this class (at least for now). /cc @aschempp

src/Resources/contao/widgets/PickerWidget.php
qzminski: This `icon` setting is defined by the developer and there is no icon provided by default. Thus if your icon is not static you can just not set this property and add it in the `label_callback`.

Also the `label_callback` is called before the evaluation of the `icon` setting, which allows you to use callback just to format the label properly but still set the static icon afterwards.

src/Resources/contao/widgets/PickerWidget.php
qzminski: I have never encountered such problem but if that's the case then also other widgets are affected. What query have you used to search the serialized data? Or did you set the field to handle data in `csv` format?

src/Resources/contao/widgets/PickerWidget.php
Toflar: I see, I can live with that :)

src/Resources/contao/widgets/PickerWidget.php
Toflar: No but maybe I want to store uuids in my case too :)

src/Resources/contao/widgets/PickerWidget.php
Toflar: Yeah it's fine for me to just use the legacy class here, but still `Database::getInstance()` instead of importing it in some global variable :)

src/Resources/contao/widgets/PickerWidget.php
rabauss: We use something like this ` WHERE (xx LIKE '%"1"%' OR xx LIKE '%"10"%') ` - it's similar to the backend-filter for members.

Yes, i guess it's also an problem for other widgets. But at the moment I don't know if there is an integer-picker in the core. the file picker uses uuid, so it's string...

qzminski: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/widgets/PageTree.php#L126

rabauss: I know that code - because I also copied that initialy for our own picker ;-)
it saves the ids correctly but you won't be able to search via LIKE...

aschempp: Imho that's not related to this pull request…

src/Resources/contao/widgets/PickerWidget.php
aschempp: 👎 for a template, please don't over-engineer old code, otherwise this will never get merged…

src/Picker/ArticlePickerProvider.php
```diff
@@ -51,10 +55,29 @@ public function getDcaTable(): string
      */
     public function getDcaAttributes(PickerConfig $config): array
     {
+        $value = $config->getValue();
         $attributes = ['fieldType' => 'radio'];
 
-        if ($this->supportsValue($config)) {
-            $attributes['value'] = str_replace(['{{article_url::', '}}'], '', $config->getValue());
+        if ('article' === $config->getContext()) {
+            $attributes = ['fieldType' => $config->getExtra('fieldType')];
```
aschempp: remove this line, it overrides the default in line 59 and is duplicate with line 65

src/Picker/ArticlePickerProvider.php
```diff
@@ -65,6 +88,10 @@ public function getDcaAttributes(PickerConfig $config): array
      */
     public function convertDcaValue(PickerConfig $config, $value): string
     {
+        if ('article' === $config->getContext()) {
+            return (string) $value;
```
aschempp: use `(int)` casting, we're picking an ID (see `PagePickerProvider`). Which means we need to remove the method's return type

src/Resources/contao/classes/Ajax.php
```diff
@@ -332,7 +333,18 @@ public function executePostActions(DataContainer $dc)
...
+
+				switch ($this->strAction) {
+					case 'reloadPagetree':
+						$strKey = 'reloadPagetree';
+						break;
+					case 'reloadFiletree':
+						$strKey = 'fileTree';
+						break;
+					default:
+						$strKey = 'picker';
```
aschempp: this changes the fallback action, but I think that's fine if noone else has any concerns

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+ * @property string         $context
+ * @property string         $labelField
+ * @property string         $icon
+ * @property boolean        $sortable
+ * @property array|callable $label_callback
+ *
+ * @author Leo Feyer <https://github.com/leofeyer>
+ * @author Kamil Kuzminski <https://github.com/qzminski>
+ */
+class PickerWidget extends \Widget
```
aschempp: The current widgets never have a class suffix. @leofeyer should decide if we want to change that here …

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+		if ($varInput == '')
+		{
+			if ($this->mandatory)
+			{
+				$this->addError(sprintf($GLOBALS['TL_LANG']['ERR']['mandatory'], $this->strLabel));
+			}
+
+			return '';
+		}
+		elseif (strpos($varInput, ',') === false)
```
aschempp: I don't think we actually need this `elseif` at all. The `else` statement is perfectly fine to handle all cases…

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+		}
+	}
+
+
+	/**
+	 * Check the selected value
+	 *
+	 * @param mixed $varInput
+	 */
+	protected function checkValue($varInput)
```
aschempp: I think this should be `validateValue($varInput)`

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+
+
+	/**
+	 * Generate the record label
+	 *
+	 * @param array $data
+	 *
+	 * @return string
+	 */
+	protected function generateRecordLabel(array $data)
```
aschempp: Can we not make `label_callback` just a mandatory argument when using this widget? Instead of multiple features that override each other?

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,240 @@
...
+		$return .= '</ul>';
+
+		if (!\System::getContainer()->get('contao.picker.builder')->supportsContext($this->context))
+		{
+			$return .= '
+	<p><button class="tl_submit" disabled>'.$GLOBALS['TL_LANG']['MSC']['changeSelection'].'</button></p>';
+		}
+		else
+		{
+			$extras = ['fieldType' => $this->fieldType];
```
aschempp: `fieldType` is an optional property, default is a radio button (see my previous comment)

aschempp:
We should also add a content element picker (either in this PR or in a subsequent one related to this).

src/Resources/contao/classes/Ajax.php
qzminski: Well not really, only three actions are allowed here (see lines 268-270) and we cover each case. This `default` can be a bit misleading though...

src/Resources/contao/widgets/PickerWidget.php
qzminski: I think you are right. Are we missing something @leofeyer or can this be dropped?

src/Resources/contao/widgets/PickerWidget.php
qzminski: I have copied it "as is" from the page tree widget. I can change the method name here but then it will be a mismatch between those two widgets. Should not be a problem but if 90% is already the same...

src/Resources/contao/widgets/PickerWidget.php
qzminski: We could, yet I find my solution more user friendly. Would love to hear opinion of the others though.

src/Resources/contao/widgets/PickerWidget.php
qzminski: Should we fix it in the page tree widget as well then? https://github.com/contao/core-bundle/blob/master/src/Resources/contao/widgets/PageTree.php#L232

src/Resources/contao/classes/Ajax.php
aschempp: I know about the three actions. What I meant is that before this change, the default was `fileTree` if `$this->strAction != 'reloadPagetree'`. Now the default is `picker`. As I said, I'm fine with this, though I would assume we should not support an unknown action anyway.

src/Resources/contao/widgets/PickerWidget.php
aschempp: Then we should keep it, sorry I didn't see that.

src/Resources/contao/widgets/PickerWidget.php
aschempp: ping @contao/developers

src/Resources/contao/widgets/PickerWidget.php
aschempp: Well I guess so. We already changed it in the picker (https://github.com/contao/core-bundle/blob/master/src/Picker/PagePickerProvider.php#L62-L67) so I guess we just missed that.

src/Resources/contao/widgets/PickerWidget.php
Toflar: For me there are only two use cases:

* You just want to output the title and have it look good. In that case you don't even want to specify the icon, you just want to have a general list icon and pass on the label field column name.
* You care about the icon, you have a different icon per row or a different label source per row. In that case you use the callback.

So for me a `labelField` (uses a predefined icon and that label column)  and `label_callback` (allows you to do everything you like) would be enough,

aschempp: I agree, but please call it `label` and `label_callback` (as it is in the DCA list config).

qzminski: Collides with the regular label field and `__set`/`__get` magic methods 🙄

src/Resources/contao/widgets/PickerWidget.php
qzminski: Changed in 072b178.

Toflar:
Ping @qzminski, this PR needs to include https://github.com/contao/core-bundle/commit/1ac36e302d41802d61a88308256677c6bfb423f1.

qzminski:
Just did that.

src/Resources/contao/dca/tl_content.php
```diff
@@ -716,9 +716,27 @@
...
-			'inputType'               => 'select',
-			'options_callback'        => array('tl_content', 'getArticleAlias'),
-			'eval'                    => array('mandatory'=>true, 'chosen'=>true, 'submitOnChange'=>true, 'tl_class'=>'w50 wizard'),
+			'inputType'               => 'picker',
+			//'options_callback'        => array('tl_content', 'getArticleAlias'),
+			'eval'                    => array(
+				'mandatory'    => true,
+				'tl_class'     => 'clr',
+
+				'foreignTable' => 'tl_article', // Foreign table for widget record fetching. @todo get from picker somehow?
```
Toflar: Imho fetching and validating the records should be done in simple, small methods so they can be easily overridden. Right now there's still to much going on in one single method. Can you split that up a little? Then that todo can be removed :)

src/Resources/contao/dca/tl_content.php
```diff
@@ -716,9 +716,27 @@
...
+				// Custom record label callback (icon is added anyway)
+				'label_callback' => function (array $row) {
+				    return $row['title'] . ' (ID: ' . $row['id'] . ')';
+				},
+
+				// Multiple sortable selection
+//				'fieldType'    => 'checkbox',
+//				'sortable'     => true,
+            ),
+			// @todo – how to handle wizard?
```
Toflar: What does the wizard do in that case? I mean why is a wizard needed at all?

qzminski: Well previously it was a select menu field, so when you have selected the record you were able to edit it in popup by clicking the edit icon. Dunno if we should drop or adjust it here.

src/Resources/contao/dca/tl_content.php
qzminski: Do you mean some custom callbacks here or just the protected methods inside the widget class?

src/Resources/contao/dca/tl_content.php
Toflar: imho that's not needed, you can add it in the callback if you like, can't you?

src/Resources/contao/dca/tl_content.php
Toflar: Protected methods :)

src/Resources/contao/dca/tl_content.php
qzminski: Dropped it in b5e3f2d. I left the method in the class for BC reasons.

src/Resources/contao/dca/tl_content.php
qzminski: See 68d9b80.

src/Resources/contao/widgets/PickerWidget.php
aschempp: 🤦‍♂️

src/Resources/contao/dca/tl_content.php
```diff
@@ -716,13 +716,25 @@
 		(
 			'label'                   => &$GLOBALS['TL_LANG']['tl_content']['articleAlias'],
 			'exclude'                 => true,
-			'inputType'               => 'select',
-			'options_callback'        => array('tl_content', 'getArticleAlias'),
-			'eval'                    => array('mandatory'=>true, 'chosen'=>true, 'submitOnChange'=>true, 'tl_class'=>'w50 wizard'),
-			'wizard' => array
-			(
-				array('tl_content', 'editArticleAlias')
```
aschempp: You should also remove this method from the class.

src/Resources/contao/widgets/PickerWidget.php
```diff
@@ -0,0 +1,250 @@
...
+		{
+			$callback = $this->label_callback;
+
+			$label = static::importStatic($callback[0])->{$callback[1]}($data, $this->objDca);
+		}
+		elseif (is_callable($callback = $this->label_callback))
+		{
+			$label = $callback($data, $this->objDca);
+		}
+		elseif ($this->labelField)
```
aschempp: Maybe check if data contains `labelField` ? I would also suggest to make the field mandatory (throw exception if it does not exist here).

src/Resources/contao/widgets/PickerWidget.php
aschempp: Why would that be necessary? With the label callback everything is possible, so I don't ever see a need to join something? If you don't have actual DB records, I would think one should override the whole method?

src/Resources/contao/dca/tl_content.php
qzminski: I have not removed it for BC reasons, some extensions could use it. Maybe instead of removing it I should throw the deprecation warning?

src/Resources/contao/widgets/PickerWidget.php
qzminski: Imo this should stay as is. What if I want to work on models and not the query? Or non-DB records like @aschempp mentioned?

src/Resources/contao/widgets/PickerWidget.php
qzminski: Changed in b853270. I have not used the `isset()` on purpose because the data property could potentially contain a `null` value.

src/Resources/contao/dca/tl_content.php
aschempp: Agreed, we should keep it. But we should add the deprecation notice.

qzminski: Changed in 0d178e6.

aschempp:
I think we need to re-think this implementation. Imho the picker provider should be able to render it's items, so the picker widget does not need to know. Picking an article should look the same in every record, it does not make sense to re-implement a label callback in every dca.

Unfortunately, this means we won't make this into 4.5 … 😞

leofeyer:
Are you planning on completing this PR?

leofeyer:
See contao/contao#714

Issue #1173, opened 2017-11-05
animated GIFs are not resized correctly with Imagick


fritzmg:
It appears Contao (or Imagick) is unable to properly resize animated GIFs, when Imagick is available and in use. Take the following example GIF: 

![f-startup-challenge-animation](https://user-images.githubusercontent.com/4970961/32414174-87ce7f80-c221-11e7-8fcc-78fa15505982.gif)

If you put that GIF into the file manager of Contao, where only GD is available, it will look normal:

![animated-gif-gd](https://user-images.githubusercontent.com/4970961/32414178-a33a1590-c221-11e7-9f37-72b2417a3f71.png)

However, if you put that GIF into the file manager of Contao where Imagick is available, it will look like this:

![animated-gif-imagick](https://user-images.githubusercontent.com/4970961/32414179-b24c9dc8-c221-11e7-9cc6-1eb1d9aaa642.png)

The file name and path of the latter image is still 
```
assets/images/2/f-startup-challenge-animation-d8ff2dde.gif
```
which means Contao indeed processed the image and probably tried to resize it to the back end size - but the resize did not really occur for some reason.

/cc @ausi

ausi:
I was able to reproduce this behavior with Imagick. I will look into it.

ausi:
I reduced it to a simple test case, and it looks like this is an Imagine or Imagick Issue. Both, resizing:

```php
(new \Imagine\Imagick\Imagine)
	->open('path/to/animated-gif.gif')
	->resize(new \Imagine\Image\Box(100, 100))
	->show('gif');
```

and cropping:

```php
(new Imagine\Imagick\Imagine)
	->open('path/to/animated-gif.gif')
	->crop(new \Imagine\Image\Point(0, 0), new \Imagine\Image\Box(100, 100))
	->show('gif');
```

doesn’t work as documented.

I think this should be reported to [avalanche123/Imagine](https://github.com/avalanche123/Imagine).

It would be great if someone can test if this issue appears with Gmagick too.

fritzmg:
May be a related ticket: https://github.com/avalanche123/Imagine/issues/402 ?

ausi:
I don’t think so. Animated GIFs should already be supported since avalanche123/Imagine#481 was merged.

PR #1175, opened 2017-11-06
set max image size in style

fritzmg:
Currently it can still happen that images in the file manager are displayed in their full size and thus break the layout (see https://github.com/contao/core-bundle/issues/1173 for example). Adding this inline style would at least prevent the image from breaking the layout.

### DIFF
--- a/src/Resources/contao/drivers/DC_Folder.php
+++ b/src/Resources/contao/drivers/DC_Folder.php
@@ -2728,12 +2728,12 @@ class DC_Folder extends \DataContainer implements \listable, \editable
 
 						if ($importantPart->getPosition()->getX() > 0 || $importantPart->getPosition()->getY() > 0 || $importantPart->getSize()->getWidth() < $objFile->width || $importantPart->getSize()->getHeight() < $objFile->height)
 						{
-							$thumbnail .= ' ' . \Image::getHtml(\System::getContainer()->get('contao.image.image_factory')->create(TL_ROOT . '/' . rawurldecode($currentEncoded), (new ResizeConfiguration())->setWidth(320)->setHeight(40)->setMode(ResizeConfiguration::MODE_BOX)->setZoomLevel(100))->getUrl(TL_ROOT), '', 'style="margin:0 0 2px 0;vertical-align:bottom"');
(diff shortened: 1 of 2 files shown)


src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2733,7 +2733,7 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 					}
 					catch (RuntimeException $e)
 					{
-						$thumbnail .= '<br><p class="broken-image" style="margin:0 0 2px -18px">Broken image!</p>';
+						$thumbnail .= '<br><p class="broken-image" style="margin:0 0 2px -18px;max-width:400px;max-height:50px;width:auto;height:auto">Broken image!</p>';
```
ausi: Why adding the style to the `<p>` tag too?

fritzmg: Search and replace error ;). I'll fix it. I'll also add something similar for the important part preview (just in case).

ausi:
@leofeyer there are already inline styles for the images, @fritzmg just added some. And I think it’s a good idea to set the max-width as inline style so that the every backend theme gets this.

leofeyer:
The existing inline styles are there for historical reasons and should be removed. I'm against adding even more, even though I get your point.

src/Resources/contao/drivers/DC_Folder.php
```diff
@@ -2717,18 +2717,18 @@ protected function generateTree($path, $intMargin, $mount=false, $blnProtected=t
 						// Inline the image if no preview image will be generated (see #636)
 						if ($objFile->height !== null && $objFile->height <= 50 && $objFile->width !== null && $objFile->width <= 400)
 						{
-							$thumbnail .= '<br><img src="' . $objFile->dataUri . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="" style="margin:0 0 2px -18px">';
+							$thumbnail .= '<br><img src="' . $objFile->dataUri . '" width="' . $objFile->width . '" height="' . $objFile->height . '" alt="" style="margin:0 0 2px -18px;max-width:400px;max-height:50px;width:auto;height:auto">';
```
leofeyer: This change seems redundant. The if condition already ensures that the image is not bigger than 400x50.

fritzmg: True, I have removed that style now.

leofeyer:
Merged in 9cc579a4c6efd417e28f25a00a9eb527ef7bf5fc. I didn't apply the fix to the `FileSelector`, because it is deprecated and does not even have the `.tl_file_manager` class.

PR #1178, opened 2017-11-06
[RTM] Only output width and height attributes if they are set

ausi:
See #1163


src/Resources/contao/templates/picture/picture_default.html5
```diff
@@ -8,7 +8,7 @@
...
+<img src="<?= $this->img['src'] ?>"<?php if ($this->img['srcset'] !== $this->img['src']) echo ' srcset="' . $this->img['srcset'] . '"'; ?><?php if (!empty($this->img['sizes'])) echo ' sizes="' . $this->img['sizes'] . '"'; elseif (!$this->sources && $this->img['width'] && $this->img['height']) echo ' width="' . $this->img['width'] . '" height="' . $this->img['height'] . '"'; ?> alt="<?= $this->alt ?>"<?php if ($this->title) echo ' title="' . $this->title . '"'; ?><?php if ($this->class || !empty($this->img['class'])) echo ' class="' . trim($this->class. ' ' . $this->img['class']) . '"'; ?><?= $this->attributes ?> itemprop="image">
```
leofeyer: Shouldn't this be `!empty($this->img['width']) && !empty($this->img['height'])`?

ausi: Done in 336603e2a497e62dc9de387d2b8717d6a612856b

ausi:
@leofeyer should we backport this to Contao 3.5?

leofeyer:
No. It is May 2018 so Contao 3.5 is in the "security fixes only" phase of the long term support. And I don't want to risk any core developer leaving the team. 😉

ausi:
Ah, I missed that 3.5 ist out of LTS now. No backport needed then 👍

PR #1181, opened 2017-11-12
[RTM] Add picture service example to DEPRECATED.md

ausi:


leofeyer:
Thank you @ausi.

@BugBuster1701 /cc

PR #1184, opened 2017-11-13
Filetree and Pagetree forces unnecessary saving of new versions

agoat:
The filetree and pagetree widgets force new versions every time the selection is empty. 

This is because the input of the order field is directly php exploded which returns always an array. An empty string will become an array with one empty entry (like `array(0=>'')`) and therefor is not equal to the order array extracted from the database in the `__construct` method which returns an empty result as an empty array.
This only happens if an `orderField` is specified in the dca.

This PR simply adds a check if the input is empty.


src/Resources/contao/widgets/FileTree.php
```diff
@@ -97,9 +97,9 @@ protected function validator($varInput)
 		}
 
 		// Store the order value
-		if ($this->orderField != '')
+		if ($this->orderField != '' && ($order = \Input::post($this->strOrderName)) !='')
```
ausi: Wouldn’t this make it impossible to deselect files if the order field in the database has already some values stored?

agoat: No. The order values will stay in the database but this does not prevent saving an empty selection itself.

When the files are selected again the saved order is applied directly. This could be a feature ;-)

agoat: But we can change this to save empty order values as well:

```php
		// Store the order value
		if ($this->orderField != '')
		{
			$arrNew = !empty($order = \Input::post($this->strOrderName)) ? explode(',', $order) : array();

			// Only proceed if the value has changed
			if ($arrNew !== $this->{$this->orderField})
			{
				$this->Database->prepare("UPDATE {$this->strTable} SET tstamp=?, {$this->orderField}=? WHERE id=?")
							   ->execute(time(), serialize($arrNew), $this->activeRecord->id);

			    $this->objDca->createNewVersion = true; // see #6285
			}
		}
```

ausi: Yes, I think saving empty order values would be better.

agoat: I am of the same opinion and changed the PR accordingly.

src/Resources/contao/widgets/FileTree.php
```diff
@@ -99,7 +99,7 @@ protected function validator($varInput)
 		// Store the order value
 		if ($this->orderField != '')
 		{
-			$arrNew = array_map('StringUtil::uuidToBin', explode(',', \Input::post($this->strOrderName)));
+			$arrNew = !empty($order = \Input::post($this->strOrderName)) ? array_map('StringUtil::uuidToBin', explode(',', $order)) : array();
```
leofeyer: I don't like the coding style here in terms of readability. What about this?

```php
$arrNew = array();

if ($order = \Input::post($this->strOrderName))
{
    $arrNew = array_map('StringUtil::uuidToBin', explode(',', $order));
}
```

src/Resources/contao/widgets/PageTree.php
```diff
@@ -95,7 +95,7 @@ protected function validator($varInput)
 		// Store the order value
 		if ($this->orderField != '')
 		{
-			$arrNew = explode(',', \Input::post($this->strOrderName));
+			$arrNew = !empty($order = \Input::post($this->strOrderName)) ? explode(',', $order) : array();
```
leofeyer: Same here.

```php
$arrNew = array();

if ($order = \Input::post($this->strOrderName))
{
    $arrNew = explode(',', $order);
}

```

src/Resources/contao/widgets/FileTree.php
agoat: Yeah, that's easier to read. I will change this tomorrow.

leofeyer: Never mind, I already have so I can merge the PR.

PR #1185, opened 2017-11-13
Copy child tables without setting tstamp if no tstamp column exist

agoat:
An SQL error occurs when copying an entry in a table that has a child table **without a tstamp column** in the 'ctable' array.

There should be a check if a tstamp column exists before setting a tstamp value in the insert dataset.

### DIFF
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -1142,7 +1142,11 @@ class DC_Table extends \DataContainer implements \listable, \editable
 					}
 
 					$copy[$v][$objCTable->id]['pid'] = $insertID;
-					$copy[$v][$objCTable->id]['tstamp'] = $time;
+
+					if ($objCTable->tstamp)
+					{
+						$copy[$v][$objCTable->id]['tstamp'] = $time;
+					}
 				}
 			}
 		}


aschempp:
Contao's DCA does not really support tables without a `tstamp` field …

leofeyer:
I also consider this a known limitation. In addition, I doubt that this one change implements the feature comprehensively. There are more places where the `tstamp` column is added.

agoat:
OK, then a tstamp column is mandatory for all child tables.
In my case I have a child table that doesn't need a tstamp. But it doesn't hurt if a tstamp colum exists.

> There are more places where the tstamp column is added.

I just stumbled over this while copying and didn't look any further.

leofeyer:
> In my case I have a child table that doesn't need a tstamp. But it doesn't hurt if a tstamp colum exists.

It is not nice, either. We should probably support tables without a `tstamp` column, however we have to find all places in the code that need to be adjusted.

agoat:
> We should probably support tables without a tstamp column, however we have to find all places in the code that need to be adjusted.

Should I open a new PR to find all `tstamp`'s in the code?

leofeyer:
I don't know. @contao/developers What do you think?

Toflar:
Many developers rely on `tstamp` being available. I'd not change that (also I don't see any reason at all why you would ever want to omit that).

PR #1186, opened 2017-11-14
Don't handle exceptions when the request it isn't a Contao request

bytehead:
Let exceptions be exceptions if the exceptions are not originating from a Contao request.

There is also a fix ready for the `develop` branch: https://github.com/bytehead/core-bundle/tree/fix/pretty-error-screen-listener


bytehead:
Already implemented in https://github.com/contao/core-bundle/commit/9abe32c9b468375ae042d81809dafe8a5ac8dd06.

leofeyer:
Thank you @bytehead, however, this has been fixed in 9abe32c9b468375ae042d81809dafe8a5ac8dd06 already. :confused:

PR #1190, opened 2017-11-15
[RTM] Simplify the fragments

aschempp:
Contao 4.5.0-beta1 added controller fragments for frontend module and content elements. After working with the implementation, I noticed it is much more complicated then necessary. Here's a draft for a rewrite. A lot of files have been deleted, simply ignore them in reviews …

### Register a frontend module

```yaml
// services.yml
app.frontend_module.foobar:
    class: App\Controller\FrontendModule\FoobarController
    tags:
        - { name: contao.frontend_module, category: application }
```

#### Additional tag options
<dl>
<dt>method</dt>
<dd>If the method argument is not present, we're assuming the <code>__invoke</code> method should be called</dd>
<dt>type</dt>
<dd>The type argument is optional and is generated from class name. A "Controller" suffix is stripped and then the class name is "underscored". <code>FoobarController => 'foobar'</code>, <code>SpecialFeatureController => 'special_feature'</code>
<dt>priority</dt>
<dd>Priority allows to sort the container result to override other controllers.</dd>
<dt>renderStrategy</dt>
<dd>Usually used with "inline" or "esi". Also useful if you want some real special features, you can create your own "kernel.fragment_renderer" and use it's name here.</dd>
</dl>

### Render the frontend module
```php
$module = \ModuleModel::findByPk(17);
$renderer = \System::getContainer()->get('contao.fragment.renderer');
echo $renderer->render(new FrontendModuleReference($module));
```

### Custom rendering options
As we are decorating the Symfony `ControllerResolver`, regular Symfony fragment rendering still works perfectly fine with our fragments.

```php
$uri = new ControllerReference(
    'contao.frontend_module.foobar',
    ['attribute' => 'value'],
    ['query' => 'value']
);
$fragmentHandler = $container->get('fragment.handler');
echo $fragmentHandler->render($uri, 'inline', ['ignore_errors' => false]);
```

(This obviously bypasses the fragment tag configuration as well as any automatic configuration done by `FrontendModuleReference`.)

## Side notes
 **Page types are no longer fragments**. They never were supposed to be, because they cannot be rendered inside a page. We should use the cmf-router at some point to map pages to actual routes.

## ToDos / Discuss:
 - [ ] Unit tests
 - [x] Should it be _renderStrategy_? Or _renderer_ (as the `render()` method argument name), or just `strategy`?


src/DependencyInjection/Compiler/MapFragmentsToGlobalsPass.php
```diff
@@ -0,0 +1,101 @@
...
+    public function process(ContainerBuilder $container): void
+    {
+        $elements = $this->getFragmentTags($container, ContentElementReference::TAG_NAME);
+        $elements = $this->createGlobalsMapForTag($elements, 'TL_CTE', ContentProxy::class);
+
+        $modules = $this->getFragmentTags($container, FrontendModuleReference::TAG_NAME);
+        $modules = $this->createGlobalsMapForTag($modules, 'FE_MOD', ModuleProxy::class);
+
+        $listener = new Definition(GlobalsMapListener::class, [array_merge_recursive($elements, $modules)]);
+        $listener->addTag('contao.hook', ['hook' => 'initializeSystem', 'priority' => 255]);
```
Toflar: Nice one :) A unit test should cover that this pass is registered before @dmolineus new hook register pass :)

src/DependencyInjection/Compiler/RegisterFragmentsPass.php
```diff
@@ -0,0 +1,172 @@
...
+use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
+use Symfony\Component\DependencyInjection\Container;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Registers Contao fragments in the registry.
+ *
+ * For custom fragment tags, just override this class and replace the process() method.
```
Toflar: Just `extend` this class, I guess :)

aschempp: Improved in 0e24990238a859115532b603e48f363859cdb614

src/DependencyInjection/Compiler/MapFragmentsToGlobalsPass.php
aschempp: Added in 0e24990238a859115532b603e48f363859cdb614

Toflar:
> Should it be renderStrategy? Or renderer (as the render() method argument name), or just strategy?

I'm fine with `renderStrategy`. I like explicit naming :)

aschempp:
@leo should be RTM

src/DependencyInjection/Compiler/MapFragmentsToGlobalsPass.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function process(ContainerBuilder $container): void
+    {
+        $elements = $this->getFragmentTags($container, ContentElementReference::TAG_NAME);
+        $elements = $this->createGlobalsMapForTag($elements, 'TL_CTE', ContentProxy::class);
+
+        $modules = $this->getFragmentTags($container, FrontendModuleReference::TAG_NAME);
+        $modules = $this->createGlobalsMapForTag($modules, 'FE_MOD', ModuleProxy::class);
+
+        $listener = new Definition(GlobalsMapListener::class, [array_merge_recursive($elements, $modules)]);
```
dmolineus: Imho `array_merge` should be enough here.

src/Fragment/Reference/FragmentReference.php
```diff
@@ -0,0 +1,58 @@
...
+    public function __construct($fragment, array $attributes = [], array $query = [])
+    {
+        parent::__construct($fragment, $attributes, $query);
+
+        if (!isset($this->attributes['scope'])) {
+            $this->attributes['scope'] = ContaoCoreBundle::SCOPE_FRONTEND;
+        }
+    }
+
+    public function isFrontend()
```
dmolineus: Return type shoud be `bool`.

src/Fragment/Reference/FragmentReference.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return $this->isScope(ContaoCoreBundle::SCOPE_FRONTEND);
+    }
+
+    public function setFrontend(): void
+    {
+        $this->attributes['scope'] = ContaoCoreBundle::SCOPE_FRONTEND;
+    }
+
+    public function isBackend()
```
dmolineus: Return type shoud be `bool`.

src/Fragment/Reference/FragmentReference.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        return $this->isScope(ContaoCoreBundle::SCOPE_BACKEND);
+    }
+
+    public function setBackend(): void
+    {
+        $this->attributes['scope'] = ContaoCoreBundle::SCOPE_BACKEND;
+    }
+
+    private function isScope(string $scope)
```
dmolineus: Return type shoud be `bool`.

agoat:
That looks good, even though I haven't been able to test it yet.

A question on the sidelines..
Not sure if this is the place and time to change it but - can we rename the `inColum` to `sections`? The layout has already custom sections (but in the module selections it's called column again). 
I think we should make this uniform (in the translation and in the code).

aschempp:
I like that idea. I have renamed `inColumn` to `section`

src/DependencyInjection/Compiler/RegisterFragmentsPass.php
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Registers the fragments.
+     *
+     * @param ContainerBuilder $container
+     * @param string           $tag
+     *
+     * @throws InvalidConfigurationException
+     */
+    protected function registerFragments(ContainerBuilder $container, string $tag): void
```
leofeyer: Do all of the methods have to be protected? We should make those methods private, which are unlikely to be required to be overwritable.

src/Fragment/FragmentConfig.php
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Constructor.
+     *
+     * @param string $controller
+     * @param string $renderer
+     * @param array  $options
+     *
+     * @see \Symfony\Component\HttpKernel\Fragment\FragmentHandler::render()
+     */
+    public function __construct(string $controller, string $renderer = 'inline', array $options = [])
```
leofeyer: Sorry for changing my mind, but `$renderer` is wrong. It suggests that one has to pass a `new FragmentRenderer()` object. We should use `$renderingStrategy` as used in Symfony.

aschempp: It's called `renderer` in Symfony! https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Fragment/FragmentHandler.php#L65

src/DependencyInjection/Compiler/RegisterFragmentsPass.php
aschempp: Yes they are intentionally protected because other people need to extend the compiler pass to register their own fragments. See the class description.

leofeyer: I have read the class description – see my comment. 😄 

> We should make those methods private, which are unlikely to be required to be overwritable

Are all methods likely to be overwritten?

src/Fragment/FragmentConfig.php
leofeyer: True.

src/DependencyInjection/Compiler/RegisterFragmentsPass.php
aschempp: Ah, yeah. They are all likely to be used/overwritten in my opinion. 

- `registerFragments` is the main entry point
- `getFragmentConfig` is necessary if you have custom options in your tag configuration
- `getControllerName` if your controllers work differently
- `addPreHandlers` to alter prehandler registration
- `getFragmentType` to alter suffix or similar stuff when generating the type

src/Resources/contao/elements/ContentProxy.php
```diff
@@ -21,31 +19,25 @@
...
-        // noop
-    }
+	/**
+	 * {@inheritdoc}
+	 */
+	public function generate()
+	{
+		$reference = new ContentElementReference($this->objModel, $this->strColumn);
+
+		if ('BE' === TL_MODE) {
```
leofeyer: This is new code, isn't it? So we should not use `TL_MODE` anymore.

src/Resources/contao/modules/ModuleProxy.php
```diff
@@ -21,31 +19,25 @@
...
-        // noop
-    }
+	/**
+	 * {@inheritdoc}
+	 */
+	public function generate()
+	{
+		$reference = new FrontendModuleReference($this->objModel, $this->strColumn);
+
+		if ('BE' === TL_MODE) {
```
leofeyer: Same here.

leofeyer:
Thank you @aschempp.

PR #1191, opened 2017-11-15
Fixed use of kernel service in assets packages pass

aschempp:
This will slightly change the behavior, we're always using a name generated from bundle class instead of using the extension alias if available. It should not make a difference for correctly configured bundles, except for the core bundle where it's even better, because previously it was `contao` and now it's `contao_core`.


leofeyer:
Thank you @aschempp.

PR #1204, opened 2017-11-27
Improved ModelArgumentResolver to find models based on type

aschempp:
I create a controller for a content element and added `public function __invoke(ContentModel $model)`. Strangely, that lead to no output. After some digging, I found the dev debug bar output that told me the controller argument could not automatically be resolved. And therefore the fragment was not rendered. In the current implementation, the argument name **must** be `$contentModel` (because `contentModel` is the attribute name). That's pretty unfortunate, so I extended the argument resolver to find an argument by type. It works for two cases:

 1. `ContentElement $contentModel` searches request argument by that name
 2. Converts the class `ContentModel` to `contentModel` and searches request argument by that name. It's a simple `lcfirst` on the class name (without namespace).

Some explanations for the additional changes:
 - The scope attribute must have our underline prefix (`_scope`) to match the one we already have in the regular controllers. Initially we agreed not to name it like that because it means one would need a `$_scope` argument to the controller function. However, I came to realize that the developer actually should use the `ScopeMatcher` and never know the attribute name at all.
 - I changed the namespace because we already have a `HttpKernel` and `ArgumentResolverInterface` is a subnamespace of `HttpKernel` in the Symfony Framework. One less folder in the bundle …
 - Contrary to the previous implementation, the argument resolver only works if a request is in Contao scope. I think that's a reasonable change though, as we don't want to boot the framework otherwise.


src/Resources/config/services.yml
```diff
@@ -9,9 +9,10 @@ services:
                 - ["setContainer", ["@service_container"]]
 
     contao.argument_resolver.model:
```
leofeyer: You also have to adjust the name to `contao.model_argument_resolver` and move the definition to the right place (services are ordered alphabetically).

src/HttpKernel/ModelArgumentResolver.php
```diff
@@ -73,11 +82,56 @@ public function resolve(Request $request, ArgumentMetadata $argument): \Generato
      */
     private function fetchModel(Request $request, ArgumentMetadata $argument): ?Model
     {
-        $id = $request->attributes->getInt($argument->getName());
+        $name = $this->findArgumentName($request, $argument);
+
+        if (null === $name) {
+            return null;
```
leofeyer: Shouldn't we throw an exception here?

leofeyer:
I'm fine with all the additional changes, however, I don't quite understand the use case for options 2 and 3. Can you elaborate?

src/Fragment/Reference/FragmentReference.php
```diff
@@ -24,8 +24,8 @@ public function __construct($fragment, array $attributes = [], array $query = []
     {
         parent::__construct($fragment, $attributes, $query);
 
-        if (!isset($this->attributes['scope'])) {
-            $this->attributes['scope'] = ContaoCoreBundle::SCOPE_FRONTEND;
```
Toflar: I agree with that change.

Toflar:
Any feedback on the use cases of 2 and 3, apparently you removed one? Why?

leofeyer:
Thank you @aschempp.

PR #1206, opened 2017-11-28
[RTM] Update to contao/image 0.3.5

ausi:


leofeyer:
Is this change necessary? Composer would update to version 0.3.5 automatically.

PR #1207, opened 2017-11-29
 [RFC] Allow fragments to render <> 200 responses

aschempp:
Replaces Contao fragment renderer with service decorating the Symfony fragment handler.

Otherwise a fragment, especially inline, is unable to redirect, e.g. if it processed a form.


src/Fragment/FragmentHandler.php
```diff
@@ -0,0 +1,134 @@
...
+     * @param bool                      $debug
+     */
+    public function __construct(ContainerInterface $renderers, BaseFragmentHandler $fragmentHandler, RequestStack $requestStack, FragmentRegistryInterface $fragmentRegistry, ContainerInterface $preHandlers, bool $debug = false)
+    {
+        $this->renderers = $renderers;
+        $this->fragmentHandler = $fragmentHandler;
+        $this->fragmentRegistry = $fragmentRegistry;
+        $this->preHandlers = $preHandlers;
+
+        parent::__construct($requestStack, [], $debug);
```
leofeyer: Any reason why this is not the first line in the method?

src/Fragment/FragmentHandler.php
```diff
@@ -0,0 +1,134 @@
...
+     * @param ContainerInterface        $renderers
+     * @param BaseFragmentHandler       $fragmentHandler
+     * @param RequestStack              $requestStack
+     * @param FragmentRegistryInterface $fragmentRegistry
+     * @param ContainerInterface        $preHandlers
+     * @param bool                      $debug
+     */
+    public function __construct(ContainerInterface $renderers, BaseFragmentHandler $fragmentHandler, RequestStack $requestStack, FragmentRegistryInterface $fragmentRegistry, ContainerInterface $preHandlers, bool $debug = false)
+    {
+        $this->renderers = $renderers;
```
leofeyer: Shouldn't we be using the `addRenderer()` method or pass the renderers to the parent constructor, which will do so?

src/Fragment/FragmentHandler.php
aschempp: no specific reason, it does not matter. And Symfony does the same in the `LazyLoadingFragmentHandler`

src/Fragment/FragmentHandler.php
aschempp: No, then it would not be lazy. See the `LazyLoadingFragmentHandler`

leofeyer: 👍

src/Fragment/FragmentHandler.php
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * @param ContainerInterface        $renderers
+     * @param BaseFragmentHandler       $fragmentHandler
+     * @param RequestStack              $requestStack
+     * @param FragmentRegistryInterface $fragmentRegistry
+     * @param ContainerInterface        $preHandlers
+     * @param bool                      $debug
+     */
+    public function __construct(ContainerInterface $renderers, BaseFragmentHandler $fragmentHandler, RequestStack $requestStack, FragmentRegistryInterface $fragmentRegistry, ContainerInterface $preHandlers, bool $debug = false)
```
leofeyer: We probably should not call the variable `$renderers`, because `$renderers` is an array in the parent class.

src/Resources/config/services.yml
```diff
@@ -101,12 +101,16 @@ services:
     contao.fragment.registry:
         class: Contao\CoreBundle\Fragment\FragmentRegistry
 
-    contao.fragment.renderer:
-        class: Contao\CoreBundle\Fragment\FragmentRenderer
+    contao.fragment.handler:
+        class: Contao\CoreBundle\Fragment\FragmentHandler
+        decorates: fragment.handler
         arguments:
+            - ""
```
leofeyer: The first argument has to implement `ContainerInterface`.

aschempp: It is set by the compiler pass (see `FragmentRendererPass`)

PR #1227, opened 2017-12-11
Use ocramius/package-versions

aschempp:
Instead of manually parsing the `installed.json`.


src/DependencyInjection/Compiler/AddPackagesPass.php
```diff
@@ -20,58 +21,17 @@
...
-     * @return array
-     */
-    private function getVersions(array $json): array
-    {
-        $packages = [];
-
-        foreach ($json as $package) {
-            if (isset($package['version'])) {
-                if (isset($package['extra']['branch-alias'][$package['version']])) {
-                    $packages[$package['name']] = $package['extra']['branch-alias'][$package['version']];
```
leofeyer: Does the `Versions` class handle branch aliases, too? Otherwise this is a BC break.

leofeyer:
Things could be even simpler if we used https://github.com/Jean85/pretty-package-versions 😄

src/DependencyInjection/Compiler/AddPackagesPass.php
aschempp: It just uses what the `composer.lock` file contains. As far as I can tell, that does handle the branch aliases.

aschempp:
Nice find with `pretty-package-versions`. Do we want to rely on that package (which does not seem widely used)?

fritzmg:
Just curios: what advantage does the usage of `package-versions` provide, rather than using `->getParameter('kernel.packages')`?

aschempp:
@fritzmg Advantages I can see:
 1. The package is already present in most cases, so we'd track the versions multiple times
 2. The versions are stored on Composer Update and not on container build
 3. We're not parsing the `installed.json` and *assuming a location* of the file, we're actually using the composer interface (or package-versions is).

leofeyer:
> Nice find with `pretty-package-versions`. Do we want to rely on that package (which does not seem widely used)?

It has [over 170,000 downloads](https://packagist.org/packages/Jean85/pretty-package-versions), so it is not directly "not widely used". But we might as well have `substr($coreVersion, 0, strpos($coreVersion, '@'));` 20 times in the code. 😄

fritzmg:
@aschempp I see. However after this PR, will you still be able to determine the Contao version via `->getParameter('kernel.packages')`? I am using that in an extension of mine. Though I'd be fine with using `jean85/pretty-package-versions` as well.

leofeyer:
@aschempp I think we can replace

```php
substr($coreVersion, 0, strpos($coreVersion, '@'));
```

with

```php
strtok($coreVersion, "@");
```

aschempp:
It is still there for BC reasons, but the new recommended way would be to use the class.

leofeyer:
I have benchmarked the replacement methods:

* `substr()`: 0.076312065124512
* `strtok()`: 0.070075988769531
* `strstr()`: 0.05613899230957
* `explode()`: 0.12011098861694

I'm afraid `strstr($coreVersion, "@", true);` is faster than `strtok()`. 😄

leofeyer:
There was another problem: `version_compare()` did not like versions with a leading "v". I have therefore added a `PackageUtil::getVersion()` method, which prepares the version number for our use cases.

leofeyer:
Thank you @aschempp.

Toflar:
Given the following example:

```php
class MyService
{
    public function doThis()
    {
        if (version_compare(PackageUtil::getVersion('contao/core-bundle'), '4.4.0', '>=')) {
            // do this for Contao 4.4+
        } else {
            // do this for Contao 4.3
        }
    }
}
```

How am I supposed to unit test that both ways work?

ausi:
@Toflar by injecting `PackageUtil` instead of calling the method statically?

aschempp:
> How am I supposed to unit test that both ways work?

Probably by running the test with both dependencies?

Toflar:
> @Toflar by injecting PackageUtil instead of calling the method statically?

Exactly, but there's no service for that. That's why I asked :)

> Probably by running the test with both dependencies?

And how do I mock them?

leofeyer:
We could make this a service, however, we would no longer be able to use it outside the application then (e.g. in a Composer plugin or the Contao Manager). But those should use the `Versions` class directly I guess, shouldn't they @aschempp?

PR #1232, opened 2017-12-12
Security Bugfixes

aschempp:
Things I found while reviewing Symfony security in Contao 4.5

**ToDo:**
- [x] Deprecate use of `tl_session` table
- [x] Remove the `tl_session` table
- [x] Remove leftover security classes from Contao 4.0-4.4
- [x] `BackendSwitch` still sets and uses session cookies
- [x] `BackendPreview` still sets and uses session cookies
- [x] `RebuildIndex` (search index maintenance screen) uses session cookies to log in the frontend user for indexing
- [x] `FrontendTemplate::hasAuthenticatedBackendUser` checks the session cookie
- [x] the `contao.security.disable_ip_check` parameter is still used but cannot have any useful functionality
- [x] ~~mark `contao.security.disable_ip_check` parameter as deprecated~~ remove the `contao.security.disable_ip_check` parameter
- [x] `UserSessionListener` (for header replay) still checks the session cookies
- [x] `BypassMaintenanceListener` still checks the session cookies
- [x] we might want to store `FE_PREVIEW` in the session (instead of a cookie) or even in the security token attributes
- [x] we should have a *central* token checker service instead of unserializing the session tokens all over the old code
- [ ] `AuthenticationFailureHandler` uses the scope matcher to set flash message, but `ContaoAuthenticationProvider` uses the firewall name
- [x] ~~The `session` service does not need to be injected if we already have the request (e.g. in `FrontendPreviewAuthenticator`)~~
- [x] The `FrontendPreviewAuthenticator` removes a frontend token from the session if the token was not authenticated. But according to `UsernamePasswordToken` that can never happen?
- [x] Use a different token than `UsernamePasswordToken` to authenticate a frontend user
- [x] ~~Possibly implement `User::eraseCredentials`~~
- [x] Check that `RebuildIndex` only allows users that are in `amg` permissions (see https://github.com/contao/core-bundle/issues/1237)
- [x] Consider reloading after `contao/preview?user=xxx` and remove the user parameter
- [x] Fix unit tests
- [x] Test contao.security.legacy_password_encoder
- [x] Test contao.security.logout_handler
- [x] Test contao.security.logout_success_handler
- [x] Test contao.security.token_checker
- [x] Test contao.security.user_checker
- [ ] It was currently possible to disable `autologin` in the localconfig


src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -76,11 +76,12 @@ public function authenticateFrontendUser($username = null): void
...
-            return;
-        }
-
-        if (null === $username || !$request->hasSession()) {
+        if (null === $username
+            || null === $request
+            || !$request->hasSession()
+            || null === $this->tokenStorage->getToken()
+            || !$this->tokenStorage->getToken()->isAuthenticated()
+        ) {
```
bytehead: multicondition `if`s are shitty to test.

bytehead:
👍  for removing, but I've never worked with the table, so I have no clue about breaking some extensions.

src/Security/Authentication/FrontendPreviewAuthenticator.php
leofeyer: Oh yes. Just fire up the PhpStorm EA Extended Plugin and it will tell you. 😄 

I am going to address the null pointer issue in the next couple of days.

aschempp: 🤔 good point, never thought of that. Should that be a general rule @contao/developers ?

Toflar: No idea what rule you're talking about so ¯\_(ツ)_/¯

Toflar:
Using the `tl_session` hack was basically the only way to authorize a user manually up until now. A lot of extensions hack this somehow which has always been stupid and should now be history 😄 I think it would be great if we could add that to `UPGRADE.md` anyway. It never was officially an API (so up until now it was never possible without hacks) but still, documenting what you should do instead as of 4.5 would be very cool (also for myself personally).

aschempp:
> We should discuss whether we want to keep the tl_session table. 

I think it depends on if we keep BC on the security stuff. Either we keep the table and model, or we remove both and also remove stuff like `ContaoToken` and other outdated security classes.

Toflar:
If we can keep BC, we should.

aschempp:
> If we can keep BC, we should.

Well it would be BC in the sense that classes and tables are still there. They obviously won't work anymore…

Toflar:
If it doesn't work, it's not BC by definition.

src/Security/Authentication/FrontendPreviewAuthenticator.php
aschempp: A general rule to not have too many conditions in an if statement, like splitting between "condition types", in this case token check and "others". I see @bytehead's point in unit test coverage which only works for separate lines.

bytehead: And it's not only about the tests. It's also bad to read and there's probably no enormous win of performance as well.

src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -96,17 +100,37 @@ public function authenticateFrontendUser($username = null): void
                 );
             }
 
-            return;
+            return false;
         }
 
-        $token = new FrontendPreviewToken($user);
+        $allowedGroups = \StringUtil::deserialize($backendUser->amg, true);
```
bytehead: `\StringUtil` in new code?

aschempp:
> What's the intention of a separate token for the preview?

Using a `UsernamePasswordToken` to authenticate a user without password seems kind of wrong. Otherwise there would not be a need for different token implementations at all. Like you would use a `FacebookToken` for facebook authentication and not just authenticate a `UsernamePasswordToken` with the facebook data, right?

src/Security/TokenChecker.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Checks if an authenticated token exists in the session.
+     *
+     * @param string $sessionKey
+     *
+     * @return bool
+     */
+    public function isAuthenticated(string $sessionKey): bool
+    {
+        return null !== $this->getToken($sessionKey);
```
bytehead: Why don't you use the `isAuthenticated` method on the token?

src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -94,17 +100,37 @@ public function authenticateFrontendUser($username = null): void
                 );
             }
 
-            return;
+            return false;
         }
 
-        $token = new UsernamePasswordToken($user, null, $providerKey, (array) $user->getRoles());
+        $allowedGroups = \StringUtil::deserialize($backendUser->amg, true);
```
bytehead: CS

src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -94,17 +100,37 @@ public function authenticateFrontendUser($username = null): void
...
+        $frontendGroups = \StringUtil::deserialize($frontendUser->groups, true);
```
bytehead: CS

src/Resources/contao/classes/BackendUser.php
```diff
@@ -136,32 +136,20 @@ public static function getInstance()
...
+			return static::loadUserByUsername($objToken->getUser()->getUsername());
 		}
 
-		/** @var SessionInterface $session */
-		$session = \System::getContainer()->get('session');
+		// Try to load possibly authenticated FrontendUser from session
+		$strUser = \System::getContainer()->get('contao.security.token_checker')->getUsername(self::SECURITY_SESSION_KEY);
 
-		if (!$session->has(self::SECURITY_SESSION_KEY))
+		if (null !== $strUser)
```
bytehead: CS

src/Resources/contao/classes/FrontendUser.php
```diff
@@ -93,32 +93,20 @@ public static function getInstance()
...
 
 		// Try to load possibly authenticated FrontendUser from session
-		if (!($token = unserialize($session->get(self::SECURITY_SESSION_KEY))) instanceof TokenInterface)
-		{
-			return parent::getInstance();
-		}
+		$strUser = \System::getContainer()->get('contao.security.token_checker')->getUsername(self::SECURITY_SESSION_KEY);
 
-		if ($token->isAuthenticated())
+		if (null !== $strUser)
```
bytehead: CS

src/Security/TokenChecker.php
aschempp: The `getToken` method already does: https://github.com/contao/core-bundle/pull/1232/files/df31357d729b02ffd8d54eac331a22c29efbb83e#diff-c81a5fe77c0d656a71ce6b60c6bf87b5R80

src/Security/Authentication/FrontendPreviewAuthenticator.php
aschempp: what CS?

src/Security/Authentication/FrontendPreviewAuthenticator.php
aschempp: what CS?

src/Security/TokenChecker.php
bytehead: Ah, missed that!

src/Security/Authentication/FrontendPreviewAuthenticator.php
bytehead: `\StringUtil` in new code.  
Shouldn't it be `Use Contao\StringUtil;` and then `StringUtil:: deserialize` ?

src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -80,7 +80,7 @@ protected function loadInternal(array $mergedConfig, ContainerBuilder $container
         $container->setParameter('contao.image.target_dir', $mergedConfig['image']['target_dir']);
         $container->setParameter('contao.image.valid_extensions', $mergedConfig['image']['valid_extensions']);
         $container->setParameter('contao.image.imagine_options', $mergedConfig['image']['imagine_options']);
-        $container->setParameter('contao.security.disable_ip_check', $mergedConfig['security']['disable_ip_check']);
+        $container->setParameter('contao.security.disable_ip_check', true);
```
leofeyer: Why? The default was `false` so far.

src/Resources/contao/classes/BackendUser.php
```diff
@@ -37,43 +37,43 @@ class BackendUser extends User
 	 * Edit page flag
 	 * @var string
 	 */
-	const CAN_EDIT_PAGE = 1;
+	public const CAN_EDIT_PAGE = 1;
```
leofeyer: Not in the old code though. 😄

src/Resources/contao/languages/sr/tl_settings.xlf
```diff
@@ -126,7 +126,7 @@
         <target>Учини IP адресе анонимним</target>
       </trans-unit>
       <trans-unit id="tl_settings.privacyAnonymizeIp.1">
-        <source>Anonymize any IP address that is stored in the database, except in the &lt;em&gt;tl_session&lt;/em&gt; table (the IP address is bound to the session for security reasons).</source>
+        <source>Anonymize any IP address that is stored in the database.</source>
```
leofeyer: You have to change `en/tl_settings.xlf` not `sr/tl_settings.xlf`. 😄

src/Resources/contao/library/Contao/System.php
```diff
@@ -716,27 +716,21 @@ public static function getFormattedNumber($varNumber, $intDecimals=2)
...
-
-		if (!$container->getParameter('contao.security.disable_ip_check'))
-		{
-			$strHash .= \Environment::get('ip');
-		}
-
-		$strHash .= $strCookie;
-
-		return sha1($strHash);
+		return sha1($session->getId().$strCookie);
```
leofeyer: You have to start the session for this to work.

src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -68,24 +61,37 @@ public function __construct(RequestStack $requestStack, SessionInterface $sessio
...
-            return;
+        $backendUser = $token->getUser();
+
+        // No back end user logged in or session not started
+        if (!$backendUser instanceof BackendUser || !$this->session->isStarted()) {
+            return false;
+        }
+
+        // Back end user does not have permission to log in front end users
+        if (!$backendUser->isAdmin && (!\is_array($backendUser->amg) && !empty($backendUser->amg))) {
```
leofeyer: You should check `empty()` before `is_array()`.

leofeyer:
A lot of reasonable and necessary changes. Thank you @aschempp.

leofeyer:
> mark `contao.security.disable_ip_check` parameter as deprecated

I'd probably remove it, too. Should not cause any problems.

> we might want to store `FE_PREVIEW` in the session (instead of a cookie) or even in the security token attributes

Oh yes.

> `AuthenticationFailureHandler` uses the scope matcher to set flash message, but `ContaoAuthenticationProvider` uses the firewall name

We should unify this.

> Possibly implement User::eraseCredentials

What would this method do?

> Check that RebuildIndex only allows users that are in amg permissions

It seems that you have already implemented this? If not, please do it. 😄 

> Consider reloading after contao/preview?user=xxx and remove the user parameter

We are using the username instead of the user ID here on purpose, so one can manually set a user in the URL. This feature was explicitly requested in the original PR.

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -250,12 +250,9 @@ public static function addToUrl($strRequest, $blnIgnoreParams=false, $arrUnset=a
 	 */
 	public function hasAuthenticatedBackendUser()
 	{
-		if (!isset($_COOKIE['BE_USER_AUTH']))
-		{
-			return false;
-		}
+		$objTokenChecker = \System::getContainer()->get('contao.security.token_checker');
```
Toflar: Why is this accessing the global `System` and the global `BackendUser`. I know it's in old code but we don't use these anymore.

leofeyer: Again, this is a "do not fix just one of 1,000 occurrences" case. 😄

src/DependencyInjection/ContaoCoreExtension.php
aschempp: Because we don't do IP checks anymore.

src/Resources/contao/classes/FrontendTemplate.php
leofeyer: We are trying to keep the legacy code untouched, so it is easier to synchronize `contao/core-bundle` with `contao/core`. After June 2019, we can change all occurrences.

src/Resources/contao/languages/sr/tl_settings.xlf
aschempp: Fixed in d518a871149eb38d5d66fec432d25ea6869ddcaf

src/Resources/contao/classes/BackendUser.php
aschempp: Fixed in d116de00ceb106bb425c771c93ed1925dbb1b59c

src/Security/Authentication/FrontendPreviewAuthenticator.php
aschempp: Fixed in 6d55b1517aa70f763b07b65bb8ca5e0b9b09e041

src/Resources/contao/classes/FrontendTemplate.php
aschempp: Because it's legacy code …

aschempp:
> > Check that RebuildIndex only allows users that are in amg permissions

> It seems that you have already implemented this? If not, please do it. 😄

It is implemented in the switch method automatically, but I have not checked that the dropdown menu in the maintenance screen is limited respectively. Also, the same must be implemented for Contao 4.4

src/Security/Authentication/FrontendPreviewAuthenticator.php
leofeyer: Don't worry about those things, I will fix them all before merging. Focus on the "real" issues. 😄

src/DependencyInjection/ContaoCoreExtension.php
leofeyer: Yeah, so just remove the parameter.

src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -68,35 +61,40 @@ public function __construct(RequestStack $requestStack, SessionInterface $sessio
...
+        if (!$backendUser->isAdmin && (empty($backendUser->amg) || !\is_array($backendUser->amg))) {
+            return false;
         }
 
         try {
-            $user = $this->userProvider->loadUserByUsername($username);
+            $frontendUser = $this->userProvider->loadUserByUsername($username);
+
+            if (!$frontendUser instanceof FrontendUser) {
+                throw new UsernameNotFoundException('User is not a front end user');
```
leofeyer: This is redundant, because `FrontendUserProvider::loadUserByUsername()` already executes the `instanceof` check.

src/Security/Authentication/FrontendPreviewAuthenticator.php
```diff
@@ -89,30 +90,18 @@ public function authenticateFrontendUser(string $username): bool
...
-            if (null !== $this->logger) {
-                $this->logger->info(
-                    sprintf('Could not find a front end user with the username "%s"', $username),
-                    ['contao' => new ContaoContext(__METHOD__, ContaoContext::ACCESS)]
-                );
-            }
-
-            return false;
-        }
+        if (null === $username) {
```
leofeyer: I don't like this. Shouldn't this logic be in a separate method? It feels weird that the method requires a username but also accepts null.

src/Security/Authentication/AuthenticationFailureHandler.php
```diff
@@ -91,7 +91,7 @@ public function onAuthenticationFailure(Request $request, AuthenticationExceptio
     private function determineTargetUrl(Request $request): string
     {
         if ($this->scopeMatcher->isBackendRequest($request)) {
-            return $request->getRequestUri();
+            return 'contao_backend_login';
```
leofeyer: This looks wrong to me. The method should return a path, shouldn't it?

Toflar: HttpUtils will use the UrlGenerator if it's a route name.

leofeyer:
Thank you @aschempp.

PR #1233, opened 2017-12-13
remove frameborder attribute from media iframes

fritzmg:
Having `frameborder="0"` on iframes produces the following validation error:
```
The frameborder attribute on the iframe element is obsolete. Use CSS instead.
```
See https://validator.w3.org/nu/?doc=https%3A%2F%2Fdemo.contao.org%2Fen%2Fmedia-elements.html

_Note:_ there are a few places in the back end where `frameborder="0"` is used on iframes as well (`be_preview.html5`, `core.js`). Those are not changed with this PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
         "contao-components/colorbox": "^1.6",
         "contao-components/colorpicker": "^1.4.0.1",
         "contao-components/compass": "^0.12",
-        "contao-components/contao": "^7.1",
+        "contao-components/contao": "^7.1.4",
         "contao-components/datepicker": "^2.2.0.2",
         "contao-components/dropzone": "^4.1.1",
         "contao-components/highlight": "^8.9",
--- a/src/Resources/contao/templates/elements/ce_vimeo.html5
+++ b/src/Resources/contao/templates/elements/ce_vimeo.html5
@@ -2,6 +2,6 @@
 
 <?php $this->block('content'); ?>
 
-  <iframe<?= $this->size ?> src="<?= $this->src ?>" frameborder="0" allowfullscreen></iframe>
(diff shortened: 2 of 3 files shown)


leofeyer:
The problem is that it does not render properly without the attribute. And that's more important than the validator.

ausi:
Then `frameborder="0"` should probably be replaced with `style="border:0"`.

fritzmg:
> Then frameborder="0" should probably be replaced with style="border:0".

But you also don't add `style="margin:0"` for every `<figure>` for example. Why should it be done for `<iframe>`?

ausi:
You are right @fritzmg. The reset.css is the better place for that.

leofeyer:
It's not about `border: 0`. I recall that some browsers – presumably IE6 to IE? – always rendered the iframe border without the attribute.

fritzmg:
I see. But hasn't the minimum IE version for the Contao CSS framework been raised to IE 8 anyway?

leofeyer:
Probably even IE11 meanwhile. Can you do a quick test which browsers might be affected?

fritzmg:
Not really :grin: I can only test IE11 (and EDGE). I don't have the facilities (VMs) to test older IE versions right now (since we do not support anything lower than IE11 either).

IE11 definitely has no problem with
```css
iframe {
    border: 0;
}
```

ausi:
I can do the test.

ausi:
Works in IE9+  
Doesn’t work in IE8-

leofeyer:
Thank you @ausi. I guess we can drop the attribute then.

leofeyer:
Thank you @fritzmg.

PR #1238, opened 2017-12-15
[RFC] Security fixes (step 2)

aschempp:
More things I found in the new security implementation.

**Tasks:**
- [x] Correctly trigger `postAuthenticate` hook on every page load (previously in `User::authenticate`)
- [x] Log if a user was not found (previously in `User::login`)
- [x] Make sure the `TokenChecker` only validates for front end and back end users
- [x] Make sure `User::authenticate` still works to keep BC
- [x] Use Symfony security (authentication utils) to get the last authentication error message
- [x] Lock user and send email on the third invalid login not when attempting a fourth login
- [x] Don't use the `security.interactive_login` event to trigger `postLogin` hook. It is executed for all firewalls and all listeners regardless of Contao configuration
- [x] Store the user object in authentication exceptions
- [x] Redirect when frontend login and logout actions are not handled by security
- [x] `AuthenticationFailureHandler` uses the scope matcher to set flash message, but `ContaoAuthenticationProvider` uses the firewall name
- [x] ~~It was currently possible to disable autologin in the localconfig~~ No longer needed as it can be done through Symfony configuration
- [x] The login screen language is switched based on the user of a failed login. Seems like a security issue to me.
- [x] Back end logout does not need a success handler (see https://github.com/contao/manager-bundle/commit/a0f96f13a34419bf0068fe4618444c0e961edda3)
- [x] Currently overrides the application-wide security authentication provider for all firewalls
   - ~~**Option 1:** Override the application-wide authentication provider, which means all `form_login` firewalls will use our provider and can't authenticate on something else (like a sub-page of the application).~~ *Not an acceptable option for me, as it would prevent regular Symfony applications from using other firewalls.*
    - ~~**Option 2:** Keep the code as is (in this PR) and decorate the firewall-specific authentication provider. That means the authentication won't work with Contao if the firewall names do not exactly match (which kinda makes the whole security configuration obsolete, we could also just hardcode everything).~~
    - ~~**Option 3:** Move everything from `ContaoAuthenticationProvider` to login success and failure handlers except the `checkCredentials` hook. Override the firewall-specific authentication provider, which means everything should work except the `checkCredentials` hook for a firewall with a different name.~~
    - ~~**Option 4:** Create our own firewall listener `contao_login` instead of `form_login` which then users our `ContaoAuthenticationProvider`. Kinda hardcore but it would also allow us to remove firewall configuration that does not make sense for Contao (like overriding the `username` and `password` fields).~~ *Not a valid option, as it would only work for our `contao_login` listener. You could not have an LDAP authentication for Contao users.*
    - **Option 5:** Move everything from `ContaoAuthenticationProvider` to login success and failure handlers except the `checkCredentials` hook and user locking. Create our own firewall listener `contao_login` instead of `form_login` which then uses our `ContaoAuthenticationProvider`.
- [x] Use authentication utils to get username, it does additional checks
- [x] ~~Possibly move legacy user configuration (`loginCount`, `lockPeriod`) to the bundle configuration~~ We'll use options for the authentication provider but currently don't need a bundle config.
- [x] ~~`$_SESSION['LAST_PAGE_VISITED']` could be replaced with `TargetPathTrait`~~ *Only works for success handler but not for failed handler*
- [x] Frontend login redirects to wrong page if the template is not updated with new target field (see https://github.com/contao/manager-bundle/commit/15fd145e654154cc4bf2fc15ed1875cc9fb77515)
- [x] Possibly reload the page after logging the user out due to inactivity
- [x] The user is still logged in after calling `/contao/login`
- [ ] ~Logout and~~ login might not validate the CSRF token? *logout is not a POST request*
- [x] Use the `LogoutUrlGenerator` to generate logout link in back end and possibly front end
- [x] Login module does not redirect back to the same page on logout
- [x] ~~The front end user can potentially see that we're Contao due to the new public post urls~~
- [x] `tl_member.autologin` is no longer needed for the cookie
- [x] `tl_settings.autologin` is no longer useable to define the autologin time
- [x] ~~Discuss the session handling on login and logout. Login = `SessionAuthenticationStrategy`, Logout = `invalidate_session` option~~
- [x] Use the existing variable to handle session expiration in `refreshUser`
- [x] Failed redirect is not the same as success redirect after login


src/Security/TokenChecker.php
```diff
@@ -25,55 +26,72 @@ class TokenChecker
...
+    private $trustResolveer;
+
+    /**
+     * @param SessionInterface                     $session
+     * @param AuthenticationTrustResolverInterface $trustResolveer
+     */
+    public function __construct(SessionInterface $session, AuthenticationTrustResolverInterface $trustResolveer)
     {
         $this->session = $session;
+        $this->trustResolveer = $trustResolveer;
```
leofeyer: I think you mean `$trustResolver`?

src/Security/TokenChecker.php
```diff
@@ -25,55 +26,72 @@ class TokenChecker
...
+        return $this->getUsername(BackendUser::class);
+    }
+
+    /**
+     * Tells whether the front end preview can show unpublished fragments.
      *
      * @return bool
      */
-    public function isPreviewMode(string $sessionKey): bool
+    public function showUnpublished(): bool
```
leofeyer: The naming is wrong. Methods returning booleans have to start with `is`, `has`, `should` etc.

src/Security/TokenChecker.php
```diff
@@ -97,6 +115,28 @@ private function getToken(string $sessionKey): ?TokenInterface
...
+    /**
+     * Gets the username of a token in the session.
+     *
+     * @param string $userClass
+     *
+     * @return string|null
+     */
+    private function getUsername(string $userClass): ?string
+    {
+        $token = $this->getToken(\constant($userClass.'::SECURITY_SESSION_KEY'));
```
leofeyer: This implies that `$userClass` always has a `SECURITY_SESSION_KEY` constant. It would be better (thinking of unit tests) to pass the session key as second argument.

src/Security/User/ContaoUserProvider.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * @var string
+     */
+    private $userClass;
+
+    /**
+     * @param ContaoFrameworkInterface $framework
+     * @param string                   $userClass
+     */
+    public function __construct(ContaoFrameworkInterface $framework, string $userClass)
```
leofeyer: I don't like this at all. There are only two types of Contao users and it should not be possible to provide anything but their two classes.

Having a back end and a front end user was fine – why did you change it?

bytehead: I don't get that as well.

src/Security/TokenChecker.php
aschempp: what would you use instead? `shouldShowUnpublished`?

src/Security/User/ContaoUserProvider.php
aschempp: Because both classes contain exactly the same code except for the class to check for. Remember the discussion we had with the AssetContext? 😉

aschempp:
@contao/developers please see the major open task and let me know your opinion (Option 1 - 4).

src/Security/TokenChecker.php
leofeyer: `isPreviewMode()` was fine with me. Similar to `Model::isPreviewMode()`.

src/Security/User/ContaoUserProvider.php
leofeyer: Except you explicitly can have multiple asset contexts but you cannot have another user than back end user and front end user.

Code duplication can also be reduced by an abstract parent class or a trait.

leofeyer:
Option 4 seems like a total overkill to me. Why don't we just write our own security component then? 😄 

Selecting one of the other options depends on which use cases we want to support.

 * Is there ever the need to add a third firewall with a `form_login` that does not use back end and front end users? If so, we should not chose option 1.

* Does it ever make sense to run our success and failure handlers in any firewall but `contao_backend` and `contao_frontend`? I don't think so, so we also should not chose option 3.

This kind of leaves us with option 2, doesn't it?

README.md
```diff
@@ -156,7 +155,7 @@ security:
             form_login:
                 login_path: contao_frontend_login
                 check_path: contao_frontend_login
-                default_target_path: contao_index
+                default_target_path: /
```
leofeyer: What's the point of changing this and returning a redirect route in the controller instead?

src/Resources/contao/classes/Frontend.php
```diff
@@ -500,20 +500,38 @@ protected function jumpToOrReload($intId, $strParams=null, $strForceLang=null)
...
+
+		if ($strCookie == 'BE_USER_AUTH' && $objTokenChecker->hasBackendUser())
+		{
+			// Disable the cache if a back end user is logged in
+			if (TL_MODE == 'FE')
+			{
+				$_SESSION['DISABLE_CACHE'] = true;
+
+				// Always return false if we are not in preview mode (show hidden elements)
+				if (!$objTokenChecker->showUnpublished())
```
leofeyer: This is a perfect example why the method should be named `isPreviewMode()`. Because the comment says "return false if we are not in preview mode" and then the code would say the same.

src/Resources/contao/controllers/BackendIndex.php
```diff
@@ -71,4 +73,32 @@ public function run()
...
+	 */
+	private function getLoginMessages()
+	{
+		$strMessages = \Message::generateUnwrapped();
+
+		$exception = \System::getContainer()->get('security.authentication_utils')->getLastAuthenticationError();
+
+		if ($exception instanceof LockedException)
+		{
+			$strMessages .= '<p class="tl_error">' . sprintf($GLOBALS['TL_LANG']['ERR']['accountLocked'], $exception->getLockedMinutes()) . '</p>';
```
leofeyer: Why not `\Message::addError()`?

src/Resources/contao/modules/ModuleLogin.php
```diff
@@ -111,13 +106,17 @@ protected function compile()
 
 		if (\System::getContainer()->get('session')->isStarted())
```
leofeyer: If you are no longer using the flash bag, you don't need to verify that the session is started.

src/Security/LogoutHandler.php
```diff
@@ -79,6 +79,8 @@ private function triggerPostLogoutHook(User $user): void
 
         @trigger_error('Using the postLogout hook has been deprecated and will no longer work in Contao 5.0. Use the contao.post_logout event instead.', E_USER_DEPRECATED);
 
+        $GLOBALS['TL_USERNAME'] = $user->getUsername();
```
leofeyer: I don't think so Tim.

src/Security/LogoutSuccessHandler.php
```diff
@@ -51,14 +45,10 @@ public function onLogoutSuccess(Request $request): RedirectResponse
...
-        }
-
-        if ($this->scopeMatcher->isBackendRequest($request)) {
-            return new RedirectResponse($this->router->generate('contao_backend_login'));
+        if (null !== $session && $session->has('_contao_logout_target')) {
+            return $this->httpUtils->createRedirectResponse($request, $session->get('_contao_logout_target'));
         }
 
-        return new RedirectResponse($this->router->generate('contao_root'));
+        return $this->httpUtils->createRedirectResponse($request, 'contao_root');
```
bytehead: Backend logout to `contao_root` seems totally wrong to me.

bytehead:
> Is there ever the need to add a third firewall with a form_login that does not use back end and front end users? If so, we should not choose option 1.

Yes there is in probably every symfony application which is not a contao standalone.

src/Security/LogoutHandler.php
aschempp: This was previously available (https://github.com/contao/core-bundle/blob/4.4/src/Resources/contao/library/Contao/User.php#L683), there's no reason to remove it if we want to keep BC with the hook.

src/Resources/contao/controllers/BackendIndex.php
aschempp: it would also render other error messages. The only advantage would be the "automatic" markup, right?

src/Resources/contao/classes/Frontend.php
aschempp: There is a distinction between *preview mode* (using the frontend preview with frames) and *showing unpublished items*.

src/Resources/contao/modules/ModuleLogin.php
aschempp: Removed in 3ca51292380adaacf625919362b34c6d017fa398

src/Security/LogoutSuccessHandler.php
aschempp: The backend logout does not need a custom success handler, this one is only for the frontend. We know the redirect target so we can simply use the configuration (see https://github.com/contao/manager-bundle/commit/a0f96f13a34419bf0068fe4618444c0e961edda3):

README.md
aschempp: The `default_target_path` is used if no other redirect target can be determined (see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Http/Authentication/DefaultAuthenticationSuccessHandler.php#L125). This is the case if the `mod_login` template is not adjusted and the `_target_path` hidden field is not submitted. Regardless of that, the target **path** can never be a route, so the configuration was incorrect in the first place.

src/Security/User/ContaoUserProvider.php
aschempp: I remember having the exact same discussion in the assets pull request. We can do parent classes as well, but that does not make a difference as you could write a third child class the same way you could register a third service with a different argument. Both ways are technically perfectly fine, I just don't see the need to write three files instead of one and use our beloved dependency injection. But I can change that if you insist…

src/Security/TokenChecker.php
aschempp: I have added a check in be0c6d4de2cafccf7bef134d31f25e8c20f4f0ae. I don't think a second argument would be good, as you could then misconfigure it (mix a `FrontendUser` class with a backend user security session key).

src/Security/TokenChecker.php
aschempp: Good catch, fixed in be0c6d4de2cafccf7bef134d31f25e8c20f4f0ae

src/Security/TokenChecker.php
aschempp: Just posting here again for reference: 
> There is a distinction between preview mode (using the frontend preview with frames) and showing unpublished items. The FrontendPreviewToken is used to login a temporary user, but the unpublished elements can still be hidden.

aschempp:
I just thought about this again in my sleep, and have reduced the *options* to 2 and 3. As @bytehead pointed out, Option 1 is not valid because the `core-bundle` could potentially not be used in a regular Symfony application anymore. Option 4 is not valid because it would not work for any other firewall listeners (like LDAP).

> Does it ever make sense to run our success and failure handlers in any firewall but contao_backend and contao_frontend? I don't think so, so we also should not choose option 3.

I think that is a little bit confusing. The *success handler* and *failure handlers* are defined per **firewall listener**, thinks like the *user provider* *user checker* are defined per **firewall**. Currently it is only designed to work if all components play together. Which means if you add a new **firewall listener** to the backend for LDAP authentication, you would need to override the *authentication provider* service of the listener for it to work. However, our *authentication provider* extends the form login, so you could not use that.

I need to further dig this, as it's not really working like this at all atm. I do have a new idea though 😄

src/Resources/contao/controllers/BackendIndex.php
leofeyer: If there are other error messages, they should probably be rendered, don't you think?

src/Resources/contao/classes/Frontend.php
leofeyer: That's probably true, however, we have not been making this distinction so far. We should rename the `Model::isPreviewMode()` method, too, then.

Still you have to use one of the prefixes. Maybe `canShowUnpublished()`?

src/Security/User/ContaoUserProvider.php
leofeyer: I don't insist.

src/Resources/contao/controllers/BackendIndex.php
aschempp: I'm not really sure, because it's the backend login screen and you're probably not supposed to see them if you're not logged in? Though that has been the case already up until now …

src/Resources/contao/classes/Frontend.php
aschempp: hah, I didn't know about that `Model` method. I guess we can argue that the preview mode is only unpublished elements, otherwise there is no difference between preview frames and just opening the page itself. I have renamed the method in 51220735aaea2435787580da126a939637efdacd to also match the model method name.

src/Resources/contao/controllers/BackendIndex.php
leofeyer: It has. And also, if you are not logged in, there will be no other messages.

src/Controller/FrontendController.php
```diff
@@ -70,16 +70,18 @@ public function shareAction(): RedirectResponse
      *
      * @Route("/_contao/login", name="contao_frontend_login")
      */
-    public function loginAction(): void
+    public function loginAction(): RedirectResponse
     {
+        return $this->redirectToRoute('contao_root');
```
leofeyer: This will potentially redirect the user to a different language than the one they were just browsing. If I e.g. log in on contao.org, switch to English and then log out, I will be redirected to the German start page.

aschempp: This response would only be triggered if someone manually opens the `_contao/login` without a POST request. In that case the firewall listener will not be triggered but the controller is. Without this, one would get an exception that the controller did not return a response. I don't think we need to handle correct locale redirection for URL manipulation?

aschempp:
@contao/developers please review, ask questions and provide feedback 😊

src/Resources/contao/controllers/BackendMain.php
```diff
@@ -293,4 +279,48 @@ protected function output()
...
+		$firewallMap = \System::getContainer()->get('security.firewall.map');
+		$request = \System::getContainer()->get('request_stack')->getCurrentRequest();
+
+		// generate exit impersonation path from current request
+		if (null === $request || null === ($firewallConfig = $firewallMap->getFirewallConfig($request)) || null === ($switchUserConfig = $firewallConfig->getSwitchUser()))
+		{
+			return;
+		}
+
+		$exitPath = $request->getRequestUri();
```
Toflar: Looks a lot like you could use the `Uri` component of `League` we already use?

src/Resources/contao/modules/ModuleLogout.php
```diff
@@ -71,14 +78,11 @@ public function generate()
 			$strRedirect = $objTarget->getAbsoluteUrl();
 		}
 
-		/** @var Session $session */
-		$session = System::getContainer()->get('session');
-		$session->set('_contao_logout_target', $strRedirect);
+		list ($strLogoutUrl, $strQuery) = explode('?', $strLogoutUrl);
```
Toflar: Same here, reuse tested components please :)

src/Security/Authentication/AuthenticationSuccessHandler.php
```diff
@@ -64,102 +67,68 @@ public function __construct(HttpUtils $httpUtils, ContaoFrameworkInterface $fram
...
 
-        @trigger_error('Using the postAuthenticate hook has been deprecated and will no longer work in Contao 5.0. Use the contao.post_authenticate event instead.', E_USER_DEPRECATED);
+        $request->setLocale($locale);
+        $this->translator->setLocale($locale);
 
-        foreach ($GLOBALS['TL_HOOKS']['postAuthenticate'] as $callback) {
-            $this->framework->createInstance($callback[0])->{$callback[1]}($user);
-        }
+        // Deprecated since Contao 4.0, to be removed in Contao 5.0
+        $GLOBALS['TL_LANGUAGE'] = str_replace('_', '-', $locale);
```
Toflar: Question: How is security related to locale handling here? Why did you have to move this?

src/Security/Authentication/Provider/ContaoAuthenticationProvider.php
```diff
@@ -29,82 +32,177 @@
...
+        $this->options = array_merge($this->defaultOptions, $options);
     }
 
     /**
      * {@inheritdoc}
      */
     public function checkAuthentication(UserInterface $user, UsernamePasswordToken $token): void
     {
+        if (!$user instanceof User) {
+            parent::checkAuthentication($user, $token);
```
Toflar: Could be shortened to `return parent::checkAuthentication($user, $token);`.

src/Controller/FrontendController.php
leofeyer: Nope.

README.md
```diff
@@ -153,14 +152,13 @@ security:
             switch_user: false
             logout_on_user_change: true
 
-            form_login:
+            contao_login:
                 login_path: contao_frontend_login
                 check_path: contao_frontend_login
-                default_target_path: contao_index
+                default_target_path: contao_root
+                failure_path: contao_root
```
leofeyer: I assume you are sure that this can be a route name instead of a path, too.

src/ContaoCoreBundle.php
```diff
@@ -57,6 +59,10 @@ public function build(ContainerBuilder $container): void
     {
         parent::build($container);
 
+        /** @var SecurityExtension $extension */
+        $extension = $container->getExtension('security');
```
leofeyer: Is this the right place to add the factory or should we do it in a separate compiler pass?

src/Controller/BackendController.php
```diff
@@ -59,6 +59,10 @@ public function loginAction(): Response
     {
         $this->container->get('contao.framework')->initialize();
 
+        if ($this->container->get('security.authorization_checker')->isGranted('IS_AUTHENTICATED_FULLY')) {
+            return new RedirectResponse($this->container->get('security.logout_url_generator')->getLogoutUrl());
```
leofeyer: Log out the user if fully authenticated?

src/EventListener/SessionExpirationListener.php
```diff
@@ -63,33 +60,37 @@ public function __construct(TokenStorageInterface $tokenStorage, ScopeMatcher $s
      */
     public function onKernelRequest(GetResponseEvent $event): void
     {
-        if (!$this->scopeMatcher->isContaoMasterRequest($event)) {
+        $request = $event->getRequest();
+        $session = $request->getSession();
+
+        if (!$event->getRequest()->hasPreviousSession()
+            || null === $session
```
leofeyer: You should handle the null pointer exception like [this](https://github.com/contao/core-bundle/blob/5e73ab0b8ebaf4608fce7a72b9d6c8f3fb0b50fa/src/Security/UserChecker.php#L150-L152) unless you are explicitly supporting the "does not have a session" case.

src/EventListener/SessionExpirationListener.php
```diff
@@ -63,33 +60,37 @@ public function __construct(TokenStorageInterface $tokenStorage, ScopeMatcher $s
...
+            || !($user = $token->getUser()) instanceof User
+        ) {
             return;
         }
 
-        $user = $token->getUser();
-
-        if (!$user instanceof User) {
+        // Logout after 1 hour
+        if ((time() - $session->getMetadataBag()->getLastUsed()) < 3600) {
```
leofeyer: I think this needs to remain configurable. Might be a compliance requirement to reduce the timeout.

src/Resources/config/services.yml
```diff
@@ -362,12 +359,22 @@ services:
...
+            - "@contao.framework"
+            - "@translator"
+            - "@logger"
 
     contao.security.backend_user_provider:
-        class: Contao\CoreBundle\Security\User\BackendUserProvider
+        class: Contao\CoreBundle\Security\User\ContaoUserProvider
         arguments:
             - "@contao.framework"
+            - "\\Contao\\BackendUser"
```
leofeyer: Can we at least pass `backend` or `frontend` and throw an exception if it is neither?

src/Resources/contao/controllers/BackendMain.php
```diff
@@ -268,6 +252,8 @@ protected function output()
 		$this->Template->systemMessagesCount = substr_count($strSystemMessages, 'class="tl_');
 		$this->Template->systemErrorMessagesCount = substr_count($strSystemMessages, 'class="tl_error"');
 
+		$this->setImporsonatedLogout();
```
leofeyer: Typo: imp**e**rsonated

src/Resources/contao/controllers/BackendMain.php
```diff
@@ -293,4 +279,48 @@ protected function output()
...
+		if (!$impersonatorUser)
+		{
+			return;
+		}
+
+		$firewallMap = \System::getContainer()->get('security.firewall.map');
+		$request = \System::getContainer()->get('request_stack')->getCurrentRequest();
+
+		// generate exit impersonation path from current request
+		if (null === $request || null === ($firewallConfig = $firewallMap->getFirewallConfig($request)) || null === ($switchUserConfig = $firewallConfig->getSwitchUser()))
```
leofeyer: Again, the null pointer handling is wrong. We expect that there is always a request, so we should throw an exception if there is not.

src/Resources/contao/languages/en/default.xlf
```diff
@@ -1115,6 +1112,9 @@ This e-mail has been generated by Contao. You can not reply to it directly.
       <trans-unit id="MSC.logoutBT">
         <source>Logout</source>
       </trans-unit>
+      <trans-unit id="MSC.switchBT">
+        <source>Back to %s</source>
```
leofeyer: Why do you not want to use "go back"?

src/Resources/contao/modules/ModuleLogin.php
```diff
@@ -33,12 +33,6 @@ class ModuleLogin extends \Module
 	 */
 	protected $strTemplate = 'mod_login';
 
-	/**
-	 * Flash type
-	 * @var string
-	 */
-	protected $strFlashType = 'contao.FE.error';
```
leofeyer: Please do not remove this but reset it to `protected $strFlashType = 'contao.' . TL_MODE . '.error';` like it is in Contao 4.4.

src/Resources/contao/modules/ModuleLogin.php
```diff
@@ -83,72 +72,81 @@ protected function compile()
...
+			// Redirect to last page visited
+			if ($this->redirectBack && $_SESSION['LAST_PAGE_VISITED'] != '')
+			{
+				$strRedirect = \Environment::get('base').$_SESSION['LAST_PAGE_VISITED'];
+			}
+
+			// Redirect home if the page is protected
+			elseif ($objPage->protected)
+			{
+				$strRedirect = \Environment::get('base');
```
leofeyer: I don't think this is correct. You should leave the decision to the "error 403" handler, which will redirect the user to e.g. the login page.

src/Resources/contao/templates/modules/mod_login.html5
```diff
@@ -13,9 +13,8 @@
       <?php endif; ?>
       <input type="hidden" name="FORM_SUBMIT" value="<?= $this->formId ?>">
       <input type="hidden" name="REQUEST_TOKEN" value="{{request_token}}">
-      <?php if ($this->targetPath): ?>
-        <input type="hidden" name="<?= $this->targetName ?>" value="<?= StringUtil::specialchars($this->targetPath) ?>">
-      <?php endif; ?>
+      <input type="hidden" name="redirect" value="<?= StringUtil::specialchars($this->redirect) ?>">
+      <input type="hidden" name="redirectBack" value="<?= (int) $this->redirectBack ?>">
```
leofeyer: Why do we need a second parameters? We could leave the `redirect` field empty if `!$this->redirectBack`.

src/Security/Authentication/AuthenticationFailureHandler.php
```diff
@@ -12,48 +12,28 @@
...
-     * @param ScopeMatcher         $scopeMatcher
-     * @param TranslatorInterface  $translator
-     * @param array                $options
-     * @param LoggerInterface|null $logger
-     */
-    public function __construct(HttpKernelInterface $httpKernel, HttpUtils $httpUtils, ScopeMatcher $scopeMatcher, TranslatorInterface $translator, array $options = [], LoggerInterface $logger = null)
+    public function __construct(HttpKernelInterface $httpKernel, HttpUtils $httpUtils, array $options = [], LoggerInterface $logger = null)
     {
-        parent::__construct($httpKernel, $httpUtils, $options, $logger);
+        $this->defaultOptions['failure_path_parameter'] = 'redirect';
```
leofeyer: Why hardcoded instead of configurable?

src/Security/Authentication/Provider/ContaoAuthenticationProvider.php
leofeyer: I don't think so.

src/Resources/contao/controllers/BackendMain.php
aschempp: Use for what?

src/Security/Authentication/AuthenticationSuccessHandler.php
aschempp: Because the `UserProvider` loads the user before the password is validated. Which caused the issue that the user language affected the locale even if the wrong password was entered (and e.g. changed the login screen language in the backend).

src/Security/Authentication/Provider/ContaoAuthenticationProvider.php
aschempp: PhpStorm would generate a warning because the `checkAuthentication` method does not have a return value …

README.md
aschempp: Yeah, everything can be routes, I was wrong about this the last time. The issue was that we didn't use `httpUtils` to generate the route but directly created our own `RedirectResponse`.

src/ContaoCoreBundle.php
aschempp: It's the same place Symfony does theirs (see https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/SecurityBundle/SecurityBundle.php#L45)

src/Controller/BackendController.php
aschempp: Yes. It was previously the case that when you accessed the login URL while already logged in, the user was logged out. I don't think we should change that as some people might have used that behavior to log a user out.

src/EventListener/SessionExpirationListener.php
aschempp: Well if there is no session we just don't have to check the session lifetime… no need for an exception.

src/EventListener/SessionExpirationListener.php
aschempp: If we add that configuration, we can't use [Symfony's pull request](https://github.com/symfony/symfony/pull/12807) without dropping the configuration variable again later on… not sure if that is ok?

src/Resources/config/services.yml
aschempp: What would that bring? We would need to `if-else` on the values to get the class?

src/Resources/contao/controllers/BackendMain.php
aschempp: Why would we throw an exception if the application works perfectly fine without it? It would just generate the regular logout link …

src/Resources/contao/languages/en/default.xlf
aschempp: First of all, I figured we could show the username the link is back to. Also, I think the word "Go back" does not necessarily work for "log back in as another user". It works perfectly in german (*Zurück*) but even english (*Go back*) it's not very obvious where to "back" is. *Back to username* is a better explanation of the link.

Btw, I also considered adding both links (a real logout and a back link), not sure if we want that?

src/Resources/contao/modules/ModuleLogin.php
aschempp: The login handler does not know about pages and their protection, and it does not generate links from page IDs or thelike. Technically it would probably be possible, but I think it would be very complicated. This is just the implementation is always was… can we maybe keep that for now and discuss whether to implement a more advanced version?

src/Resources/contao/templates/modules/mod_login.html5
aschempp: Unfortunately we cannot. The process is rather complex:

 1. if *redirect back* is enabled and a URL is present, pass that URL and always use it in the success handler
 2. if *redirect back* is **not enabled** but a forward page is set, pass that URL
 3. the SuccessHandler redirects according to 1. if `redirectBack` is set. If not, it first checks for user groups and redirects to them, otherwise falls back to `redirect`

src/Security/Authentication/AuthenticationFailureHandler.php
aschempp: We're hardcoding everything that is Contao specific, otherwise the template would need to be synchronized with the firewall configuration and I think that's highly overkill… It's our own firewall listener after all, so there is no need to have parameters that are not to be configured.

src/Resources/contao/controllers/BackendMain.php
aschempp: Fixed in 2741c66253622f16f2eea2f85ab0a78f48d2ffd7

src/Resources/contao/modules/ModuleLogin.php
aschempp: Fixed in 6ad2a9ce87ea617a5e78ce463e5049df81008cff

src/Resources/contao/controllers/BackendIndex.php
aschempp: Changed in 596453b420c72e7192e0712a955b638fe3e3dd63

src/Controller/BackendController.php
leofeyer: We have changed this in Contao 4 though and we should not revert it now. 😄

src/EventListener/SessionExpirationListener.php
leofeyer: In which case is there no session?

src/EventListener/SessionExpirationListener.php
leofeyer: We need to have it configurable, so yes, it is ok to deprecate the variable later.

Of course, we could also name the variable `session_expiration.max_idle_time`, then it would be forward compatible with Symfony 4.1.

src/Resources/config/services.yml
leofeyer: It would make it impossible to use the class with anything else than `backend` and `frontend`. You recall the initial discussion?

src/Resources/contao/controllers/BackendMain.php
leofeyer: How can there ever not be a request?

src/Resources/contao/languages/en/default.xlf
leofeyer: We do not. 😄

src/Resources/contao/modules/ModuleLogin.php
leofeyer: Just remove the `$objPage->protected` check that you have added in your PR. It worked fine before.

src/Security/Authentication/AuthenticationFailureHandler.php
leofeyer: Agreed.

src/Controller/BackendController.php
aschempp: No we havent changed that, I explicitly tested it on demo.contao.org and you‘re logged out of the backend if you access the login screen.

src/EventListener/SessionExpirationListener.php
aschempp: I have no idea why a session can be `null`, probably if session management is disabled in the application?

src/Controller/BackendController.php
leofeyer: Ok, but now that we have a `/contao/logout` route, this is no longer necessary. I can change it in another PR though.

src/EventListener/SessionExpirationListener.php
leofeyer: Then Contao would not work anymore. As you see `null` is not an expected state.

src/EventListener/SessionExpirationListener.php
aschempp: It would not be forward compatible, because the configuration is per firewall, not a bundle configuration.

src/Resources/config/services.yml
aschempp: Yeah, so in that case I can also check the two classes instead of introducing arbitrary names?

src/Resources/contao/controllers/BackendMain.php
aschempp: I don't think that case will ever happen. It's just shorter code 😄

src/Resources/contao/modules/ModuleLogin.php
aschempp: Uhm, I added that because it existed before? https://github.com/contao/core-bundle/blob/4.4/src/Resources/contao/modules/ModuleLogin.php#L142

leofeyer: Sorry, I missed this. My bad.

src/Security/Authentication/FrontendAuthenticationSuccessHandler.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function determineTargetUrl(Request $request)
+    {
+        if (!$this->user instanceof FrontendUser) {
+            return parent::determineTargetUrl($request);
+        }
+
+        if (ParameterBagUtils::getRequestParameterValue($request, '_always_use_target_path')
```
leofeyer: You are using an internal method here.

src/Security/Logout/FrontendLogoutSuccessHandler.php
```diff
@@ -0,0 +1,37 @@
...
+use Symfony\Component\Security\Http\ParameterBagUtils;
+
+class FrontendLogoutSuccessHandler extends DefaultLogoutSuccessHandler
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function onLogoutSuccess(Request $request): RedirectResponse
+    {
+        if ($targetUrl = ParameterBagUtils::getRequestParameterValue($request, 'redirect')) {
```
leofeyer: Again using an internal method.

src/Security/Authentication/FrontendAuthenticationSuccessHandler.php
aschempp: I know. I'm using the same method Symfony does in their implementation…

leofeyer: Except its their method so they are "allowed" to use it. Is there no other way to retrieve the value?

src/EventListener/AuthenticationListener.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if (!$exception instanceof UsernameNotFoundException) {
+            return;
+        }
+
+        $username = $exception->getUsername() ?: $event->getAuthenticationToken()->getUsername();
+
+        $this->logger->info(
+            sprintf('Could not find user "%s"', $username),
+            ['contao' => new ContaoContext(__METHOD__, ContaoContext::ACCESS, $username)]
```
leofeyer: Does using the invalid username in the context really make sense?

leofeyer:
Thank you @aschempp.

PR #1247, opened 2017-12-19
[RFC] Remove obsolete fulltext index

ausi:
AFAIK we don’t use the fulltext index but only `WHERE text REGEXP ?` in the search module.


leofeyer:
We might as well require MySQL 5.6, as there is no actively maintained OS which supports only MySQL 5.5 and neither MySQL 5.6 nor 5.7:

https://www.mysql.com/de/support/supportedplatforms/database.html

leofeyer:
Removed in 62f3f546aa9f867da4675a1418660e78566d9e68 and d1772e32bb1f353f694452f75b1ba6843f9530f4.

PR #1249, opened 2017-12-19
Check if package contains slash before splitting

aschempp:


leofeyer:
I don't see why we need this change. The method only handles packages with the vendor name `contao-components`, so unless there is a vendor-less package named `contao-components`, the check is useless.

aschempp:
I've encountered this issue, because there apparently can be a `__root__` package 😉

leofeyer:
Then `$vendor` would be `__root__` and since `'contao-components' !== '__root__'`, the package would be ignored in line 90.

aschempp:
Okay, so two things:
 1. we could exit/break/continue if the package name does not have a slash
 2. The check **is** needed because otherwise `list` will complain about only having one value for two parameters

aschempp:
This should not be closed

leofeyer:
I cannot reproduce it.

aschempp:
I could swear there was a root package name without a slash, but I can't find any related code. Let's keep it closed until we have the issue again.

ausi:
I can reproduce this error if I open *app_dev.php* with an empty cache. The last key in `PackageVersions\Versions::VERSIONS` is set to `__root__` in my setup.

> Then `$vendor` would be `__root__` and since `'contao-components' !== '__root__'`, the package would be ignored in line 90.

The error occurs already in line 83:

```php
[$vendor, $packageName] = explode('/', $name, 2);
```

Because PHP tries to access the second element of the array to set it to `$packageName`. This results in the error `Notice: Undefined offset: 1`.

The error could be fixed in my case by moving the `contao-components` check to the top of the `foreach`:

```php
foreach (Versions::VERSIONS as $name => $version) {
    if (strncmp('contao-components/', $name, 18) !== 0) {
        continue;
    }

    $packageName = substr($name, 18);
    ...
```

leofeyer:
I don't have a `__root__` package in my `Versions` class but I have still implemented @ausi's fix in 9aeff7e0082bc5c624ebca1bc0c951948e39ad6b.

PR #1256, opened 2017-12-20
[WIP] Security changes (step 3)

aschempp:
More things we should add to the new security implementation.

Tasks:
 - [x] Use persistent remember-me functionality 
 - [ ] Decide on the default remember-me duration
 - [x] Remove use of `AdvancedUserInterface` (see https://github.com/symfony/symfony/pull/23508)
 - [ ] ~~Logout and~~ login might not validate the CSRF token? *logout is not a POST request*
 - [x] Enable ORM and use it for persistent remember-me (see https://github.com/contao/manager-bundle/pull/51)
 - [x] Do not ignore non-`tl_` prefixed tables in the install tool (see https://github.com/contao/installation-bundle/pull/79)
 - [ ] Do not set the locale from user objects in the front end

This PR is just an early draft for tasks and will be rebased once #1238 is merged.


src/Security/UserChecker.php
```diff
@@ -91,7 +91,7 @@ private function checkIfAccountIsLocked(User $user): void
      */
     private function checkIfAccountIsDisabled(User $user): void
     {
-        if (false !== $user->isEnabled()) {
+        if (!$user->disable) {
```
leofeyer: What about `(!$user->start || $user->start < $time) && (!$user->stop || $user->stop > $time)`?

aschempp: That's already handled in https://github.com/contao/core-bundle/pull/1256/commits/e534341f23d6aae78e43a492c4f34c51627fb3cb#diff-3965f81a275b864b36b992503eea7bd5R126, apparently because the log messages are different.

leofeyer:
Since we cannot agree on how to implement ORM, let's just write our own token provider that uses `tl_remember_me` instead of `rememberme_token`. I can do the work if you want.

aschempp:
I thought we're gonna discuss this tomorrow on the public Mumble call?

Issue #706, opened 2017-02-14
prepareMetaDescription reduce max length to 160


christianromeni:
Since the prepareMetaDescription is meant for Meta Descriptions, it should resemble the currently recommended max lengths for search engines.
I know that they vary, but almost all of them are below 160 symbols.. So I guess the Line 595 in Frontend.php should be 

$strText = \StringUtil::substr($strText, 160);

https://github.com/contao/core-bundle/blob/7f746f41156f81867aefbb812f5e086e543d5bdd/src/Resources/contao/classes/Frontend.php#L595

leofeyer:
I don't see any reason to enforce a length limit on people here, so -1 from me.

christianromeni:
Well, you are already doing it by limiting it to 180 .. So why not limit it to some sort of standard then?
Otherwise, remove the limit completely... Following your logic.

leofeyer:
Ah, I got you wrong, sorry. Reducing the limit might be ok. Let's see what other people think about it, ok? @contao/developers /cc

discordier:
Ich kenne mich mit dem Thema nicht wirklich aus und muss mich daher relativ enthalten, [sistrix](https://www.sistrix.de/frag-sistrix/onpage-optimierung/meta-element-meta-tag/wie-sieht-die-optimale-meta-description-aus/) hingegen empfiehlt:
> - nicht mehr als 156 Zeichen (inkl. Leerzeichen)
>   - mehr als 156 Zeichen werden durch (…) abgeschnitten
> - verfügt die Webseite über ein “veröffentlicht am” Datum oder eine Produktauflistung (1 – 25 von 250) so wird dies mit Wahrscheinlichkeit als Teil der Meta-Description vorangestellt angezeigt
>   - sollte dies der Fall sein, so empfiehlt es sich, nicht mehr als 139 Zeichen (inkl. Leerzeichen) zu verwenden

Und:
> Technisch gesehen wird die maximale Länge einer Meta-Description seitens Google nicht durch die Anzahl der Zeichen bestimmt, sondern auf Pixelbasis berechnet. Auf einer Breite (engl. width) von ~ 500 Pixel lässt sich die Meta-Description unterbringen.

Das werden wir aber sicher nicht berechnen (koennen und wollen).

Ich bin also mit beidem zufrieden, ob wir es aendern oder belassen.

ausi:
[...]

PR #1258, opened 2017-12-20
[RFC] increase meta description length

fritzmg:
Since around November 30th Google has been showing longer snippets in the search results. According to [Rank Ranger](https://www.rankranger.com/google-serp-features) the average length has been increased from ~175 to almost 230. Other sites quote the new length being 320 characters, but I could not find a reliable source for that.

Some posts discussing this:

* https://www.inc.com/john-lincoln/googles-new-meta-description-length-what-you-need-to-know.html
* https://www.seroundtable.com/google-extends-snippets-320-characters-24862.html
* https://searchengineland.com/google-officially-increases-length-snippets-search-results-287596

I have used a character limit of 230 in this PR. I am not sure what it should be in the end though. May be this should actually also be configurable?

### DIFF
--- a/src/Resources/contao/classes/Frontend.php
+++ b/src/Resources/contao/classes/Frontend.php
@@ -581,7 +581,7 @@ abstract class Frontend extends \Controller
 		$strText = $this->replaceInsertTags($strText, false);
 		$strText = strip_tags($strText);
 		$strText = str_replace("\n", ' ', $strText);
-		$strText = \StringUtil::substr($strText, 160);
+		$strText = \StringUtil::substr($strText, 230);
 
 		return trim($strText);
 	}


ausi:
It was reduced from 180 to 160 in #706 🙂

fritzmg:
Yeah. I don't know why it was reduced to 180 in the first place, even back then the maximum observed snippet length was around 180, as far as I know. But doesn't matter anymore now anyhow ;).

leofeyer:
I'm against merging this. As @ausi said, we have just reduced the maximum length for the reasons mentioned in #706.

bekanntmacher:
I'm with @fritzmg 's merge! Currently 320 (maximal) characters are supported.

official source:
danny-sullivan-description-320-zeichen

We should increase to 320 characters.

fritzmg:
Link to tweet: https://twitter.com/dannysullivan/status/936780855019581440

ausi:
I think we should ask @christianromeni here too, he requested the change to 160.

fenepedia:
I would remove this limit over all and only give a hint in the description. Maybe google is testing something and in 2 or 3 weeks there are X characters allowed.

fritzmg:
Contao should probably trim to some sort of maximum, e.g. 320. Then the user can control what length he wants to use in their news and event teasers and extensions like [derhaeuptling/contao-seo-serp-preview](https://github.com/derhaeuptling/case the-serp-preview) can show recommendations.

Why should it trim at all? Because there are some use cases where the teaser text is used as a sort of detailed text and in that case the meta description probably should not contain the whole "detailed" text.

Metis77:
I support the idea, that the limit should be set to 320 or even better removed at all.

For now Google will trim at 320 and my change it some day.
So there is no need for contao to trim, as google  or other search engines do anyway.

fritzmg:
> So there is no need for contao to trim

As explained in the previous post, some users might use the _teaser_ as a sort of detailed text, potentially containing hundreds of words. If Contao does not trim, the whole content would be put into the `description`, which is unnecessary.

Using the currently reported maximum of `320` seems sensible to me.

Metis77:
@fritzmg true, the event/news/article teaser do not match the needs of a meta description.
This is because they are usually used as real visible content. Also they do offer formatted text.
Those Teaser should not be used as meta description at all.

In my opinion this is a different issue, that should be solved  by adding a separate description field below teaser texts.
Such a description field should, if empty, inherit content from the teaser field.

leofeyer:
As discussed in Mumble on January 18th, 2018, we want to increase the length to 320.

leofeyer:
Fixed in d868f699e1e5af8aad8e01a79c34bdbc1043fe70.

PR #1260, opened 2017-12-20
[RTM] Always lazy initialize session for BC

Toflar:


src/Framework/ContaoFramework.php
```diff
@@ -297,6 +297,13 @@ private function initializeFramework()
         $this->includeHelpers();
         $this->includeBasicClasses();
 
+        // Start the session because that used to be the case. For BC reasons
+        // we always have to make sure the session is started because people
+        // are relying on $_SESSION to be present/working.
+        if (null !== ($session = $this->request->getSession())) {
+            $session->start();
```
leofeyer: This can only be fixed in Contao 4.4. Contao 4.5 explicitly supports using the front end without a session.

Toflar: As I said. We will forever need this in Contao 4.4 because everybody uses `$_SESSION` directly and it will always remain `null` if we don't start it explicitly here.

leofeyer:
Here is the solution. I'd say it is "advanced PHP usage" to avoid the word "hacky", but it works. 😄 

```php
class LazySession implements \ArrayAccess
{
    public function offsetExists($offset)
    {
        $this->start();

        return isset($_SESSION[$offset]);
    }

    public function offsetGet($offset)
    {
        $this->start();

        return $_SESSION[$offset];
    }

    public function offsetSet($offset, $value)
    {
        $this->start();

        $_SESSION[$offset] = $value;
    }

    public function offsetUnset($offset)
    {
        $this->start();

        unset($_SESSION[$offset]);
    }

    private function start()
    {
        @trigger_error('Using the $_SESSION variable has been deprecated … Use the Symfony session instead.', E_USER_DEPRECATED);

        session_start();
    }
}

$_SESSION = new LazySession();

// at this point $_SESSION is a LazySession object

unset($_SESSION['foo']);
$_SESSION['TL_LANGUAGE'] = 'de';

// after the first time $_SESSION has been accessed, PHP has replaced
// $_SESSION with the actual session content

var_dump($_SESSION);

// array('TL_LANGUAGE' => 'de)
```

This allows us to automatically start the session if someone accesses `$_SESSION`.

leofeyer:
It seems your approach was too complicated: https://3v4l.org/RQ17G

Toflar:
I tried simulating the session service injection ;)

leofeyer:
Which is not necessary. As soon as Symfony starts the session, `$_SESSON` will be overridden.

Toflar:
I know, which is why I tested injecting the service.

aschempp:
I don't like messing with the global objects, we already had issues with `$GLOBALS`. Also, you approach would not work with `count()`, `isset()` or any other methods that could potentially be used on a valid array.

m-vo:
It's a hack for sure, but `__isset()` and the `Countable` interface could be implemented as well. Or were these only examples of a large list?

Toflar:
Everything should be feasible I think.

ausi:
> This is from this morning: 3v4l.org/83irp

Seems that id didn’t work because `$value` was misspelled in `offsetSet($offset, $vaue)` :)

Toflar:
lol ok 😄

leofeyer:
@Toflar Are you going to update the pull request accordingly or do you want me to do the work?

Toflar:
Working on it right now.

src/EventListener/LocaleListener.php
```diff
@@ -66,10 +66,6 @@ private function getLocale(Request $request): string
             return $this->formatLocaleId($request->attributes->get('_locale'));
         }
 
-        if (null !== ($session = $request->getSession()) && $session->has('_locale')) {
```
leofeyer: Removing this is wrong. But we should check `$session->isStarted()` and only execute the logic then.

src/Session/LegacySessionAccess.php
```diff
@@ -0,0 +1,102 @@
...
+        return count($this->session->all());
+    }
+
+    /**
+     * Ensures a session is initialized as soon as someone accesses
+     * $_SESSION and we can store certain attributes accordingly.
+     */
+    private function ensureSessionStarted()
+    {
+        var_dump(debug_backtrace(0, 5));
```
leofeyer: Forgotten debug statement.

UPGRADE.md
```diff
@@ -272,6 +272,9 @@ been removed in Contao 4.0.
 
 The `$_SESSION['TL_USER_LOGGED_IN']` flag has been removed.
 
+### $_SESSION['TL_LANGUAGE']
+
+The `$_SESSION['TL_LANGUAGE']` flag has been removed.
```
leofeyer: You don't have to add this here. The variable has been deprecated since Contao 4.0.

https://github.com/contao/core-bundle/blob/4.4/DEPRECATED.md#globalstl_language-and-_sessiontl_language

src/Session/LegacySessionAccess.php
Toflar: It's wip

src/EventListener/LocaleListener.php
Toflar: I don't see why the locale needs to be stored in the session, makes absolutely no sense to me.

aschempp: It's a topic to discuss. It is/was there so the locale is available before routing and security. I do agree that we should remove that, but there are a lot more places to handle.

src/Session/LazySessionAccess.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Ensures a session is initialized as soon as someone accesses
+     * $_SESSION and we can store certain attributes accordingly.
+     */
+    private function ensureSessionStarted()
+    {
+        @trigger_error('Accessing $_SESSION directly is deprecated and support will be dropped with Contao 5.0. Use the Symfony request instead to work with the session.', E_USER_DEPRECATED);
+
+        if ($this->session->isStarted()) {
```
leofeyer: This case can never occur, because as soon as the session is started, the `$_SESSION` variable is replaced and the `LazySessionAccess` object is destroyed.

leofeyer:
It seems that we are not handling one of possible three cases.

**Session started at `$framework->initialize()`**

This case is fine. The framework will add the Contao session bags to `$_SESSION`.

**Session not started in framework, later started by accessing `$_SESSION`**

This case is also fine. The `LazySessionAccess` object will add the Contao session bags to `$_SESSION`.

**Session not started in framework, later started by Symfony**

Who will be adding the Contao session bags to `$_SESSION` in this case? The framework has been initialized already and the `LazySessionAccess` object has been destroyed already.

leofeyer:
You are absolutely right.

leofeyer:
Thank you @Toflar.

PR #1262, opened 2017-12-20
[RTM] Quote reserved words in database queries

ausi:
This is the port of contao/core#8813.

Also see #1106 and contao/installation-bundle#70

### DIFF
--- a/src/Resources/contao/library/Contao/Database.php
+++ b/src/Resources/contao/library/Contao/Database.php
@@ -235,7 +235,7 @@ class Database
 			$varSet = $this->resConnection->quote($varSet);
 		}
 
-		return "FIND_IN_SET(" . $strKey . ", " . $varSet . ")";
+		return "FIND_IN_SET(" . \Database::quoteColumnName($strKey) . ", " . $varSet . ")";
 	}
 
 
@@ -490,7 +490,7 @@ class Database
 	 */
 	public function isUniqueValue($strTable, $strField, $varValue, $intId=null)
 	{
-		$strQuery = "SELECT * FROM $strTable WHERE $strField=?";
+		$strQuery = "SELECT * FROM $strTable WHERE ".static::quoteColumnName($strField)."=?";
 
 		if ($intId !== null)
 		{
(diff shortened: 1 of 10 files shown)


aschempp:
Is there a reason we don't copy the reserved words from Contao 3.5 to 4.4? I mean it should be very little effort to also fix that version?

ausi:
IIRC we didn’t want to add support for using Contao 4.4 with PHP 5 together with MariaDB 10.2.4+, see https://github.com/contao/installation-bundle/pull/70#issuecomment-353210582

src/Resources/contao/library/Contao/Database.php
```diff
@@ -706,6 +706,32 @@ public function getUuid()
...
+	/**
+	 * Quote the column name if it is a reserved word
+	 *
+	 * @param string $strName
+	 *
+	 * @return string
+	 */
+	public static function quoteColumnName($strName)
+	{
+		return \System::getContainer()->get('database_connection')->getDatabasePlatform()->quoteSingleIdentifier($strName);
```
leofeyer: I wonder if we should just return ```'`' . $strName . '`'``` instead? Contao does not work with anything but MySQL and `quoteColumnName()` is potentially executed often, so we might be able to optimize performance here?

leofeyer:
Thank you @ausi.

leofeyer:
> IIRC we didn’t want to add support for using Contao 4.4 with PHP 5 together with MariaDB 10.2.4+

Even though we are now quoting database fields, the changes in MariaDB 10.2 also [broke Doctrine](https://github.com/doctrine/doctrine2/issues/6565). And Doctrine will only fix the issue in version 2.7, which requires PHP 7.1+, so this effectively means that Contao 4 with PHP 5 will never be compatible with MariaDB 10.2.7+.

src/Resources/contao/library/Contao/Database.php
ausi: I think we should definitely use `quoteSingleIdentifier()` from the doctrine database platform, but maybe we can store the DatabasePlatform object in a static variable so that we don’t call the whole `\System::getContainer()->get('database_connection')->getDatabasePlatform()` every time?

leofeyer: We can simply use `quoteIdentifier()` instead of `quoteSingleIdentifier()`.

https://github.com/contao/core-bundle/commit/130fc001365bff41662c3dfea56b8a2961ae8969

Then we don't have to fix the `findInSet()` method, either.

ausi: That makes sense 👍

leofeyer:
From the current ER2 outage I have learned that some developers are manually quoting identifiers in their code. I had to add this fix:

https://github.com/contao/core-bundle/commit/541a8ca5a673bd561c49f5472f99287d34751438

PR #1263, opened 2017-12-21
Never merge cache-control headers

Toflar:
This is the quick fix for Contao 4.4.
I'm working on a better solution for Contao 4.5 (which is not possible to merge in 4.4 because of security and request token handling).


leofeyer:
Thank you @Toflar.

Issue #1264, opened 2017-12-21
4.5-RC2 Setting the backend language doesn’t work anymore


ausi:
Changing the language in backend on the profile page doesn’t work.

aschempp:
Is it not saved or not applied?

ausi:
It gets saved, but no applied.

ausi:
If I log out after changing the language and then login again it gets applied correctly.

leofeyer:
See #1266.

PR #1266, opened 2017-12-22
Set locale from backend user after authentication

aschempp:
Fixed https://github.com/contao/core-bundle/issues/1264

### DIFF
--- a/src/Security/Authentication/AuthenticationSuccessHandler.php
+++ b/src/Security/Authentication/AuthenticationSuccessHandler.php
@@ -21,7 +21,6 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Http\Authentication\DefaultAuthenticationSuccessHandler;
 use Symfony\Component\Security\Http\HttpUtils;
-use Symfony\Component\Translation\TranslatorInterface;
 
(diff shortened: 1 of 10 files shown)


src/Security/Authentication/AuthenticationSuccessHandler.php
```diff
@@ -110,23 +98,4 @@ private function triggerPostLoginHook(User $user): void
...
-     * @param string  $locale
-     */
-    private function setLocale(Request $request, string $locale): void
-    {
-        if (null !== ($session = $request->getSession()) && $session->isStarted()) {
-            $session->set('_locale', $locale);
-        }
-
-        $request->setLocale($locale);
-        $this->translator->setLocale($locale);
```
leofeyer: I understand that we must not initialize the translator with the user's language in the front end, because the page language is the only relevant language. But where are we initializing the translator with the page language now?

leofeyer:
Thank you @aschempp.

src/Security/Authentication/AuthenticationSuccessHandler.php
aschempp: https://github.com/contao/core-bundle/blob/master/src/Resources/contao/pages/PageRegular.php#L70

leofeyer: Yeah, I meanwhile figured this out, too (see #1282).

PR #1275, opened 2017-12-29
Security Changes (Step 4)

aschempp:
- [x] Correctly trigger the Symfony authentication entry point
- [x] Bypass RememberMe Token update in header replay mode
- [x] Update unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### DEV
 
+ * Correctly handle authentication through the Symfony firewall entry point (see #1275).
  * Make services public that we need to access directly.
 
 ### 4.5.0 (2017-12-28)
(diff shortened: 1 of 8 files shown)


aschempp:
@Toflar we currently have to bypass the remember token update because the response of the header replay is not processed. We should discuss whether the cookies of a header replay response should be processed and appended to the actual request.

leofeyer:
Checking for the `T42-Replay-Headers` content type is certainly not the correct solution, because it would not work with Varnish.

I think that our approach with stopping event propagation in the preflight request is wrong (see terminal42/header-replay-bundle@7da33c3). It will prevent the response listeners from running, one of which adds the updated remember me cookie from the request attributes to the response.

We should therefore discuss a different solution for contao/core-bundle#1009 and other related issues. Maybe we should not execute certain listeners upon HEAD requests?

leofeyer:
Thank you @aschempp.

PR #1289, opened 2018-01-06
RFC DcaSchemaProvider: use default collation when setting custom charset

m-vo:
If an extension sets a character set in the table options without a collation ([like Haste does](https://github.com/codefog/contao-haste/blob/master/library/Haste/Model/Relations.php#L507)) this might lead to an invalid data definition statement in the install tool like the following

`SQLSTATE[42000]: Syntax error or access violation: 1253 COLLATION 'utf8mb4_unicode_ci' is not valid for CHARACTER SET 'utf8'`

The character set and collation are set by the config defaults and then overwritten individually if the respective table options are set. So setting one without unsetting the other (or resetting to default) is probably a bad thing.

If setting a collation without a character set is a valid use case this should be handled as well.
If setting a charset without a collation isn't a valid use case and this is no bug we should document that somewhere. :-)

### DIFF
--- a/src/Doctrine/Schema/DcaSchemaProvider.php
+++ b/src/Doctrine/Schema/DcaSchemaProvider.php
@@ -96,6 +96,8 @@ class DcaSchemaProvider
 
                 if (preg_match('/DEFAULT CHARSET=([^ ]+)/i', $definitions['TABLE_OPTIONS'], $match)) {
                     $table->addOption('charset', $match[1]);
+                    // reset table option to default collation to prevent conflicts 
+                    $table->addOption('collate', 'DEFAULT');
                 }
 
                 if (preg_match('/COLLATE ([^ ]+)/i', $definitions['TABLE_OPTIONS'], $match)) {


m-vo:
:disappointed:  setting `DEFAULT` isn't really supported in the installer as well. 
https://github.com/contao/installation-bundle/blob/4.5/src/Database/Installer.php#L221.

So this is no more than part of a fix.

aschempp:
It sure should be added to the installer if Doctrine Schema supports it.

m-vo:
I think part of the problem is that we're *not* using Doctrine Schema for the update of the charset/collation/.. but are manually adding `ALTER` queries in a second step.

(btw for new tables this also allows executing the `ALTER` statements without the `CREATE` statements which should not be possible)

aschempp:
> I think part of the problem is that we're not using Doctrine Schema for the update of the charset/collation/.. but are manually adding ALTER queries in a second step.

I totally agree. Unfortunately, Doctrine Schema does not support charset/collation/.. updates as far as I know. So either we do it or we create a pull request for Doctrine and wait for an update…

m-vo:
I just tested it by manually defining a collation in one of my entities and the schema diff picks those up. So maybe this is already working?

m-vo:
...but interestingly enough not with the current master branch. I'll have to find out what has changed.

For the sake of documentation: With `contao/installation-bundle:4.5.1` i can for instance add  `options={"collate"="abc", "charset"="xyz"}` to my entities table definition and the install tool will generate a statement `ALTER TABLE tl_my_table CONVERT TO CHARACTER SET xyz COLLATE abc`. This should be applicable for the dca generated parts of the schema as well.


`

m-vo:
Update:

Doctrines schema comparator looks at the plattform options but ignores them if they are empty:
https://github.com/doctrine/dbal/blob/2.6/lib/Doctrine/DBAL/Schema/Comparator.php#L495-L502

When the SQL is built (https://github.com/doctrine/dbal/blob/2.6/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L604-L605) it seems that charset and collation should be in there: 
https://github.com/doctrine/dbal/blob/2.6/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L2218-L2247

m-vo:
Ok, this is way more f*cked up, than I thought it would be. :-)

Here are my findings (some of the following points maybe only apply to MySQL but could be abstracted):

### Where to look at?
There are multiple places where options can be set: server level, database level, table level, column level. These settings are hierarchical and and cascade down (if nothing is set). Let's focus on the last two.

(So btw something like `SELECT CHARACTER_SET_NAME AS CharacterSet, COLLATION_NAME AS Collation FROM information_schema.COLUMNS WHERE [...]` won't always show the actual value.)

Doctrine Entities can be annotated with options per table and/or per field. Contao's DCAs likewise can have SQL containing attributes for the table and/or fields. 

### How to migrate?
For a correct migration imo 3 things have to work:
1. db schema and future schema (entities / DCA) must contain all options for tables and columns (keep in mind that there are also default values)
2. the comparer has to detect the differences
3. the relevant SQL has to be generated

Right now the db schema misses the table options completely (see [here](https://github.com/doctrine/dbal/blob/2.6/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L288), empty array) but contains column options (`plattformOptions`) if set. The future schema has the values set but table options always contain the default values if not set.

Doctrines comparer only checks the columns for differences. Basically the option arrays get intersected and only the result is checked for differing values (see [Comparer](https://github.com/doctrine/dbal/blob/2.6/lib/Doctrine/DBAL/Schema/Comparator.php#L495-L502)). So if there is no value in either one of the schemes nothing will happen.

Funny enough: The relevant SQL to be generated if 1. and 2. are in the right conditions works, collations and charsets are [explicitly taken in account](https://github.com/doctrine/dbal/blob/2.6/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L2225-L2229).

### What are our options?

In my opinion doctrine is broken and should support this. However by the look of how many open and unresolved issues and pull requests are on that topic (just search for e.g. `collation` in the repo) i don't think there will be a quick solution if any.

What I tried roughly and what could work in principle is the following: We could ignore the option hierachy and set everything on a column basis. This maybe sounds like a bad idea but the behaviour is broken anyhow because the default values are beeing set everytime on table level and therefore mask the intention (see above).  So after creating the schemas (and before diffing) we would have to...

* ... enrich the `fromSchema` by querying the database for table and column options and setting the effective value on the schema (`$column->setPlatformOption`) *[btw, trap: the option key is 'collation' on column level, but 'collate' on table level]*. 
* ... walk through the `toSchema` setting the column options from the table options if not explicitly specified.

And of course, for the future: If we use a dbal we should get rid of those general SQL options and just support those options that our dbal supports. 


### A word on collation/characterset

If I understand correctly the collation *always* contains the character set it's valid with (part before the first underscore), e.g. '**utf8mb4**_unicode_ci' / 'utf8mb4'. If that's true we could at least derive the charset from the collation if no charset is specified. But thats a) an edge case and b) should as well be handled by Doctrine. (And I'm wondering if it's true :grin: as a lot of db tools set the charset explicitly...)

leofeyer:
Fixed in f0383d0c4bd3d9f28f351aad49b6f09c17262bb8. This is a legacy case (`database.sql`), so we can simply append `_general_ci` here.

m-vo:
@leofeyer Why does this have anything to do with `database.sql`?

Isn't the problem, that there should not be a characterset set without a collation? If I'd set charset `binary` for instance this would break because there is no collation called `binary_general_ci`.

leofeyer:
No, `ENGINE=MyISAM  DEFAULT CHARSET=utf8` is perfectly fine (taken from your Haste example).

m-vo:
But now ` ENGINE=MyISAM DEFAULT CHARSET=binary` or  ` ENGINE=MyISAM DEFAULT CHARSET=latin5` would crash.

m-vo:
Part of my monologe above was about having doctrine handle those things. :-) So basically if setting a charset is a valid solution this should work too for entities (which it doesn't):
`@ORM\Table(name="tl_someTable", options={"charset"="utf8"})` 

(`@ORM\Table(name="tl_someTable", options={"collate"="utf8_general_ci", "charset"="utf8"})` does.) But there is still the problem of generating  `CREATE` and `ALTER` statements for the same operation. And allowing to select them individually can cause a crash as well.

Issue #1288, opened 2018-01-04
Application state is not stored


leofeyer:
It seems that in Contao 4.5 the application state is not stored in the session anymore. Open palettes, active filters etc. are reset after each logout.

leofeyer:
@aschempp @bytehead I guess this is also related to Symfony security? Maybe our session save handler is triggered after the session has been closed by the logout handler?

leofeyer:
We have to add the user session data in `AuthenticationSuccessHandler::onAuthenticationSuccess()`.

leofeyer:
Fixed in 2b627d4e42fcdffe7aed690875d5e83dc5fc3e8e.

aschempp:
I'm not sure this is entirely correct. It means the session is only loaded from database on the actual login. Not sure it will be done on `remember_me` and it certainly will not be done on each request. And I'm pretty sure that was previously the case.

leofeyer:
This is just one additional step, I did not remove the existing `UserSessionListener`.

leofeyer:
Because the firewall redirects after a successful login so any listener below priority 8 is never executed.

aschempp:
Why would that matter? After the redirect the session should be loaded again? The only reason I can think of is that the save handler is still executed and overwrites the user data. However, if that is the case, then the handling is wrong because the session must never be saved to the user if it was not loaded before. The redirect could come from anywhere…

leofeyer:
You may be right. Feel free to provide a better implementation.

leofeyer:
Ok, so the problem is that `$token->getUser()` in

https://github.com/contao/core-bundle/blob/e4f0d29c314d22ca2ac235ff938027f48aaf985b/src/EventListener/UserSessionListener.php#L79-L85

only returns the incomplete, unserialized `User` object upon login. Therefore `$user->session` is empty and the existing session data is overwritten.

aschempp:
Why would the user not be complete? And why is the listener (priority 0) even executed when the firewall redirects on priority 8?

leofeyer:
I might have been wrong about the redirect. You may want to debug this yourself.

PR #1293, opened 2018-01-08
Do not store user session if it has not already been initialized

aschempp:
see https://github.com/contao/core-bundle/issues/1288

### DIFF
--- a/src/EventListener/UserSessionListener.php
+++ b/src/EventListener/UserSessionListener.php
@@ -46,6 +46,11 @@ class UserSessionListener
     private $scopeMatcher;
 
     /**
+     * @var bool
+     */
+    private $registered = false;
+
+    /**
      * @param Connection                           $connection
      * @param TokenStorageInterface                $tokenStorage
      * @param AuthenticationTrustResolverInterface $authenticationTrustResolver
@@ -87,6 +92,8 @@ class UserSessionListener
         if (\is_array($session)) {
             $this->getSessionBag($event->getRequest())->replace($session);
         }
+
+        $this->registered = true;
     }
 
     /**
@@ -96,7 +103,7 @@ class UserSessionListener
      */
     public function onKernelResponse(FilterResponseEvent $event): void
     {
-        if (!$this->scopeMatcher->isContaoMasterRequest($event)) {
+        if (!$this->registered || !$this->scopeMatcher->isContaoMasterRequest($event)) {
             return;
         }
 


leofeyer:
I am pretty sure that the fix is wrong.

https://github.com/contao/core-bundle/blob/e4f0d29c314d22ca2ac235ff938027f48aaf985b/src/EventListener/UserSessionListener.php#L79-L85

The user object is incomplete upon login, because it has not been created via `User::loadUserByUsername()` but via `User::unserialize()`. That's why `$user->session` is empty.

* My approach ensures that `$user->session` is set upon login.

* Your approach ensures that the empty session is not stored in the database.

So in your approach, the session will be empty for the login request! And we don't know if there are other listeners (e.g. storing the referer?) that also rely on the session not being empty?

aschempp:
> The user object is incomplete upon login, because it has not been created via User::loadUserByUsername() but via User::unserialize().

This cannot be true. First of all, the login process obviously does not have a token in the session, otherwise a login would not be required. So there is nothing to be unserialized or incomplete. **After** a login the user is unserialized from the session, but the `refreshUser` method will fetch the database object again before the firewall is successful.

leofeyer:
I have dumped the user object in line 86:

https://github.com/contao/core-bundle/blob/e4f0d29c314d22ca2ac235ff938027f48aaf985b/src/EventListener/UserSessionListener.php#L86

```
Array
(
    [id] => 1
    [tstamp] => 1514978172
    [username] => k.jones
    [name] => Kevin Jones
    [email] => k.jones@example.com
    [language] => de
    [backendTheme] => flexible
    [uploader] => FileUpload
    [showHelp] => 1
    [thumbnails] => 1
    [useRTE] => 1
    [useCE] => 1
    [password] => $2y$10$mASa6zAiFYfIWirQbLepx.q5B4t52o9KU8NwOIGBY6mfF6..emjbK
    [pwChange] => 
    [admin] => 1
    [groups] => 
    [inherit] => 
    [modules] => 
    [themes] => 
    [pagemounts] => Array
        (
        )

    [alpty] => 
    [filemounts] => Array
        (
            [0] => a:0:{}
        )

    [fop] => 
    [forms] => 
    [formp] => 
    [disable] => 
    [start] => 
    [stop] => 
    [session] => Array
        (
        )

    [dateAdded] => 1257428510
    [lastLogin] => 1515481605
    [currentLogin] => 1515481617
    [loginCount] => 3
    [locked] => 0
    [calendars] => 
    [calendarp] => 
    [calendarfeeds] => 
    [calendarfeedp] => 
    [news] => 
    [newp] => 
    [newsfeeds] => 
    [newsfeedp] => 
    [newsletters] => 
    [newsletterp] => 
    [faqs] => 
    [faqp] => 
    [imageSizes] => 
    [amg] => 
    [fullscreen] => 
)
```

You are right that the object is not incomplete (e.g. `dateAdded` is not serialized). But why is there no session data then?

aschempp:
Maybe the session data is simply not unserialized?

leofeyer:
The session data is never serialized so it cannot be unserialized, can it?

aschempp:
It sure is serialized, `tl_user.session` is a database field?

leofeyer:
I'm talking about this:

https://github.com/contao/core-bundle/blob/e4f0d29c314d22ca2ac235ff938027f48aaf985b/src/Resources/contao/library/Contao/User.php#L642-L645

leofeyer:
I have added the changes (with unit test) in 5a721c6205a2de7b529c8fcd723ad43557eb886c. Although it works, I don't like that the listener is now stateful. Is this ok for us?

aschempp:
The only other option I can think of is to only register the `kernel.response` event if `kernel.request` is called. That's how the firewall works (e.g. `ContextListener`).

leofeyer:
Brilliant. Changed in 82b22c993f1d008158d5255a6085e31022426a19.

PR #1300, opened 2018-01-10
Use the BINARY flag instead of COLLATE *_bin

### LINKED ISSUE #1286: Allow %s_bin in SQL field definitions
<a href="https://github.com/fritzmg"></img></a> [Comment](https://github.com/contao/newsletter-bundle/issues/13#issuecomment-355225031) by @fritzmg
January 4th, 2018, 09:29 GMT

Btw. why do all the `alias` fields use a hardcoded `utf8mb4_bin` collation now anyway? Even if you set
```yml
doctrine:
    dbal:
        connections:
            default:
                default_table_options:
                    charset: utf8
                    collate: utf8_unicode_ci
```
the collation of these fields would still be `utf8mb4_bin`.

leofeyer:
This PR fixes #1286 as discussed with @aschempp. It handles the following cases:

* `'sql' => "varchar(128) BINARY NOT NULL default ''"`
* `'sql' => ['type' => 'string', 'length' => 128, 'not_null' => false, 'default' => '', 'customSchemaOptions' => ['case_sensitive' => true]]`

### DIFF
--- a/src/Doctrine/Schema/DcaSchemaProvider.php
+++ b/src/Doctrine/Schema/DcaSchemaProvider.php
@@ -63,6 +63,8 @@ class DcaSchemaProvider
      * Adds the DCA data to the Doctrine schema.
      *
      * @param Schema $schema
+     *
+     * @throws \LogicException
      */
     public function appendToSchema(Schema $schema): void
     {
@@ -78,6 +80,10 @@ class DcaSchemaProvider
 
                     // Use the binary collation if the "case_sensitive" option is set
                     if ($this->isCaseSensitive($config)) {
+                        if (isset($options['platformOptions']['collation'])) {
+                            throw new \LogicException('Cannot handle both a collation and the case_sensitive flag.');
+                        }
+
                         $options['platformOptions']['collation'] = $this->getBinaryCollation($table);
                     }
 
@@ -176,6 +182,8 @@ class DcaSchemaProvider
      * @param Table  $table
      * @param string $columnName
      * @param string $sql
+     *
+     * @throws \LogicException
      */
     private function parseColumnSql(Table $table, string $columnName, string $sql): void
     {
@@ -207,7 +215,11 @@ class DcaSchemaProvider
         }
 
         // Use the binary collation if the BINARY flag is set (see #1286)
-        if (preg_match('/\bbinary\b/i', $def, $match)) {
+        if (0 === strncasecmp($def, 'binary ', 7)) {
+            if (null !== $collation) {
+                throw new \LogicException('Cannot handle both a collation and the binary flag.');
+            }
+
             $collation = $this->getBinaryCollation($table);
         }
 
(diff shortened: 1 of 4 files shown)


src/Doctrine/Schema/DcaSchemaProvider.php
```diff
@@ -190,6 +206,11 @@ private function parseColumnSql(Table $table, string $columnName, string $sql):
             $collation = $match[1];
         }
 
+        // Use the binary collation if the BINARY flag is set (see #1286)
+        if (preg_match('/\bbinary\b/i', $def, $match)) {
```
aschempp: Should this be supported for non-VARCHAR fields? Also, if a collation is already set, I assume it should not be overridden?

src/Doctrine/Schema/DcaSchemaProvider.php
```diff
@@ -387,8 +408,8 @@ private function getSqlDefinitions(): array
     private function getMaximumIndexLength(Table $table, string $column): int
     {
         $indexLength = $this->getDefaultIndexLength($table);
+        $collation = $table->getOption('collate');
```
aschempp: Are you sure this is correct? The documentation says this should be `collation` (http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html#vendor-specific-options). Not sure if that's for fields only though.

leofeyer: Yes, I am sure. The default table options are merged:

src/Doctrine/Schema/DcaSchemaProvider.php
leofeyer: > Should this be supported for non-VARCHAR fields?

Yes, of course. It should be supported for every field that supports collations (usually all text fields).

> Also, if a collation is already set, I assume it should not be overridden?

I don't think so. Having both `COLLATE` and `BINARY` is a configuration error.

src/Doctrine/Schema/DcaSchemaProvider.php
m-vo: I'd stumbled across this as well - ``collation`` for fields ``colate`` for tables, extremely unintuitive... :-/

src/Doctrine/Schema/DcaSchemaProvider.php
aschempp: But then we should invert the if's, so that `COLLATE` overrides `BINARY` imho.

leofeyer: No. If the `BINARY` flag is set, it means "use the binary collation". Doesn't matter what collation is given IMHO.

But why don't we just throw an exception? It is a configuration error.

leofeyer: Changed in d8f4653d71baba848b2b4b598dd2dc7d5a341e14.

leofeyer: Nope, we should not throw an error message but do the same as MySQL does if we execute the following query:

```sql
ALTER TABLE `tl_article` ADD `test` VARCHAR(255) BINARY COLLATE utf8mb4_unicode_ci NOT NULL;
```

This is how the field is created:



The `binary` flag has precedence over the `collate` function, so the initial code was correct.

Issue #1305, opened 2018-01-11
Add a service for alias generation


m-vo:
There is a save callback `generateAlias` registered in `tl_article`, `tl_page` and `tl_form` with quite some repeated code. And I keep repeating myself aswell when I'm using aliases in my  extensions. :-)

What do you think about adding a generic service as a replacement? I think that could be handy.

```yaml
   services:
        contao.datacontainer.generate_alias:
            class: Contao\DataContainer\AliasGenerator
            arguments: ['@contao.slug.generator', 'title']
```

There would be the need to differentiate a bit of course (e.g. `generatePageAlias` vs. `generateArticleAlias`...) or a inspection of the dca parameter. The source field could be injected as a constructor parameter. If someone needs custom functionality it's easy to extend the class and register a new service. 

(Btw.: I'm pro moving the callback code out of the dca files. Are there any BC breaks involved in doing so?)

aschempp:
You can‘t remove existing code from DCA‘s for BC reasons, but we can deprecate and stop using it.

m-vo:
I see. So the callbacks are considered API?

fritzmg:
Someone might use
```php
class tl_page_foo extends tl_page
{
    …
}
```
for example. Thus you cannot remove these classes and functions.

m-vo:
(Ok, what about moving all classes out of the dca files into their respective files in the `DataContainer` namespace (e.g. `..\DataContainer\Page.php`). And then for BC we alias them and trigger `class_exist` in the original location or extend the old ones to use the new? Imho this would make further refactoring easier. If we want that I can make a PR but I'm afraid that I'm overlooking something.)

But for this topic: What do you think about creating services that'll replace the callback functions?

fritzmg:
> But for this topic: What do you think about creating services that'll replace the callback functions?

I think this is the recommended way now anyway. Any extension should do it this way. It's just that the core does not for backwards compatibility.

m-vo:
I know, but can we safely exchange `'save_callback' => array (array('tl_page', 'generateAlias'))` with `'save_callback' => [['contao.datacontainer.page', 'generateAlias']]`  + service declaration? Of course replacing the old function body with a service call + adding a deprecation notice if the service declaration does not point to the same file as before.

Or do we need to cater for the case that someone relies on the exact representation of the callback func array?

fritzmg:
[...]

PR #1314, opened 2018-01-17
[POC] Refactor DCA classes

m-vo:
Here is a draft of how a refactoring of the old DCA classes could look like. I'd like to see the actual code that sits in the DCA files right now (mainly callbacks) being moved into it's dedicated listeners, maybe refactoring it on the go (take this as a follow up to the discussion here #1305).

To see how this could work out, I made this POC for the `tl_theme` class as this is a rather small one. If we like the concept I'm happy to add the other ones as well. :-) 

Here is what I did:
 - rewrote the DCA definition to use an event listener (`contao.listener.datacontainer.theme`)
 - moved the implementations to a new class `EventListener\DataContainer\Theme`
 - deprecated the `tl_theme` class and it's functions
 - rewired the `tl_theme`  functions to use the new listener funtions (BC)
(I did not move the class out of the dca file yet to prevent the diff being more messed up as it is already. But I think the best would be to put it somewhere else like in a `deprecated.php` so that people reading the code won't get confused.)

I also did refactor the code because I think that should be done before re-releasing legacy code in the wild (however this might be better suited in another PR) and might not yet be enough/up to best practice/ [...].

Nevertheless, that's what I did:
 - got rid of the `Backend` base class
 - injected dependencies where possible
 - did some refactoring for simplicity/clarity
    - removed duplicate code (rendering buttons)
    - flattened logic by inverting ifs / removing ternary if
    - replaced large string concats with `sprintf()`

Still there are some things that would need some action like the call to `updateStyleSheets` that either should be made static or injectable. Or the former protected method `doGetTemplateFolders` that either needs to have it's code copied to the respective deprecated method or needs to be made public (BC).

Another question would be if adding type constraints to the methods is considered a breaking change. If so there will be some logic / casts needed in the rewired (deprecated) functions as a compat layer.

But all in all it's a good bit cleaner, isn't it? Wdyt? Tests now should be feasible, I guess. And it would be one step further to config-only DCAs..

---

(btw., as you see: I'm proof-of-concepting myself through the codebase until I've learned enough or anyone tells me to please stop because I annoy the hell out of you. :grimacing: )

### DIFF
--- a/src/EventListener/DataContainer/Theme.php
+++ b/src/EventListener/DataContainer/Theme.php
@@ -15,7 +15,7 @@ namespace Contao\CoreBundle\EventListener\DataContainer;
 use Contao\Backend;
 use Contao\BackendUser;
 use Contao\CoreBundle\Exception\AccessDeniedException;
-use Contao\CoreBundle\Image\ImageFactory;
+use Contao\CoreBundle\Image\ImageFactoryInterface;
 use Contao\FilesModel;
 use Contao\Image;
 use Contao\StringUtil;
@@ -35,8 +35,8 @@ class Theme
     /** @var BackendUser */
     protected $user;
 
-    /** @var ImageFactory */
-    protected $image;
+    /** @var ImageFactoryInterface */
+    private $imageFactory;
 
     /**
      * Theme constructor.
@@ -44,18 +44,18 @@ class Theme
      * @param RequestStack          $requestStack
      * @param SessionInterface      $session
      * @param TokenStorageInterface $tokenStorage
-     * @param ImageFactory          $image
+     * @param ImageFactoryInterface $imageFactory
      */
     public function __construct(
         RequestStack $requestStack,
         SessionInterface $session,
         TokenStorageInterface $tokenStorage,
-        ImageFactory $image
+        ImageFactoryInterface $imageFactory
     ) {
         $this->requestStack = $requestStack;
         $this->session      = $session;
         $this->user         = $tokenStorage->getToken()->getUser();
-        $this->image        = $image;
+        $this->imageFactory = $imageFactory;
     }
 
     /**
@@ -108,7 +108,7 @@ class Theme
             return $label;
         }
 
-        $imageUrl = $this->image
+        $imageUrl = $this->imageFactory
             ->create(TL_ROOT . '/' . $objFile->path, [75, 50, 'center_top'])
             ->getUrl(TL_ROOT);
 
--- a/src/EventListener/DataContainer/Theme.php
+++ b/src/EventListener/DataContainer/Theme.php
@@ -1,0 +1,379 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2018 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\EventListener\DataContainer;
+
+use Contao\Backend;
+use Contao\BackendUser;
+use Contao\CoreBundle\Exception\AccessDeniedException;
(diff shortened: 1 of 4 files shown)


src/EventListener/DataContainer/Theme.php
```diff
@@ -0,0 +1,379 @@
...
+     *
+     * @throws AccessDeniedException
+     */
+    public function onCheckPermission(): void
+    {
+        if ($this->user->isAdmin) {
+            return;
+        }
+
+        $request = $this->requestStack->getCurrentRequest();
```
Toflar: Can be `null` e.g. on command line when building DCA cache etc.

src/EventListener/DataContainer/Theme.php
```diff
@@ -0,0 +1,379 @@
...
+     *
+     * @return string
+     */
+    public function onAddPreviewImage(array $row, string $label): string
+    {
+        if (!$row['screenshot']) {
+            return $label;
+        }
+
+        $objFile = FilesModel::findByUuid($row['screenshot']);
```
Toflar: In case you want to add unit tests, you need to be able to mock this. Inject `ContaoFrameworkInterface $framework` (you can also use setter injection because if possible we'd like to drop that one day): 
`$this->framework->getAdapter('Contao\FilesModel')->findByUuid($row['screenshot']);`.

src/EventListener/DataContainer/Theme.php
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+
+    /**
+     * Check for modified style sheets and update them if necessary
+     */
+    public function onUpdateStyleSheet(): void
+    {
+        if ($this->session->get('style_sheet_update_all')) {
+            // todo: make updateStyleSheets() static
```
Toflar: Why static? In the best case, the Stylesheets handling would reside in its own services too 😄

src/Resources/contao/dca/tl_theme.php
```diff
@@ -230,260 +230,293 @@
 /**
  * Provide miscellaneous methods that are used by the data configuration array.
  *
- * @author Leo Feyer <https://github.com/leofeyer>
+ * @author     Leo Feyer <https://github.com/leofeyer>
+ *
+ * @deprecated Deprecated since Contao 4.6, to be removed in Contao 5.0
+ *             Use the datacontainer.theme listener instead.
```
Toflar: Not sure if we should write this. To me, an event listener is not public API. I never rely on a listener to be reusable for me. No idea why one would rely on this to be honest.

fritzmg: Consider the following example: https://github.com/inspiredminds/contao-sibling-navigation/blob/1.0.1/src/Resources/contao/dca/tl_module.php#L41-L49
This extension overrides the `options_callback` of the `news_order` field of `tl_module`, returns custom options based on certain conditions and otherwise executes the original callback function.

If that is not possible then you would always have to copy the original code into your callback, which seems unnecessary.

src/EventListener/DataContainer/Theme.php
m-vo: You're right! I'll add a check. I guess the same holds true for the user and I should better use `$user = $this->token->getUser();` in the functions?

Toflar: Yes :)

aschempp:
I would generally agree with your rewrite concept, however: we agreed not to rewrite the old code, because it does not really make any sense. Same as we agreed not to add events for all hooks. There is no advantage in the rewrite, except moving stuff around and possibly breaking things (which we can't really test). The callbacks are also related to the old `DC_Table`, so if we ever refactor everything we should first work on a new DataContainer that allows actual event listeners etc...

m-vo:
So I guess I'm closing this, then?

PR #1315, opened 2018-01-17
remove mandatory from html

fritzmg:
I often use `ce_html_*` templates for custom PHP code or HTML that is inserted on just one page as a content element. However, since the field is mandatory, I also always have to insert some content into the `html` field, which will then never be rendered in the front end anyway in the custom templates (unless you specifically do that of course).

In `tl_module` the field isn't mandatory either. Wouldn't it be more consistent to not make it mandatory in `tl_content` too?

### DIFF
--- a/src/Resources/contao/dca/tl_content.php
+++ b/src/Resources/contao/dca/tl_content.php
@@ -314,7 +314,7 @@ $GLOBALS['TL_DCA']['tl_content'] = array
 			'exclude'                 => true,
 			'search'                  => true,
 			'inputType'               => 'textarea',
-			'eval'                    => array('mandatory'=>true, 'allowHtml'=>true, 'class'=>'monospace', 'rte'=>'ace|html', 'helpwizard'=>true),
+			'eval'                    => array('allowHtml'=>true, 'class'=>'monospace', 'rte'=>'ace|html', 'helpwizard'=>true),
 			'explanation'             => 'insertTags',
 			'sql'                     => "mediumtext NULL"
 		),


leofeyer:
I'd rather consider it a bug that it is not mandatory in `tl_module`. What's the point of adding an HTML element without any code?

fritzmg:
> What's the point of adding an HTML element without any code?

In order to integrate a `mod_html_*` or `ce_html_*` template. Please don't make it mandatory in tl_module ;)

leofeyer:
You could also write `{{file::my_custom_html_content.html5}}` into the HTML field instead of choosing a template from the select menu.

> Please don't make it mandatory in tl_module ;)

I cannot promise …

fritzmg:
> You could also write {{file::my_custom_html_content.html5}} into the HTML field instead of choosing a template from the select menu.

No, in that case `my_custom_html_content.html5` is included as a _PHP script_ and not _parsed as a template _in the template object context. It is not the same thing.

m-vo:
I actually like adding elements as placeholders and fill them later. Being forced to enter something was against my workflow ever since.

ausi:
I think the use case of @fritzmg sounds interesting. How about adding a new content element and module that’s called something like “Template”? We could also add a key/value wizard to set variables for the template.

asaage:
i use the content to provide a description of what the custom-template will show/do
sometimes i even end up adding an empty article with a custom-template instead of a content-element (for un/publishing convenience)
A new content-element of type "Template" wouldn't help in this case but i like the idea of @ausi.

fritzmg:
> I think the use case of @fritzmg sounds interesting.

I was under the impression this was widely used anyway since Contao 3.5 (or whichever version introduced the element/module template selection), at least under the more advanced users/devs ;). Contao 4 even improved this by only allowing templates to be selected that start with `mod_html` or `ce_html`, so that it's more clear which custom templates are at your disposal.

> How about adding a new content element and module that’s called something like “Template”? We could also add a key/value wizard to set variables for the template.

Yes, before Contao 3.5 came out, I wanted to make such a feature request. However now that you are able to integrate a custom template this way, I don't see much reason to do so anymore. Because if you need additional variable input, I always use your RSCE extension instead.

Though I think it wouldn't hurt to add a "Template" module and content element, where you have a simple `optionsWizard` or something like that for key value pairs that you can pass to the template.

> i use the content to provide a description of what the custom-template will show/do

Me too. Sometimes I find it unnecessary or superfluous though.

leofeyer:
Merged in contao/contao@add20822d016d6413eb9f65fde87f83f8b27185d.

PR #1323, opened 2018-01-18
[RTM] Enable access to nested keys in ArrayAttributeBag

ausi:
Fixes write operations on nested keys, e.g. `$_SESSION['FE_DATA']['foo']['bar'] = 'x';`.

Related #1316


leofeyer:
Thank you @ausi.

PR #1332, opened 2018-01-21
change isset to array_key_exists

dominixdesign:
isset returns _false_ if the array key value is _NULL_. This is an issue by updating fields with value _NULL_, because they don't get marked as modified.
At this point it is only necessary to check if this key is present in those arrays. The value doesn't matter.

Bonus: No NOTICE logs for undefined array index.

### DIFF
--- a/src/Resources/contao/library/Contao/Model.php
+++ b/src/Resources/contao/library/Contao/Model.php
@@ -394,7 +394,7 @@ abstract class Model
 	 */
 	public function markModified($strKey)
 	{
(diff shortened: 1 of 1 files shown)


fritzmg:
You should probably prepend ith with `\` for performance reasons.

leofeyer:
This has already been fixed (see https://github.com/contao/core-bundle/commit/08a04c96aac825145090adbd3814994b346b5a97).

PR #1336, opened 2018-01-24
add download attribute to download elements

frontendschlampe:
prevent errors like:
Resource interpreted as Document but transferred with MIME type application/pdf: "*?file=*.pdf". (Chrome) or [Error] Failed to load resource: Das Laden des Frames wurde unterbrochen (*, line 0) *?file=*.pdf (Safari)
see more https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download

### DIFF
--- a/src/Resources/contao/templates/elements/ce_download.html5
+++ b/src/Resources/contao/templates/elements/ce_download.html5
@@ -3,7 +3,7 @@
 <?php $this->block('content'); ?>
 
   <p class="download-element ext-<?= $this->extension ?>">
-    <a href="<?= $this->href ?>" title="<?= $this->title ?>"><?= $this->link ?> <span class="size">(<?= $this->filesize ?>)</span></a>
+    <a href="<?= $this->href ?>" title="<?= $this->title ?>" download><?= $this->link ?> <span class="size">(<?= $this->filesize ?>)</span></a>
   </p>
 
 <?php $this->endblock(); ?>
--- a/src/Resources/contao/templates/elements/ce_downloads.html5
+++ b/src/Resources/contao/templates/elements/ce_downloads.html5
@@ -5,7 +5,7 @@
   <ul>
     <?php foreach ($this->files as $file): ?>
       <li class="download-element ext-<?= $file['extension'] ?>">
-        <a href="<?= $file['href'] ?>" title="<?= $file['title'] ?>"><?= $file['link'] ?> <span class="size">(<?= $file['filesize'] ?>)</span></a>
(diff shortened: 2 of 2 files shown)


frontendschlampe:
should be included to Contao 4.4, 4.5 and future versions

ausi:
Where do you get this error?
We are using the `Content-Disposition` header in https://github.com/contao/core-bundle/blob/5600248fdbeebc87bf068458facfc0ddb6954d97/src/Resources/contao/library/Contao/File.php#L788-L793 which should prevent such errors.

frontendschlampe:
> Where do you get this error?

What do you mean? The error is browserconsole after/while downloading the element. Without the download-attribut, I got this error. With download-attribut there's no error. I also checked with @discordier on his Chrome browser and he has the same problems.

ausi:
I was not able to reproduce it with Chrome 63.
Do you geht this error on the contao demo too?

frontendschlampe:
> Do you geht this error on the contao demo too?

yes

`[Error] Failed to load resource: Das Laden des Frames wurde unterbrochen (file-elements.html, line 0) https://demo.contao.org/en/file-elements.html?file=files/contaodemo/media/content-images/DSC_5276.jpg`

asaage:
Cant't reproduce this either with chrome or firefox on windows.
I did however get an error more than a year ago https://github.com/contao/core/issues/8375#issuecomment-265978339
Imo the download elements (and attachments) should be served in response with content-disposition: attachment to a form-post-request to prevent hotlinking and indexing.

ausi:
> `[Error] Failed to load resource: Das Laden des Frames wurde unterbrochen (file-elements.html, line 0)`

This seems to be a bug in Safari: [#46491](https://bugs.webkit.org/show_bug.cgi?id=46491)

leofeyer:
@frontendschlampe 681 files changed? Please rebase the PR. Or close it, since it seems to be a Safari bug.

frontendschlampe:
o.O ... sorry ... my fault. I will rebase.

I don't know, if it's a bug or a feature. @discordier and @ausi discussed it in Mumble this week, but I don't know the final conclusion.

leofeyer:
Before you rebase, let's wait what the discussion on Thursday brings, ok?

frontendschlampe:
> Before you rebase, let's wait what the discussion on Thursday brings, ok?

yes ... no problem, I have already rebased! ;-)

leofeyer:
As discussed in Mumble on Thursday 15th, we don't want to add the download attribute, because the error seems to be caused by a bug in Chromium on Linux (see https://groups.google.com/forum/#!topic/gsutil-discuss/Dye3xalLb9A).

Toflar:
BTW, just found this: https://bugs.chromium.org/p/chromium/issues/detail?id=9891
So looks like a debug toolbar issue of Chromium.

PR #1341, opened 2018-01-26
require terminal42/header-replay-bundle ^1.3 instead of 1.3.*

fritzmg:
See https://github.com/composer/composer/issues/7042

Why was it changed to `1.3.*` anyway?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "symfony/swiftmailer-bundle": "^2.3",
         "symfony/yaml": "^3.3",
         "tecnickcom/tcpdf": "^6.0",
-        "terminal42/header-replay-bundle": "1.3.*",
+        "terminal42/header-replay-bundle": "^1.3",
         "true/punycode": "^2.1",
         "twig/twig": "^1.26",
         "webmozart/path-util": "^2.0"


m-vo:
That's related to the reverts with the session issue I guess. IIRC version 1.4 fixed the problems with security (making the request a HEAD request), 1.4.1 reissues a request for invalid responses.

fritzmg:
Hm, I see. Then the `contao/managed-edition` for example should probably reintroduce the
```
"conflict": {
    "contao/core-bundle": "<4.4.13"
}
```
into its `composer.json`, so that it prevents https://github.com/composer/composer/issues/7042

Why was _that_ removed anyway? 😁

leofeyer:
> Why was it changed to `1.3.*` anyway?

Because version 1.4 is not compatible with Contao 4.4.

> Why was that removed anyway?

That should have been removed in Contao 4.5 only.

leofeyer:
Re-added in 3742b022cd3ffbd882da7b37869ee1514eb066f5.

fritzmg:
Shouldn't it be
```
"conflict": {
    "contao/core-bundle": "<4.4.13"
}
```
?

leofeyer:
Why should it? The security issue has been fixed in version 4.4.9.

fritzmg:
Well the issue was that even with the most recent Contao Version two installable sets of packages were possible, thus a composer updates switches back and forth between 4.4.12 and 4.4.13. If the managed-edition 4.4.14 only added
```
"conflict": {
    "contao/core-bundle": "<4.4.9"
}
```
wouldn't composer still switch between 4.4.12 and 4.4.14?

leofeyer:
This should be fixed in a540d1fb9c904bc4b18faa7aa784a256ca3c0eee now (also see terminal42/header-replay-bundle#8).

Issue #1366, opened 2018-02-12
Cannot install contao 4.5 alone


baumannsven:
`
    - contao/core-bundle 4.5.0 requires terminal42/header-replay-bundle ^1.3 -> satisfiable by terminal42/header-replay-bundle[1.3.0, 1.4.0, 1.4.1].
    - contao/core-bundle 4.5.1 requires terminal42/header-replay-bundle ^1.3 -> satisfiable by terminal42/header-replay-bundle[1.3.0, 1.4.0, 1.4.1].
    - contao/core-bundle 4.5.2 requires terminal42/header-replay-bundle ^1.4 -> satisfiable by terminal42/header-replay-bundle[1.4.0, 1.4.1].
    - contao/core-bundle 4.5.3 requires terminal42/header-replay-bundle ^1.4 -> satisfiable by terminal42/header-replay-bundle[1.4.0, 1.4.1].
    - terminal42/header-replay-bundle 1.3.0 requires friendsofsymfony/http-cache 2.* -> satisfiable by friendsofsymfony/http-cache[2.0.0, 2.0.1, 2.0.2, 2.1.0, 2.1.1, 2.1.2].
    - terminal42/header-replay-bundle 1.4.0 requires friendsofsymfony/http-cache 2.* -> satisfiable by friendsofsymfony/http-cache[2.0.0, 2.0.1, 2.0.2, 2.1.0, 2.1.1, 2.1.2].
    - terminal42/header-replay-bundle 1.4.1 requires friendsofsymfony/http-cache 2.* -> satisfiable by friendsofsymfony/http-cache[2.0.0, 2.0.1, 2.0.2, 2.1.0, 2.1.1, 2.1.2].
    - friendsofsymfony/http-cache 2.1.2 requires php-http/client-implementation ^1.0.0 -> no matching package found.
    - friendsofsymfony/http-cache 2.1.1 requires php-http/client-implementation ^1.0.0 -> no matching package found.
    - friendsofsymfony/http-cache 2.1.0 requires php-http/client-implementation ^1.0.0 -> no matching package found.
    - friendsofsymfony/http-cache 2.0.2 requires php-http/client-implementation ^1.0.0 -> no matching package found.
    - friendsofsymfony/http-cache 2.0.1 requires php-http/client-implementation ^1.0.0 -> no matching package found.
    - friendsofsymfony/http-cache 2.0.0 requires php-http/client-implementation ^1.0.0 -> no matching package found.
[...]

Issue #1364, opened 2018-02-11
cant find Route contao/install 


easycalib-git:
Hi:
i hope you can help me,
after installing core-bundle 4.5.3 on my Symfony 3.4 applications, and followed all the instructions given in the main page, i am unable to access the route contoa/install, i checked the debug routes, and it doenst show up
![image](https://user-images.githubusercontent.com/33333099/36077917-c0c60fd2-0f70-11e8-972a-53b0d535bc26.png)

fritzmg:
You also need the contao/installation-bundle

aschempp:
If you want to use the install tool, you need to add the installation-bundle. Otherwise you can update the database using the Doctrine command or add Doctrine Migrations to your application. The Contao database changes are added to the Doctrine Schema tool (even though we're not actually using Doctrine entities).

bytehead:
This should be mentioned as installation/configuration optionin the `README.md`.

PR #1367, opened 2018-02-12
[RFC] Updated installation documentation

bytehead:
See discussions/questions in https://github.com/contao/core-bundle/issues/1364 and https://github.com/contao/core-bundle/issues/1366.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,23 @@ Edit your `composer.json` file and add the following:
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


fritzmg:
I think a description on how to update the database without the Contao Install Tool (installation-bundle) would be nice too :)

leofeyer:
@bytehead I would prefer it the other way round. Recommend all three by default and add notes that another client implementation can be used and that the install tool can be omitted (sometime in the future).

Do you want me to change the text accordingly?

bytehead:
👍

leofeyer:
Thank you @bytehead.

Issue #1246, opened 2017-12-18
http_cache not working


Metis77:
after updating to 4.4.9 var/cache/prod/http_cache stopped working.
with 4.4.6 it still worked.

"Systemwartung" say 0 cached files and the folder http_cache stays empty.

tested in multiple websites, but same server.

bytehead:
I've encountered the same.

Toflar:
Can you test with 4.4.8 or does "with 4.4.6 it still worked" mean, that's the last one it worked with?

Metis77:
just tested:
4.4.9 - not working
4.4.8 - not working
4.4.7 - I dont know
4.4.6 - worked

Toflar:
Ok, I'll let you find out if 4.4.7 works then.

Metis77:
well, updating from a Contao 4.4.4 version to 4.4.7 specifically failed using contao manager oO.
maybe another issue.

Without beeing able to install specific version with contao manager, I cant test right now.

Toflar:
Okay after an hour of debugging, I found the issue. It was introduced by @aschempp in https://github.com/contao/core-bundle/commit/cae98ce3ec038982f47167264ab1c3464ee3b282#diff-6c590ef7e538a1f4a9115824540acb51. 
I think merging the `Cache-Control` header needs special handling. 

In my case, the `$response` correctly contained the `Cache-Control` header `public, s-maxage=10800`. In `$this->headers` the `Cache-Control` header contains `Cache-Control: max-age=0, private, must-revalidate`. The end result is `max-age=0, must-revalidate, private, public, s-maxage=10800` which is considered to be `private` and thus nothing is cached.
I have no idea why the `Cache-Control` header is even part of `headers_list()`. I debugged using `headers_list()` with the two arguments to debug who set the header but the result was an empty string and `0` which tells me it's coming natively from the server. 

There's certainly some kind of bug in Symfony because when merging the `Cache-Control` headers it shouldn't be `private` and `public` at the same time. But also I wonder if it ever makes sense to merge a cache header?

Toflar:
It's wrong that `public` is still there but basically it does the right thing. It merges the `private` instruction from `headers_list()` onto the response which turns it `private`. I just have no clue why there's a default `Cache-Control` header present 😄

Toflar:
[...]

Issue #1283, opened 2018-01-03
Session handling in Contao 4.4


leofeyer:
I am referring to #1260, which has solved the legacy session access problem for Contao 4.5. However, it seems we have not fixed it in Contao 4.4 yet, have we?

@Toflar /cc

Toflar:
I think we have to wait for Symonfy because there are yet more changes to that. Currently I guess all of the reverse proxy caching is just simply broken and we cannot do anything about it until the situation is clarified.

leofeyer:
This ticket is not about the broken HTTP cache caused by the session changes in Symfony. It is about legacy session handling (starting the session as soon as someone accesses `$_SESSION`).

Don't we have to always start the session when the framework is initialized in Contao 4.4?

aschempp:
We should, but I think it was only by accident.

leofeyer:
As discussed in Mumble on Januar 18th, 2018, we want to add a response listener that makes the response public depending on the page settings and that is triggered after the one that automatically makes the response private. The listener should also check if the response is cacheable.

leofeyer:
The listener needs to be removed again in Contao 4.5.

leofeyer:
> The listener needs to be removed again in Contao 4.5.

After further analyzing of the session usage in Contao, I believe we also need the listener in version 4.5. There is a lot of

https://github.com/contao/comments-bundle/blob/bc9c347dbfe722352082605fad17f9bab08cfe5d/src/Resources/contao/classes/Comments.php#L285-L296

and ignoring our cache logic – may it be good or bad – would break BC.

Toflar:
The logic is also needed for FOSHttpCacheBundle's UserContextListener. I'll check with @dbu what we can do here.

leofeyer:
I have reverted the changes, because they feel wrong. We have to thoroughly discuss cache and session handling on the next developer's meeting.

aschempp:
So it will stay broken for another month?

leofeyer:
The reverted changes do not fix the issue, either, so yes, sadly.

leofeyer:
The best intermediate fix is to require Symfony in version `3.3.*` instead of `^3.3`. The breaking changes were introduced in Symfony 3.4.

aschempp:
> The reverted changes do not fix the issue, either, so yes, sadly.

😞  can you explain why?

[...]

PR #1368, opened 2018-02-13
Fix http_cache of issue 1246

joeherold:
@leofeyer 
@Toflar 

This pull-request should fix the issue, with the not working http_cache.

Tested in my local environment and works like a charm.

The implementation also preserves the ability, to cache inserttags individually and do not cache insert tags with the pipe uncached or refresh.

This should fix the problem from issue http_cache not working #1246
(https://github.com/contao/core-bundle/issues/1246)

http_cache

### DIFF
--- a/src/EventListener/FinalCacheHeadersListener.php
+++ b/src/EventListener/FinalCacheHeadersListener.php
@@ -1,0 +1,220 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2018 Leo Feyer
+ *
+ * @contribute Johannes Pichler <j.pichler@webpixels.at>
+ *
+ * @description This Listener handles the new SessionListener handling from Symfony, what results into a bug
+ *              in Contao, because the creation of the Cache Files is Broken.
+ *
+ *              The listener is registered with a priority of -2038, results in beeing executed just before the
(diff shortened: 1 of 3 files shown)


leofeyer:
This is basically the same approach as #1283, which we have decided against.

joeherold:
@leofeyer great, @Toflar asked me to do so...
https://github.com/contao/core-bundle/issues/1246#issuecomment-365219991

> Yeah we might need to do so. But also I would like to make this dependent on the global $objPage, so essentially moving the FrontendTemplate::setCacheHeaders() into that listener. Would you like to work on a Pull Request for the Core-Bundle for that? That would be awesome

leofeyer:
@joeherold We have a core developer's meeting next week and this topic is on our agenda. Is it ok if we delay this discussion until after the meeting?

joeherold:
Sure. To not have stress I made the 3rd Party Bundle as a quick fix workaround. https://packagist.org/packages/pliigo/pagecacheenabler-bundle.

So anyone needs it now, can help himself by installing it via manager.

I just made it, because yannick asked to. feel free to discuss.

joeherold:
@Toflar Auf den Kaffee od. Bier freue ich mich dennoch :)

leofeyer:
See #1388 now.

PR #1376, opened 2018-02-18
[RTM] Draft of Contao controllers

aschempp:
Abstract controllers that implement the same features as the abstract `ContentElement` and `Module` classes did.

### DIFF
--- a/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/src/Controller/ContentElement/AbstractContentElementController.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2018 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace App\Controller\ContentElement;
+
+use Contao\CoreBundle\Controller\AbstractFragmentController;
+use Contao\ModuleModel;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+abstract class AbstractContentElementController extends AbstractFragmentController
+{
+    /**
+     * @param Request     $request
+     * @param ModuleModel $module
+     * @param string      $section
+     *
+     * @return Response
+     */
+    public function __invoke(Request $request, ModuleModel $module, string $section)
+    {
+        // TODO: define a permission name
+//        $this->denyAccessUnlessGranted('', $module);
+
+        $template = $this->createTemplate($module, 'ce_');
+
+        $template->inColumn = $section;
+
+        return $this->getResponse($template, $module, $request);
+    }
+}
--- a/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/src/Controller/ContentElement/AbstractContentElementController.php
@@ -21,10 +21,12 @@ use Symfony\Component\HttpFoundation\Response;
 abstract class AbstractContentElementController extends AbstractFragmentController
 {
     /**
+     * Invokes the controller.
+     *
      * @param Request      $request
      * @param ContentModel $model
      * @param string       $section
-     * @param string|null  $classes
+     * @param array|null   $classes
      *
      * @return Response
      */
@@ -41,6 +43,8 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
     }
 
     /**
+     * Returns the response.
+     *
      * @param Template|\stdClass $template
      * @param ContentModel       $model
      * @param Request            $request
--- a/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/src/Controller/ContentElement/AbstractContentElementController.php
@@ -21,14 +21,14 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
 {
     /**
      * @param Request      $request
-     * @param ContentModel $module
+     * @param ContentModel $model
      * @param string       $section
      *
      * @return Response
      */
-    public function __invoke(Request $request, ContentModel $module, string $section)
+    public function __invoke(Request $request, ContentModel $model, string $section)
     {
-        $template = $this->createTemplate($module, 'ce_');
+        $template = $this->createTemplate($model, 'ce_');
 
         $template->inColumn = $section;
 
@@ -36,6 +36,6 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
             $template->class .= ' '.implode(' ', $classes);
         }
 
-        return $this->getResponse($template, $module, $request);
+        return $this->getResponse($template, $model, $request);
     }
 }
--- a/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/src/Controller/ContentElement/AbstractContentElementController.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file is part of Contao.
  *
@@ -31,7 +33,7 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
         $template->inColumn = $section;
 
         if (is_array($classes = $request->attributes->get('classes'))) {
-            $template->class .= ' ' . implode(' ', $classes);
+            $template->class .= ' '.implode(' ', $classes);
         }
 
         return $this->getResponse($template, $module, $request);
--- a/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/src/Controller/ContentElement/AbstractContentElementController.php
@@ -33,6 +33,10 @@ abstract class AbstractContentElementController extends AbstractFragmentControll
 
         $template->inColumn = $section;
 
+        if (is_array($classes = $request->attributes->get('classes'))) {
+            $template->class .= ' ' . implode(' ', $classes);
+        }
+
         return $this->getResponse($template, $module, $request);
     }
 }
--- a/src/Controller/ContentElement/AbstractContentElementController.php
+++ b/src/Controller/ContentElement/AbstractContentElementController.php
@@ -10,7 +10,7 @@ declare(strict_types=1);
  * @license LGPL-3.0+
  */
 
-namespace App\Controller\ContentElement;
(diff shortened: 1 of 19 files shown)


leofeyer:
The PR should be targeted against the master branch.

aschempp:
I have just updated this PR to make our fragments actually useful. The problem of inline fragment renderer is that it creates a subrequest without any of the current request information. It create a GET request even if the master request is POST, respectively drops POST parameters and much more.

Our fragments work different than the Symfony core, that's why it makes sense to inherit the master request in our case. It is still very much possible to keep the existing behavior by simply overriding the renderer in the tag configuration.

Toflar:
> Our fragments work different than the Symfony core, that's why it makes sense to inherit the master request in our case. It is still very much possible to keep the existing behavior by simply overriding the renderer in the tag configuration.

That's correct. Reviewed and concept approved. But tests need to be fixed 😄

aschempp:
@contao/developers any other feedback? Can we agree that we want this implementation before I spend more time on fixing tests?

Toflar:
I've got it on my list, will review today :)

src/Controller/ContentElement/AbstractContentElementController.php
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @return Response
+     */
+    public function __invoke(Request $request, ContentModel $module, string $section)
+    {
+        $template = $this->createTemplate($module, 'ce_');
+
+        $template->inColumn = $section;
+
+        if (is_array($classes = $request->attributes->get('classes'))) {
```
Toflar: `\is_array()`

src/Controller/FrontendController/AbstractFrontendModuleController.php
```diff
@@ -0,0 +1,77 @@
...
+    protected function getBackendWildcard(ModuleModel $module, Request $request)
+    {
+        $href = $this->get('router')->generate(
+            'contao_backend',
+            ['do' => 'themes', 'table' => 'tl_module', 'act' => 'edit', 'id' => $module->id]
+        );
+
+        $template = new BackendTemplate('be_wildcard');
+
+        $template->wildcard = '### ' . strtoupper($GLOBALS['TL_LANG']['FMD'][$this->getType()][0]) . ' ###';
```
Toflar: Can't we use our translator here?

src/DependencyInjection/Compiler/RegisterFragmentsPass.php
```diff
@@ -77,6 +78,10 @@ protected function registerFragments(ContainerBuilder $container, string $tag):
                     $preHandlers[$identifier] = $reference;
                 }
 
+                if (is_a($definition->getClass(), AbstractFragmentController::class, true)) {
```
Toflar: That just kills the possibility to have a custom controller that needs to know the fragment options but does not want to extend the `AbstractFragmentController`. I mean, the purpose of the fragments is that they are fully independent controllers so I think we should work with a `FragmentOptionsAwareInterface` here, wdyt?

src/Fragment/ForwardFragmentRenderer.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Contao\CoreBundle\Fragment;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Fragment\InlineFragmentRenderer;
+
+class ForwardFragmentRenderer extends InlineFragmentRenderer
```
Toflar: Would be great if you could add some docs here and explain why this is needed.

dmolineus:
It's a great simplification of such attract controllers would be provided.

I'd prefer not to depend on the Contao template classes though. We could extend the symfony template engine wrapping the Contao templates. Then it's up to the developer to decide which template to use. Migrating later to twig templates would be really easy then.

For example following template syntax could be used:

 - contao:be:be_main.html5
 - contao:fe:fe_page.html5

I know it's off topic here but it would be nice to have this feature included when merging this pr.

If you like the idea I'd love to contribute the code. (I already do it in my toolkit library, https://github.com/netzmacht/contao-toolkit/blob/master/src/View/Template/ToolkitTemplateEngine.php)

src/Controller/ContentElement/AbstractContentElementController.php
qzminski: Should be a task of php-cs-fixer, no?

src/Controller/FrontendController/AbstractFrontendModuleController.php
```diff
@@ -0,0 +1,77 @@
...
+        return $this->get('contao.routing.scope_matcher')->isBackendRequest($request);
+    }
+
+    /**
+     * @param ModuleModel $module
+     * @param Request     $request
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    protected function getBackendWildcard(ModuleModel $module, Request $request)
```
qzminski: This method should actually be also available for content elements, as some of them may display dynamic data as well.

aschempp: It should not be in the parent class for sure, because the generated URL is different for different types. Also, none of the core elements have wildcards. What's the reason not to show dynamic data?

qzminski: And what's the reason to show backend wildcard for frontend modules?

aschempp: ¯\_(ツ)_/¯

I guess it's also a quick link to edit the module, which does not make sense for an element. There are content elements that change their appearance in the backend, but never to a wildcard. The YouTube element does not generate an iframe but a link to the video for example.

src/DependencyInjection/Compiler/RegisterFragmentsPass.php
discordier: I second that. We should use small(!) interfaces for these tasks.
Brings also the benefit of being able to deprecate one way (old interface) and introduce a new way without changing the API. I guess we will have to alter the mechanics in the future which is easier by adding another interface than having version_compares() and the like.

aschempp:
I have updated the PR. Most important change is to disable `ignore_errors` by default. If errors are ignores, it is not possible to handle `ResponseException` and fragments simply don't show up in the response. I noticed this because I had my own element that tried so use `Controller::sentFileToBrowser`.

Toflar:
Good find!

src/Controller/FrontendController/AbstractFrontendModuleController.php
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Contao\CoreBundle\Controller\FrontendModule;
+
+use Contao\BackendTemplate;
+use Contao\CoreBundle\Controller\AbstractFragmentController;
+use Contao\ModuleModel;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+abstract class AbstractFrontendModuleController extends AbstractFragmentController
```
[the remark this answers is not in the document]
aschempp: I dont think so, because there is also a backend module. Imho the old naming is „wrong“.

leofeyer:
Thank you @aschempp.

christian-kolb:
@leofeyer @aschempp @fritzmg @Toflar @qzminski @dmolineus Has someone of you already implemented a frontend module with this or the previous controller in a public repository? I think it would be great to have this linked from the 4.6.0 release blog post. This way more people would get a kickstart and we might see more bundles using it faster 🙂

bytehead:
Something like this?
```yml
# app/config/services.yml

services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false

    # Content Elements
    app.content_element.acme_element:
        class: AppBundle\ContentElement\AcmeElement
        tags:
            - { name: contao.content_element, category: myCategory }
        calls:
            - [ setContainer,[ '@service_container' ] ]

    # Modules
    app.frontend_module.acme_module:
        class: AppBundle\Module\AcmeModule
        tags:
            - { name: contao.frontend_module, category: myCategory }
        calls:
            - [ setContainer,[ '@service_container' ] ]
```

```php
// src/AppBundle/ContentElement/AcmeElement.php

<?php

declare(strict_types=1);

namespace AppBundle\ContentElement;

use Contao\CoreBundle\Controller\ContentElement\AbstractContentElementController;
use Contao\ModuleModel;
use Contao\Template;
use Symfony\Component\HttpFoundation\Request;

class AcmeElement extends AbstractContentElementController
{
    protected function getResponse(Template $template, ContentModel $model, Request $request): Response
    {
        $template->title = "AcmeElement";

        return new Response($template->parse());
    }
}
```

```php
// src/AppBundle/Module/AcmeModule.php

<?php

declare(strict_types=1);

namespace AppBundle\Module;

use Contao\CoreBundle\Controller\FrontendModule\AbstractFrontendModuleController;
use Contao\ModuleModel;
use Contao\Template;
use Symfony\Component\HttpFoundation\Request;

class AcmeModule extends AbstractFrontendModuleController
{
    protected function getResponse(Template $template, ModuleModel $model, Request $request): Response
    {
        $template->title = "AcmeModule";

        return new Response($template->parse());
    }
}
```

qzminski:
@bytehead looks good and similar to mine, I am just not sure if `setContainer` call in `services.yml` is really needed. I would rather use service injection here 🤔

bytehead:
Yep, you're right. Probably somehow a leftover...

bytehead:
They are required indeed: https://github.com/contao/contao/blob/master/core-bundle/src/Controller/FrontendModule/AbstractFrontendModuleController.php#L26

qzminski:
You sure? The class extends `Controller` which should already have container in it https://github.com/contao/contao/blob/master/core-bundle/src/Controller/AbstractFragmentController.php#L23

bytehead:
Yes, just tested. Probably because `AbstractFrontendModuleController` is `abstract`?

PR #1379, opened 2018-02-19
[RFC] Fix namespaced paths in Contao DataCollector

sheeep:
The Contao DataCollector used an outdated method for referencing templates which
seen to have been remove recently. Update to the modern @-prefixed way of referencing
templates.

Fixes the following error (occures when opening the debugger, or even just show the debug toolbar).

```
The profiler template "ContaoCore:Collector:contao.html.twig" for data collector "contao" does not exist.

UnexpectedValueException
in vendor/symfony/symfony/src/Symfony/Bundle/WebProfilerBundle/Profiler/TemplateManager.php (line 108)
```

> In the past, Symfony used a different syntax to refer to templates. This format, which uses colons (:) to separate each template path section, is less consistent and has worse performance than the Twig syntax. (https://symfony.com/doc/3.4/templating/namespaced_paths.html)

> [...] but you must prefix the @ character when using it in templates (that's how Twig can differentiate namespaces from regular paths). (https://symfony.com/doc/current/templating/namespaced_paths.html)

There is one open question (thus tagged as RFC). It does seem to work in the managed edition. Any idea why?

### DIFF
--- a/src/Resources/config/services.yml
+++ b/src/Resources/config/services.yml
@@ -61,7 +61,7 @@ services:
         arguments:
             - "%kernel.packages%"
         tags:
-            - { name: data_collector, template: "ContaoCoreBundle:Collector:contao", id: "contao" }
+            - { name: data_collector, template: "@ContaoCore/Collector/contao.html.twig", id: "contao" }
 
     contao.doctrine.schema_provider:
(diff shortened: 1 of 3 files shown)


leofeyer:
Thank you @sheeep.

Issue #1274, opened 2017-12-29
Distinguish between 401 and 403


leofeyer:
We should distinguish between 401 (not authenticated) and 403 (authenticated but insufficient permissions). This implies adding a new page type 401.

In this context, we might want to remove the logout page type, as this is now handled by Symfony.

aschempp:
We need to register an `access_denied_handler` in the firewall that renders the 403 page type, and replace the existing `error_403` handling with the 401 page type.

leofeyer:
See #1381.

PR #1381, opened 2018-02-19
[RTM] Distinguish between error 401 and 403

leofeyer:
This PR implements #1274.


src/Controller/FrontendController.php
```diff
@@ -79,7 +81,7 @@ public function loginAction(): Response
     {
         $this->get('contao.framework')->initialize();
 
-        if (!isset($GLOBALS['TL_PTY']['error_403']) || !class_exists($GLOBALS['TL_PTY']['error_403'])) {
+        if (!isset($GLOBALS['TL_PTY']['error_401']) || !class_exists($GLOBALS['TL_PTY']['error_401'])) {
             return $this->redirectToRoute('contao_root');
```
aschempp: You should throw the `UnauthorizedHttpException` here as well

src/Resources/contao/controllers/FrontendIndex.php
```diff
@@ -260,17 +261,22 @@ public function renderPage($pageModel)
...
-		if ($objPage->protected && !\System::getContainer()->get('security.authorization_checker')->isGranted('ROLE_MEMBER'))
-		{
-			throw new AccessDeniedException('Access denied: ' . \Environment::get('uri'));
-		}
-
-		// Check the user groups if the page is protected
+		// Authenticate the user if the page is protected
 		if ($objPage->protected)
 		{
+			if (!\System::getContainer()->get('contao.security.token_checker')->hasFrontendUser())
```
aschempp: This service should not be used in this case. It's only meant to check users across firewalls (because it unserialized the token from the session instead of using the token storage).

src/Resources/contao/models/PageModel.php
```diff
@@ -433,6 +433,34 @@ public static function findFirstPublishedRegularByPid($intPid, array $arrOptions
...
+	 *
+	 * @param integer $intPid     The parent page's ID
+	 * @param array   $arrOptions An optional options array
+	 *
+	 * @return PageModel|null The model or null if there is no 401 page
+	 */
+	public static function find401ByPid($intPid, array $arrOptions=array())
+	{
+		$t = static::$strTable;
+		$arrColumns = array("$t.pid=? AND $t.type='error_401'");
```
aschempp: so this only works if the page is a direct child of the root page? Is this already the case for 403?

src/Resources/contao/pages/PageError401.php
```diff
@@ -0,0 +1,118 @@
...
+	 *
+	 * @return PageModel
+	 *
+	 * @throws InsufficientAuthenticationException
+	 *
+	 * @internal Do not call this method in your code. It will be made private in Contao 5.0.
+	 */
+	protected function prepare($objRootPage=null)
+	{
+		// Use the given root page object if available (thanks to Andreas Schempp)
```
aschempp: I think you can remove that 😂

src/Resources/contao/models/PageModel.php
leofeyer: Yes.

PR #1382, opened 2018-02-19
Use the modern @-notation when referencing twig templates for Menu

sheeep:
Details available in [this PR](https://github.com/contao/core-bundle/pull/1379).

> Update to the modern @-prefixed way of referencing templates.


leofeyer:
Thank you @sheeep.

PR #1383, opened 2018-02-19
[RFC] Make all services public that we are retrieving directly from the container

sheeep:
I searched the Contao source code for instances of `getContainer()->get('...');` occurrences, filtered out `contao` prefixed services and "uniquified" the resulting array.

```
assets.packages
database_connection
doctrine.dbal.default_connection
event_dispatcher
filesystem
fragment.handler
lexik_maintenance.driver.factory
monolog.logger.contao
request_stack
router
security.authentication_utils
security.authorization_checker
security.firewall.map
security.logout_url_generator
security.token_storage
session
swiftmailer.mailer
translator
```

After intersecting this array to the one found in this ticket (future public services) [this ticket](https://github.com/symfony/symfony/pull/24104), we end up with:

```
assets.packages
database_connection
doctrine.dbal.default_connection
fragment.handler
lexik_maintenance.driver.factory
monolog.logger.contao
security.firewall.map
security.logout_url_generator
swiftmailer.mailer
```

I added those to our `MakeServicesPublicPass` and made sure it will also find aliases instead of definitions only.

Unfortunately I'm not sure if I really caught all direct service references. Do you have any other ideas?


leofeyer:
I think if we merge #1381, we also need `security.authentication.trust_resolver`, don't we?

Toflar:
Shouldn't we make that pass a bit more generic? We'll have more and probably also others need that when they want to access a service from within the legacy code.
What about

```yaml
service_name_that_needs_to_be_accessible:
   class: foobar
   tags:
      - { name: contao.allow_legacy_container_access }
```

which will basically ensure they are all `public`?

leofeyer:
Thank you @sheeep.

PR #1384, opened 2018-02-19
Fixed argument types for AutomatorCommand in symfony/console ^4.0

sheeep:
`InputArgument` enforces its third parameter (`description`) to be of type `string`, so our nice way of assuring the lazy loading of the description (providing `$this` with a `__toString`) method does not work anymore in `symfony/console ^4.0`.

I refactored it to use a service argument.


leofeyer:
Thank you @sheeep.

PR #1388, opened 2018-02-20
[RTM] Do not make the response private when saving the session

### LINKED ISSUE #1283: Session handling in Contao 4.4

leofeyer:
This PR fixes #1283.

Symfony has recently changed their session listener to always make the response private if the session has been started. Although we agree with the change, it renders the HTTP cache unusable, because Contao always starts a session (e.g. to store the user's language). This PR decorates the Symfony session listener and circumvents Symfony's changes by not making the response private if the request is a Contao front end request.

Of course, our long term goal should be to adjust Contao so it only starts a session if it really has to.


src/EventListener/SessionListener.php
```diff
@@ -0,0 +1,90 @@
...
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Contao\CoreBundle\Routing\ScopeMatcher;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\EventListener\AbstractSessionListener;
+use Symfony\Component\HttpKernel\EventListener\SessionListener as BaseSessionListener;
+
+/**
+ * Decorates the default session listener.
```
Toflar: There's no explanation why this listener is present, what it does and when it will be removed again. I think that should be added because it's important information (it should also go into the PR description).

leofeyer: Done.

src/EventListener/SessionListener.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        if (!$this->framework->isInitialized() || !$this->scopeMatcher->isFrontendMasterRequest($event)) {
+            $this->inner->onKernelResponse($event);
+
+            return;
+        }
+
+        $session = $event->getRequest()->getSession();
```
Toflar: @leofeyer There's an issue here. Either we have to call ` $this->inner->onKernelResponse($event);` at the end of the method again or we can leave out the whole `save()` part here because we're overriding the method.

leofeyer: No, we must not call `$this->inner->onKernelResponse($event)` otherwise we have the same problem (private response) again.

Toflar: This does not make sense to me. If you do not forward to the decorated service, why do you call `save()` then?

leofeyer: So the follow up listener, which checks `$session->isStarted()`, is not run.

Toflar: Which follow up listener? I think you confuse stuff here. The one that checks for this is the one you decorate, no?

leofeyer: I'm not confusing anything. We have analyzed the process and set up the existing solution during the developer's meeting. Which issues does it cause?

PR #1390, opened 2018-02-20
Make sure insert tags get the same client cache time as the page

Toflar:
Otherwise the main request will never allow for client cache as it takes the lowest value of all fragments (which is 0 at the moment for us).


leofeyer:
Thank you @Toflar.

PR #1391, opened 2018-02-20
Fix MooTools version for googleapis.

richardhj:
MooTools gets not loaded in the frontend because the ajax.googleapis.com server replied with an 404.

1. There is obviously a typo / copy-paste error.
2. I had to trim version 1.6.0.5 (package version of contao-components/mootools) to 1.6.0 in order to have a valid URI.
3. This is a Contao 4.5 exclusive bug.

### DIFF
--- a/src/Resources/contao/pages/PageRegular.php
+++ b/src/Resources/contao/pages/PageRegular.php
@@ -428,7 +428,9 @@ class PageRegular extends \Frontend
 			{
 				try
 				{
-					$version = PackageUtil::getVersion('contao-components/jquery');
+                    $version = PackageUtil::getVersion('contao-components/mootools');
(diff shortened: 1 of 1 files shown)


xchs:
Could you also target the PR at the 4.5 branch?

richardhj:
The diff then consists of more commits.

xchs:
It's because of the [Trunk-based development](https://contao.org/de/news/wechsel-von-gitflow-zu-trunk-based-development.html) strategy  that was recently switched to.

leofeyer:
Fixed in 6f67345ad675e446182a927e9133c4dab00509c9. Thank you @richardhj for debugging the issue.

PR #1393, opened 2018-02-21
Allow login with username or email address

bytehead:
Additional feature mentioned in #685.

### DIFF
--- a/src/Resources/contao/library/Contao/User.php
+++ b/src/Resources/contao/library/Contao/User.php
@@ -346,6 +346,30 @@ abstract class User extends System implements UserInterface, EncoderAwareInterfa
 	}
 
 	/**
+	 * Find a user in the database based on username or email
+	 *
+	 * @param string  $usernameOrEmail  The username or email
+	 *
+	 * @return boolean True if the user was found
+	 */
+	public function findByUsernameOrEmail($usernameOrEmail)
+	{
+		$objResult = $this->Database->prepare("SELECT * FROM " . $this->strTable . " WHERE username=? OR email=?")
+			->limit(1)
+			->execute($usernameOrEmail, $usernameOrEmail);
+
+		if ($objResult->numRows > 0)
+		{
+			$this->arrData = $objResult->row();
+
+			return true;
+		}
+
+		return false;
+	}
+
+
+	/**
 	 * Update the current record
 	 */
 	public function save()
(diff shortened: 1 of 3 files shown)


leofeyer:
The problem is that `tl_member.email` is not unique. It does have the `eval.unique` flag, but we could not set a real unique database index, because emails were not unique in the past and there are non-unique records.

leofeyer:
Related: https://github.com/contao/core/issues/3421#issuecomment-45525809 and https://github.com/contao/core/pull/7176

ausi:
We could add another column `emailHash` that holds a hash of the email address, this column can have an unique index on the database.

ausi:
> Why would a hash solve the problem of having multiple accounts with the same email address?

You are right, it would not :)
I misread the commend and thought that the index length of the unique index is the problem.

bytehead:
I've missed that the email is not unique. I'll refactor this.

leofeyer:
Are you planning to complete this PR?

bytehead:
Is it somehow possible without breaking BC?

aschempp:
I doubt it… maybe we could add a configuration to the bundle, so to have it explicitly enabled.

leofeyer:
Yes, that would work. I am already using unique e-mail addresses in my installations, so I could use the feature. And I probably would. 😄

bytehead:
So i'm gonna refactor this.

aschempp:
Yeah, I simply use https://github.com/terminal42/contao-mailusername 😉

richardhj:
Just for the record:

> For the future imo the email address is better suited as the primary login identifier than a username. This way it's easy to normalize logins from different sources that also verify the email address (e.g. facebook) without having the user to enter anything (plus the login name does not have to be remembered by the user).

Even though I prefer email addresses over usernames due to UX concerns, this change may not be correct.

A username is a representation of one distinct **user account**, while an email address is one **identity**. Identities may be:
- an email account
- an OAuth-Provider

This implies two things:
- there is a trust relationship between the site and the identity provider (e.g., we trust that the email provider correctly authorizes the account owner)
- a user account may have multiple identities (e.g., multiple email addresses)

This PR allows using an email address or the username for login while preserving the username.
This violates both aforementioned points:

- We may not trust the email provider to correctly authorize the user (this is important in case this PR automatically allows password reset by email (?)
- The 1:n (user-account/identity) problem

Further, it comes with additional major security implications because it allows using email addresses to log in that may be publically known.

Same arguments against the [contao-mailusername extension](https://github.com/terminal42/contao-mailusername) (even though I use for every project ;-))

Towards a solution:

- We should support the 1:n account:identity concept by allowing multiple email addresses.
- We should make using the username optionally for the log-in (we still have the user ID).
- We should make the login options configurable.

leofeyer:
See https://github.com/contao/contao/issues/6354

PR #1394, opened 2018-02-21
[RFC] Drag and drop for file manger

ausi:
Enables Drag and Drop for files and folders.


leofeyer:
Thank you @ausi. Great feature!

may17:
@ausi You should ignore right (context menu) clicks with an check like:
``
if (event.righClick || event.control) {
}
``

PR #1400, opened 2018-02-28
[RTM] Added insert tag for translator

aschempp:
Contao 4.5 supports the Symfony translator, but there is no insert tag to use the Symfony translator. Here's a fix for that.

 - `{{trans::foo_bar}}` => Replace `foo_bar` with the default message domain (*messages*), mostly not useful
 - `{{trans::foo_bar::app}}` => Replace `foo_bar` from `app` message domain
 - `{{trans::foo_bar::app::param1:param2}}` => Replace `foo_bar` from `app` message domain with parameters `param1` and `param2`. The parameters are separated by single colons, like they are for the `label` insert tag.

### DIFF
--- a/src/EventListener/InsertTags/TranslationListener.php
+++ b/src/EventListener/InsertTags/TranslationListener.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2018 Leo Feyer
+ *
+ * @license LGPL-3.0+
+ */
+
+namespace Contao\CoreBundle\EventListener\InsertTags;
+
+use Symfony\Component\Translation\TranslatorInterface;
+
+class TranslationListener
+{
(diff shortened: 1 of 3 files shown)


Toflar:
Unit tests are missing but otherwise I'm +1 on that!

src/EventListener/InsertTags/TranslationListener.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Contao.
+ *
+ * Copyright (c) 2005-2018 Leo Feyer
+ *
+ * @license LGPL-3.0+
```
xchs: LGPL-3.0-or-later

aschempp:
Unit tests added. I couldn't run the php-cs-fixer for some reason so the license etc. hasn't been changed.

leofeyer:
Thank you @aschempp.

PR #1416, opened 2018-03-06
Remove duplicated license file

bytehead:
License file already here: https://github.com/contao/core-bundle/blob/master/LICENSE. 
Probably incorrect merge.


leofeyer:
Thank you @bytehead.

Issue #1421, opened 2018-03-12
SEO: Make Frontend-Breadcrumb accessible for Google


Xendiadyon:
In the Frontend-Breadcrumb navigation, the BreadcrumbList-Schema is used, which is perfectly fine.
But: Google does never generate a Breadcrumb-Path in the search results for our pages.
We think that the whole issue is complicated as soon as we are using https.

This Contao page does not generate a "breadcrumb-able" search result:
![image](https://user-images.githubusercontent.com/2657385/37274292-7a8dace6-25dc-11e8-9d97-fb4eb6aaf6b9.png)

But it is obviously possible to make Google generate a breadcrumb, even when using https. Let me just show you a generic example page:
![image](https://user-images.githubusercontent.com/2657385/37273610-a2f0a83e-25da-11e8-80f7-f525bb0decc8.png)

Code:
![image](https://user-images.githubusercontent.com/2657385/37273724-ee938856-25da-11e8-9ca4-db89f8d8c419.png)

Do you have experience how we could bring Google to parse/implement our breadcrumb menus to the Google search results?

Do we explicitely need use use itemprop="breadcrumb"? Or is the type="v:Breadcrumb" the catcher? Is an ol instead of ul neccessary?

Literature:
https://www.trustagents.de/wissen/breadcrumb
In Fact, Google recognizes our BreadcrumbList:
https://search.google.com/structured-data/testing-tool/

fritzmg:
See also https://developers.google.com/search/docs/data-types/breadcrumb

CMSworker:
> We think that the whole issue is complicated as soon as we are using https

No, it's not related to HTTPS. The only difference is that Google doesn't show a short name for the Website as first breadcrumb path element, but instead the URL to indicate that it has HTTPS:

Example without https:
`Contao Official Demo > Content Elements > Content slider`

Example with https would look like:
`https://demo.contao.org > Content Elements > Content slider`

> Do we explicitely need use use itemprop="breadcrumb"? Or is the type="v:Breadcrumb" the catcher?

[...]

PR #1429, opened 2018-03-15
[WIP] Improve breadcrumb microdata

CMSworker:
Belongs to issue #1421

I added temporarily an out-commented long php version with many comments so you can see whats happening and why I changed the code or added elements.

If you are ok with the code structure I'll remove the long php part again and you can merge it if you want.

Is there an additional pull request required for 4.5/4.6?


leofeyer:
Why did you change the list from unordered to ordered?

> Is there an additional pull request required for 4.5/4.6?

No.

fritzmg:
> Why did you change the list type from unordered to ordered?

See https://github.com/contao/core-bundle/issues/1421#issuecomment-373210871

> `<ol>` is recommended and used in the examples by Google (see link from @fritzmg) and also semantically correct, because the path has an order `1 > 2 > 3` to indicate the depth or nesting levels - So you should use an ordered list.

src/Resources/contao/templates/modules/mod_breadcrumb.html5
```diff
@@ -2,14 +2,163 @@
...
+        <?php if ($item['isActive']): ?>
+          <li class="active'<?php if ($item['class']): ?> <?= $item['class'] ?><?php endif; ?> last"><?= $item['link'] ?></li>
+        <?php else: ?>
+          <li<?php if ($item['class']): ?> class="<?= $item['class'] ?>"<?php endif; ?>><a href="<?= $item['href'] ?>" title="<?= $item['title'] ?>"><?= $item['link'] ?></a></li>
+        <?php endif; ?>
       <?php else: ?>
-        <li<?php if ($item['class']): ?> class="<?= $item['class'] ?>"<?php endif; ?> itemscope itemtype="http://schema.org/ListItem" itemprop="itemListElement"><a href="<?= $item['href'] ?>" title="<?= $item['title'] ?>" itemprop="url"><span itemprop="name"><?= $item['link'] ?></span></a></li>
+        <?php if ($item['isActive']): ?>
+          <li itemprop="itemListElement" itemscope="" itemtype="http://schema.org/ListItem" class="active<?php if ($item['class']): ?> <?= $item['class'] ?><?php endif; ?> last"><span itemprop="item" itemscope="" itemtype="http://schema.org/Thing"><span itemprop="name"><?= $item['link'] ?></span><link itemprop="url" href="<?= $item['href'] ?>" /></span><meta itemprop="position" content="<?= $path_number ?>" /></li>
```
leofeyer: How will browsers render the `<meta>` tag?

fritzmg: They don't render it at all. `<meta>` tags are allowed in the `<body>` in HTML5. A lot of examples on http://schema.org take advantage of that.

leofeyer:
Ok, but changing the list type from `<ul>` to `<ol>` will definitely require manual CSS adjustments, so we should only add this in Contao 4.6. Although there is no BC promise for templates, we are trying to avoid changes that visually break a large number of sites in bugfix releases.

asaage:
Please check for the ```<strong>```  https://github.com/contao/core-bundle/commit/acd841e33a50347487009da8508c931ae6a9a03e

fritzmg:
@CMSworker isn't the `<span>` surrounding the `<a>` still missing?

CMSworker:
@asaage Is added now, was missing because I used 4.4 branch as base

CMSworker:
@fritzmg No it is added at the breadcrumbs with microdata (line 15 and 17, or in the long commented php version line 101 and 132). As stated in the original issue the first element (link to homepage) should not be part of the microdata breadcrumb to avoid having it double shown by Google (first microdata element is a category/subpage, not the homepage). The microdata breadcrumb content/order is different to the html ol (ol has one more element). But that shouldn't be a problem because the microdata information is only used by services like Google and not relevant for or used by browsers or screen readers...

leofeyer:
Why is there no micro-data on the first element (`$path_number == 0`)?

fritzmg:
@leofeyer see @CMSworker previous' comment:

> the first element (link to homepage) should not be part of the microdata breadcrumb to avoid having it double shown by Google (first microdata element is a category/subpage, not the homepage)

src/Resources/contao/templates/modules/mod_breadcrumb.html5
```diff
@@ -2,14 +2,165 @@
...
 
-  <ul itemscope itemtype="http://schema.org/BreadcrumbList">
-    <?php foreach ($this->items as $item): ?>
-      <?php if ($item['isActive']): ?>
-        <li class="active<?php if ($item['class']): ?> <?= $item['class'] ?><?php endif; ?> last" itemscope itemtype="http://schema.org/ListItem" itemprop="itemListElement"><span itemprop="name"><?= $item['link'] ?></span></li>
+  <ol itemscope="" itemtype="http://schema.org/BreadcrumbList">
+    <?php $path_number = 0; foreach ($this->items as $item): ?>
+      <?php if ( $path_number == 0 ): ?>
+        <?php if ($item['isActive']): ?>
+          <li class="active'<?php if ($item['class']): ?> <?= $item['class'] ?><?php endif; ?> last"><strong><?= $item['link'] ?></strong></li>
```
fritzmg: There is a superfluous `'` instead of a space. 
```
active'<?php
```

leofeyer:
This seems a little strange to me. Do other CMS handle it this way, too?

fritzmg:
The Yoas WordPress Breadcrumb plugin for example _doesn't_. Here is some discussion about this: https://webmasters.stackexchange.com/questions/95231/should-the-first-breadcrumb-not-be-the-home-page

leofeyer:
IMHO we should use semantically correct markup instead of mangling the schema tags to work around a problem that Google should fix in the first place (and probably will in the future).

CMSworker:
It looks weird, but semantically it is still correct:
* HTML semantic: It is a regular, valid ordered list with the homepage as starting point. The full path will be interpreted by browsers and read out correctly by screen readers etc. The nesting level for the path elements is semantically indicated through the use of an ordered list `<ol>`.
* Microdata semantic: It is a valid breadcrumb list with the "category" as starting point (link to homepage is ignored). The microdata breadcrumb order (and so the nesting level) is defined through the microdata markup.

Microdata structure/semantic is not (and doesn't have to be) equal to HTML structure/semantic. Microdata and HTML are two different technologies. In fact, microdata requires often more complex/splitted data structure so machines can interpret the provided content correctly and built correct relations between the single elements.

Of course it would be great to have the same semantic structure and content for both microdata and HTML. But in this case it would end up Google showing the Homepage as first child of the URL.

And yes, Google is doing it wrong here (or behaving unexpected). But ignoring Google and hoping they are going to fix this soon doesn't really make sense: The main reason to implement microdata (at lease from my point of view, as well as in the original issue #1421 ) is that Google shows the breadcrumb in search results. If we are ignoring Google, why should we implement microdata at all?

fritzmg:
> The main reason to implement microdata (at lease from my point of view, as well as in the original issue #1421 ) is that Google shows the breadcrumb in search results. If we are ignoring Google, why should we implement microdata at all?

It's not the point of Microdata to cater to Google's specific search result implementation. Microdata provides a way to convey information to any platform that might want to extract data from your web service.

And in general you should not design the syntactic and semantic implemenation of a web service so that it "best fits Google". You should adher to standards and do sound semantic implementation. It's the goal of Google to provide the user with the best search result for their search keyword. And the more they can rely on well established and defined standards, the easier it is for Google (or any other service) to get the right information in order to achieve that goal.

CMSworker:
Here you go...

leofeyer:
As discussed in Mumble on May 17th, @ausi will check if Google really requires this markup (ordered list with position).

ausi:
See #1527

leofeyer:
@CMSworker Can you please check if #1527 works for you, too?

leofeyer:
See #1527 and #1561.

PR #1435, opened 2018-03-19
[RFC] Index pages other than regular pages

agoat:
This PR adds the ability to index (and search) more not just regular pages.

Only regular pages are currently indexed : 
https://github.com/contao/core-bundle/blob/10efd81c1f99094af52f518f227e7a6c9db1e5a6/src/Resources/contao/classes/Frontend.php#L618-L625

When using a global variable to set the page types to be indexed, a developer would be able to make custom page types searchable. Simply by extending the global variable in the `config.php` file:
```php
$GLOBALS['TL_INDEX_PAGES'][] = 'mypagetobeindexed';
```

Since this should also be backward compatible, it would be desirable to integrate it into the 4.4 LTS version as well.

### DIFF
--- a/src/Resources/contao/classes/Frontend.php
+++ b/src/Resources/contao/classes/Frontend.php
@@ -616,7 +616,7 @@ abstract class Frontend extends \Controller
 		}
 
 		// Index page if searching is allowed and there is no back end user
-		if (\Config::get('enableSearch') && $objPage->type == 'regular' && !BE_USER_LOGGED_IN && !$objPage->noSearch)
+		if (\Config::get('enableSearch') && \in_array($objPage->type, $GLOBALS['TL_INDEX_PAGES']) && !BE_USER_LOGGED_IN && !$objPage->noSearch)
 		{
 			// Index protected pages if enabled
 			if (\Config::get('indexProtected') || (!FE_USER_LOGGED_IN && !$objPage->protected))
--- a/src/Resources/contao/config/config.php
+++ b/src/Resources/contao/config/config.php
@@ -413,7 +413,7 @@ $GLOBALS['TL_AUTO_ITEM'] = array('items', 'events');
 
 
 /**
- * Index only the following page types
+ * Register the pages to be indexed
  */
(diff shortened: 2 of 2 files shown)


m-vo:
Can this be done without introducing yet another superglobal? :-)

agoat:
> Can this be done without introducing yet another superglobal? :-)

I just followed the way, how global settings currently done in contao (like the other settings e.g. `$GLOBALS['TL_NOINDEX_KEYS']`).

leofeyer:
> it would be desirable to integrate it into the 4.4 LTS version as well

Of course not. New feature == next minor version.

agoat:
> Of course not. New feature == next minor version.

That's not good. I have to find an extra way to search the pages of an extension with version 4.4.

But I respect the version policy.

leofeyer:
We should probably just change `$objPage->type == 'regular'` to something like this:

https://github.com/contao/core-bundle/blob/e0a0411e639234d2dac230bbbe1c7c3670420970/src/Resources/contao/models/PageModel.php#L391

agoat:
Ah, just the other way around. Not a bad idea.

dmolineus:
Dealing with custom page types needs more than the requirement to index them, f.e. if articles are allowed or not or providing pages which should be handled like error pages (see the page model example linked above).

I'd prefer introducing a page type registry where such information could be stored.

leofeyer:
The same problem exists here as well BTW: https://github.com/contao/core-bundle/blob/10efd81c1f99094af52f518f227e7a6c9db1e5a6/src/Resources/contao/classes/Backend.php#L659

leofeyer:
As discussed in Mumble on April 12th, we want to replace the `$objPage->type == 'regular'` check with a `$objResponse->getStatusCode() == 200` check in the `Frontend::indexPageIfApplicable()` method.

We don't want to adjust the `Backend::findSearchablePages()` method though to avoid BC breaks.

leofeyer:
Changed in e9add0c63f2da2ab45f747aecf8a456c3848655c.

PR #1451, opened 2018-04-03
Added a few unit test for legacy classes

aschempp:
Also fixes a test for outdated PHPUnit class


leofeyer:
It seems you broke the unit tests. 😉

leofeyer:
Thank you @aschempp.

Issue #1403, opened 2018-02-28
Drag and drop file manager


leofeyer:
I have noticed two issues with the new drag and drop implementation for the file manager.

1. Clicking an element without draging it shows the clone:

<img width="551" alt="" src="https://user-images.githubusercontent.com/1192057/36785581-9891cc2a-1c83-11e8-9269-2096467ce4a3.png">

2. All other draggable elements in the back end use a drag handle. For reasons of consistency, I think we should add a drag handle in the file manager, too.

Toflar:
Can we try highlighting the file name with a different color like on OSX? It really helps UX a lot. I would love to get some screenshots so we can decide which one looks best 😎

leofeyer:
Also see https://github.com/contao/core-bundle/pull/1394#issuecomment-377721950

> @ausi You should ignore right (context menu) clicks with an check like:
> ``
> if (event.rightClick || event.control) {
> }
> ``

ausi:
Yeah, drag handle seams reasonable. Should I add it as the last action button like it is in the content elements view?

leofeyer:
Yes please.

leofeyer:
I have found another small issue: the clone is also created upon selecting a row.

https://www.dropbox.com/s/f3ydj23i2r61907/1403.mp4?dl=0

I guess this will resolve itself when we switch to using a drag handle?

ausi:
Yes, I think so too.

PR #1456, opened 2018-04-04
[RFC] implement drag and drop for site structure and articles

may17:
This pull request ships the ability for simple drag and drop support (inspired by @ausi on #1394) for the contao site structure and the articles overview. In the sense of simplicity, there are just two placement types:

a) inside on top when it`s an folder.
b) after an element what would be the most used case.

This ticket is related to issue: https://github.com/contao/core/issues/874

Preview Video: https://drive.google.com/open?id=1NGwH-gCTpmEljNiYVpcoiR9jGpBkVkJE


aschempp:
The preview video does not load for me 😞

bytehead:
@aschempp use Chrome 😜

bezin:
Really nice 😍 I was just wondering if the styling could need some improvement? I think it would be more precisely, if the hover state would be just the bottom border, rather than the whole rectangle?

src/Resources/contao/dca/tl_page.php
```diff
@@ -28,7 +28,8 @@
 			array('tl_page', 'addBreadcrumb'),
 			array('tl_page', 'setRootType'),
 			array('tl_page', 'showFallbackWarning'),
-			array('tl_page', 'makeRedirectPageMandatory')
+			array('tl_page', 'makeRedirectPageMandatory'),
+			array('tl_page', 'addListDragAndDrop')
```
Toflar: Can you explain why this is needed here? Can't we find a general solution for `DC_Table` so that code is just added for any DCA with the tree view?

may17: I decided this way because the DCTable monster is used everywhere and the selectors differs a little bit. It was the fastest way to bring up the feature and discuss it. I prefer the better done than perfect philosophy :)

Perhaps we can add an option for the DCA`s to configure it?

Toflar: Yeah it's just a problem right now for my own tree view DCA's. They all would need to implement that callback. What's specific there? The `listingContainer`?

may17:
@bezin @Toflar thanks for your feedback. I will think on it and improve the styling.

Toflar:
But definitely keep on working on that! Together with the already merged PR's Contao 4.6 could get a lot of new nice drag n' drop UX, people are going to love it 🎉

src/Resources/contao/dca/tl_page.php
may17: Just the droppables. Because of the current markup everything is an tl_folder but on the articles we just wanna allow to drop into pages or after articles inside a pages. So that .level_0 selector comes up :/

Toflar: Yeah so maybe we can think about some general - new - droppable container with the correct class and just make sure the `DC_Table` adds that. Some data attribute or so, which will then work everywhere. That would be excellent :)

may17: Also the current url is generated here to trigger the location.href.

may17: @Toflar i like to implement it as option of the dca list like $GLOBALS['TL_DCA']['tl_page']['list']['dragdrop'] = [ ..options ] or underneath sorting. What did you think? What @leofeyer think?

Toflar: I don't understand what needs to be configurable yet. To me, whenever a configuration can be omitted, it should :)

may17: I understand your intention but the feature was just created as quick win to work for site structure and article overview. I like the idea to implement into any tree but we didn`t know if it works in any case. So i think it should be explicit activated via dca.

By the way, i thankfully realized with this discussion that i`ve missed to check the user rights ☠️. Only if the user has the right to move items their could use this feature. This maybe belongs to #1394 too. @ausi did you add an check for the file drag and drop?

Maybe we should talk about this topics via mumble in the upcoming days?

leofeyer: Yes, we should. The next call is next week on Thursday, April 12th.

leofeyer:
Very nice, thank you @may17! 👍

We also need to discuss another topic that I have already mentioned in contao/core-bundle#1403. Right now we are using drag handles wherever drag and drop is possible and I think we should keep things consistent here. Also using a drag handle still allows to copy and paste text from the draggable item.

may17:
@Toflar @bezin @everyone here are my new suggestion to improve the ux https://drive.google.com/file/d/1TTkfkGDL1jETtK-t6qyhFpZwZ8ql5kQ5/view

Toflar:
I really like this idea. I especially like the green contrast and the osx-like drag clone. I would like to see that for the file manager too, wdyt @ausi

ausi:
Yes, we should adapt the styling for the file manager to get the same look and feel. But the green bar shown in the video is not possible for the file manager, because there you can only drop into folders, not before or after files.

Toflar:
Yeah I was talking about the drag clone and the green color in general. I really like the contrast it gives, I wouldn't have come up with green in the first place but I think it looks excellent.

may17:
@Toflar @ausi maybe we need this "drop into" style for general drag and drop too. We should talk about that next week.

leofeyer:
As discussed in Mumble on April 12th:

* The code shall go into the `DC_Table` class, however it must be enabled in the DCA.
* We want to add drag handles like in the parent view (e.g. content elements).
* Drag permissions can be checked by checking for the icons.
* Drop permissions need to be checked in an Ajax request that opens the tree in paste mode and converts the returned HTML in an AST tree or similar so it can analyze the paste buttons.

We have agreed that checking drop permissions is complex and should be done in a second step.

may17:
Sorry i was on vacation for 2 and a half weeks. I will continue on it next week.

ausi:
We discussed this feature in the recent hackathon. Here are some screenshots of how we think it should look and work:

Dropping into another page:
dnd-into

Dropping between two pages:
dnd-between

Dropping before a page:
dnd-before

Dropping after a page:
dnd-after

Which action gets triggered should depend on the vertical drop position relative to the row:

- 0%-25% (1-7 pixel): drop before
- 25%-75% (8-21 pixel): drop into
- 75%-100% (22-28 pixel): drop after

may17:
„Should look and work“ means you already implemented it or it‘s up for implementation?

Anyway i really like it. the only thing i didn‘t like is the unnatural Position of the drag handler on the right side.

may17:
Alright. Next question 🙈😊.

Am i right assuming that i should implement that? I ask those questions because of some conversations on the Contao Conference Salzburg and i thought @ausi would finalize both Drag and drop Features.

Anyway just say go and i will keep up the work :)

leofeyer:
We had planned to implement it at the hackathon, however we have not had enough time. @may17 @ausi You should probably get into direct contact before you both end up implementing this. 😄

ausi:
> Am i right assuming that i should implement that?

That would be great!

> Anyway just say go and i will keep up the work :)

Go 😊
(I do not have the time currently)

Just as a side note: Drag and drop in the file manager is completed now (in the master branch). The look and feel should be similar where possible.

leofeyer:
@may17 Are you still working on this one?

may17:
Yes but i had still problems with my contao dev setup for core development :( Hope we could find a solution on the contao camp?

Toflar:
I've documented how it works here: https://github.com/contao/contao#development
Anything that does not work for you? I can adjust the docs then :)

PR #1462, opened 2018-04-06
[RTM] allow multi root and multi page search

fritzmg:
These changes allow you to expand/limit the search to multiple pages or website roots. This enables you to:

* limit the search to multiple pages and its children - instead of just one page and its children
* expand the search over multiple website roots - instead of being restricted to the current website root

The changes are fairly simple and straight forward. I tagged it as RFC since I am unsure whether the use of a new field called `searchPages` is really necessary. I decided to not use the existing `pages` field, since the sort order does not matter. Though may be I should simply alter the DCA in a `load_callback`, similar to `tl_content::setSingleSrcFlags`? Then the `pages` field could be reused for this.

### DIFF
--- a/src/Resources/contao/dca/tl_module.php
+++ b/src/Resources/contao/dca/tl_module.php
@@ -115,7 +115,7 @@ $GLOBALS['TL_DCA']['tl_module'] = array
 		'lostPassword'                => '{title_legend},name,headline,type;{config_legend},reg_skipName,disableCaptcha;{redirect_legend},jumpTo;{email_legend:hide},reg_jumpTo,reg_password;{template_legend:hide},customTpl;{protected_legend:hide},protected;{expert_legend:hide},guests,cssID',
 		'closeAccount'                => '{title_legend},name,headline,type;{config_legend},reg_close;{redirect_legend},jumpTo;{template_legend:hide},customTpl;{protected_legend:hide},protected;{expert_legend:hide},guests,cssID',
 		'form'                        => '{title_legend},name,headline,type;{include_legend},form;{protected_legend:hide},protected;{expert_legend:hide},guests,cssID',
-		'search'                      => '{title_legend},name,headline,type;{config_legend},queryType,fuzzy,contextLength,totalLength,perPage,searchType;{redirect_legend:hide},jumpTo;{reference_legend:hide},searchPages;{template_legend:hide},searchTpl,customTpl;{protected_legend:hide},protected;{expert_legend:hide},guests,cssID',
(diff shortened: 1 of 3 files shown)


aschempp:
Do I understand correctly that child pages of the selected ones are searched automatically? The field description does not mention that, maybe that would be useful?

Generally, I like the idea, have you tested whether the search result module also generates absolut links to the potential foreign domains?

fritzmg:
> Do I understand correctly that child pages of the selected ones are searched automatically? The field description does not mention that, maybe that would be useful?

Yes, child pages are searched automatically - as it's already the case in the current implementation, when you select a reference page. The current field description doesn't mention it either:

> **Reference page**
>
> Please choose the reference page from the site structure.

However that field is shared with multiple modules, so the description has to be generic. We can choose a more descriptive one here, since it's its own field. Unless we switch to the `pages` field.

> Generally, I like the idea, have you tested whether the search result module also generates absolut links to the potential foreign domains?

Yes :). This already worked anyway (if you set the reference page to a website root of another domain for example).

leofeyer:
I'm pro reusing the `pages` field if possible.

Toflar:
> I'm pro reusing the pages field if possible.

Same here. Also I don't really think it is a problem if you can order pages. Who cares 😄

fritzmg:
Alright, I'll update the PR accordingly :)

fritzmg:
PR updated. I had to use a `load_callback` anyway, because the `pages` field is mandatory by default (which it should not be for the search module).

src/Resources/contao/dca/tl_module.php
```diff
@@ -1053,4 +1057,29 @@ public function setMultiSrcFlags($varValue, DataContainer $dc)
...
+	 *
+	 * @return mixed
+	 */
+	public function setPagesFlags($varValue, DataContainer $dc)
+	{
+		if ($dc->activeRecord)
+		{
+			switch ($dc->activeRecord->type)
+			{
+				case 'search':
```
leofeyer: This is unnecessarily complex. Why not the following?

```php
if ($dc->activeRecord && $dc->activeRecord->type == 'search')
{
    $GLOBALS['TL_DCA'][$dc->table]['fields'][$dc->field]['eval']['mandatory'] = false;
    unset($GLOBALS['TL_DCA'][$dc->table]['fields'][$dc->field]['eval']['orderField']);
}
```

BTW, is it really necessary to unset the order field?

leofeyer:
Thank you @fritzmg.

PR #1463, opened 2018-04-06
use www.youtube-nocookie.com domain

fritzmg:
YouTube calls this the "privacy-enhanced mode" in the embed dialogue.

> When you turn on privacy-enhanced mode, YouTube won't store information about visitors on your website unless they play the video.

It doesn't do all that much, but I think there is no reason _not_ to use it.

### DIFF
--- a/src/Resources/contao/elements/ContentYouTube.php
+++ b/src/Resources/contao/elements/ContentYouTube.php
@@ -108,7 +108,7 @@ class ContentYouTube extends \ContentElement
 			$params[] = 'end=' . (int) $this->youtubeStop;
 		}
 
-		$url = 'https://www.youtube.com/embed/' . $this->youtube;
+		$url = 'https://www.youtube-nocookie.com/embed/' . $this->youtube;
 
 		if (!empty($params))
 		{


frontendschlampe:
we already added it in our PR https://github.com/contao/core-bundle/pull/1348 ;-)

fritzmg:
Ah cool, I haven't noticed :)

aschempp:
Can this PR be closed then?

fritzmg:
One problem is that this change would be hard coded into the system, while #1348 would make it _optional_ (and **off** by default!).

Toflar:
Is there any reason **not** to use this by default? Because I don't see one either atm.

frontendschlampe:
> Is there any reason not to use this by default? Because I don't see one either atm.

It's the decision of editor, which implementation he want to use - not ours. It's like copying newsletter recipients.

fritzmg:
Shouldn't Contao follow the "privacy by default" directive? That would imply using `youtube-nocookie.com` by default as well.

frontendschlampe:
> Shouldn't Contao follow the "privacy by default" directive? That would imply using youtube-nocookie.com by default as well.

Of course ... but than we also have to discuss about newsletter recipients and somer other things, I think. For example delete the default support for Google Analytics and Piwik.

fritzmg:
> Of course ... but than we also have to discuss about newsletter recipients and somer other things, I think. For example delete the default support for Google Analytics and Piwik.

Google Analytics and Piwik are optional, hence it's still "privacy by default". And even if you do enable those services, Contao uses the IP anonymization setting by default as well. So it's again "privacy by default".

The YouTube domain would be the other way around. You need to activate that option for _every single YouTube content element_ (if you do not configure your own defaults in a DCA file). In my opinion this would not be "privacy by default".

frontendschlampe:
> My question was if we even need to be able to use any different domain?

I don't understand, what you mean? ;-)

> I see no reason to make it optional either.

Why not? We can activate it by default, but it should be optional, because we don't know, what to clients want to do or why they want to use the normal You Tube url. Another thing is, that the "no-cookie" option only says, that no cookies will be used, but we don't know if there're any other problems or restrictions by using it. We also don't know, if it's ok for GDPR.

Toflar:
Yeah, makes absolutely no sense if it's not needed. Maybe you can only use the API over the regular domain?

frontendschlampe:
> I think he's simply also questioning why we should make the usage of youtube-nocookie.com optional ;).

Because we can! I understand to activate it by default, but if somebody because of any reason want to deactivate it, why should he don't activate it? It's the same with the anonymization of Google Analytics IP - if somebody want to deactivate, he can do!

fritzmg:
Well, right now there is no way in Contao 4 to change the domain of the YouTube Content Element anyway either. And I guess there are a lot more things that users possibly would want to do but (currently) cannot in Contao.

frontendschlampe:
You're right, but I don't see the problem. Using the no-cookie domain by default is ok for me, but make it hardcoded is not an option, I think. Why we should delete an option if we have it already? We just want to add an option, because we don't have it. If this option is activated by default is no problem.

aschempp:
I'm not sure scripts like mediaelement work with the nocookie domain, so there might be a reason not to use it and not to just change it in a patch release…

fritzmg:
True.

Btw. I accidentally based this to the 4.5 branch - I meant to base it against the master branch for Contao 4.6.

frontendschlampe:
> Btw. I accidentally based this to the 4.5 branch - I meant to base it against the master branch for Contao 4.6.

For Contao 4.6 our PR will be active! We can change the no-cookie option to default if we want to do.

fritzmg:
> For Contao 4.6 our PR will be active!

I know, just saying that I did not mean for this change to be a bug fix for 4.4 and 4.5.

leofeyer:
Closed in favor of #1348.

PR #1465, opened 2018-04-07
Use a double pipe as logical OR for the version range

xchs:
The single pipe `|` operator is considered deprecated but retained for backwards compatibility. For the logical OR version comparison, it is recommended to use the double pipe `||` operator in the `composer.json` as per the [official Composer documentation](https://getcomposer.org/doc/articles/versions.md#version-range).


aschempp:
There is no need for a pull request for every branch, version 4.4 changes will be merged into 4.5, so this PR is obsolete.

PR #1470, opened 2018-04-09
[RFC] InsertTags as fragments

Toflar:
This adds insert tags support on top of our new fragment registry.
I've moved the `{{trans::*}}` insert tag to the new way of writing insert tags to illustrate what needs to be done and what the advantages are.

Main differences to the current behaviour:

* Of course we use services and not hooks anymore
* An insert tag is now a controller that has to return a response instead of a string which allows to set proper cache headers.
* Insert tag flags are an event instead of a hook that can access the whole response which allows them not only to modify the content if needed but again, also the caching headers which is important if the flags e.g. behave differently based on the request (e.g. Vary headers etc.).

So registering an insert tag for `{{trans::*}}` now requires the following steps:

```yaml
// services.yaml
service_name:
    class: Vendor\Controller\SuperController
    tags:
        - { name: 'contao.insert_tag', type: 'trans' }
```

As parameters in your controller you can type hint (regular argument resolver):

* `insertTag` - contains `trans` (should not be needed, really)
* `parameters` - contains everything after `trans::` until `}}` so you can do whatever you like with that
* `flags` - contains the insert tag flags

And of course the other ones like `Request` etc.

I've added an abstract insert tag controller to dispatch the insert tag flags event. I think it should be optional for any insert tag to allow for insert tag flags to be applied or not.

ToDo's:

* [ ] Agree on the concept (although 90% is already given by our registry concept)
* [ ] Unit tests


aschempp:
Some questions and feedback:
 1. I think flags should be applied each on their own in a loop, not one event for all. Because the order might be important, which was already the case afaik?
 2. How are the existing insert tag flags applied to new insert tags?
 3. I'd drop the *apply* in flags. So `InsertTagFlagEvent` and `contao.insert_tag_flag` event name etc.
 4. I still see the potential issue of resolving insert tags with unknown/prefixed names. It would be possible to register for `link` and `link_url` and `link_title` etc, but I don't know if you can do `page_xxx` to get any random attribute of a page?

The big advantage would be that we can actually validate insert tags at some point.

PS: why would it not make sense to cache the translation insert tags for you?

Toflar:
> I think flags should be applied each on their own in a loop, not one event for all. Because the order might be important, which was already the case afaik?

The order can be controlled using the priorities in the event listener.

> How are the existing insert tag flags applied to new insert tags?

I would move all the logic from `InsertTags` to a core listener.

> I'd drop the apply in flags. So InsertTagFlagEvent and contao.insert_tag_flag event name etc.

I could do that, yeah.

> I still see the potential issue of resolving insert tags with unknown/prefixed names. It would be possible to register for link and link_url and link_title etc, but I don't know if you can do page_xxx to get any random attribute of a page?

I've always considered this to be a design flaw. It should be `link::url` and `link::title` etc. Same for `page::foobar`. 

> PS: why would it not make sense to cache the translation insert tags for you?

It would mean I have to make it an ESI request and I think it would actually hurt the performance rather than helping it. The additional roundtrip is nonsense because translation is a very lightweight operation. Setting cache headers only makes sense for ESI insert tags and as such only for heavy-lifting stuff or insert tags that should never be cached (such as `{{request_token}}`). See what I mean?

leofeyer:
I like your approach, although I'm not sure about the insert tag flags. Have we not agreed on not adding new hooks for old stuff? Is the `InsertTags` class part of how we want to handle insert tags in the future?

Toflar:
It's not a new hook for old stuff? It's a new concept?

> Is the InsertTags class part of how we want to handle insert tags in the future?

No, the class should be deprecated and replaced by a proper service. What name would you like to have for such a service? `InsertTagsHandler`?

aschempp:
> No, the class should be deprecated and replaced by a proper service. What name would you like to have for such a service? InsertTagsHandler?

I don't think that should be part of this PR to keep things simple.

What we have not discussed is whether we want to keep insert tags as-is? Like should we keep the syntax (even though it conflicts with Twig) or is there anything else that always bothered use and we might be able to fix on a re-implementation?

Toflar:
> I don't think that should be part of this PR to keep things simple.

Agreed.

> What we have not discussed is whether we want to keep insert tags as-is? Like should we keep the syntax (even though it conflicts with Twig) or is there anything else that always bothered us and we might be able to fix on a re-implementation?

I think that's not really a discussion to me. We won't be able to use Twig everywhere in 4.* but still, we want to be able to cache them properly. We can talk about a different concept for v5.

Toflar:
All the ToDo's have been addressed and caused quite some refactoring so I still haven't worked on unit tests yet because we should review that again.

src/Fragment/FragmentHandler.php
```diff
@@ -68,9 +68,11 @@ public function __construct(ContainerInterface $renderers, BaseFragmentHandler $
     /**
      * {@inheritdoc}
      */
-    public function render($uri, $renderer = 'inline', array $options = []): ?string
+    public function render($uri, $renderer = '', array $options = []): ?string
     {
-        if (!$uri instanceof FragmentReference) {
+        if (!$uri instanceof FragmentReference || '' !== $renderer) {
+            $renderer = $renderer ?: 'inline';
```
leofeyer: Does this need to become `forward`, too?

leofeyer:
I have reviewed and I'm still fine with the concept. I recall there was one use case that will not work anymore once we merge this, right?

src/Fragment/FragmentHandler.php
Toflar: Yes, once the other PR is merged.

leofeyer: It already is. 😃

aschempp: No it does not! Because we are decorating the original Symfony service, we must keep it's default arguments!

Toflar:
I still want to change a few things like extracting the whole parsing logic into its own service so you can use it elsewhere. This should clean up things even more.

PR #1473, opened 2018-04-11
remove 'files'=>true for pages

fritzmg:
While working on #1462 I noticed that `tl_modules.pages` has `'files'=>true` set in its `eval`. That does not make any sense, does it?


leofeyer:
Thank you @fritzmg.

Issue #1370, opened 2018-02-14
Missing request token in form when maintenance mode is enabled


netzarbeiter:
Tested on Contao 4.4.13 and 4.4.14.

- Set up a form with several mandatory fields and disable HTML5 validation.
- Include the form on a page
- Go to the page and view the source code, the request token is empty 
(`<input name="REQUEST_TOKEN" value="" type="hidden">`)

Disable the maintenance mode and it works like a charm.

leofeyer:
Does the issue occur in Contao 4.5 as well?

netzarbeiter:
I have now set up two new blank installations 4.4.14 and 4.5.4, in both the value of the request token remains empty when the maintenance mode is enabled.

ausi:
It seems that the ESI fragment for the request token returns an empty string, maybe because the maintenance mode changes the status code to something different than `200`?

netzarbeiter:
I get status `200` with and without maintenance mode enabled.

Toflar:
Yes, the maintenance mode listener returns a `503` for any request. It also does so for ESI requests and that's why the data is hidden. That listener should be disabled for fragments.

The config.yml should be adjusted from 

```yaml
lexik_maintenance:
    authorized:
        path: ^/contao($|/)
        attributes:
           _bypass_maintenance: true
```
to this
```yaml
lexik_maintenance:
    authorized:
        path: '^(/contao($|/))|(%fragment.path%)'
        attributes:
           _bypass_maintenance: true
```

Completely untested but you get the point.

leofeyer:
Can you please try again to reproduce the issue with the latest `4.4` branch? I have merged the HTTP cache fixes and I made sure that the page is cached:

```php
<input type="hidden" name="REQUEST_TOKEN" value="<?php echo $this->surrogate->handle($this, '/_fragment?_path=insertTag%3D%257B%257Brequest_token%257D%257D%26_format%3Dhtml%26_locale%3Den%26_controller%3Dcontao.controller.insert_tags%253ArenderAction&clientCache=900&pageId=62&request=en%2Fapplications.html&_hash=qZBuTWT1ZeiB61Aq2KtbEISyMmAwHroZwSH8%2BkACbhU%3D', '', true) ?>
```

Then I have enabled the maintenance mode, logged in to the back end and opened the front end preview. The request token was never blank. 🤔

netzarbeiter:
[...]

PR #1476, opened 2018-04-12
Use the master request's IP to determine maintenance mode

aschempp:
fix for  #1370


leofeyer:
Important: The fix must only be applied to Contao 4.4!

leofeyer:
Thank you @aschempp.

PR #1477, opened 2018-04-13
Implement UserInterface to work around Symfony core changes

aschempp:
see https://github.com/symfony/symfony/pull/26788#issuecomment-381040718

Obviously needs to be reverted/not be merged in Contao 4.5


leofeyer:
As discussed this PR does not fix the problem, therefore I'm closing it.

PR #1478, opened 2018-04-13
[RTM] Introduced auto cache tag invalidation on DCAs for better DX

Toflar:
Automatically invalidates DCA entries when users edit them in the back end.
For testing purposes I've added corresponding tags on the page response.

Depends on https://github.com/contao/manager-bundle/pull/67 which itself depends on other PR's again. Leaving it here for reference and to be updated.


Toflar:
ping @contao/developers 

This is ready to be reviewed. The goal here is that we invalidate tags based on a scheme (`contao.dca.<table>.<id>`). We automatically tag all pages, modules, articles and content elements. This means, you can now configure page cache settings to e.g. 30 days (that's the max as of today, we might want to adjust that to allow higher values?) and when you edit this page in the backend or an article, module or content element that's output on that page, the cache entry automatically gets purged. That means no more manual cache deletion for the users, just cache the page for as long as you can and Contao will automatically purge it when needed 😀 

@aschempp you might want to include the response tagger service on your abstract fragments PR 😎

Toflar:
Of course, still depends on https://github.com/contao/manager-bundle/pull/67 which is ready to be merged though, so would be glad if that happens soon @leofeyer so we can continue the work on master 😄

leofeyer:
Shouldn't we also tag pages?

leofeyer:
Ah, I see. It is done in the `FrontendTemplate` class instead of in the `Page*` classes.

src/Resources/contao/classes/FrontendTemplate.php
```diff
@@ -389,6 +390,16 @@ private function setCacheHeaders(Response $response)
 			$response->setSharedMaxAge($objPage->cache);
 		}
 
+		// Tag response
+		if (System::getContainer()->has('fos_http_cache.http.symfony_response_tagger'))
```
leofeyer: I think this code does not belong here. We cannot not assume that the return value of `FrontendTemplate::getResponse()` is always a page, can we? The code should go into the `PageRegular` class IMHO.

Toflar: I think it's correct here. It's the old `FrontenTemplate::output()` that was only called once and sent the response to the browser. It's the same thing for `getResponse()` now and we also set all the other caching headers regarding the page here (like disabling if FE preview is active and stuff).

leofeyer: The old `output()` method still exists and the `getResponse()` method is not its replacement! You might as well use a front end template for an API response, therefore we cannot be sure it is always a page.

Toflar: So why are all the other caching headers set there?

leofeyer: Is this a serious question? An API or ESI response can have cache headers, which does not make them a page.

Toflar: Of course it's a serious question, I don't have time for non-serious questions. ESI requests should certainly not use `FrontendTemplate::getResponse()`. It would set caching headers from **the page configuration** which makes no sense. An ESI fragment should do exactly the opposite.

aschempp: I'm using `FrontendTemplate::getResponse` a lot, so @leofeyer is right with his concern. Though it looks like the method is incorrect and should not add the page cache information…

leofeyer: > It would set caching headers from the page configuration which makes no sense

That's true …

Toflar: So I'm moving the relevant ones to `PageRegular`? Nobody can really re-use that but this concept has to be reworked for CMF routing anyway.

leofeyer: No, you don't have to move them. It is all based on `$objPage`, so either we change this or we might as well leave everything as it is.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -1176,6 +1177,50 @@ protected function panel()
...
+	 * ids.
+	 *
+	 * @param string $table
+	 * @param array $ids
+	 *
+	 * @return array
+	 */
+	protected function getCacheTags($table, array $ids = array())
+	{
+		$ns = 'contao.dca.';
```
leofeyer: As discussed on the hackathon, we want to rename this to `contao.db.`.

src/Resources/contao/classes/DataContainer.php
```diff
@@ -1176,6 +1177,50 @@ protected function panel()
...
+		$table = preg_replace('/^tl_/', '', $table, 1);
```
leofeyer: As also discussed, we don't want to strip the tüdelü prefix.

Toflar:
Updated according to latest discussions. This PR is RTM now imho.

leofeyer:
Thank you @Toflar.

PR #1480, opened 2018-04-13
Update default.xlf

poddus:
fix grammatical errors in the language fallback error messages

### DIFF
--- a/src/Resources/contao/languages/en/default.xlf
+++ b/src/Resources/contao/languages/en/default.xlf
@@ -174,13 +174,13 @@
         <source>Invalid redirect target (circular reference)!</source>
       </trans-unit>
       <trans-unit id="ERR.noFallbackEmpty">
-        <source>None of the active website root pages without an explicit DNS setting has the language fallback option set.</source>
+        <source>None of the active website root pages without an explicit DNS setting have the language fallback option set.</source>
       </trans-unit>
       <trans-unit id="ERR.noFallbackDns">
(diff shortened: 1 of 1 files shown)


leofeyer:
I am pretty sure that the first two changes are wrong. Can you please explain your changes?

poddus:
"pages", being plural, "have" instead of  "has".
reduced sentence:
"none of the pages have the option set"
vs.
"none of the pages has the option set"

leofeyer:
You are right. Reducing the sentence makes it easier to see. 👍

gmpf:
> "pages", being plural, "have" instead of "has".

This reasoning is not correct though. The question is whether the subject of the sentence is singular or plural, and "pages" alone is not the subject of the sentence. The way I understand the grammar in this case, "none" would be the determinant, and "none" can be singular or plural depending on the sense of the sentence.

Compare this to the German version, where it is
>Keiner der aktiven Website-Startpunkte ohne explizite DNS-Einstellungen wurde als Sprachen-Fallback definiert.

with "Keiner" and "wurde" indicating singular use.
>Keine der aktiven Website-Startpunkte ohne explizite DNS-Einstellungen wurden als Sprachen-Fallback definiert.

would also be correct, and it would be the equivalent of this suggested change. Note the slight change in meaning however.

asaage:
https://www.englishforums.com/English/NoneHasOrNoneHave/cqjgq/post.htm

leofeyer:
Thank you @poddus.

PR #1486, opened 2018-04-18
Optimize the StringUtil::deserialize() method

leofeyer:
This PR uses a simple regex to filter strings which are definitely not serialized strings. It noticeably reduces the number of `unserialize()` calls and speeds up the method by up to 70% (see https://3v4l.org/F3gdM and https://3v4l.org/6oSrr).

### DIFF
--- a/src/Resources/contao/library/Contao/StringUtil.php
+++ b/src/Resources/contao/library/Contao/StringUtil.php
@@ -1050,6 +1050,7 @@ class StringUtil
 		}
 
 		// Return if definitely not a serialized string
+		// See https://github.com/php/php-src/blob/master/ext/standard/var_unserializer.c#L711-L722
 		if (!preg_match('/^[CONRSabdiors]:[0-9]+:/', $varValue))
 		{
 			return $blnForceArray ? array($varValue) : $varValue;
--- a/src/Resources/contao/library/Contao/StringUtil.php
+++ b/src/Resources/contao/library/Contao/StringUtil.php
@@ -1049,6 +1049,12 @@ class StringUtil
 			return $blnForceArray ? array() : '';
 		}
 
+		// Return if definitely not a serialized string
+		if (!preg_match('/^[CONRSabdiors]:[0-9]+:/', $varValue))
+		{
+			return $blnForceArray ? array($varValue) : $varValue;
+		}
+
 		// Potentially including an object (see #6724)
 		if (preg_match('/[OoC]:\+?[0-9]+:"/', $varValue))
 		{
--- a/src/Resources/contao/library/Contao/StringUtil.php
+++ b/src/Resources/contao/library/Contao/StringUtil.php
@@ -1049,9 +1049,8 @@ class StringUtil
 			return $blnForceArray ? array() : '';
 		}
 
-		// Return if definitely not a serialized string
-		// See https://github.com/php/php-src/blob/master/ext/standard/var_unserializer.c#L711-L722
-		if (!preg_match('/^[CONRSabdiors]:[0-9]+:/', $varValue))
+		// Not a serialized array (see #1486)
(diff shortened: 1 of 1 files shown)


src/Resources/contao/library/Contao/StringUtil.php
```diff
@@ -1049,6 +1049,12 @@ public static function deserialize($varValue, $blnForceArray=false)
 			return $blnForceArray ? array() : '';
 		}
 
+		// Return if definitely not a serialized string
+		if (!preg_match('/^[CONRSabdiors]:[0-9]+:/', $varValue))
```
Toflar: I've checked the PHP C source and here's the result:

* `C` and `O` are for classes and objects
* `N` is `null`
* `R` and `r` are references
* `a` is an `array`
* `b` is a `boolean`
* `d` is a `double`
* `i` is an `integer`
* `s` is a `string`

I did not find `S` and `o`, what do they stand for?

leofeyer: I don't know. I just know that they exist: https://github.com/php/php-src/blob/master/ext/standard/var_unserializer.c#L711-L722

Toflar: Okay, then please leave that reference as a comment in the source code so that we still know in the future. Otherwise this PR is just fine 👍

leofeyer: Added in 9974fd7e02f836b93c9fed95f79ce14863d720cf. Still I would like to know what `S` and `o` stand for. 🤔 @ausi ?

leofeyer:
Do we want to backport this to Contao 4.4 or even 3.5?

Aybee:
IMHO 4.4 yes. 3.5 not necessary but would be nice, why not?

src/Resources/contao/library/Contao/StringUtil.php
ausi: `S` is for escaped strings, e.g. `S:1:"\5A";` computes to `"Z"`.

`o` is for `stdClass` objects, e.g. `o:1:"s:3:"foo";s:3:"bar";}` computes to an object with a `foo` property.

Both do not get generated by `serialize()` (anymore) AFAIK.

leofeyer:
I have implemented the `strncmp()` check in abcd5be96a90313b308d3ec615d947e34ebdc31c.

Toflar:
BTW: I'm against backporting this. It's a core method used in thousands of places in Contao. It's not a wise decision to bring that to older versions. It's not a bugfix (if performance was horrible, I'd consider it a bugfix but nobody complained so far) so it should go into 4.6.

aschempp:
I am very much not sure about the content always being an array...

Toflar:
Yeah, it would imply a BC break. Until now this worked:
```php
$v = serialize(true);
$v = deserialize($v);
```

It won't be anymore with this change.

ausi:
@Toflar your Code does also not work in the current version for me. I dont see a BC Break here

Toflar:
Ah, I missed line https://github.com/leofeyer/core-bundle/blob/abcd5be96a90313b308d3ec615d947e34ebdc31c/src/Resources/contao/library/Contao/StringUtil.php#L1068. So yes, it can only handle arrays now. PR is good to merge.

leofeyer:
I still think that there is a BC break:

```php
$v = serialize(true);
$v = deserialize($v);
```

In Contao 4.4 `$v` would be `array(true)` and in Contao 4.6 it would be `b:1;`, wouldn't it?

ausi:
No, it would not AFAIK. Did you test this?

leofeyer:
I have now and you are right. There is no BC break. 🙄

Issue #1506, opened 2018-04-30
DropZone cannot handle errors


fritzmg:
If `$this->blnHasError` is set to true for `Contao\DropZone` during `uploadTo(…)` for example, there will simply be a `500` response and the error will not be visible in the DropZone. Instead it will show the following:

![dropzone-error](https://user-images.githubusercontent.com/4970961/39418798-a406ad42-4c5c-11e8-8186-7c4d94d17c6f.png)

In this example I simulated a file with an invalid file name that would trigger
```php
// Invalid file name
if (!\Validator::isValidFileName($file['name']))
{
	\Message::addError($GLOBALS['TL_LANG']['ERR']['filename']);
	$this->blnHasError = true;
}
```
within `FileUpload::uploadTo`. This can be fixed by adding
```diff
	// Redirect or reload
	if (!$objUploader->hasError())
	{
		if ($blnIsAjax)
		{
			throw new ResponseException(new Response(\Message::generateUnwrapped(), 201));
		}

		// Do not purge the html folder (see #2898)
		if (isset($_POST['uploadNback']) && !$objUploader->hasResized())
		{
			\Message::reset();
			$this->redirect($this->getReferer());
		}

		$this->reload();
	}
+	elseif ($blnIsAjax)
+	{
+		throw new ResponseException(new Response(\Message::generateUnwrapped(), 500));
+	}
```
in `DC_Folder::move`. However, then it will still look like this:

![dropzone-error-2](https://user-images.githubusercontent.com/4970961/39418923-75ec1eaa-4c5d-11e8-86b2-3cdff58caacc.png)

Since the `Message` class automatically wraps the messages with HTML and DropZone expects the server response to be pure text.

leofeyer:
See #1506.

PR #1507, opened 2018-04-30
[RFC] disallow upload of large images

fritzmg:
Currently Contao uses the system settings `gdMaxImgWidth` and `gdMaxImgHeight` to prevent memory limit errors while processing large images. If an image is larger than any of those values, the image simply will not be processed and shown as is in the front end.

Despite instructing editors not to upload any large images it still happens that they do. This is problematic for the front end - because then you will have large images that a visitor has to download. And the image could break the layout, if the image is expected to be a certain size or aspect ratio.

With this PR you can disallow the back end upload of images larger than the maximum GD size setting.

disallow

This PR depends on fixing the issue of #1506.


aschempp:
I generally like the idea, however we should not add new options to `tl_settings` as it is kind of deprecated. Such flags should be added to the bundle configuration in my opinion. There already is a `contao.image` configuration namespace.

fritzmg:
Fair enough. What is the general plan with `tl_settings` anyway? People are already complaining that they cannot change certain settings via the system settings in the back end anymore. Will this be done via the Contao Manager in the future? But what's the point of always having to go to the Contao Manager to change certain settings of the Contao installation?

aschempp:
At some point, system configuration and maintenance should all happen in the Manager. Only the Manager is able to configure a non-running Symfony application and rebuild the cache. Also, it's not a task that should be done multiple times, but on initial system setup.

fritzmg:
I have updated PR, using `contao.image.disallow_large_image_upload` now.

fritzmg:
I have renamed the parameter.

Question: should the same apply to front end uploads in the form generator, when enabled? There is already a check like that though - which uses the `imageWidth` and `imageHeight` settings (which is used for automatically resizing the images in the back end upload).

Without any further changes you can achieve the same effect in the front end by setting the `imageWidth` and `gdMaxImgWidth` and `imageHeight` and `gdMaxImgHeight` respectively to the same value.

leofeyer:
Thank you @fritzmg.

PR #1508, opened 2018-04-30
Fixed numeric aliases 404 handling

Toflar:
When you don't have an url suffix configured or it does not start by a `.`, the current regex does not match and thus the 404 page is not shown but instead the content of the page is loaded.

Steps to reproduce:

1. Configure `urlSuffix` to be an empty string `''`.
2. Add some regular content page with the alias `foobar` and note your id (e.g. `42`).
3. Call `/42`.

Expected:

I expect to be redirected to the configured 404 page because the page id 42 has a configured alias `foobar`.

What happens:

I can see the content of page ID `42`.


leofeyer:
Thank you @Toflar.

PR #1509, opened 2018-05-01
[RTM] fix DropZone showing HTML in errors

### LINKED ISSUE #1506: DropZone cannot handle errors

fritzmg:
This fixes #1506. Not the most elegant solution though. Fix is needed for #1507.

### DIFF
--- a/src/Resources/contao/classes/FileUpload.php
+++ b/src/Resources/contao/classes/FileUpload.php
@@ -94,6 +94,7 @@ class FileUpload extends \Backend
 		$maxlength_kb_readable = $this->getReadableSize($maxlength_kb);
 		$arrUploaded = array();
 		$arrFiles = $this->getFilesFromGlobal();
+		$strErrorType = $this instanceof DropZone ? 'TL_RAW' : 'TL_ERROR';
 
 		foreach ($arrFiles as $file)
 		{
@@ -104,7 +105,7 @@ class FileUpload extends \Backend
 			}
 			catch (\InvalidArgumentException $e)
 			{
-				\Message::addError($GLOBALS['TL_LANG']['ERR']['filename']);
+				\Message::add($GLOBALS['TL_LANG']['ERR']['filename'], $strErrorType);
 				$this->blnHasError = true;
 
 				continue;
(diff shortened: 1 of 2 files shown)


xchs:
@fritzmg The PR should be targeted against the [4.4 branch](https://github.com/contao/core-bundle/tree/4.4), shouldn't it?

fritzmg:
Hm, true, I mis-targeted because I have done this for a new feature.

fritzmg:
I forcefully changed the base now... not sure if that was a good idea ;)

leofeyer:
Wouldn't a simple `strip_tags()` do the job, too?

fritzmg:
Well, the `ResponseException` is thrown in `DC_Folder`, however there I have no information about whether the response is for the DropZone or not. Unless we can assume that any AJAX Request wants the messages from the response in pure text instead of HTML?

leofeyer:
> Unless we can assume that any AJAX Request wants the messages from the response in pure text instead of HTML?

I guess we can safely assume this. 😉

fritzmg:
@leofeyer this needs to be merged too - otherwise #1507 won't work properly :)

leofeyer:
Since you did not like the `strip_tags()` approach, I have implemented something else in 8c99c38fdd5b435037ee546b26abe6f4fefd916d. Do you like it?

leofeyer:
I have backported the changes in a53201e0cbd96c177495b92fca298a4b75154ded.

Issue #1518, opened 2018-05-08
Using the core bundle with the configuration described in the README does not work


christian-kolb:
There is information missing to make the bundle run. Different parameters and missing configuration of a few components.

christian-kolb:
I attached the pull request to fix it. Please let me know if there is anything missing.

christian-kolb:
Rest of discussion can happen in the pull request. Will close this ticket in the meantime.

PR #1519, opened 2018-05-08
Add missing and fix incorrect information in README

christian-kolb:
Pull request for issue #1518

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,18 @@ security:
                 authenticator: contao.security.authenticator
 ```
 
-Edit your `app/config/config.yml` file and add the following:
+Edit your `app/config/config.yml` file edit the following parameters:
+
+```yml
+parameters:
+    prepend_locale: false # Set to true if you like, just has to be set
+
+framework:
+    esi: { enabled: true }
+    translator: { fallbacks: ['%locale%'] }
+```
+
+Then add the following to end of the `app/config/config.yml`:
 
 ```yml
 # Contao configuration
@@ -134,7 +145,7 @@ contao:
 
     # Optional parameters
     web_dir:              "%kernel.project_dir%/web"
-    encryption_key:       "%kernel.secret%"
+    encryption_key:       "%secret%"
     url_suffix:           .html
     upload_path:          files
     csrf_token_name:      contao_csrf_token
--- a/README.md
+++ b/README.md
@@ -39,7 +39,9 @@ Edit your `composer.json` file and add the following:
 
 Then run `php composer.phar update` to install the vendor files.
 
-Note that you can exchange the `php-http/guzzle6-adapter` package with any other [HTTP client implementation][4]. If you already have an HTTP client implementation, you can omit the package entirely.
+Note that you can exchange the `php-http/guzzle6-adapter` package with any
+other [HTTP client implementation][4]. If you already have an HTTP client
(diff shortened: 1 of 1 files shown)


christian-kolb:
There is information missing to make the bundle run. Different parameters and missing configuration of a few components. Fixed in this PR.

leofeyer:
Does this refer to the managed edition or did you add the bundle to a Symfony app?

christian-kolb:
This does refer to the core-bundle used in a Symfony app.

README.md
```diff
@@ -124,7 +124,18 @@ security:
                 authenticator: contao.security.authenticator
 ```
 
-Edit your `app/config/config.yml` file and add the following:
+Edit your `app/config/config.yml` file:
+
+```yml
+parameters:
+    prepend_locale: false # Set to true if you like, just has to be set
```
leofeyer: Shouldn't the setting be added to the `parameters.yml.dist` file?

christian-kolb: Yes, that makes even more sense if it's a mandatory parameter.

fritzmg: It's not mandatory. But I prefer to set it in `app/config/config.yml` via
```yml
contao:
    prepend_locale: true
```
because usually I want this to be the same in all environments of the site.

christian-kolb: Ok, then it's ok to put it like this into the config.yml. The issue at the moment is, that it's using a parameter that is not defined. We could also do it with a fixed value and without a parameters entry. If nobody has any issues with this, @fritzmg will you adapt the pull request?

fritzmg: The parameter is already documented in the README:
```yml
# Contao configuration
contao:
    # Required parameters
    prepend_locale: "%prepend_locale%"
```
This means, that you can either override the config directly in your `config.yml` or use a parameter called `prepend_locale` in your `parameters.yml`.

christian-kolb: Yes, but this parameter is not in the `parameters.yml.dist` file. So if you just add the package, the build will just crash with a missing parameter which is not really best practice. So we ether add it to the `parameters.yml.dist` or add a static one into the `config.yml`. I don't really care which way, but an error when adding the core-bundle as a symfony bundle is not nice at all : )

fritzmg: I see what you mean now. Then I agree it should be added to the `parameters.yml.dist`.

However I still disagree that this configuration should be a parameter in the first place. What is the use case where you want this to be a parameter of the environment?

christian-kolb: Yes you're right. Lets remove the parameter and make it a static setting. @leofeyer Do you know why this is a parameter in the first place?

leofeyer: It was in the `parameters.yml.dist` file of the standard edition, so the Incenteev parameters handler would ask you during install. But since the standard edition is deprecated, we should remove the parameter.

leofeyer: @aschempp /cc

leofeyer: See https://github.com/contao/manager-bundle/pull/76

christian-kolb: @leofeyer Does this pull request effect the core-bundle?

leofeyer: Only the `README.md` file.

leofeyer:
Merged in contao/contao@9ca95e2bb86030aaa180e09e271878bd8665e368. Thank you @christian-kolb.

PR #1524, opened 2018-05-11
Fix the potential scandir() error in the helper functions

qzminski:
We have the below exception logged in the Sentry. In fact the `scandir()` can return also a `false` according to the documentation:

```
Returns an array of filenames on success, or FALSE on failure. If directory is not a directory, then boolean FALSE is returned, and an error of level E_WARNING is generated.
```

Exception:

```
ErrorException: Invalid argument supplied for foreach()
#18 vendor/contao/core-bundle/src/Resources/contao/helper/functions.php(73): handleError
#17 vendor/contao/core-bundle/src/Resources/contao/helper/functions.php(73): scan
#16 vendor/contao/core-bundle/src/Resources/contao/drivers/DC_Folder.php(2502): generateTree
#15 vendor/contao/core-bundle/src/Resources/contao/drivers/DC_Folder.php(2455): ajaxTreeView
#14 vendor/contao/core-bundle/src/Resources/contao/classes/Ajax.php(195): executePostActions
#13 vendor/contao/core-bundle/src/Resources/contao/classes/Backend.php(412): getBackendModule
#12 vendor/contao/core-bundle/src/Resources/contao/controllers/BackendMain.php(132): run
#11 vendor/contao/core-bundle/src/Controller/BackendController.php(55): mainAction
#10 vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(151): handleRaw
#9 vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(68): handle
#8 vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php(202): handle
#7 vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php(462): forward
#6 vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/HttpCache/HttpCache.php(57): forward
#5 vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php(234): pass
#4 vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php(251): invalidate
#3 vendor/friendsofsymfony/http-cache/src/SymfonyCache/EventDispatchingHttpCache.php(126): invalidate
#2 vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php(177): handle
#1 vendor/friendsofsymfony/http-cache/src/SymfonyCache/EventDispatchingHttpCache.php(98): handle
#0 web/app.php(33): null
```

### DIFF
--- a/src/Resources/contao/helper/functions.php
+++ b/src/Resources/contao/helper/functions.php
@@ -67,10 +67,16 @@ function scan($strFolder, $blnUncached=false)
 		return $arrScanCache[$strFolder];
 	}
 
+	// Return an empty array if the directory can not be read
+	if (($arrScan = scandir($strFolder)) === false)
+	{
+		return array();
+	}
+
 	$arrReturn = array();
 
 	// Scan directory
-	foreach (scandir($strFolder) as $strFile)
+	foreach ($arrScan as $strFile)
 	{
 		if ($strFile == '.' || $strFile == '..')
 		{


leofeyer:
Why is the function called with an invalid argument? Shouldn't we fix the actual problem in this case?



The image is taken from [MonkeyUser](http://www.monkeyuser.com/2018/root-cause/).

qzminski:
We could add the `is_dir()` check before… but still if for some unknown reason `scandir()` will not like the given path it will return `false` which must be considered. Thus I think the proposed PR is sufficient.

leofeyer:
True, but still we do not know what actually caused the problem. Do you have the corresponding request URL?

qzminski:
Hmm no, not really. We've caught this exception in Sentry only once or maybe a few times but were unable to reproduce it later on.

qzminski:
Ah, we apparently still have it the Sentry. Here is the curl:

```
curl \
 -X POST \
 --compressed \
 -H "Accept: application/json" \
 -H "Accept-Encoding: gzip, deflate, br" \
 -H "Accept-Language: de-DE,de;q=0.9,en-US;q=0.8,en;q=0.7" \
 -H "Connection: close" \
 -H "Content-Length: 159" \
 -H "Content-Type: application/x-www-form-urlencoded; charset=UTF-8" \
 -H "Host: domain.tld" \
 -H "Origin: https://domain.tld" \
 -H "Referer: https://domain.tld/" \
 -H "User-Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/66.0.3359.170 Safari/537.36" \
 -H "X-Accel-Internal: /internal-nginx-static-location" \
 -H "X-Real-Ip: 46.140.81.190" \
 -H "X-Request: JSON" \
 -H "X-Requested-With: XMLHttpRequest" \
 --data "level=2&action=loadFileManager&state=1&REQUEST_TOKEN=yYhq26U4GFjGqX7iD4MAvjM9Di8GtFqNee9LscS8eBc&folder=files%2Fbynder_assets%2Fg%EF%BF%BD&id=filetree_0d617dd2" \
 "https://domain.tld/contao?do=files&ref=jqxgXnCX"
```

Toflar:
The PR is certainly okay. The function should not cause warnings but instead return false. But @leofeyer is right that the root cause should be fixed as well. @qzminski can you fix that as well? It's quite obvious (`DC_Folder.php(2502): generateTree`).

leofeyer:
I don't think it is obvious. It seems that the `Ajax` class is called with an invalid argument:

https://github.com/contao/core-bundle/blob/8459f477b77c83f1631a34cdbbb0ddf0b5c5f6f2/src/Resources/contao/classes/Ajax.php#L193-L195

But the request is made by our `core.js` file, so `\Input::post('folder', true)` should always be a folder, shouldn't it?

leofeyer:
Taken from the request URL: `&folder=files%2Fbynder_assets%2Fg%EF%BF%BD`

Is `folder=files` correct?

Toflar:
The folder `files/bynder_assets` exists, yes.

leofeyer:
The decoded path would be `files/bynder_assets/g�` though. I guess that is the problem.

Toflar:
Ah, you're absolutely right. That was an issue on my side which I've fixed. That's a follow up issue then that's not related to Contao. Thanks a lot for helping! Can be closed.

PR #1527, opened 2018-05-17
[RFC] Fix breadcrumb schema

### LINKED ISSUE #1421: SEO: Make Frontend-Breadcrumb accessible for Google

ausi:
This should fix #1421 and #1429.

I think the problem why Google does not show the breadcrumb in the search results is because of a missing link between the webpage and the breadcrumb. This can be fixed by adding `itemprop="breadcrumb"` to mod_breadcrumb and removing the header, main and footer `itemscope`s from fe_page.

breadcrumbs

@Xendiadyon can you test in one of your pages if the changes from this pull request fix the issue?

### DIFF
--- a/src/Resources/contao/templates/frontend/fe_page.html5
+++ b/src/Resources/contao/templates/frontend/fe_page.html5
@@ -34,7 +34,7 @@
 
       <?php $this->block('header'); ?>
         <?php if ($this->header): ?>
-          <header id="header" itemscope itemtype="http://schema.org/WPHeader">
+          <header id="header">
             <div class="inside">
               <?= $this->header ?>
             </div>
@@ -48,7 +48,7 @@
         <div id="container">
 
           <?php $this->block('main'); ?>
-            <main id="main" itemscope itemtype="http://schema.org/WebPageElement" itemprop="mainContentOfPage">
+            <main id="main">
               <div class="inside">
                 <?= $this->main ?>
               </div>
@@ -83,7 +83,7 @@
 
       <?php $this->block('footer'); ?>
         <?php if ($this->footer): ?>
-          <footer id="footer" itemscope itemtype="http://schema.org/WPFooter">
+          <footer id="footer">
             <div class="inside">
               <?= $this->footer ?>
             </div>
--- a/src/Resources/contao/templates/modules/mod_breadcrumb.html5
+++ b/src/Resources/contao/templates/modules/mod_breadcrumb.html5
@@ -2,7 +2,7 @@
 
 <?php $this->block('content'); ?>
 
-  <ul itemscope itemtype="http://schema.org/BreadcrumbList">
+  <ul itemprop="breadcrumb" itemscope itemtype="http://schema.org/BreadcrumbList">
     <?php foreach ($this->items as $item): ?>
       <?php if ($item['isActive']): ?>
         <li class="active<?php if ($item['class']): ?> <?= $item['class'] ?><?php endif; ?> last" itemscope itemtype="http://schema.org/ListItem" itemprop="itemListElement"><span itemprop="name"><?= $item['link'] ?></span></li>


CMSworker:
Did somebody test this successfully in an active live installation?

In my case it didn't solve the problem of Google showing the homepage link twice. Still looking like this:

`https://www.domain.de > Homepage > Subpage`

instead of

`https://www.domain.de > Subpage`

I thougt the problem could possibly be the wrong id of the webpage, when the breadcrumbs are now a microdata child element of the whole page.

Like in your screenshot, right side, 2. line:

`@id: https://.../top`

The id "top" is set in line 28 in fe_page.html5 together with itemtype="http://schema.org/WebPage" :

`<body id="top" class="...`

Because of the anchor id there is a difference in the URL of the webpage element and the URL in the breadcrumb element. So probably machines see this as two different entities/webpages and the breadcrumb and page-id-url will never be identical?

I changed it for testing and added an explizit microdata itemid, so the url of the webpage-element is the same as within the breadcrumb link:

`<body id="top" itemid="{{env::path}}{{env::request}}" class="...`

Didn't work.

What are your findings?

leofeyer:
Thank you @ausi.

@CMSworker I have created a new issue for the problem that Google is showing the homepage link twice (see #1561).

CMSworker:
@leofeyer Thanks for merging. Might have been too quick as nobody else tested it? But anyway, now it's merged and we'll see if it works in active installations.

For the record: In my test I just adapted the microdata nesting from @ausi so that the breadcrumb list is a child element of the webpage element - and removed WPHeader, mainContentOfPage and WPFooter.

But I did'n remove "position", item "Thing" from my previous suggestion in #1429 because those elements are clearly defined as "required" by Google: https://developers.google.com/search/docs/data-types/breadcrumb#type_definitions

Well, let's wait and see what happens after rollout...

ausi:
> But I did'n remove "position" and item "Thing" from my previous suggestion in #1429 because those elements are clearly defined as "required" by Google: https://developers.google.com/search/docs/data-types/breadcrumb#type_definitions

As you can see in my screenshot from above, the validator doesn’t show errors or warnings, so I think it is a documentation error.

CMSworker:
I'd like to repeat my question: **Did somebody test this successfully in an active live installation?**

The merged fix from @ausi should be included in the online demo already, correct?

If so: Google doesn't show any breadcrumbs for pages of [demo.contao.org](https://demo.contao.org) yet.

Here some screenshots: (tested with page https://demo.contao.org/en/link-elements.html )

screenshot2

screenshot1

screenshot3

screenshot4

As @fritzmg [mentioned here in #1429](https://github.com/contao/core-bundle/pull/1429#issuecomment-380572283), you are not trying to implement it in a way it "best fits Google". Unfortunately I don't know any other service besides search engines, that are interpreting breadcrumb microdata. Could you test it anywhere successfully, if this fix is working correctly in the way you expected it to work?

From my point of view, this fix didn't solve #1421 (or #1429), so I vote to reopen them ;-)

PR #1536, opened 2018-05-31
fix mode 1 sorting by flag for multiple fields

patrickjDE:
WIth the following dca configuration
```
    ...
    'list' => [
        'sorting' => [
            'mode' => 1,
            'fields' => ['date_start', 'date_stop'],
            'flag' => 8, // sort descending by month
        ],
        ...
```
I'd expect the records to be sorted first descending by `date_start` then descending by `date_stop`, instead they are sorted ascending by `date_start` and descending by `date_stop`.

This is caused by https://github.com/contao/core-bundle/blob/4.4/src/Resources/contao/drivers/DC_Table.php#L4664 where the SQL for the sorting direction is only added once at the end, instead of being added to every field.

This PR fixes this issue by setting the sorting for each field according to the chosen flag, except where it was set explicitly (e.g. `'fields' => ['date_start', 'date_stop ASC'],`).

It also fixes an SQL syntax error when neither sorting fields nor sorting flag are given (`DESC` was added to the query string, although there was no `ORDER BY ...` part).
Futher it fixes another SQL syntax error, when an even flag or no flag at all is given, and the last field has its sorting set explicitly (`DESC` was added twice in this case).

### DIFF
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -4606,6 +4606,14 @@ class DC_Table extends \DataContainer implements \listable, \editable
 			{
 				list($key, $direction) = explode(' ', $v, 2);
 
+				if ($direction === null)
+                {
+                    if ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 1 && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] % 2) == 0)
+                    {
+                        $direction = "DESC";
+                    }
+                }
+
 				if ($GLOBALS['TL_DCA'][$this->strTable]['fields'][$key]['eval']['findInSet'])
 				{
 					if (\is_array($GLOBALS['TL_DCA'][$this->strTable]['fields'][$key]['options_callback']))
@@ -4659,11 +4667,6 @@ class DC_Table extends \DataContainer implements \listable, \editable
 			{
 				$query .= " ORDER BY " . implode(', ', $orderBy);
 			}
-		}
-
-		if ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 1 && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] % 2) == 0)
-		{
-			$query .= " DESC";
 		}
 
 		$objRowStmt = $this->Database->prepare($query);
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -4606,11 +4606,21 @@ class DC_Table extends \DataContainer implements \listable, \editable
 			{
 				list($key, $direction) = explode(' ', $v, 2);
 
-				if ($direction === null)
+				if ($direction === null && $GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 1)
                 {
-                    if ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 1 && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] % 2) == 0)
+                    if (isset($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag']))
                     {
-                        $direction = "DESC";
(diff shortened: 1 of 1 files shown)


leofeyer:
You are right that the current logic is incorrect. However, your fix does not seem to fix it.

Contao considers two flags, the `list.sorting.flag` and the `fields.<field>.flag`. If you want to sort by both fields descending, you have to make sure that both `date_start` and `date_stop` have `'flag' => 8` set. Then the query will be:

```sql
… ORDER BY date_start DESC, date_stop DESC
```

So now the question remains why there is an additional `list.sorting.flag` in sorting mode 1? I don't recall. @contao/developers Do you?

patrickjDE:
You're right, as my test cases were dates (and thus had `fields.<field>.flag` set for formatting reasons), I didn't notice the value wasn't applied if the field has no date sorting flag. I also noticed another problem with my fix (with no flag at all would sort DESC).

I've updated the PR to also take `fields.<field>.flag` into account and apply sorting with the following priority:
1. explicit in `list.sorting.fields` (e.g. `'fields' => ['date_start', 'date_stop ASC'],`).
2. `fields.<field>.flag`
3. `list.sorting.flag` (can simply be removed, if the flag is deemed superfluous)


Additionally, thinking a bit about it, I'm not sure why this is limited to mode 1, shouldn't mode 2 (and mode 3, but that is broken beyond repair, afair) also respect the sorting directions defined in the dca?

leofeyer:
> shouldn't mode 2 also respect the sorting directions defined in the dca?

Probably yes.

Regarding the priority in your implementation: Shouldn't the `list.sorting.flag` override the `fields.<field>.flag`? So you can override the default sorting that is e.g. used in the filter menus with a custom sorting order?

@contao/developers /cc

patrickjDE:
> Regarding the priority in your implementation: Shouldn't the list.sorting.flag override the fields.<field>.flag? So you can override the default sorting that is e.g. used in the filter menus with a custom sorting order?

I'm not sure. In https://github.com/contao/core-bundle/blob/4.4/src/Resources/contao/drivers/DC_Table.php#L4367 and  https://github.com/contao/core-bundle/blob/4.4/src/Resources/contao/drivers/DC_Table.php#L4855 `fields.<field>.flag` also takes precedence over `list.sorting.flag`.
Also it feels just natural to me, that a setting that applies to a single field (`fields.<field>.flag`) overrides the setting that applies to all fields (`list.sorting.flag`).

patrickjDE:
Well, the actual use-case is simple: Sort a mode 1 dca by two fields, both descending. IMO a valid use-case, especially for dates.

I do realize that it is currently possible to achieve the desired result with explicitly defining the sorting SQL, but both ways to get there are counter-intuitive and IMO semantically wrong:
1. Set DESC for every field, but the last:
`'fields' => ['date_start DESC', 'date_stop'],`
2. Set DESC for both fields, but also add `list.sorting.flag` with an odd value (which means sort ascending):
`'fields' => ['date_start DESC', 'date_stop DESC'], 'flag' => 7,`

aschempp:
are we gonna merge/fix this at all or close the PR?

leofeyer:
@aschempp That depends on whether or not you (or rather we) consider Patrick's use case to be valid or not. 🤓

leofeyer:
> Set DESC for every field, but the last:
> `'fields' => ['date_start DESC', 'date_stop'],`

As discussed in Mumble on June 6th, this is a bug. We have to add an `isset($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'])` check here:

https://github.com/contao/core-bundle/blob/master/src/Resources/contao/drivers/DC_Table.php#L4704-L4707

patrickjDE:
That would fix the sql syntax bugs, but setting `['list']['sorting']['flag']` would still only apply to the last item, instead of all, or am I missing something?

leofeyer:
You are right. We have to move the whole `if ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['mode'] == 1 && ($GLOBALS['TL_DCA'][$this->strTable]['list']['sorting']['flag'] % 2) == 0)` condition into the foreach loop, so the sorting flag is added to every field. But we still have to add the `isset()` check.

leofeyer:
Eventually fixed in fc9a68639b4cab27f9951fb2071fe82a6b2dc1b3. Thank you @patrickjDE.

Issue #1479, opened 2018-04-13
purpose of try catch block in addImageToTemplate?


fritzmg:
What is the purpose of
```php
try
{
    $objFile = new \File($arrItem['singleSRC']);
}
catch (\Exception $e)
{
    $objFile = new \stdClass();
    $objFile->imageSize = false;
}
```
at the beginning of [`Controller::addImageToTemplate`](https://github.com/contao/core-bundle/blob/4.5.7/src/Resources/contao/library/Contao/Controller.php#L1451-L1459)?

If the file does not exist for example, the code will fail [at the following point](https://github.com/contao/core-bundle/blob/4.5.7/src/Resources/contao/library/Contao/Controller.php#L1548-L1553) anyway:
```php
// Image dimensions
if ($objFile->exists() && ($imgSize = $objFile->imageSize) !== false)
{
    $objTemplate->arrSize = $imgSize;
    $objTemplate->imgSize = ' width="' . $imgSize[0] . '" height="' . $imgSize[1] . '"';
}
```
```
Attempted to call an undefined method named "exists" of class "stdClass".
```

leofeyer:
@ausi /cc

ausi:
I can’t remember. The related ticket is contao/core#7383 I think, but I don’t know why the call to `new \File()` should be in a `try` block.

fritzmg:
In that PR there is no `$objFile->exists()` yet, so it would not fail at least.

leofeyer:
As discussed in Mumble on May 17th, we have to check what the method does if the image does not exist. Then we have to make sure it does the same if the exception is thrown.

leofeyer:
Ok, so here is what happens if the image does not exist:

```
Warning: filemtime(): stat failed for files/contaodemo/media/content-images/xDSC_5276.jpg
  at vendor/contao/core-bundle/src/Resources/contao/library/Contao/File.php:234
  at Contao\File->__get('mtime')
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/File.php:259)
  at Contao\File->__get('imageSize')
     (vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:1435)
```

However, if `singleSRC` points to a directory:

```
Attempted to call an undefined method named "exists" of class "stdClass".
  at vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:1521
```

Both cases are not handled appropriately if you ask me. 😄

[...]

PR #1538, opened 2018-06-04
Fix a potential problem with image object class ambiguity in Controller::addImageToTemplate method

qzminski:
In the `Controller::addImageToTemplate` method if the `File` cannot be initialized it is replaced by `stdClass` instance:

https://github.com/contao/core-bundle/blob/4.4/src/Resources/contao/library/Contao/Controller.php#L1431

Thus it can result in an error because the further code assumes it is always the `File` instance and happily tries to call its methods. We have caught this error using Sentry, here is the log:

```
ErrorException: Uncaught Symfony\Component\Debug\Exception\UndefinedMethodException: Attempted to call an undefined method named "exists" of class "stdClass". in /var/www/vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php:1521
Stack trace:
#0 /var/www/src/AppBundle/Resources/contao/templates/elements/rsce_slider.html5(62): Contao\Controller::addImageToTemplate(Object(stdClass), Array)
#1 /var/www/vendor/contao/core-bundle/src/Resources/contao/library/Contao/TemplateInheritance.php(96): include('/var/www/...')
#2 /var/www/vendor/contao/core-bundle/src/Resources/contao/library/Contao/Template.php(268): Contao\Template->inherit()
#3 /var/www/vendor/contao/core-bundle/src/Resources/contao/classes/FrontendTem
#1 vendor/contao/core-bundle/src/Resources/contao/library/Contao/Controller.php(1521): handleFatalError
#0 vendor/sentry/sentry/lib/Raven/ErrorHandler.php(0): null
```

### DIFF
--- a/src/Resources/contao/library/Contao/Controller.php
+++ b/src/Resources/contao/library/Contao/Controller.php
@@ -1518,7 +1518,7 @@ abstract class Controller extends \System
 		}
 
 		// Image dimensions
-		if ($objFile->exists() && ($imgSize = $objFile->imageSize) !== false)
+		if (is_a($objFile, \File::class) && $objFile->exists() && ($imgSize = $objFile->imageSize) !== false)
 		{
 			$objTemplate->arrSize = $imgSize;
 			$objTemplate->imgSize = ' width="' . $imgSize[0] . '" height="' . $imgSize[1] . '"';


m-vo:
possibly related to #1479

qzminski:
True, sorry I haven't searched for it. Can close then because of duplicate.

PR #1544, opened 2018-06-09
[RFC] Fix abandoned packages on composer install/update

bytehead:
```
Package guzzle/guzzle is abandoned, you should avoid using it. Use guzzlehttp/guzzle instead.
Package satooshi/php-coveralls is abandoned, you should avoid using it. Use php-coveralls/php-coveralls instead.
```
Don't know if we can use `php-coveralls/php-coveralls` in version `2.x`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,4 +43,4 @@ script:
   - php vendor/bin/phpunit $COVERAGE
 
 after_script:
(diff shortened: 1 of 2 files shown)


leofeyer:
Thank you @bytehead.

PR #1545, opened 2018-06-09
[RFC] two-factor authentication 

bytehead:
This is a working draft to add a two-factor authentication mechanism [2FA](https://en.wikipedia.org/wiki/Multi-factor_authentication) with [TOTP](https://en.wikipedia.org/wiki/Time-based_One-time_Password_algorithm) for backend users.
It uses [`scheb/two-factor-bundle`](https://github.com/scheb/two-factor-bundle) to extend the symfony security authentication mechanism, [`spomky-labs/otphp`](https://github.com/Spomky-Labs/otphp) for generating and validating one-time passwords and [`bacon/bacon-qr-code`](https://github.com/Bacon/BaconQrCode) to render QR codes as SVGs.

Changes to `security.yml`:
```yaml
security:
    firewalls:
        contao_backend:
            two_factor:
                auth_form_path: contao_backend_two_factor
                check_path: contao_backend_two_factor_check
                auth_code_parameter_name: verify

    access_control:
        - { path: ^/contao/two-factor, roles: [IS_AUTHENTICATED_2FA_IN_PROGRESS, ROLE_USER] }
```

Changes to `AppKernel.php`:
```php
/**
     * {@inheritdoc}
     */
    {
        $bundles = [
  			// ...
            new Scheb\TwoFactorBundle\SchebTwoFactorBundle(),
        ];
		// ...
    }
```

How to enforce 2FA for backend users via `config.yml`:
```yaml
contao:
    authentication:
        two_factor:
            enforce_backend: true    #default: false
```

### ToDos:

- [x] Move QR code image to user profile
- [x] Unit tests
- [x] Option to enforce 2FA for all backend users
- [x] Icon in the backend list view of `tl_user` indicating which users have 2FA enabled

### Possible new features (future PRs):
- 2FA also for frontend users


richardhj:
🙌

I'm really looking forward to see this feature.

- You aren't using the Google Authenticator already part of `scheb/two-factor-bundle` because it is not that generic than the implementation of `spomky-labs/otphp`?
- We might want an icon in the backend list view of tl_user indicating which users have 2FA enabled.
- We might want to force users to use 2FA.
- 2FA must only get activated after the user successfully confirmed the setup with a token generated on his 2FA device.

And currently im getting the `Unrecognized option "two_factor" under "security.firewalls.contao_backend"` error. Do I need to alter the configuration in the DI extension?

bytehead:
> You aren't using the Google Authenticator already part of `scheb/two-factor-bundle` because it is not that generic than the implementation of `spomky-labs/otphp`?

Yes I don't use it. It didn't work with all of my apps I've tried upfront. That's why I decided to choose a library that supports actually the RFCs.

> We might want an icon in the backend list view of tl_user indicating which users have 2FA enabled.

I like the idea, @contao/developers what do you think about?

> We might want to force users to use 2FA.

IMHO the users should enable 2FA by themselves.

> 2FA must only get activated after the user successfully confirmed the setup with a token generated on his 2FA device.

I will implement this, already discussed 👍

> And currently im getting the Unrecognized option "two_factor" under "security.firewalls.contao_backend" error. Do I need to alter the configuration in the DI extension?

Did you miss to enable the `SchebTwoFactorBundle`?

leofeyer:
>> We might want an icon in the backend list view of tl_user indicating which users have 2FA enabled.
>
> I like the idea, @contao/developers what do you think about?

Go for it.

>> We might want to force users to use 2FA.
>
> IMHO the users should enable 2FA by themselves.

I also think we should not force them.

fritzmg:
> > > We might want to force users to use 2FA.
> 
> > IMHO the users should enable 2FA by themselves.
> 
> I also think we should not force them.

I think @richardhj meant that you should _optionally_ be able to enforce 2FA for one Contao Installation for everyone. Not that it should be enforced in general.

richardhj:
> I think @richardhj meant that you should optionally be able to enforce 2FA for one Contao Installation for everyone. Not that it should be enforced in general. This makes sense imho.

Optionally for some users or user groups, in order to comply with some TOMs  :-) 
Fore sure it is not necessary.

bytehead:
> I think @richardhj meant that you should optionally be able to enforce 2FA for one Contao Installation for everyone. Not that it should be enforced in general. This makes sense imho.

I don't get this one.

If somebody else enforces 2FA in general the question will be how the users will initially setup the app in this case?

aschempp:
> > We might want an icon in the backend list view of tl_user indicating which users have 2FA enabled.

> I like the idea, @contao/developers what do you think about?

I like it as well. Maybe a badge on the user/admin icon like the page publishing options?


> If somebody else enforces 2FA in general the question will be how the users will initially setup the app in this case?

Something like forcing the setup after login attempt?

leofeyer:
Please don't do this. None of the apps I am using 2FA with does it – and I am using a lot of them.

patrickjDE:
As an _option_ to enforce 2FA upon users would be, well.. _optional_, I don't see a problem there. It'd work analogous to "Password change required". Upon first/next login the user is required to change his password/setup 2FA, whatever the admin's reasons for that decision might be.

If 2FA becomes a core feature, so should the _option_ to enforce its usage upon all or specific users.

fritzmg:
I agree with @patrickjDE . A company should be able to decide whether to force all editors of their website two use 2FA in order to increase security.

@leofeyer you are probably talking about a front end context there, right? But this is about the back end functionality. If your website has 2 administrators for the back end and only one uses 2FA, the other account is still (maybe) more vulnerable. Hence you might want to decide that every back end user (or specific back end user groups - or just administrators) should use 2FA.

bytehead:
> We might want an icon in the backend list view of tl_user indicating which users have 2FA enabled.

@frontendschlampe do you want to create an icon for it? 😎

src/Resources/contao/dca/tl_user.php
```diff
@@ -985,4 +1019,155 @@ public function toggleVisibility($intId, $blnVisible, DataContainer $dc=null)
...
+				continue;
+			}
+
+			// Don't show 2FA options, if it is not the logged in user itself and 2FA is disabled.
+			if ($dc->id !== $this->User->id && (int) $activeRecord->use2fa !== '1')
+			{
+				continue;
+			}
+
+			$GLOBALS['TL_DCA']['tl_user']['palettes'][$palette] = str_replace
```
Toflar: Why not use the `PaletteManipulator` here? Would make things easier 😄

src/Resources/contao/dca/tl_user.php
```diff
@@ -985,4 +1019,155 @@ public function toggleVisibility($intId, $blnVisible, DataContainer $dc=null)
...
+	protected function generate2faSecret(DataContainer $dc)
+	{
+		$secret = $dc->activeRecord->secret;
+
+		if ($secret === null)
+		{
+			// Generate 1024 bit secret
+			$secret = random_bytes(128);
+
+			$this->Database->prepare("UPDATE tl_user SET secret=? WHERE id=?")
```
Toflar: Shouldn't that be something like this?

```php
$user = BackendUser::getInstance();
$user->secret = random_bytes(128);
$user->save();
```

src/Resources/contao/dca/tl_user.php
```diff
@@ -985,4 +1019,155 @@ public function toggleVisibility($intId, $blnVisible, DataContainer $dc=null)
...
+		/** @var \Symfony\Component\HttpFoundation\Request $request */
+		$request = System::getContainer()->get('request_stack')->getCurrentRequest();
+
+		$class = $GLOBALS['TL_DCA'][$dc->table]['fields'][$dc->field]['eval']['tl_class'] . ' 2fa-qr-code';
+
+		return '
+<div class="' . $class . ' widget">
+  <div id="ctrl_' . $dc->field . '" class="">
+    <h3><label for="ctrl_' . $dc->field . '">' . $GLOBALS['TL_LANG']['tl_user']['2faQrCode'][0] . '</label></h3>
+    <img src="data:image/svg+xml;base64,' . base64_encode($twoFactorAuthenticator->getQrCode($this->User, $request)) . '" />
```
Toflar: Can we use `BackendUser::getInstance()` everywhere instead of `$this->User` that's stored somewhere in nowhere?

src/Resources/contao/dca/tl_user.php
```diff
@@ -985,4 +1019,155 @@ public function toggleVisibility($intId, $blnVisible, DataContainer $dc=null)
...
+		if (false === $twoFactorAuthenticator->validateCode($this->User, $varValue))
+		{
+			// Disable 2FA, otherwise 2FA stays enabled if the user leaves the window
+			$this->Database->prepare("UPDATE tl_user SET use2fa='' WHERE id=?")
+				->execute($dc->id);
+
+			throw new Exception($GLOBALS['TL_LANG']['ERR']['invalidTwoFactor']);
+		}
+
+		return 1;
```
Toflar: Imho `true` should work.

src/Security/TwoFactor/ContaoTwoFactorAuthenticator.php
```diff
@@ -0,0 +1,84 @@
...
+{
+    /**
+     * @param User   $user
+     * @param string $code
+     *
+     * @return bool
+     */
+    public function validateCode(User $user, string $code): bool
+    {
+        // The 2FA app from Google (Google authenticator) does not strictly confirm to RFC 4648 [1] (they confirm to the old RFC 3548 [2]).
```
Toflar:
```php
private function getUpperUnpaddedScretForUser(User $user): string
```

Then you need to add the google comment only once :-P

src/Resources/contao/dca/tl_user.php
bytehead: Definitely! 😎

src/Resources/contao/dca/tl_user.php
bytehead: Yep. Can't see the forest for the trees.

src/Resources/contao/dca/tl_user.php
bytehead: 👍

src/Security/TwoFactor/ContaoTwoFactorAuthenticator.php
bytehead: True that.

frontendschlampe:
> @frontendschlampe do you want to create an icon for it?

yes ... :-) Today is icon day :-D

frontendschlampe:
What do you think?

2fa

richardhj:
Love it. 
Should be on the left side since it is no action button though.

bytehead:
@frontendschlampe Yeah, I like! I'll place it on the left as @richardhj already stated.

bytehead:
Here is the implemented version:

bildschirmfoto 2018-06-20 um 07 33 06

bytehead:
Here is my new screenshot with the smaller icons:

bildschirmfoto 2018-06-20 um 07 33 06

bytehead:
So I'll wait with the finalization until #1575 is merged.

leofeyer:
I would actually prefer a lock overlay instead of a separate icon. 🙈

frontendschlampe:
> I would actually prefer a lock overlay instead of a separate icon.

What do you mean with "lock overlay"? The user icon with lock - like secure pages?

leofeyer:
Yes.

frontendschlampe:
2fa_2

bytehead:
Nice, I like it more this way!

frontendschlampe:
@leofeyer how should I name the files?

- `user_2fa.svg`
- `user_2fa_.svg`
- `admin_2fa.svg`
- `admin_2fa_.svg`

bytehead:
You would not see "2FA", even on its own (bigger) icon it's rarely readable.

frontendschlampe:
> Maybe it would be too small to read though…

yes ... I tried ... it was to small ... with the big lock it was hard to read and with the small one ;-)

leofeyer:
> @leofeyer how should I name the files?

@frontendschlampe Exactly like you did. 😄

frontendschlampe:
ok ... update in work :-)

frontendschlampe:
I added the icons in my PR https://github.com/contao/core-bundle/pull/1575 - I've changed the position of the lock because of the info circle in inactive status of user/admin icon. Ok for you?

bytehead:
Updated preview:

bildschirmfoto 2018-06-22 um 10 00 27
bildschirmfoto 2018-06-22 um 10 00 49

leofeyer:
Icons have been merged in f0ff185eada20b9bf259420c865d4550e2be6b78.

src/DependencyInjection/ContaoCoreExtension.php
```diff
@@ -81,6 +81,7 @@ protected function loadInternal(array $mergedConfig, ContainerBuilder $container
         $container->setParameter('contao.image.valid_extensions', $mergedConfig['image']['valid_extensions']);
         $container->setParameter('contao.image.imagine_options', $mergedConfig['image']['imagine_options']);
         $container->setParameter('contao.image.reject_large_uploads', $mergedConfig['image']['reject_large_uploads']);
+        $container->setParameter('contao.authentication.2fa.enforce_backend', $mergedConfig['authentication']['2fa']['enforce_backend']);
```
leofeyer: Shouldn't we re-use the `contao.security` domain here?

src/Resources/config/services.yml
```diff
@@ -430,6 +430,26 @@ services:
...
 
+    contao.security.two_factor.authenticator:
+        class: Contao\CoreBundle\Security\TwoFactor\ContaoTwoFactorAuthenticator
+        public: true
+
+    contao.security.two_factor.backend.form_renderer:
+        class: Contao\CoreBundle\Security\TwoFactor\ContaoBackendTwoFactorFormRenderer
+        arguments:
+            - "@router"
+        public: false
```
leofeyer: This is not needed, as services are private by default.

src/Controller/BackendController.php
```diff
@@ -238,4 +242,47 @@ public function pickerAction(Request $request): RedirectResponse
...
+        }
+
+        return $this->loginAction();
+    }
+
+    /**
+     * Dummy action for symfony security
+     * See https://symfony.com/doc/current/reference/configuration/security.html#check-path.
+     *
+     * @Route("/contao/2fa_check", name="contao_backend_2fa_check")
```
aschempp: are you sure this route is needed? We currently never use check routes, we just use the regular route and it will match for POST requests…

bytehead: I could not get it to work without. See [here](https://symfony.com/doc/current/reference/configuration/security.html#check-path) and [here](https://github.com/scheb/two-factor-bundle/blob/master/Resources/doc/configuration.md#configuration).

src/DependencyInjection/ContaoCoreExtension.php
bytehead: We could. I've talked about the naming of this with @Toflar.

src/Resources/config/services.yml
bytehead: 👍

src/DependencyInjection/ContaoCoreExtension.php
leofeyer: We should. 😄

src/Controller/BackendController.php
bytehead: I even don't get it how it works without a check route for the current security.

leofeyer:
Once you have set up 2FA, the QR code remains visible in the back end profile. None of the 2FA apps I am using (including GitHub) does this though.





I think we should not show the QR code once 2FA has been activated.

bytehead:
I'm fine with that. I'll update the PR accordingly.

src/Resources/contao/controllers/BackendIndex.php
```diff
@@ -68,10 +76,30 @@ public function run()
...
 
 		/** @var BackendTemplate|object $objTemplate */
 		$objTemplate = new \BackendTemplate('be_login');
+		$objTemplate->action = ampersand(\Environment::get('request'));
+		$objTemplate->headline = $GLOBALS['TL_LANG']['MSC']['loginBT'];
+
+		/** @var Request $request */
+		$request = $container->get('request_stack')->getCurrentRequest();
+
+		if ($request->getRequestUri() === $router->generate('contao_backend_2fa'))
```
leofeyer: This seems wrong to me. The `contao_backend_2fa` route must not be publicly callable by just adding `/contao/2fa` to the URL. The same is true for the `contao_backend_2fa_check` route.

src/Controller/BackendController.php
```diff
@@ -238,4 +242,47 @@ public function pickerAction(Request $request): RedirectResponse
...
+
+        if (!$user instanceof BackendUser) {
+            return new RedirectResponse($router->generate('contao_backend_login'));
+        }
+
+        return $this->loginAction();
+    }
+
+    /**
+     * Dummy action for Symfony security.
```
leofeyer: Are we 100% sure that this dummy action is required? @aschempp /cc

src/Resources/contao/controllers/BackendIndex.php
bytehead:
```yaml
#security.yml
    access_control:
        - { path: ^/contao/2fa, role: IS_AUTHENTICATED_2FA_IN_PROGRESS }
```

bytehead: See here: 
https://github.com/bytehead/core-bundle/blob/b3109397897e918c40aa48b57e9ec4fa15125a07/src/Controller/BackendController.php#L251-L261

leofeyer: Ah, true.

src/Controller/BackendController.php
aschempp: https://github.com/symfony/symfony-docs/issues/6126

BTW, that someone at SymfonyCon was me 😂

bytehead: Then I get `ERR_TOO_MANY_REDIRECTS`.

src/Security/TwoFactor/BackendProvider.php
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+namespace Contao\CoreBundle\Security\TwoFactor;
+
+use Contao\User;
+use Scheb\TwoFactorBundle\Security\TwoFactor\AuthenticationContextInterface;
+use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\TwoFactorFormRendererInterface;
+use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\TwoFactorProviderInterface;
+
+class BackendProvider implements TwoFactorProviderInterface
```
aschempp: This class looks like it's a generic provider that would work with the front end as well?

leofeyer: True.

src/Resources/contao/languages/en/tl_user.xlf
```diff
@@ -293,6 +293,27 @@
...
+        <source>Use two-factor authentication</source>
+      </trans-unit>
+      <trans-unit id="tl_user.use2fa.1">
+        <source>Use two-factor authentication to log into the Contao back end.</source>
+      </trans-unit>
+      <trans-unit id="tl_user.2faQrCode.0">
+        <source>QR code</source>
+      </trans-unit>
+      <trans-unit id="tl_user.2faQrCode.1">
+        <source>Please scan the QR code with a 2FA/TOTP app of your choice (e.g. 1password).</source>
```
aschempp: `1Password` is written with uppercase P. We should also add Google Authenticator, or maybe a link to a tool site?

src/Controller/BackendController.php
```diff
@@ -238,4 +242,47 @@ public function pickerAction(Request $request): RedirectResponse
...
+    public function twoFactorAuthenticationAction(): Response
+    {
+        $this->get('contao.framework')->initialize();
+
+        /** @var RouterInterface $router */
+        $router = $this->get('router');
+        $token = $this->get('security.token_storage')->getToken();
+
+        if (!$token instanceof TwoFactorToken) {
+            return new RedirectResponse($router->generate('contao_backend_login'));
```
aschempp: could use `$this->redirectToRoute(…)`

src/Security/TwoFactor/BackendProvider.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        $this->authenticator = $authenticator;
+        $this->renderer = $renderer;
+        $this->enforce2fa = $enforce2fa;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function beginAuthentication(AuthenticationContextInterface $context): bool
```
aschempp: The `AuthenticationContextInterface` also provides the firewall name and token. I think we should validate for `UsernamePasswordToken` and `backend` firewall (or match the user object). We only want 2FA for the actual login route, never for any other means of authentication, should the firewall be configured differently.

src/Security/TwoFactor/Authenticator.php
```diff
@@ -0,0 +1,101 @@
...
+     * @param User   $user
+     * @param string $code
+     *
+     * @return bool
+     */
+    public function validateCode(User $user, string $code): bool
+    {
+        $totp = TOTP::create($this->getUpperUnpaddedSecretForUser($user));
+
+        return $totp->verify($code, time());
```
aschempp: The `time()` argument is optional, default is `time()` 😉

src/Security/TwoFactor/BackendProvider.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getFormRenderer(): TwoFactorFormRendererInterface
+    {
+        return $this->renderer;
```
aschempp: Instead of creating a "useless" renderer class that does nothing else but redirect on a call, we could redirect on this method call directly and omit the whole `BackendFormRenderer` class. It's not happening anyway because the `FormController` is not used.

src/Resources/contao/dca/tl_user.php
```diff
@@ -985,4 +1023,154 @@ public function toggleVisibility($intId, $blnVisible, DataContainer $dc=null)
...
+		if (!$twoFactorAuthenticator->validateCode($user, $varValue))
+		{
+			// Disable 2FA, otherwise 2FA stays enabled if the user leaves the window
+			$user->use2fa = false;
+			$user->save();
+
+			throw new RuntimeException($GLOBALS['TL_LANG']['ERR']['invalidTwoFactor']);
+		}
+
+		return true;
```
aschempp: This method should return `1` and not a boolean to be stored in the database.

leofeyer:
> If I open the `contao/2fa` route manually, I should be redirected to `contao/login`. Currently get an Access Denied exception.

That is because you have logged in already and your role is `IS_AUTHENTICATED_2FA_IN_PROGRESS`. If you log out and try again, you will be redirected. 😄

src/Resources/contao/languages/en/tl_user.xlf
bytehead: AFAIK it was Google Authenticator, maybe @leofeyer changed?

bytehead: https://github.com/bytehead/core-bundle/commit/b3109397897e918c40aa48b57e9ec4fa15125a07#diff-fecc377bceddaab13c32ad1270a29c99L309

leofeyer: Indeed, because Google Authenticator is dead. 😄

aschempp: If Google Authenticator is dead, why do we lower our security standard to what it supports?

bytehead: It's not dead, that's just Leo's opinion 😉

aschempp:
> > If I open the contao/2fa route manually, I should be redirected to contao/login. Currently get an Access Denied exception.

> That is because you have logged in already and your role is IS_AUTHENTICATED_2FA_IN_PROGRESS. If you log out and try again, you will be redirected. 😄

I know that. But why am I redirected when accessing the Contao backend then?

bytehead:
> If I open the contao/login page manually while in 2fa process, the login screen appears but the debug bar say's I'm logged it (using TwoFactorToken). I should be redirected to 2fa

Fixed in 4e36d0003.

bytehead:
> Make DCA/DB fields consistent. tl_user.2faEnabled, tl_user.2faCode, tl_user.2faSecret, tl_user.2faConfirmed. (Not sure field starting with number are allowed, but you get the point)

I've renamed it according to the bundle (fields starting with numbers has some issues) in 59ec21f76.

bytehead:
> If I disable 2FA and enable it again, do I understand correctly that there is no new secret generated? I think it should?

I'm not sure about this? What do @contao/developers think?

src/Controller/BackendController.php
bytehead: It's required because of the two-factor bundle. See https://github.com/scheb/two-factor-bundle/pull/145, once it gets fixed we can remove this.

src/Controller/BackendController.php
bytehead: It's required because of the two-factor bundle. See scheb/two-factor-bundle#145, once it gets fixed we can remove this.

bytehead:
>>> If I open the contao/2fa route manually, I should be redirected to contao/login. Currently get an Access Denied exception.

>>That is because you have logged in already and your role is IS_AUTHENTICATED_2FA_IN_PROGRESS. If you log out and try again, you will be redirected. 😄

>I know that. But why am I redirected when accessing the Contao backend then?

We have to adjust the `access_control`:
```
access_control:
        - { path: ^/contao/two-factor, roles: [IS_AUTHENTICATED_2FA_IN_PROGRESS, ROLE_USER] }
```

bytehead:
> If 2fa is enforced, I am required to setup 2fa on the first login (🎉). However, the checkbox in my user/profile stays disabled.

Fixed in bd26db610.

bytehead:
> Only show the 2fa checkbox on the profile page, but not in the regular palettes. Or show them for all users, so I can potentially help someone create an access code for their backend. And I could disable 2FA for users that lost their authentication.

You can do exactly that if you're an admin (only disable, not enable, enable only globally via configuration).

src/Resources/contao/languages/en/tl_user.xlf
leofeyer: 🙈

aschempp:
With the change in https://github.com/contao/core-bundle/commit/4e36d0003a4f90dbd64adb8837d1c283328be970, so we actually still need a different route for the 2FA or can we just apply 2FA on the login route?

bytehead:
@aschempp see here: https://github.com/contao/core-bundle/pull/1545#discussion_r201745468

src/Security/TwoFactor/Provider.php
```diff
@@ -0,0 +1,100 @@
...
+
+        if (!$user instanceof User) {
+            return false;
+        }
+
+        if (!$user->secret) {
+            return false;
+        }
+
+        if (!$this->enforceTwoFactor && !$user->useTwoFactor) {
```
aschempp: So if the user does not have a secret (which is the default, if he never edited his data), then we don't enforce 2FA?

leofeyer: This seems correct to me. Otherwise the user would be asked to enter their verification code although they never had a chance to set up 2FA.

bytehead: In fact that's wrong. If there is no secret and 2FA is enforced, a secret should be generated, so the user gets the QR code to verify on the login screen.

src/Security/TwoFactor/Provider.php
```diff
@@ -57,11 +57,13 @@ public function beginAuthentication(AuthenticationContextInterface $context): bo
             return false;
         }
 
-        if (!$user->secret) {
-            return false;
+        // Generate a secret if 2FA is enforced and the user has not yet enabled it
+        if ($this->enforceTwoFactor && !$user->secret) {
+            $user->secret = random_bytes(128);
```
bytehead: Isn't here a `$user->save();` missing?

leofeyer: IDK. Can we be sure that `$context->getUser()` returns an object with a `save()` method? Also it seems to work without `save()`.

bytehead: Yes we can, it returns a `Contao\User` object.

leofeyer: Added in 9f3b5df0068aea47d3958bb96a82c703618619cf.

bytehead: 👍

src/Security/TwoFactor/Provider.php
bytehead: Fixed in eac180e.

src/Security/TwoFactor/Provider.php
```diff
@@ -0,0 +1,103 @@
...
+        if (!$user instanceof User) {
+            return false;
+        }
+
+        if (!$this->authenticator->validateCode($user, $authenticationCode)) {
+            return false;
+        }
+
+        // 2FA is now confirmed, save the user flag
+        if ($this->enforceTwoFactor && !$user->confirmedTwoFactor) {
```
leofeyer: Why do we only set this if 2FA is enforced?

bytehead: Because this only happens if 2FA is enforced by config. Otherwise it gets saved [here](https://github.com/bytehead/core-bundle/blob/bbdb15422b01d04b6dff4afe8dc6cf12795fa54d/src/Resources/contao/dca/tl_user.php#L1169-L1185).

leofeyer:
Thank you big time @bytehead!

BoozieBadd:
https://github.com/contao/core-bundle/pull/1545#

PR #1546, opened 2018-06-09
[RFC] Add filetree drag handle

### LINKED ISSUE #1403: Drag and drop file manager

ausi:
Followup to #1394, fixes #1403, related to #1456


src/Resources/contao/dca/tl_files.php
```diff
@@ -113,6 +113,20 @@
 				'href'                => 'act=source',
 				'icon'                => 'editor.svg',
 				'button_callback'     => array('tl_files', 'editSource')
+			),
+			'uploadButton' => array
```
leofeyer: If we renamed this to `move` we could re-use the existing translation, couldn't we?

ausi: No, the old upload button from DC_Folder uses `$GLOBALS['TL_LANG'][$this->strTable]['move'][0]` and `$GLOBALS['TL_LANG']['tl_files']['uploadFF']` which is “File upload” and “Upload files to folder "%s"”.

PR #1549, opened 2018-06-09
reduce memory consumption during sitemap generation

fritzmg:
See https://github.com/contao/news-bundle/pull/37


leofeyer:
Thanks a lot for debugging this @fritzmg. The fix might be a one-liner only, but finding the problem took several hours.

Issue #1389, opened 2018-02-20
Tasks to fix and improve caching


Toflar:
As pointed out in other issues already, the caching system still needs a lot of love. We are making good progress but there are still a lot of things to do to make Contao become the CMS with the best native caching support out there. We should not forget that our requirements are very complex and it will remain an ongoing topic that will accompany us for another few versions of Contao.
This issue should serve as a central place to keep track of all the TODO's around the HTTP cache.

### Bugfixes

* [x] For Contao versions 4.4 and 4.5, we must force `public` responses even though the session was started (see #1388). @leofeyer 

* [x] The Symfony `ResponseCacheStrategy` does not differentiate between `private` responses and `no-cache` cache headers. We must provide a PR to fix that. (https://github.com/symfony/symfony/pull/26532) @aschempp 

* [x] Our `InsertTagsController` must set `max-age` (client cache) settings based on page settings for BC insert tags (see #1390). @toflar 

* [x] `FrontendTemplate::addCacheHeadersToResponse()` must be fixed because it only sets `private`. The `no-cache` stuff is missing. @leofeyer 

### Features

* [x] We must ensure, we're not starting the session if we don't need it. Starting the session implicitly means that a response is `private` and may not be cached as it may contain personal information. As long as the core still requires stuff like this, we need to force `public` responses.

   * [x] `$_SESSION['FORM_DATA']` is a real killer here because we never unset the information stored in that value. We must introduce a config option in the form settings to disable that (or enable explicitly). It should be removed in version 5.

    * [x] `hasPreviousSession()` should be added everywhere we access the session to make sure the session is not started if there's no cookie (`Input` class) (https://github.com/contao/core-bundle/pull/1471) @toflar

*  [ ] We need insert tags as fragments so every insert tag can decide for its own, how it would like to be rendered (and set the correct caching headers) (https://github.com/contao/core-bundle/pull/1470) @toflar 

*  [x] Add support for tagging responses so that we can cache stuff "forever" and invalidate using cache tags. (https://github.com/contao/manager-bundle/pull/60) @toflar

leofeyer:
@contao/developers I need your feedback in order to fix the last open bugfix. These are the headers we used to send:

```php
if (!headers_sent())
{
	if ($intCache > 0 && (\Config::get('cacheMode') == 'both' || \Config::get('cacheMode') == 'browser'))
	{
		header('Cache-Control: private, max-age=' . ($intCache - time()));
[...]

PR #1550, opened 2018-06-09
[RTM] Auto-clear the session form data

leofeyer:
See #1389.

### DIFF
--- a/src/Resources/config/listener.yml
+++ b/src/Resources/config/listener.yml
@@ -44,7 +44,7 @@ services:
     contao.listener.clear_form_data:
         class: Contao\CoreBundle\EventListener\ClearFormDataListener
         tags:
-            # The priority must be lower than the one of the Symfony save session listener (defaults to -1000)
+            # The priority must be higher than the one of the Symfony save session listener (defaults to -1000)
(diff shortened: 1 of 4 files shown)


src/Resources/config/listener.yml
```diff
@@ -41,6 +41,12 @@ services:
             # and lower than the header replay header (defaults to 7)
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest, priority: 6 }
 
+    contao.listener.clear_form_data:
+        class: Contao\CoreBundle\EventListener\ClearFormDataListener
+        tags:
+            # The priority must be lower than the one of the Symfony save session listener (defaults to -1000)
```
aschempp: the priority must be **higher** than...

leofeyer: Fixed in 6847201f50a61d2fc69a8907d67074a7377d4396.

aschempp:
To make it all work, do we need to start the session when form data is submitted and actually store the data?

leofeyer:
Right now the session is started automatically thanks to our `LazySessionAccess` class. And since using `$_SESSION` is deprecated anyway, I would leave it this way.

PR #1553, opened 2018-06-11
[RFC] Enable package auto sorting for composer

bytehead:
Enables automatic sorting for packages in `composer.json` (see https://getcomposer.org/doc/06-config.md#sort-packages).


leofeyer:
In case you wonder what "composer normalization" is: https://github.com/localheinz/composer-normalize 😉

leofeyer:
Thank you @bytehead.

PR #1558, opened 2018-06-12
Pass search result data to the search template

dmolineus:
At the moment it's not possible to access enriched search result data in the `search_*` template. Such data could be added using the `indexPage` hook, f.e. a preview image or additional descriptions. 

This pull requests provides access to all data being stored in the `tl_search` dataset.


leofeyer:
Thank you @dmolineus.

Issue #1552, opened 2018-06-11
Subpalettes won't work on users profile page


bytehead:
I would like to add a subpalette to the users profile page (`/contao?do=login`) for #1545.
But the ajax request always fails with a `403` thrown on line 1808 here: https://github.com/contao/core-bundle/blob/186b649afa90e008d8414b6d64c153f530cbcee4/src/Resources/contao/drivers/DC_Table.php#L1788-L1809

`intId` has value `false` in this case.

The value gets set on line 413 here:
https://github.com/contao/core-bundle/blob/186b649afa90e008d8414b6d64c153f530cbcee4/src/Resources/contao/classes/Ajax.php#L384-L416

As already discussed with @Toflar, he will try to shed some light/investigate this.

Toflar:
@leofeyer you can assign that to me.

Toflar:
PR is ready https://github.com/contao/core-bundle/pull/1559

bytehead:
Thank you @Toflar! 🎉

PR #1559, opened 2018-06-12
Added support for subpalettes in personal profile

Toflar:
The problem is that `Backend` executes `$this->objAjax->executePostActions($dc);` before the `callback` of the registered module is called. Actually not even before but it doesn't call the callback at all if there are ajax requests. So basically no ajax operation can work which is why I removed the callback which was not necessary anyway and put it into a `onload_callback` which is what I would do in a custom module anyway.

This means anything that's possible in a regular DC is now possible in the back end user profile as well.

Fixes https://github.com/contao/core-bundle/issues/1552


src/Resources/contao/classes/Backend.php
```diff
@@ -302,11 +302,11 @@ protected function getBackendModule($module, PickerInterface $picker = null)
 		// Dynamically add the "personal data" module (see #4193)
 		if (\Input::get('do') == 'login')
 		{
-			$arrModule = array('tables'=>array('tl_user'), 'callback'=>'ModuleUser');
+			$arrModule = array('tables'=>array('tl_user'));
 		}
 
 		// Check whether the current user has access to the current module
-		elseif ($module != 'undo' && !$this->User->hasAccess($module, 'modules'))
+		if ($module != 'undo' && !$this->User->hasAccess($module, 'modules'))
```
aschempp: Strange that `undo` is here but `login` is now in `hasAccess`. But I guess that's up to another refactor.

Toflar:
Fixed in https://github.com/contao/core-bundle/pull/1559/commits/52cca135d4bab35a03d4918011fafb0842496874.

src/Resources/contao/classes/Backend.php
leofeyer: I also find this strange and would prefer to handle the login module check here instead of hardcoded in the `BackendUser::hasAccess()` method. @Toflar Do you think it is possible?

Toflar: Like so? fd47c7224868466fcc1db0f09cbe5834062ab752

Toflar: Hmm let me check if we can get that even nicer

Toflar: Well, it's not really possible without adding more keys to the `$GLOBALS['BE_MOD']`. In fact we could extend that array with an `doNotDisplay` and `disablePermissionCheck` key. That way everything would be supported, also for third party modules.

Toflar: Here we go, that way it's also supported for any third party module which is even better I find :) https://github.com/contao/core-bundle/pull/1559/commits/4531f7eaaf99d0e6e49bc83860c4b20ffa137dbf

Toflar:
Yeah I thought about that but I just couldn't come up with any valid use case why one would have ever done that. And even if we'd leave the class there, nothing would happen because the callback is not called anymore so...

bytehead:
Ok, fine for me then.

src/Resources/contao/classes/Backend.php
leofeyer: I agree. :)

leofeyer:
Thank you @Toflar.

PR #1562, opened 2018-06-13
[RFC] Quick draft of event helper

aschempp:
Just a quick idea I came up with to deal with the repetitive code when calling hooks and callbacks.

Example usage:

Call simple hook (https://github.com/contao/core-bundle/blob/fa05410fbd75cca7655599e59972475fe371ca0e/src/Resources/contao/pages/PageRegular.php#L77-L85)
```php
if (isset($GLOBALS['TL_HOOKS']['getPageLayout']) && \is_array($GLOBALS['TL_HOOKS']['getPageLayout']))
{
	$helper->triggerAll(
		$GLOBALS['TL_HOOKS']['getPageLayout'],
		[$objPage, $objLayout, $this]
	);
}
```

Call hook with return value (https://github.com/contao/core-bundle/blob/60403e47690ffcc253096750719a7d310f5f4561/src/Resources/contao/classes/FrontendTemplate.php#L56-L64)
```php
// HOOK: add custom parse filters
if (isset($GLOBALS['TL_HOOKS']['parseFrontendTemplate']) && \is_array($GLOBALS['TL_HOOKS']['parseFrontendTemplate']))
{
	$strBuffer = $helper->triggerAll(
		$GLOBALS['TL_HOOKS']['parseFrontendTemplate'],
		[$strBuffer, $this->strTemplate],
		0
	];
}
```

Call hook with return validation (https://github.com/contao/core-bundle/blob/5a411a7383e7ef24289bc6a739269ae572a52db1/src/Resources/contao/library/Contao/InsertTags.php#L985-L999)
```php
if (isset($GLOBALS['TL_HOOKS']['replaceInsertTags']) && \is_array($GLOBALS['TL_HOOKS']['replaceInsertTags']))
{
	$varValue = $helper->triggerAll(
		$GLOBALS['TL_HOOKS']['replaceInsertTags'],
		[$tag, $blnCache, $arrCache[$strTag], $flags, $tags, $arrCache, $_rit, $_cnt],
		null,
		function ($result) { return $varValue !== false; }
	);

	if ($varValue !== false)
	{
		$arrCache[$strTag] = $varValue;
		break 2;
	}
}
```

The same works for every DCA callback. We could also add a quick helper to call `isset` and `is_array` on hooks, but that won't work for callbacks.

### DIFF
--- a/src/DataContainer/EventHelper.php
+++ b/src/DataContainer/EventHelper.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Contao\CoreBundle\DataContainer;
+
+use Contao\CoreBundle\Framework\ContaoFrameworkInterface;
+use Contao\System;
+
+class EventHelper
+{
+    /**
+     * @var System
+     */
+    private $system;
+
+    /**
+     * Constructor.
+     *
+     * @param ContaoFrameworkInterface $framework
+     */
+    public function __construct(ContaoFrameworkInterface $framework)
+    {
+        $this->system = $framework->getAdapter(System::class);
+    }
+
+    /**
+     * Triggers a callback whilte handling singleton and service classes.
+     *
+     * @param array|callable $callback
+     * @param array          $arguments
+     *
+     * @return mixed
+     */
+    public function trigger($callback, array $arguments = [])
+    {
+        if (\is_array($callback)) {
+            return call_user_func_array(
+                [$this->system->importStatic($callback[0]), $callback[1]],
+                $arguments
+            );
+        }
+
+        if (\is_callable($callback)) {
+            return call_user_func_array($callback, $arguments);
+        }
+
+        throw new \InvalidArgumentException(sprintf('Cannot trigger "%s"', var_export($callback, true)));
+    }
+
+    /**
+     * Triggers an array of callback.
+     *
+     * @param array         $callbacks
+     * @param array         $arguments
+     * @param mixed|null    $returnArgument
+     * @param callable|null $returnIf
+     *
+     * @return mixed
+     */
+    public function triggerAll(array $callbacks, array $arguments = [], $returnArgument = null, callable $returnIf = null)
+    {
+        $result = null;
+
+        foreach ($callbacks as $callback) {
+            $result = $this->trigger($callback, $arguments);
+
+            if (null !== $returnIf && true === $returnIf($result)) {
+                return $result;
+            }
+
+            if (null !== $returnArgument) {
+                $arguments[$returnArgument] = $result;
+            }
+        }
+
+        return $result;
+    }
+}
--- a/src/Resources/config/services.yml
+++ b/src/Resources/config/services.yml
@@ -459,3 +459,6 @@ services:
             - "@contao.routing.scope_matcher"
         tags:
             - { name: twig.extension }
+
+    Contao\CoreBundle\DataContainer\EventHelper:
+        arguments: ['@contao.framework']


src/DataContainer/EventHelper.php
```diff
@@ -0,0 +1,77 @@
...
+     * Triggers an array of callback.
+     *
+     * @param array         $callbacks
+     * @param array         $arguments
+     * @param mixed|null    $returnArgument
+     * @param callable|null $returnIf
+     *
+     * @return mixed
+     */
+    public function triggerAll(array $callbacks, array $arguments = [], $returnArgument = null, callable $returnIf = null)
```
m-vo: I like this in general. The ``$returnArgument`` index looks a bit unintuitive to me however. Maybe use a callback instead to select the input for the next round or even merge with the ``$returnIf`` one?

leofeyer:
The code for "hook with return validation" is more complex than the original code, so I'm not sure about this.

src/DataContainer/EventHelper.php
aschempp: How would a callback work?

aschempp:
> The code for "hook with return validation" is more complex than the original code

Why is it more complex? The whole point is to get rid of the loop and `System::importStatic` in every implementation… This would also very much simplify things like a third option instead of array and callbacks (like triggering an event or whatever stupid thing we come up with 😁 )

src/DataContainer/EventHelper.php
m-vo: Something like this:
```PHP
$arguments = $selectorFunction($arguments, $result);
```

Which you could consume like so:
```PHP
triggerAll(
    [$callback1, $callback2, $callback3, $callback4]
    [$a, $b, $c],
    function($args, $result) { $args[0] = true; $args[1] = $result; return $args;  }
);
```

Or directly asking for arguments each run (then there could be an exception thrown to abort):
```PHP
$args = [ $enabled, $this, $that ];
triggerAll(
    [$callback1, $callback2, $callback3, $callback4]
    function($index, $previousResult) use $args { 
        if($index > 0) { $args[1] = $previousResult; }
        if('dangerous' === $previousResult) { throw new StopPropagationException(); }
        return $args;
    }
);
```
But that's not that elegant as well to be honest. It all depends on the primary usecase. If that's piping something through all callbacks, then your approach is probably the best. But maybe consider naming the parameter `returnArgumentIndex` or something to make clear that it relates to `arguments`?

m-vo:
... and it would allow making changes/optimizing at a central place. Right now the code segments have no common semantic.

src/DataContainer/EventHelper.php
aschempp: `$args` is an array, so it will not be passed by reference (variant 2 won't work). Variant 1 *might*, if you pass `$args` as reference (`&$args`). Haven't done that with a closure though.

m-vo: I'm not sure if we mean the same thing? https://3v4l.org/Z0Fmh

aschempp: The example does not qualify. Contrary to your code, the `triggerAll` method in the implementation does not live within the same class, so you can't `use` the same variable.

m-vo: The use statement makes a closure that binds & extends lifetime of the local variables. It's the same as putting the `$args` in my example inside the function and omitting the `use`.

aschempp: You can't `use` `$args` because it does not exist in the place where you create the closure 😉

m-vo: Sure, the closure gets created on the consumer/caller side where the args are available (and it would work without `use` if you put `$args` directly into the anonymous function). So I'm still not sure if we are talking about the same thing. But I think this discussion misses the point anyways. :grinning:

aschempp: Here's an example: https://3v4l.org/ZCNdN

1. passing `$args` by reference does not work in closures
2. `use`ing the original `$arguments` does not work by default
3. `use`ing the original `$arguments` by reference allows you to modify the original arguments, but that does not pass them on the multiple callbacks
4. The only thing that _does_ work is if your closure returns the arguments array that should be used for the successive call.

aschempp:
@leofeyer @Toflar @ausi would anyone agree this is a nice feature, or should we close the PR?

Toflar:
I don't see the use case, sorry. It's only for hooks and we don't want to extend any logic for hooks imho.

leofeyer:
I also vote to close the PR, because

> The code for "hook with return validation" is more complex than the original code

PR #1563, opened 2018-06-13
Correctly support singleton classes when calling hooks

aschempp:


leofeyer:
You also have to adjust the test, I'm afraid. 😉

aschempp:
Tests are fixed and I actually found another incorrect class. The tests were pretty wrong in my opinion…

leofeyer:
Thank you @aschempp.

PR #1568, opened 2018-06-16
[RTM] Add page picker for meta fields in file manager

ausi:
As requested in contao/core#7169

metadata-picker


aschempp:
Great idea, but are you sure we don't want to keep the existing design and have the icon after the field (by shortening the Link field)?

leofeyer:
I actually like the new design a lot more. We should adjust the other fields as well. 😄

ausi:
> are you sure we don't want to keep the existing design and have the icon after the field (by shortening the Link field)?

This way it was much easier to implement because of the HTML structure of the meta wizard. ☺️

leofeyer:
Changing it everywhere only works for text fields:





It does not work for drop down menus:



So now we can either

a) change the meta wizard only
b) change all text fields
c) leave everything as it is (and adjust the meta wizard).

@contao/developers What do you prefer?

Toflar:
> c) leave everything as it is (and adjust the meta wizard).

ausi:
c

aschempp:
c

leofeyer:
Thank you @ausi.

leofeyer:
I have tried to implement c) but it looks even worse:



I therefore leave this as is for the moment.

frontendschlampe:
I agree that it looks not perfect, but in this case the functionality is much more important, than the design. It's another point on to do list for backend redesign. ;-)

PR #1569, opened 2018-06-17
[RFC] Fix: PHP files do not work in app/Resources/contao/languages

mvscheidt:
See #740 

I've replaced the old implementation which had two foreach Loops (one for php, one for xliff) with Symfonys LoaderResolver and LoaderDelegator.

The Loader Resolver is configured to Resolve xliff files with the existing `Contao\CoreBundle\Config\Loader\XliffFileLoader` - for php files I created a new Loader Class `Contao\CoreBundle\Config\Loader\PhpFileIncluder` which basically wraps the `include $file` statement that used to be in the System Class.

I didn't use the existing `Contao\CoreBundle\Config\Loader\PhpFileLoader` because it works differently (it strips some stuff from php files and writes the result to the cache files)

Technically we could turn the supported endings/loaders into a yml based configuration, but I'd say that would be a new feature request if it would be needed.

----

As I've written [here](https://github.com/contao/core-bundle/issues/740#issuecomment-384211116) this fix only needs to apply if you did NOT warmup the cache.

Warming up the cache bypasses the problem for both entrypoints (`app.php` and `app_dev.php`).

Reason for this is that the warmup command does something similar like I did here already.

### DIFF
--- a/src/Config/Loader/PhpFileIncluder.php
+++ b/src/Config/Loader/PhpFileIncluder.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of Contao.
+ *
+ * (c) Leo Feyer
+ *
+ * @license LGPL-3.0-or-later
+ */
+
+namespace Contao\CoreBundle\Config\Loader;
+
+use Symfony\Component\Config\Loader\Loader;
+
+/**
+ * Includes the given php file, e.g. for old php/GLOBALS based translations
+ *
+ * @author Mike vom Scheidt <https://github.com/mvscheidt>
+ */
+class PhpFileIncluder extends Loader
+{
+    /**
+     * Includes a php file
+     *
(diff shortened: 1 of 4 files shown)


aschempp:
I generally like the idea, but I think we should try to reduce duplicate code. I don't know why the CacheWarmer works completely different. Also, are we sure this should be targeted at 4.4 (it being a bugfix rather a "known limitation")?

leofeyer:
This is definitely not a bug. The PR should be targeted against the `master` branch.

mvscheidt:
I may have time this friday - I will redo the pull request based on the `master` branch of https://github.com/contao/contao

I'll also take a look about the duplicate code issue aschempp mentioned at the same time.

leofeyer:
Closed in favor of contao/contao#110.

PR #1567, opened 2018-06-16
Fix drag and drop issues

ausi:
Fixes #1556

PR #1573, opened 2018-06-18
[RTM] Only show the clone once dragging has started

ausi:
Follow up to #1567


leofeyer:
Thank you for you patience @ausi.

PR #1576, opened 2018-06-20
[RFC] CS fixes

bytehead:
Fix CS introduced in #1559. See also https://github.com/contao/core-bundle/commit/52aa6a001c813a3485610cb95ccee87f15705f72#r29337990.


leofeyer:
Thank you @bytehead.

Issue #1565, opened 2018-06-14
UX improvement: Focus last edited field after saving


patrickjDE:
Inspired by #1541

When editing an item in the backend and saving, the page reloads. The scroll position is then restored, but the keyboard focus is not.

It should be possible with some javascript to also restore the focus to the last edited field. I think this would make for a nicer UX for people who save often. Any other thoughts on this?

aschempp:
I like the idea, can you provide some Javascript for this?

patrickjDE:
I'll have a closer look at it this weekend.

leofeyer:
See #1582.

PR #1582, opened 2018-06-21
[WIP] Focus last edited field after saving

patrickjDE:
Basic implementation of #1565.

Supports restoring the focus on default `input`, `select` & `textarea` elements, plus chosen-enhanced `select`s.

Chosen support looks a bit hackish, due to the version used by the backend not supporting the `chosen:ready` event.

ToDo:
- [ ] tinyMCE
- [ ] ACE


leofeyer:
> due to the version used by the backend not supporting the `chosen:ready` event.

Maybe we should update our version and implement the event?

@ausi @aschempp What about the general concept?

aschempp:
Please don't use cookies, we're trying to get rid of them as far as possible. Should use the `sessionStorage` and be all-client-side imho.

leofeyer:
We are not trying to get rid of cookies. We are trying to avoid using the session. Cookies are ok though, aren't they?

aschempp:
We are trying to minimize cookies because of Http Cache. Also, it does not make sense to use cookies for client-side-only implementations (and this is a perfect case of sessionStorage)

Toflar:
It would be a perfect use case for `localStorage`, not `sessionStorage`. I think it should keep working when I open a link in a new tab 😄 But I agree, it makes no sense to use cookies. Cookies are sent to the server and the server never needs to care about this. It's client side only.

aschempp:
> It would be a perfect use case for localStorage, not sessionStorage. I think it should keep working when I open a link in a new tab 

It's about saving a form, there is no new tab imho …

aschempp:
On the contrary, I think it **must** be `sessionStorage` to (potentially) work on multiple tabs.

Toflar:
Ah I see, it's only for "save" when you stay on the same page. `sessionStorage` would be fine then, yes. But then the implementation is very weird because we already have code that searches the first input field when you edit an entry. And none of this code in this PR seems to re-use any of that existing code?

patrickjDE:
Sorry for the late reply, but he feedback is much appreciated.

Regarding cookies: I'm perfectly fine with using localStorage, just tried to stick to the existing behaviour exhibited by BackendScrollOffset. (Might want to refactor that one? It's also Frontend only, afaics)

> But then the implementation is very weird because we already have code that searches the first input field when you edit an entry.
That's done with one single function which only handles `input`s.
> And none of this code in this PR seems to re-use any of that existing code?
I "borrowed" the disabled and visible check from that function, but as it is one single function, there's nothing reusable there.
But, I've added a check to make sure the two functions don't interfere with each other.

leofeyer:
@ausi /cc

aschempp:
Generally looks good to me, though the coding style is a bit "broken" 😆

ausi:
Wouldn’t it be much simpler to use an onsubmit event and store the `document.activeElement`?

Also, we should use `sessionStorage` here as noted in https://github.com/contao/core-bundle/pull/1582#issuecomment-402625233

patrickjDE:
> though the coding style is a bit "broken" 😆

Probably due to using spaces instead of tabs, cs should be fixed now.

> Wouldn’t it be much simpler to use an onsubmit event and store the document.activeElement?

Doesn't work, as activeElement is always the submit button, when saving.

> Also, we should use sessionStorage here as noted in #1582 (comment)

Agreed, done.

leofeyer:
@patrickjDE Are you still working on this PR?

PR #1583, opened 2018-06-22
[WIP] Phpstan

Toflar:
Here we go, let's bring Contao's code quality to yet another level by using [phpstan](https://github.com/phpstan/phpstan) for static analysis 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ install:
 script:
   - if [[ $COVERAGE ]]; then mkdir -p build/logs; fi
   - php vendor/bin/phpunit $COVERAGE
+  - php vendor/bin/phpstan analyse src tests --level=0
 
 after_script:
   - if [[ $COVERAGE ]]; then php vendor/bin/php-coveralls; fi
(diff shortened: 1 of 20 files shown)


tests/Fixtures/library/Model/Collection.php
```diff
@@ -6,7 +6,7 @@ class Collection implements \ArrayAccess, \Countable, \IteratorAggregate
 {
     private $models = [];
 
-    public function __construct(array $models)
+    public function __construct(array $models, $strTable)
```
leofeyer: I doubt that this is correct?

Toflar: It is. But that's also a dummy that should be removed completely.

.travis.yml
```diff
@@ -47,6 +47,7 @@ install:
 script:
   - if [[ $COVERAGE ]]; then mkdir -p build/logs; fi
   - php vendor/bin/phpunit $COVERAGE
+  - php vendor/bin/phpstan analyse src tests --level=0
```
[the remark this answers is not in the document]
Toflar: Jup but I wanted to be specific. Level `0` is not very helpful anyway, we'll want to raise it step by step :)

leofeyer:
Thank you @Toflar. Should we also add Phpstan to Contao 4.4 or at least backport the fixes found here?

Toflar:
Nah, I don't see any real advantage in backporting that now. Fixes are only related to tests so...no real value.

PR #1586, opened 2018-06-22
[RFC] Deprecate maxImageWidth

ausi:
Followup to #1585

`maxImageWidth` doesn’t work correctly if using an image size from `tl_image_size` and I think this configuration doesn’t make sense anymore since we have responsive images now.

For the use of it in the backend (see contao/core#7875) I moved to a `max-width` solution via CSS.


src/Resources/contao/dca/tl_settings.php
```diff
@@ -205,12 +205,6 @@
 				array('tl_settings', 'checkTemplateFiles')
 			)
 		),
-		'maxImageWidth' => array
```
leofeyer: If we remove the field, how are people supposed to reset it without having to adjust the `localconfig.php` file?

ausi: I think adjusting the `localconfig.php` should be OK. But we could also keep it in tl_settings and add “(deprecated)” to the Label.

leofeyer: @contao/developers What do you think?

Toflar: Removing is fine. People are quite used to editing the `localconfig.php` anyway as there are other settings that have never been configurable in the UI. I'm okay with that change.

PR #1587, opened 2018-06-25
Invoke callbacks on label/thumbnail creation for files and folders in DC_Folder

markotodic:
Invoke defined callbacks on folder label creation
Move file label and thumbnail creation in separate method
Invoke defined callbacks on file label/thumbnail creation

### DIFF
--- a/src/Resources/contao/drivers/DC_Folder.php
+++ b/src/Resources/contao/drivers/DC_Folder.php
@@ -2569,6 +2569,7 @@ class DC_Folder extends DataContainer implements \listable, \editable
 			$md5 = substr(md5($folders[$f]), 0, 8);
 			$content = scan($folders[$f]);
 			$currentFolder = \StringUtil::stripRootDir($folders[$f]);
+			$objFolder = \FilesModel::findByPath($currentFolder);
 			$session['filetree'][$md5] = is_numeric($session['filetree'][$md5]) ? $session['filetree'][$md5] : 0;
 			$currentEncoded = $this->urlEncode($currentFolder);
 			$countFiles = \count($content);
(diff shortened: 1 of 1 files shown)


aschempp:
Looks like this somewhat got lost in the _old_ repository.
@markotodic can you explain the use case?

markusmilkereit:
@aschempp 
This request works nicely in tandem with #1588 to give Contao the only thing I like about TYPO3, a per file usage overview and counter :)

The main purpose is to keep an eye on how often a file is used in the website (as content, download etc.), which includes the info if it is used at all. On bigger, more complex websites, running a couple of years you usually collect a couple of files, that are no longer used. Outdated downloads, replaced images etc - but it's really hard to delete them without making sure that they're not missing in the content afterwards.

Secondary use case is for organizing a relaunch - we sometimes copy all the old files (images, downloads) into the system as placeholders, which then need to be replaced with the updated (usually: bigger) images, and in many cases updated downloads, for example because the Corporate Design has been changed. The goal of course is for none of the old files to be used at the end - they should all have been replaced (usage count 0). If you can see in the file manager how often a file is used, that's a very handy information, much easier than looking at the resulting pages.

Third use case: sometimes we need to find out where a single image is used - and if it is used at all. For example if a certain employee has left the company and only signed a partial release form (only as long as he is in the company his image can be used). If the files itself are nicely organized, that's the best / easiest place to clean up. Especially like in bigger companies and universities, where teams of people are doing the content management, it's really hard to "just know" where certain images might be used. Hence: info in the file manager would be quite handy.

m-vo:
Images/files can also be used by InsertTags, extensions, external tools or e.g. programmatically like in your build process... So there is no way the system could know a thing is still used (other than maybe delegating everything via a route + tracking - but that's not really an option and wouldn't work for new, unseen, content).

leofeyer:
We have discussed this in Mumble on June 6th. Isn't your `create_folder_label_callback` the same as the `group_callback` in the DC_Table and the `create_file_label_thumbnail_callback` the same as the `label_callback` in the DC_Table?

markusmilkereit:
More or less - there are some differences: in DC_Table almost everything is more dynamically created, in DC_Folder there is a lot more "static" code. 

The `group_callback` and `create_folder_label_callback` are similar, at a very similar place in the logic - we just added the "array of callbacks" logic, instead of having a single callback.

For the `create_file_label_thumbnail_callback` which relates to the `label_callback` we moved the whole label creating code into a separate method to help with readability. The label callback could alternatively just be included at the end of the original method.

We could also rewrite names and implementation to be closer to the original, that would be a small step :)

Toflar:
If possible, I would like to have the exact names and same parameters (as far as possible) 😄

markusmilkereit:
We can do that - should be a small change. Will come in soon, today or monday - I will be really happy if this gets approved, so we can convince even more clients to switch away from TYPO3 :)

markusmilkereit:
We rewrote the code, but still have to test properly in one of the projects we already use it. 

We also came across the main reason for renaming the callbacks: the variables used are very different from DC_Table, mainly because the whole label-building is done in a different way. The new version will have the same position, the same callback name, just one level of callbacks (all like DC_Table), but a pretty different list of parameters.

markusmilkereit:
@leofeyer I don't see a connection with the commit and our request - we're asking for Hooks in DC_Folder, the commit is improving the Results handling. We're still very much interested in having a more dynamic DC_Folder handling. The extension is ready, we just don't have the necessary connection points in the core :)

leofeyer:
Since contao/core-bundle has become read-only, you would have to recreate the PR in contao/contao. We cannot automatically transfer pull requests, I'm afraid. 🙈

PR #1588, opened 2018-06-25
Backend popup hook to add custom rows

markotodic:
Enable client logic to populate "additional" rows inside backend popup

### DIFF
--- a/src/Resources/contao/controllers/BackendPopup.php
+++ b/src/Resources/contao/controllers/BackendPopup.php
@@ -166,6 +166,42 @@ class BackendPopup extends Backend
 		$objTemplate->charset = \Config::get('characterSet');
 		$objTemplate->labels = (object) $GLOBALS['TL_LANG']['MSC'];
 		$objTemplate->download = \StringUtil::specialchars($GLOBALS['TL_LANG']['MSC']['fileDownload']);
+		$objTemplate->arrAdditionalRows = [];
+
+        /**
+         * Call user defined callbacks to insert additional rows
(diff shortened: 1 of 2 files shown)


Toflar:
Hey, thanks for contributing! Why would you need that if I may ask?

markotodic:
Hello @Toflar ,
Currently I'm writing contao bundle that counts "usage" of each file, but also on which pages they are used (article, news...).
Need entrance there for popup so my bundle can display links to corresponding places, etc.
Thanks for the reply!

aschempp:
Looks like this somewhat got lost in the _old_ repository.
I'm not really sure about this feature, it looks very specific to me. In every case I ever needed to adjust the popup, I just added a custom popup/button to my rows. Wouldn't that work as well @markotodic ?

markusmilkereit:
The main reason for the information beeing in the "info" popup is because we have a couple of clients who like that exact view in TYPO3 - the info popup contains all the important infos about the file. Example attached as screenshot. 

For me having that compact overview in the popup and the information included in pull request #1587 makes Contao complete. The file usage overview / usage count is the only element clients are really missing when switching over from TYPO3.

Without that competing CMS comparison, a new / standalone icon would also work.
file-popup-typo3

aschempp:
If you want to implement a custom popup view, that is certainly possible. Simply use `key=info` instead of `act=info` on the operation, and implement your own view/controller that shows all the details you want. Wouldn't that be much more "efficient" than adding *additional rows* that could mean nothing? I mean you still can't add an image on top with *additional rows* etc. right?

PR #1590, opened 2018-06-26
Fixed DoctrineMigrationsPass not working with latest Symfony 3.4+

Toflar:
The changes in https://github.com/symfony/symfony/pull/27272 broke our support for doctrine migrations support.

### DIFF
--- a/tests/ContaoCoreBundleTest.php
+++ b/tests/ContaoCoreBundleTest.php
@@ -60,10 +60,16 @@ class ContaoCoreBundleTest extends TestCase
         $bundle->build($container);
 
         $classes = [];
+        $beforeRemovingClasses = [];
 
         foreach ($container->getCompilerPassConfig()->getPasses() as $pass) {
             $reflection = new \ReflectionClass($pass);
             $classes[] = $reflection->getName();
+        }
+
+        foreach ($container->getCompilerPassConfig()->getBeforeRemovingPasses() as $pass) {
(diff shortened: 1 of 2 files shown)


tests/ContaoCoreBundleTest.php
```diff
@@ -60,16 +60,23 @@ public function testAddsTheCompilerPaths()
...
+        foreach ($container->getCompilerPassConfig()->getBeforeRemovingPasses() as $pass) {
+            $reflection = new \ReflectionClass($pass);
+            $beforeRemovingClasses[] = $reflection->getName();
+        }
+
         $this->assertContains(AddPackagesPass::class, $classes);
         $this->assertContains(AddSessionBagsPass::class, $classes);
         $this->assertContains(AddResourcesPathsPass::class, $classes);
         $this->assertContains(AddImagineClassPass::class, $classes);
         $this->assertContains(DoctrineMigrationsPass::class, $classes);
```
leofeyer: Shouldn't we delete this line?

Toflar: It should be in both arrays 😄

leofeyer: We probably should change `getPasses()` to `getBeforeOptimizationPasses()` to be more explicit then.

leofeyer:
Thank you @Toflar.

PR #1597, opened 2018-07-03
Added extra values to composer.json section in the installation documentation

sheeep:
In order to let the `ScriptHandler` guess where the `console` binary is found, we need to add at least the following two extra values.

```
"symfony-bin-dir": "bin",
"symfony-var-dir": "var"
```

Otherwise it will default to `app/console` and any composer process which involves `run-script` will return an error:

```
> post-install-cmd: Contao\CoreBundle\Composer\ScriptHandler::addDirectories
Could not open input file: app/console
Script Contao\CoreBundle\Composer\ScriptHandler::addDirectories handling the post-install-cmd event terminated with an exception
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,20 @@ Edit your `composer.json` file and add the following:
 "config": {
     "component-dir": "assets"
 },
-"post-install-cmd": [
-    "Contao\\CoreBundle\\Composer\\ScriptHandler::addDirectories",
-    "Contao\\CoreBundle\\Composer\\ScriptHandler::generateSymlinks"
-],
-"post-update-cmd": [
-    "Contao\\CoreBundle\\Composer\\ScriptHandler::addDirectories",
-    "Contao\\CoreBundle\\Composer\\ScriptHandler::generateSymlinks"
-]
+"scripts": {
+    "post-install-cmd": [
+        "Contao\\CoreBundle\\Composer\\ScriptHandler::addDirectories",
+        "Contao\\CoreBundle\\Composer\\ScriptHandler::generateSymlinks"
+    ],
+    "post-update-cmd": [
+        "Contao\\CoreBundle\\Composer\\ScriptHandler::addDirectories",
+        "Contao\\CoreBundle\\Composer\\ScriptHandler::generateSymlinks"
+    ],
+}
+"extra": {
+    "symfony-bin-dir": "bin",
+    "symfony-var-dir": "var",
+}
 ```
 
 Then run `php composer.phar update` to install the vendor files.


aschempp:
Why would this be necessary for a bundle? The config is only used from the root package, so this setting would need to go into the managed edition?

sheeep:
> Why would this be necessary for a bundle?

Because `Contao\CoreBundle\Composer\ScriptHandler` uses those values to determine the console binary directory. The defaults only work for a Symfony 3 directory structure. Better be verbose but correct, I guess.

```php
    /**
     * Returns the bin directory.
     *
     * @param Event $event
     *
     * @return string
     */
    private static function getBinDir(Event $event): string
    {
        $extra = $event->getComposer()->getPackage()->getExtra();

        // Symfony assumes the new directory structure if symfony-var-dir is set
        if (isset($extra['symfony-var-dir']) && is_dir($extra['symfony-var-dir'])) {
            return $extra['symfony-bin-dir'] ?? 'bin';
        }

        return $extra['symfony-app-dir'] ?? 'app';
    }
```

m-vo:
But this should still be changed for the managed edition, shouldn't it?

aschempp:
> Or maybe the title of this PR is a bit misleading. I guess you thought I changed the composer.json of this bundle, which I didn't. I changed README.md, and added those entries to the composer.json section of the installation process documentation. Sorry for that.

ah yes, this confused me.

> But this should still be changed for the managed edition, shouldn't it?

Very likely, yes

aschempp:
on the other hand, I don't think this is correct. The core-bundle readme has nothing to do with how you set up your distribution. You can place the files anywhere you like if you don't use the managed edition, don't you?

leofeyer:
> The core-bundle readme has nothing to do with how you set up your distribution

Correct. The core bundle README describes how to add the core bundle to a Symfony app.

aschempp:
I just discussed this with @sheeep. The problem seems to be that the `ScriptHandler` does not correctly detect a Symfony 4 installation and use it's default directory structure. So this needs to be updated for Contao 4.6

Also, the README describes a lot of things in relation to Symfony 3, which is probably incorrect as well (e.g. parameters instead of env variables).

aschempp:
If I remember correctly, our original code was inspired by https://github.com/sensiolabs/SensioDistributionBundle/blob/master/Composer/ScriptHandler.php

leofeyer:
As discussed in Mumble on July 5th, we want to adjust our script handler according to Symfony's script handler, which can handle both the Symfony 3 and Symfony 4 file structure without configuration.

leofeyer:
Our conclusion was wrong unfortunately. Our script handler already supports the new file structure if `symfony-bin-dir` and `symfony-var-dir` are set in the `composer.json` file. So the only thing we might be able to change is to use the Symfony 3 structure by default (instead of the Symfony 2 structure, which is the current default if `symfony-var-dir` is not set).

aschempp:
So for every Symfony installation I **have to** set `symfony-var-dir` if I want to support the new file structure? Why is this not the case in https://github.com/symfony/skeleton/blob/master/composer.json then?

leofeyer:
Because that's Symfony 4 and they are no longer using the distribution bundle.

leofeyer:
@sheeep Since you have overworked the `README.md` file a while ago and the current `README.md` is correct, I think this can be closed.

PR #1601, opened 2018-07-05
[RTM] Improve file/folder API for public/protected resources

m-vo:
see  https://github.com/contao/core-bundle/pull/1594#issuecomment-402662622

This PR attempts the following for files and folders:
* invert API logic: protected <-> public to mimic the actual behaviour in the FS*
* add endpoints to check if a certain file/folder is public

*) IMO this is neccessary to achieve a consistent API because the *public* property is inherited and protecting a subfolder in a public folder isn't supported.

I'd introduce a small behaviour change for removing public access so that an exception gets thrown if someone attempts to remove public access for a folder that is only implicitly public (by a `.public` file inside a parent folder) to avoid an inconsisten API. See https://github.com/contao/core-bundle/pull/1594#issuecomment-402672064 for an example.

Todo:
- [x] Agree on the general concept
- [x] Rewrite the few usages (4) of the `protect()` and `unprotect()` functions inside `DC_Folder` and `tl_files` to use the new API functions
- [x] Add `isPublic()` function for `File` as well that uses `Folder::isPublic()`
- [x] Consume `Folder->isPublic()` instead of separate implementation in `DC_Folder`
- [x] CS


src/Resources/contao/library/Contao/Folder.php
```diff
@@ -319,27 +319,93 @@ public function copyTo($strNewName)
...
+            return;
+        }
+
+        if (!file_exists(TL_ROOT . '/' . $this->strFolder . '/.public')) {
+            throw new \RuntimeException(
+                sprintf('Can\'t remove public access for implicitly public folder "%s".', $this->strFolder)
+            );
+        }
+
+        $objFile = new \File($this->strFolder . '/.public');
```
Toflar: Should not use global namespace but Contao namespace instead.

Toflar:
What about `allowPublicAccess()` and `denyPublicAccess()`?

ausi:
As discussed on Mumble: we should also throw an exception if `protect()` is called on a public subfolder. IMO this is not a BC break because `protect()` cannot make a subfolder of a public folder protected, so this was a bug since Contao 4.0.

Maybe we should also backport the exception for Contao 4.4?

Toflar:
Looks good to me (apart from some coding style issues), I think the other todos can be tackled with that concept 👍

m-vo:
> Maybe we should also backport the exception for Contao 4.4?

Probably, if we consider this to be a bug. @leofeyer What do you think?

m-vo:
**Notes:**

I've deprecated `DC_Folder->protect()` instead of rewriting it because I could not find any usage of this function. Can anyone shed some light on wether this is the right way?

I also added a function `getParent()` to `Folder` to make working with `isPublic()` less cumbersome in case you want to know if *only* the parent is public. Right now it wouldn't be possible to differentiate with only using the API because `isPublic(includeParents: false)` and `isPublic(includeParents: true)` would return true in both cases (parent public, folder public / parent not public / folder public). This functionality is acutally used in `tl_files`*.

*) Sidenote: I noticed a small inconsistency between API/FS and UI: It's possible to allow public access to a folder A and it's subfolder B by adding a `.public` file or via the API by calling `allowPublicAccess()` and I think this is the right behaviour (it allows to make a subfolder public first and then deny access for the parent for example). The backend UI howecer disallows making a folder public if a parent folder is already public (disabled checkbox). Personally I think this is a legit way to keep things simple but wanted to point this out.

leofeyer:
I'm fine with throwing the exception in the `protect()` method when applied to a public subfolder, however I don't like any of the rest. We have agreed on not adding new features to the Contao 3 framework unless absolutely necessary – and neither inverting the logic nor renaming the `protect()` and `unprotect()` methods is necessary IMHO.

m-vo:
I do not agree - in my eyes the current API is broken. And that's the reason for this PR:

Problem 1: You cannot protect a folder (this won't work with the way files are symlinked right now). But the API suggest this can be done. Also it's totally inconsistent with the filesystem as there are `.public` files and not `.protected` files.

Problem 2: the API is missing the essential function to check if a folder is accessible. There is no way to get this information as a consumer. So every bit of code using the Folder API would need to know about the mechanism itself (e.g. checking for `.public` files). That's very wrong IMHO. I'd rather see this added to `Folder` (where it belongs logically) instead of duplicating FS code elsewhere.

leofeyer:
> I’m not sure about `File::isPublic()` and `Folder::getParent()`, but the rest should be added IMO.

What would be the rest then?

> Alternative names for allow/denyPublicAccess we discussed after the last Mumble call are publish/unpublish (if you like them better @leofeyer ?)

I do not see a reason for renaming methods at all. The old Contao 3 framework will most likely not make it into Contao 5, therefore there is no point in renaming methods and triggering deprecation notices saying the new methods will work in Contao 5. Because chances are they will not!

m-vo:
I added `File::isPublic()` as a convenient method. That's not needed but eases working with the attribute IMHO. But `Folder::isPublic()` is somewhat essential - or what would be an alternative?

I could as well have added `Folder::isProtected` without changing the rest. But than we'd have yet another method with the wrong intuition.

leofeyer:
> But than we'd have yet another method with the wrong intuition.

Who cares? It is the **legacy** Contao 3 framework!

m-vo:
> Who cares? It is the legacy Contao 3 framework!

The ones that need to interact with the API? It doesn't really help that it's still the old framework as long as there is no replacement.

What what do you propose instead? 

1. adding `isProtected()` instead
2. adding `isPublic()` but keeping `protect()` and `unprotect()`
3. closing the PR and ignoring the need to know if a folder is public/protected
 (= no abstraction, no shared code)
4. drop the File/Folder API completely and replace it with a new concept that then (besides other things) will be capable to tell if a folder is public
5. anything else

src/Resources/contao/library/Contao/Folder.php
```diff
@@ -319,27 +319,88 @@ public function copyTo($strNewName)
...
+            $path = \dirname($path);
+        } while ($includeParents && '.' !== $path);
+
+        return false;
+    }
+
+	/**
+	 * Allow public access by adding a .public file.
+	 */
+    public function allowPublicAccess(): void
```
leofeyer: This method should be named `unprotect` (no unnecessary renaming).

src/Resources/contao/library/Contao/Folder.php
```diff
@@ -319,27 +319,88 @@ public function copyTo($strNewName)
...
+
+        File::putContent($this->strFolder.'/.public', '');
+    }
+
+	/**
+     * Deny public access by deleting the .public file.
+     *
+     * @throws \Exception if folder is only implicitly public by a public parent
+     */
+    public function denyPublicAccess(): void
```
leofeyer: This method should be named `protect` (no unnecessary renaming).

src/Resources/contao/library/Contao/Folder.php
```diff
@@ -319,27 +319,88 @@ public function copyTo($strNewName)
...
 	}
 
+	/**
+	 * Check if the folder (or any parent) is public by checking if the .public file is set.
+	 *
+	 * @param bool $includeParents
+	 *
+	 * @return bool
+	 */
+    public function isPublic(bool $includeParents = true): bool
```
leofeyer: This method should be named `isUnprotected` (as discussed in Mumble on October 18th).

m-vo:
@leofeyer I changed the API names as discussed on the last call.

I also exchanged the words "public" with "unprotected" in the descriptions to be consistent - only place where this is a bit odd is maybe with `isUnprotected` where it's now *"Check if the folder (or any parent) is **unprotected** by checking if the **.public** file is set"*. But well...

leofeyer:
IMHO the help text for the user does not have to match the method names, so we should just keep _public_ there.

leofeyer:
Merged in contao/contao@578044764b17db7cbec78c94d0beae589dc6dbd6. Thank you @m-vo.

PR #1602, opened 2018-07-06
Remove support for deprecated user password hashes

leofeyer:
This PR implements contao/core#8889 for Contao 4.6.


README.md
```diff
@@ -109,9 +109,6 @@ security:
         default:
```
aschempp: do we even need to define the default encoder? I think Symfony uses bcrypt if nothing is defined automatically.

leofeyer: I'm not sure.

src/Resources/contao/library/Contao/User.php
```diff
@@ -502,19 +496,8 @@ public static function loadUserByUsername($username)
...
-			else
-			{
-				$blnAuthenticated = password_verify($request->request->get('password'), $user->password);
-				$blnNeedsRehash = password_needs_rehash($user->password, PASSWORD_DEFAULT);
-			}
+			$blnAuthenticated = password_verify($request->request->get('password'), $user->password);
+			$blnNeedsRehash = password_needs_rehash($user->password, PASSWORD_DEFAULT);
 
 			// Re-hash the password if the algorithm has changed
 			if ($blnAuthenticated && $blnNeedsRehash)
```
aschempp: Only partially related to this PR, but why does the User object rehash the password? That seems very wrong to me… should be a security component…

README.md
Toflar: `->scalarNode('algorithm')->cannotBeEmpty()->end()`

src/Resources/contao/library/Contao/User.php
aschempp: It's still wrong in this place, just because we can use the request…

src/Resources/contao/library/Contao/User.php
```diff
@@ -588,22 +571,7 @@ public function setSalt($salt)
 	 */
 	public function getEncoderName()
```
aschempp: We should remove this method and the `EncoderAwareInterface`

README.md
aschempp: I digged some code. `default` here only works because our user object implements `EncoderAwareInterface` and returns this value. The _correct_ approach would be to define the encoder by class name, so it should look like this:

```
security:
    encoders:
        Contao\User:
            algorithm: bcrypt
```

leofeyer: Changed in 01bb96ee64808027bb6bc15d5cdc02911d5f9d8b.

src/Resources/contao/library/Contao/User.php
leofeyer: Removed in 01bb96ee64808027bb6bc15d5cdc02911d5f9d8b.

src/Resources/contao/library/Contao/User.php
aschempp: FYI: https://github.com/symfony/symfony/pull/26050

PR #1605, opened 2018-07-09
[RFC] change lock file directory

fritzmg:
Due to the [ongoing problems](https://github.com/contao/core-bundle/issues/1551) with the `AbstractLockedCommand` on Strato, I propose at least the following change: use a `/var/locks` directory within the project instead of the `sys_get_temp_dir()` location.

Since we need a read and write file handle for file locks on Solaris, the file permissions of the lock files need to be changed from `0444` to `0666` (see https://github.com/symfony/symfony/pull/27903). Any lock files generated prior to this change will still cause problems. Thus, if the lock files are at least within the user's reach (i.e. in the project directory) such errors would be easier to handle.

Also there is a small bug in PHP that causes the `sys_get_temp_dir()` function to return the temp path with a trailing slash on Solaris. So all the lock file paths will contain a double slash. But this is just a cosmetic discrepancy and doesn't affect the functionality.

### DIFF
--- a/src/Command/AbstractLockedCommand.php
+++ b/src/Command/AbstractLockedCommand.php
@@ -29,7 +29,7 @@ abstract class AbstractLockedCommand extends ContainerAwareCommand
     {
         $lock = new LockHandler(
             $this->getName(),
-            sys_get_temp_dir().'/'.md5($this->getContainer()->getParameter('kernel.project_dir'))
+            $this->getContainer()->getParameter('kernel.project_dir').'/var/locks'
         );
 
         if (!$lock->lock()) {


leofeyer:
I like the idea. @contao/developers What about you?

Toflar:
I don't like this. I would rather have a general `tmp` Directory to store stuff into. So a replacement for `sys_get_temp_dir()`. It's not related to `locks` at all imho 😄

fritzmg:
One reason not to do it would be because then the users would be able to shoot himself in the foot by deleting one or all lock files while they are needed ;)

Toflar:
Related: https://github.com/symfony/symfony/issues/23354

ausi:
I’m not sure if I should like this.

If I understood the linked issue about Strato correctly, it was about opening the file handle with write permissions and setting the correct file rights. Wouldn’t these issues have been the same in the `/var/locks` directory?

As @Toflar already wrote in https://github.com/contao/core-bundle/issues/1551#issuecomment-396559056 other dependencies might rely on `sys_get_temp_dir()` too and a `/var/locks` directory wouldn’t fix that.

IMO if `sys_get_temp_dir()` doesn’t work on a specific hosting environment, we should try to fix that in the manager if possible.

> Any lock files generated prior to this change will still cause problems.

This could also be fixed by changing the hash to `md5("v2\0".$this->getContainer()->getParameter('kernel.project_dir'))` I think.

leofeyer:
Given that symfony/symfony#27903 has been merged, this ticket can be closed.

Issue #1273, opened 2017-12-29
Require fully authenticated to change passwords


leofeyer:
If a member is authenticated based on a remember me cookie, they should be required to re-authenticate before changing their password (and maybe more?).

aschempp:
We simply need to throw an `AccessDeniedException` which will be handled by the firewall listener (see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Http/Firewall/ExceptionListener.php#L120)

leofeyer:
We should check the `IS_FULLY_AUTHENTICATED` role before throwing the exception. 😉

leofeyer:
> We simply need to throw an `AccessDeniedException` which will be handled by the firewall listener

Unfortunately this does not work. I'm getting stuck here:

https://github.com/contao/core-bundle/blob/5e8f65ba154b6fef6b0d2b1701ff980b2c855d1c/src/Controller/FrontendController.php#L100-L102

leofeyer:
Never mind, I forgot to create a 401 page in the site structure.

leofeyer:
@contao/developers Of course the login form shows "You are logged in as …" in this case, because the member is logged in (just not fully authenticated). Do we want to log out the member or do we just show the login form? The latter implies that the member can navigate to another menu item and keep being logged in (based on the remember me cookie).

leofeyer:
See #1610.

PR #1610, opened 2018-07-10
[WIP] Require fully authenticated to change passwords

leofeyer:
This PR implements #1273.

- [ ] Do we need the check in the "close account" module?
- [ ] Do we really want to show the login module where the member could log in as a different user?

### DIFF
--- a/src/Resources/contao/modules/ModuleChangePassword.php
+++ b/src/Resources/contao/modules/ModuleChangePassword.php
@@ -11,6 +11,7 @@
 namespace Contao;
 
 use Patchwork\Utf8;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 
 /**
  * Front end module "change password".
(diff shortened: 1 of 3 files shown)


src/Resources/contao/modules/ModuleChangePassword.php
```diff
@@ -53,6 +54,12 @@ public function generate()
 			return '';
 		}
 
+		// Require the member to be fully authenticated (see #1273)
+		if (!\System::getContainer()->get('security.authorization_checker')->isGranted('IS_AUTHENTICATED_FULLY'))
+		{
+			throw new AccessDeniedException('Please authenticate');
```
aschempp: As this message is not shown to the user, I think it should rather be something like `Full authentication is required to change the password`.

src/Resources/contao/modules/ModuleCloseAccount.php
```diff
@@ -53,6 +54,12 @@ public function generate()
 			return '';
 		}
 
+		// Require the member to be fully authenticated (see #1273)
+		if (!\System::getContainer()->get('security.authorization_checker')->isGranted('IS_AUTHENTICATED_FULLY'))
+		{
+			throw new AccessDeniedException('Please authenticate');
```
aschempp: Same here

leofeyer:
> Do we need to be fully authenticated for the password change module? It requires to enter the current password anyway.

That's a very good point. And since it also requires to enter the password to close an account, we might not need this check at all.

ausi:
How does this work with two factor authentication?

aschempp:
> How does this work with two factor authentication?

We don't have two-factor authentication in the front end (yet)

PR #1611, opened 2018-07-10
Add longer cache durations

Toflar:
Now that Contao 4.6 supports cache tagging and invalidation by cache tags, we can increase the caching settings. This is especially true for the reverse proxy cache. Not sure if we also want to allow it for the `clientCache`. I haven't added this because I think a client should probably get a new response every month.

This is an intermediary step. The ultimate goal for Contao would be to disable the settings for the server cache entirely and always cache forever (aka 1 year) on the reverse proxy. Developers should actively tag responses and invalidate the cache entries when something happens in the back end so that a user doesn't need to care about that anymore but always gets maximum performance 😄


leofeyer:
Thank you @Toflar.

PR #1612, opened 2018-07-10
Cache tags for new abstract controllers

Toflar:
Added support for automatically tagging the elements created using the new abstract controllers. That was still missing 😄

### DIFF
--- a/src/Controller/AbstractFragmentController.php
+++ b/src/Controller/AbstractFragmentController.php
@@ -99,6 +99,20 @@ abstract class AbstractFragmentController extends Controller implements Fragment
     }
 
     /**
+     * Tags the current response with given cache tags.
+     *
+     * @param array $tags
+     */
+    protected function tagResponse(array $tags): void
+    {
+        if (!$this->has('fos_http_cache.http.symfony_response_tagger')) {
+            return;
+        }
+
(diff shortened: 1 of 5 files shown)


Toflar:
Ping for review @aschempp

aschempp:
Should we tag for the current page ID (and possibly article) as well, as a change to them might affect the response? Or is this done in another place?
I assume the `tagResponse` method can be called multiple times? I don't really like the name though, as it's not obvious we're tagging the _final_ response and not the one generated by this controller…

Toflar:
> Should we tag for the current page ID (and possibly article) as well, as a change to them might affect the response? Or is this done in another place?

This should be done in the respective places. The idea of tags is that you don't have to worry about in what context it is used. If someone else adds another tag to your response then that's just fine. So nothing to do here.

> I assume the tagResponse method can be called multiple times?

Yes, that's why it's not called `setTags()` or similar.

> I don't really like the name though, as it's not obvious we're tagging the final response and not the one generated by this controller…

You're not. Tags are added to all master requests (so ESI as well) during the `kernel.response` event. So not necessarily the "final response".

leofeyer:
Thank you @Toflar.

Issue #616, opened 2016-11-06
AddImagineClassPath must not thrown an exception


leofeyer:
If the CLI PHP version does not have an image extension, `composer update` will fail, because we are throwing an exception in the `AddImagineClassPath` class. A PHP image extension is not required to run `composer update` (although it is to e.g. run the tests), therefore we should probably not throw the exception?

@contao/developers @DanielSchwiperich /cc

leofeyer:
Also, the Imagine classes seem to have their own sanity checks throwing an `Imagine\Exception\RuntimeException`, therefore I think we do not need our exception.

aschempp:
I don't get the issue. How is Imagine related to Composer?

DanielSchwiperich:
via compiler pass the image class gets set if I'm not wrong. 
And this gets triggered during composer script execution like `post-install-cmd` for example. 

The same happens not only for composer, also if you do a simple `php bin/console cache:clear`

ausi:
IMO it is a requirement that one of the PHP image extensions (GD, Imagick or Gmagick) exists. If none of them are installed the `contao/core-bundle` cannot be used.

Also `ext-gd` is [still a requirement](https://github.com/contao/core-bundle/blob/4176f0f00367245dde0b35e1a7d23ff2a8d91849/composer.json#L15) of `contao/core-bundle` so I’m not sure how `composer update` could work in the first place. But I may be misunderstanding the issue.

If this needs to be fixed, I think we should `return $class;` instead of throwing the exception in [AddImagineClassPass.php:61](https://github.com/contao/core-bundle/blob/4176f0f00367245dde0b35e1a7d23ff2a8d91849/src/DependencyInjection/Compiler/AddImagineClassPass.php#L61) so that `Imagine\Gd\Imagine` gets returned if no implementation is available.

Toflar:
I agree with @ausi. It's a requirement for Contao, so I don't think it's something we need to fix. Especially because we cannot add a `require` that represents "either gd, imagick or gmagick".

ausi:
> we cannot add a `require` that represents "either gd, imagick or gmagick".

I think this would be possible by creating three empty packages that require one of the image extensions and `provide` a virtual package that is than required by `contao/core-bundle`, but that would be too much overhead IMO and should – if at all – be done by the Imagine package itself. Related: composer/composer#751

leofeyer:
One of the PHP image extensions is required to **run** Contao but not to **install** it.

[...]

PR #1615, opened 2018-07-11
Allow to configure the imagine class

Toflar:
Right now you cannot configure the desired `Imagine` class. This has two major drawbacks:

1. If e.g. `Imagick` is installed on your system but it's broken somehow, you are lost. You cannot switch to `Gd` without adding another compiler pass.
2. If you add your own `Imagine` implementation you cannot use that without adding another compiler pass.

This PR allows you to configure the class very easily:

```yaml
contao:
    image:
        imagine_service: 'contao.image.imagine.gd'
```


src/DependencyInjection/Configuration.php
```diff
@@ -166,6 +171,31 @@ function (string $value): string {
...
+    {
+        static $magicks = ['Gmagick', 'Imagick'];
+
+        foreach ($magicks as $name) {
+            $class = 'Imagine\\'.$name.'\Imagine';
+
+            // Will throw an exception if the PHP implementation is not available
+            try {
+                new $class();
+            } catch (RuntimeException $e) {
```
m-vo: What's the benefit of catching instead of letting it fail?

bytehead: It will fall back to `return Imagine::class; // see #616`

src/DependencyInjection/Configuration.php
```diff
@@ -145,6 +147,9 @@ function (string $value): string {
                         ->arrayNode('imagine_options')
                             ->addDefaultsIfNotSet()
                             ->children()
+                                ->scalarNode('class')
```
ausi: The `imagine_options` array gets directly passed to `setImagineOptions()` https://github.com/contao/image/blob/dc9e2dc78533c6da1d7d0cf010b0b89b2b5ffd06/src/ResizeOptionsInterface.php#L29 so we should probably not add the class name here.

IMO it would be better to use `contao.image.imagine_class` instead.

aschempp: Agrred

aschempp:
@ausi 's suggestion looks reasonable to me

Toflar:
Nope, I disagree here. The compiler pass was wrong from day one that's why I changed it. The class you would like to use is a *configuration* value. It's exactly the right place to do so and it doesn't need any compiler pass at all. But yes, I can change the key to `contao.image.imagine_class`.

ausi:
But what if the custom Imagine class you want to use needs a constructor parameter or other injections?

IMO the services definition ist the right place for replacing services in your application.

aschempp:
> But what if the custom Imagine class you want to use needs a constructor parameter or other injections?

That means you would need dependency injection and not just set the class as an argument. A completely new (maybe valid) case. Maybe we should generate a service on the fly and assign it as argument to that method?

Toflar:
I have a better idea, give me a few minutes :)

Toflar:
Okay, so now you can configure any service you like. This will allow to define your own service in the container (that has other arguments) and configure it here. If you don't configure anything at all, the best solution is chosen for you and aliased.
For convenience, I've added `gd`, `imagick` and `gmagick` so it's easy to configure if you want to use one of these three, see the updated initial example.

src/DependencyInjection/Configuration.php
```diff
@@ -142,6 +142,10 @@ function (string $value): string {
                             ->prototype('scalar')->end()
                             ->defaultValue(['jpg', 'jpeg', 'gif', 'png', 'tif', 'tiff', 'bmp', 'svg', 'svgz'])
                         ->end()
+                        ->scalarNode('imagine_service')
+                            ->info('By default Contao tries to take the best Imagine service out of Gmagick, Imagick and Gd (in this order). If you would like to use a different service, configure the service id here.')
```
leofeyer: When will this be displayed?

Toflar: `config:dump-reference`  :)

leofeyer: Nice. However we don't have this for any other parameter, therefore I would prefer to remove this from this PR and to create a separate PR which adds descriptions for all parameters. WDYT?

Toflar: Nah, it belongs to this PR. Just add the info in another commit :)

leofeyer:
Ok. Let's see what @ausi says and then merge this.

ausi:
I still think that the services definition is the right place for replacing services in your application. We do that in Contao for example with the translator service. Every developer that is familiar with Symfony knows how to do that:

```yaml
services:
    contao.image.imagine:
        class: Imagine\Gd\Imagine
```

To get knowledge of the configuration value you need to read the documentation first.

My suggestion from https://github.com/contao/core-bundle/pull/1615#issuecomment-404230609 would solve your issues too, right?

Toflar:
I don't like that. I don't want to **configure** things in my `services.yml`. I already configure the imagine options in `contao.image`. So I would need to do things in 2 places.

ausi:
But how do you configure which translator class should be used be the Symfony translator for example, or any other service?

Toflar:
No idea, it's not part of my problem at the moment 😄 The way this PR implements the service you choose is the same as it's done in thousands of bundles. I don't see the issue here. There's always many ways to achieve the same thing but this one is the one I like best.

aschempp:
That is not the same to me. Overriding the application is one thing, setting configuration is a different thing.

Manually defining your Imagine class is a configuration that makes sense, because people might prefer one over the other. Therefore, there should be a configuration so set it. It's like settings the JPEG quality, you could do that using a compiler pass and change the service argument, but that's just unnecessarily cumbersome. We **want** people to configure their Imagine service, if they know which one they can use on their server.

leofeyer:
Thank you @Toflar.

PR #1618, opened 2018-07-13
Fixed the SQL query error when fetching members of multiple groups for search index rebuild

qzminski:
If you try to access the Maintenance backend module as a non-admin which has access to multiple member groups, you will run into the SQL query error:

```
Doctrine\DBAL\Exception\SyntaxErrorException: An exception occurred while executing 'SELECT id, username FROM tl_member WHERE (GROUPS LIKE '%"3"%' OR1"%' OR4"%') AND disable!='1' AND (start='' OR start<='1531467943') AND (stop='' OR stop>'1531468003') ORDER BY username':

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'OR1"%' OR4"%') AND disable!='1' AND (start='' OR start<='1531467943') AND (stop=' at line 1
```

This query is executed to fetch the available members that can be selected to rebuild the search index.

### DIFF
--- a/src/Resources/contao/classes/RebuildIndex.php
+++ b/src/Resources/contao/classes/RebuildIndex.php
@@ -66,7 +66,7 @@ class RebuildIndex extends \Backend implements \executable
 
 			if (!empty($amg) && \is_array($amg))
 			{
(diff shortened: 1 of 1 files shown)


leofeyer:
Thank you @qzminski.

PR #1619, opened 2018-07-13
Drop the unnecessary 2FA login route

aschempp:
This is what I meant in https://github.com/contao/core-bundle/pull/1545#issuecomment-404282149


leofeyer:
Thank you @aschempp.

PR #1621, opened 2018-07-16
[minor] SCA

kalessil:
Micro-optimizations and a case where array keys fragmentation (due to array_filter) could cause side-effects.


leofeyer:
Hey, thank you @kalessil. 👍 

I guess we should backport the array keys fragmentation fixes to Contao 4.4, shouldn't we @contao/developers? We cannot backport the null coalescing operator changes due to PHP 5.6 support though.

kalessil:
You are welcome @leofeyer =)

PR #1629, opened 2018-07-22
[RFC] Use the Intl extension to split words

ausi:
As discussed in contao/core#2769.

This is a work in progress.

One problem I encountered is that it is currently possible to search in pages with different languages but we have to know the language in advance to know how to split the serch term. How should we do that?

### DIFF
--- a/src/Resources/contao/library/Contao/Search.php
+++ b/src/Resources/contao/library/Contao/Search.php
@@ -237,35 +237,12 @@ class Search
 		$strText = preg_replace(array('/- /', '/ -/', "/' /", "/ '/", '/\. /', '/\.$/', '/: /', '/:$/', '/, /', '/,$/', '/[^\w\'.:+-]/u'), ' ', $strText);
 
 		// Split words
-		$arrWords = preg_split('/ +/', Utf8::strtolower($strText));
+		$arrWords = self::splitIntoWords(Utf8::strtolower($strText), $arrData['language']);
 		$arrIndex = array();
 
 		// Index words
 		foreach ($arrWords as $strWord)
 		{
-			// Strip a leading plus (see #4497)
-			if (strncmp($strWord, '+', 1) === 0)
-			{
-				$strWord = substr($strWord, 1);
-			}
-
-			$strWord = trim($strWord);
-
-			if (!\strlen($strWord) || preg_match('/^[\.:,\'_-]+$/', $strWord))
-			{
-				continue;
-			}
-
-			if (preg_match('/^[\':,]/', $strWord))
-			{
-				$strWord = substr($strWord, 1);
-			}
-
-			if (preg_match('/[\':,.]$/', $strWord))
-			{
-				$strWord = substr($strWord, 0, -1);
-			}
-
 			if (isset($arrIndex[$strWord]))
 			{
 				$arrIndex[$strWord]++;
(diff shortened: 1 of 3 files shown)


leofeyer:
You should probably base this on the 4.4 branch as it is a bugfix?

ausi:
Is it? I think it’s a feature: adding support for non-latin languages to the search. Also, the behaviour will probably change slightly so I would prefer to add it in 4.7.

aschempp:
> we have to know the language in advance to know how to split the serch term. How should we do that?

well we do know the page language, don't we?

ausi:
> well we do know the page language, don't we?

Yes, we do. But currently it is possible to search across muliple languages at once, but if splitting words depends on the language this is no longer possible this way.

aschempp:
You mean splitting on the input field, not on the page content to index?

ausi:
Yes. The search input has to be splitted the same way, otherwise many results wouldn’t be found.

aschempp:
I see. Well I guess we can only use the current page language …

src/Resources/contao/library/Contao/Search.php
```diff
@@ -233,39 +233,13 @@ public static function indexPage($arrData)
 
 		unset($arrSet);
 
-		// Remove special characters
-		$strText = preg_replace(array('/- /', '/ -/', "/' /", "/ '/", '/\. /', '/\.$/', '/: /', '/:$/', '/, /', '/,$/', '/[^\w\'.:+-]/u'), ' ', $strText);
-
 		// Split words
-		$arrWords = preg_split('/ +/', Utf8::strtolower($strText));
+		$arrWords = self::splitIntoWords(Utf8::strtolower($strText), $arrData['language']);
```
leofeyer: Should we make this `static::splitIntoWords()` and set the visibility of the method to `protected` so people can overwrite it if needed?

src/Resources/contao/library/Contao/Search.php
```diff
@@ -298,6 +272,27 @@ public static function indexPage($arrData)
 		return true;
 	}
 
+	/**
+	 * @return string[]
+	 */
+	private function splitIntoWords(string $strText, string $strLocale): array
```
leofeyer: CS: no return types in the legacy code.

src/Resources/contao/library/Contao/Search.php
```diff
@@ -298,6 +272,27 @@ public static function indexPage($arrData)
...
+	{
+		$iterator = \IntlBreakIterator::createWordInstance($strLocale);
+		$iterator->setText($strText);
+
+		$words = [];
```
leofeyer: CS: use the long array syntax.

src/Resources/contao/library/Contao/Search.php
```diff
@@ -349,23 +343,29 @@ public static function searchFor($strKeywords, $blnOrSearch=false, $arrPid=array
 				case '"':
 					if ($strKeyword = trim(substr($strKeyword, 1, -1)))
 					{
-						$arrPhrases[] = '[[:<:]]' . str_replace(array(' ', '*'), array('[^[:alnum:]]+', ''), $strKeyword) . '[[:>:]]';
+						$arrPhrases[] = str_replace(' ', '[^[:alnum:]]+', preg_quote($strKeyword));
```
leofeyer: Do we no longer need the word boundaries?

ausi: They don’t work for many cases like `"Google+"` or `"使用"`.
Also `"oogle+"` would still not match because the query first searches for `oogle` in `tl_search_index` before it does the regex search.

src/Resources/contao/library/Contao/Search.php
ausi: I’m not sure if this is needed. What is the use case for extending the `Search` class?

src/Resources/contao/library/Contao/Search.php
ausi: Fixed in 8e546398628310081e5f8adfb32a58ccda76face

src/Resources/contao/library/Contao/Search.php
ausi: Fixed in 8e546398628310081e5f8adfb32a58ccda76face

src/Resources/contao/library/Contao/Search.php
leofeyer: Ok. 👍

leofeyer:
Thank you @ausi.

PR #1630, opened 2018-07-23
[RTM] Make response with non-session cookies uncacheable

aschempp:
This disables http cache if a response has cookies.

### DIFF
--- a/src/EventListener/SessionListener.php
+++ b/src/EventListener/SessionListener.php
@@ -15,6 +15,7 @@ use Contao\CoreBundle\Routing\ScopeMatcher;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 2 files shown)


src/EventListener/SessionListener.php
```diff
@@ -112,4 +116,31 @@ public static function getSubscribedEvents()
...
+            }
+
+            if ($response->isCacheable()) {
+                $response
+                    ->setPrivate()
+                    ->setMaxAge(0)
+                    ->headers->addCacheControlDirective('must-revalidate');
+            }
+
+            break;
```
ausi: If the session cookie is the third cookie it would not get removed because of this line. Is that on purpose?

aschempp: You're absolutely right, fixed in add696b0246abf2d8134747d67f8ead3e197a9bd

aschempp:
I have also added unit tests and fixed minor bugs. This should be RTM now.

aschempp:
Done in c6980e2d5ff9bcfa14234a1d9289848f6ed8fbf3

leofeyer:
Thank you @aschempp.

PR #1631, opened 2018-07-23
[RFC] Check if service or alias exists before trying to make it public

sheeep:
The `lexik/LexikMaintenanceBundle` is only required by `require-dev`, which is therefore not installed when omitting the `--dev` in the installation process. Without a check in the`MakeServicePublicPass` the `cache:clear` command will fail with an exception.

```
You have requested a non-existent service "lexik_maintenance.driver.factory
```

By checking if the service exists beforehand, we can skip non existent services and aliases.


leofeyer:
Thank you @sheeep.

PR #1634, opened 2018-07-24
Fix phpdoc (missing brackets)

bytehead:


tests/Contao/ValidatorTest.php
```diff
@@ -37,7 +37,7 @@ public function testValidatesEmailAddresses(string $email, $expected): void
     }
 
     /**
-     * @return (string,int)[]
+     * @return array(string,int)[]
```
ausi: I think this should be `(string,int)[][]`?

tests/EventListener/StoreRefererListenerTest.php
```diff
@@ -60,7 +60,7 @@ public function testStoresTheReferer(Request $request, ?array $currentReferer, ?
     }
 
     /**
-     * @return array<string,(Request|array<string,string>[]|null)[]>
+     * @return array<string,(Request|array<string,string>|null)>[]
```
ausi: Shouldn’t this be `array<string,(Request|array|null)[]>`?

tests/Contao/ValidatorTest.php
bytehead: will change it to `(string|int)[][]`

leofeyer:
Thank you @bytehead.

PR #1635, opened 2018-07-24
Delete parent tag on create and edit

Toflar:
This is because a new entry (or an unpublished entry) might have not been shown and thus not tagged in the FE. We need to invalidate the direct parent to ensure these entries appear after changing them.

### DIFF
--- a/src/Resources/contao/drivers/DC_Table.php
+++ b/src/Resources/contao/drivers/DC_Table.php
@@ -1753,7 +1753,7 @@ class DC_Table extends DataContainer implements \listable, \editable
 							   ->execute($row[1]['sorting'], $row[0]['id']);
 
 				// Invalidate cache tags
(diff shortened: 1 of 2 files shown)


src/Resources/contao/drivers/DC_Table.php
```diff
@@ -1753,7 +1753,7 @@ public function move()
 							   ->execute($row[1]['sorting'], $row[0]['id']);
 
 				// Invalidate cache tags
-				$this->invalidateCacheTags($this->getCacheTags($this->strTable, array($row[1]['id'], $row[0]['id'])));
+				$this->invalidateCacheTags($this->getCacheTags($this->strTable, array($row[1]['id'], $row[0]['id']), (string) $this->ptable, (int) $this->activeRecord->pid));
```
leofeyer: Why did you cast the values? The `invalidateCacheTags()` method seems to handle numeric strings as parent ID and there are no type-hints that would enforce a certain type.

PR #1636, opened 2018-07-24
[RFC] Use PHPStan on level 2

bytehead:
PHPStan on level 2 needs to know about return types from `$container->get()` calls.

To extend the features of PHPSTan I've created an extension for PHPStan: [oneup/phpstan-contao](https://github.com/1up-lab/phpstan-contao).
This extension parses the `services.yml` file of the `contao/core-bundle` to provide the right return types for PHPStan.

Moreover this extension also requires `phpstan/phpstan-symfony` and provides the XML file containing the container definitions from Symfony to the Symfony extension for PHPStan.


leofeyer:
Excellent work. Thank you @bytehead.

PR #1643, opened 2018-07-26
[RFC] Include the src/Resource/contao in contao.resource_pathes

sheeep:
Using Symfony Flex you do not have an AppBundle anymore. Therefore it is impossible to load (for example) templates from `src/Resource/contao`. These changes will add the `src/Resource/contao` directory hard-coded to the `AddResourcePathsPass`, like `app/Resource/contao` is already.

### DIFF
--- a/src/DependencyInjection/Compiler/AddResourcesPathsPass.php
+++ b/src/DependencyInjection/Compiler/AddResourcesPathsPass.php
@@ -46,6 +46,10 @@ class AddResourcesPathsPass implements CompilerPassInterface
             $paths[] = $rootDir.'/app/Resources/contao';
         }
 
+        if (is_dir($rootDir.'/src/Resources/contao')) {
+            $paths[] = $rootDir.'/src/Resources/contao';
+        }
+
         return $paths;
     }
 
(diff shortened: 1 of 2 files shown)


leofeyer:
According to the Symfony best practices, you are not supposed to use a `src/(AppBundle)/Resources` folder but only the `app/Resources` folder:

https://symfony.com/doc/3.4/best_practices/templates.html#template-locations

leofeyer:
As briefly discussed, this is the best option since Symfony 4 does not have an `app` folder anymore.

leofeyer:
Thank you @sheeep.

aschempp:
I don't agree with this, why are the files not in the config folder? Or, following the concept of templates and config in root, why not a `contao` folder in the root? There is no `Resources` folder supposed to be in the `src` dir afaik?

aschempp:
BTW I changed my mind 😇 
https://stackoverflow.com/questions/45239350/how-can-i-override-resources-for-third-party-bundles-in-symfony-4

leofeyer:
Good find. 👍

Issue #222, opened 2015-04-17
Illegal mix of collations


xchs:
The Contao 4.0.0-beta1 shows an exception when creating a new folder:
```
An exception occurred while executing 'SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid UNION ALL SELECT UNHEX(REPLACE(UUID(), '-', '')) AS uuid':

SQLSTATE[HY000]: General error: 1270 Illegal mix of collations (utf8_general_ci,COERCIBLE), (utf8_unicode_ci,COERCIBLE), (utf8_unicode_ci,COERCIBLE) for operation 'replace'
```

![collation_mix](https://cloud.githubusercontent.com/assets/1218766/7210601/95f1b198-e553-11e4-9740-db6e8dae5035.png)

The database collation is set to `utf8_general_ci`.

leofeyer:
I cannot reproduce the issue here.

xchs:
Never mind, I have now tried the latest develop version and could not reproduce the issue again. Sorry for any inconveniences.

PR #1649, opened 2018-07-30
Adjust the icon and styles for the 2FA menu entry

xchs:
The styles for the mobile view of the 2FA menu entry were missing:

contao_be_mobile

This PR adds the corresponding styles and adjustments.

However, the anchor text for the 2FA menu entry ("_Two-factor authentication_") is still a bit too long for the mobile view. Maybe we should only show "_2FA_" in the mobile view?

contao_be_2fa

[Contao 4.6.x-dev]

### DIFF
--- a/src/Resources/contao/themes/flexible/icons/shield.svg
+++ b/src/Resources/contao/themes/flexible/icons/shield.svg
@@ -1,1 +1,1 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24"><path d="M12 22s8-4 8-10V5l-8-3-8 3v7c0 6 8 10 8 10z" fill="none" stroke="#91979c" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>
+<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24"><path d="M12 22s8-4 8-10V5l-8-3-8 3v7c0 6 8 10 8 10z" fill="none" stroke="#fff" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>


src/Resources/contao/themes/flexible/icons/shield.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24"><path d="M12 22s8-4 8-10V5l-8-3-8 3v7c0 6 8 10 8 10z" fill="none" stroke="#91979c" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" viewBox="0 0 24 24"><path d="M12 22s8-4 8-10V5l-8-3-8 3v7c0 6 8 10 8 10z" fill="none" stroke="#fff" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></svg>
```
bytehead: Good catch! 👍

leofeyer: The fix is wrong though. The stroke color needs to be `#91979c` in the regular menu and `#fff` in the mobile menu, therefore we need two separate icons (similar to `profile.svg` and `profile_dark.svg`).

xchs: There're already two separate icons. See the dark version of the shield icon: https://github.com/contao/core-bundle/edit/master/src/Resources/contao/themes/flexible/icons/shield_dark.svg

All other dark icons also have the color `#222`:
* `exit_dark.svg`: https://github.com/contao/core-bundle/edit/4.6/src/Resources/contao/themes/flexible/icons/exit_dark.svg
* `profile_dark.svg`: https://github.com/contao/core-bundle/edit/4.6/src/Resources/contao/themes/flexible/icons/profile_dark.svg

leofeyer: You are right, sorry.

xchs:
However, the problem with the too long text label is still open:

> However, the anchor text for the 2FA menu entry ("Two-factor authentication") is still a bit too long for the mobile view. Maybe we should only show "2FA" in the mobile view?

How do we address this?

leofeyer:
See #1661 for the other problem ("two-factor authentication" too long).

PR #1652, opened 2018-08-06
Services must be public to be used in hooks

aschempp:
see https://github.com/contao/managed-edition/issues/38

### DIFF
--- a/src/DependencyInjection/Compiler/RegisterHookListenersPass.php
+++ b/src/DependencyInjection/Compiler/RegisterHookListenersPass.php
@@ -54,6 +54,8 @@ class RegisterHookListenersPass implements CompilerPassInterface
             foreach ($tags as $attributes) {
                 $this->addHookCallback($hooks, $serviceId, $attributes);
             }
+
+            $container->findDefinition($serviceId)->setPublic(true);
         }
 
         return $hooks;
--- a/src/DependencyInjection/Compiler/RegisterHookListenersPass.php
+++ b/src/DependencyInjection/Compiler/RegisterHookListenersPass.php
@@ -51,6 +51,10 @@ class RegisterHookListenersPass implements CompilerPassInterface
         $serviceIds = $container->findTaggedServiceIds('contao.hook');
 
         foreach ($serviceIds as $serviceId => $tags) {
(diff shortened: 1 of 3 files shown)


src/Resources/contao/library/Contao/System.php
```diff
@@ -164,6 +165,10 @@ protected function import($strClass, $strKey=null, $blnForce=false)
 			{
 				$this->arrObjects[$strKey] = $container->get($strClass);
 			}
+			elseif (strpos($strClass, '.') !== false)
+			{
+				throw new ServiceNotFoundException($strClass, null, null, [], sprintf('The service "%s" was not found or is not public. See https://symfony.com/doc/current/service_container.html#public-versus-private-services', $strClass));
```
leofeyer: You also have to add `@throws ServiceNotFoundException` to the phpDoc.

src/DependencyInjection/Compiler/RegisterHookListenersPass.php
```diff
@@ -54,6 +54,8 @@ private function getHooks(ContainerBuilder $container): array
             foreach ($tags as $attributes) {
                 $this->addHookCallback($hooks, $serviceId, $attributes);
             }
+
+            $container->findDefinition($serviceId)->setPublic(true);
```
leofeyer: Are you aware that `findDefinition()` no longer includes aliases? We therefore check services and aliases separately in the [MakeServicesPublicPass](https://github.com/contao/core-bundle/blob/4.6/src/DependencyInjection/Compiler/MakeServicesPublicPass.php) class.

aschempp: I am not aware of that… But can an alias be tagged and would it show up with the original service name or the alias name in `findTaggedServiceIds`?

leofeyer: I don't know.

aschempp: Fixed in a8fbc7f

leofeyer:
Thank you @aschempp.

PR #1658, opened 2018-08-09
Article section field is mandatory

aschempp:


leofeyer:
We should also add `'includeBlankOption'=>true` if the field is mandatory.

aschempp:
No I don't think so, there is a default value on the field. The mandatory is only useful to have correct accessibility and the red asterisk.

leofeyer:
Thank you @aschempp.

PR #1668, opened 2018-08-13
Add removeField to the palette manipulator

dmolineus:
This PR adds the ability to remove fields using the palette manipulator.

Implemented behaviour:

- The removal task can be limited to a parent (=legend). So the field would be only removed if it's in a specific legend. Useful if you want to "move" a field to another legend.
- The fields are removed after fields are added. Refer to a removed field as parent keep working in `addField`.

Removing a whole legend does not make sense to me, so it's not part of the PR.

### DIFF
--- a/src/DataContainer/PaletteManipulator.php
+++ b/src/DataContainer/PaletteManipulator.php
@@ -93,7 +93,7 @@ class PaletteManipulator
     }
 
     /**
-     * If $parent is empty the field is removed everythere.
+     * Parent refers to the legend. If $parent is empty the field is removed everywhere.
      *
      * @param string|array               $name
      * @param string|array               $parent
--- a/src/DataContainer/PaletteManipulator.php
+++ b/src/DataContainer/PaletteManipulator.php
@@ -33,6 +33,11 @@ class PaletteManipulator
      */
     private $fields = [];
 
+    /**
+     * @var array
+     */
+    private $removes = [];
+
     public static function create(): self
     {
         return new static();
@@ -88,6 +93,22 @@ class PaletteManipulator
     }
 
     /**
+     * If $parent is empty the field is removed everythere.
+     *
+     * @param string|array               $name
+     * @param string|array               $parent
+     */
+    public function removeField($name, $parent = []): self
+    {
+        $this->removes[] = [
+            'fields' => (array) $name,
+            'parents' => (array) $parent
+        ];
+
+        return $this;
+    }
+
+    /**
      * @param string $name
      */
     public function applyToPalette($name, string $table): self
@@ -133,6 +154,10 @@ class PaletteManipulator
 
         foreach ($this->fields as $field) {
             $this->applyField($config, $field, $skipLegends);
+        }
+
+        foreach ($this->removes as $remove) {
+            $this->applyRemove($config, $remove);
         }
 
         return $this->implode($config);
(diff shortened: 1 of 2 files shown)


src/DataContainer/PaletteManipulator.php
```diff
@@ -87,6 +92,22 @@ public function addField($name, $parent, string $position = self::POSITION_AFTER
         return $this;
     }
 
+    /**
+     * If $parent is empty the field is removed everythere.
```
Toflar: `everywhere`. Also, can you add a description on what `$parent` should represent?

dmolineus: Done with https://github.com/contao/core-bundle/pull/1668/commits/ea46a4996a5b91fd5d9c963608bade3021f91341

src/DataContainer/PaletteManipulator.php
```diff
@@ -87,6 +92,22 @@ public function addField($name, $parent, string $position = self::POSITION_AFTER
         return $this;
     }
 
+    /**
+     * Parent refers to the legend. If $parent is empty the field is removed everywhere.
+     *
+     * @param string|array               $name
+     * @param string|array               $parent
+     */
+    public function removeField($name, $parent = []): self
```
aschempp: To be consistent, `$parent` should not have a default value

dmolineus: That doesn't really make sense to me. When adding a field, you always want to have a parent. Maybe a legend or another field depending on the position value.

The logic behind removing a field is different. Furhtermore as a parent there is only a legend accepted. Removing a field without knowing the legend makes sense to me. But I'd happy with renaming the param to `$legend` instead to make it more explicit.

aschempp: You're right. Shouldn't the default value then be `null`?

dmolineus: As it's converted to an array later on I already used an empty array as default value.

https://github.com/contao/core-bundle/blob/c120c7a76069927a80a1afc1061af7c9e1763cc4/src/DataContainer/PaletteManipulator.php#L105

But I don't mind to change it. It's imho a matter of taste anyway.

aschempp: I agree, I don't really like any of the options 😄 
Thinking of it, do we really need an option to actually set a palette? I mean would you ever want to a remove field if it's in palette A but not if it's in palette B? That sounds very strange to me. So maybe we can just drop that argument?

dmolineus: As I explained in the opening post, it's required if I want to move a field from one legend to another (Yeah, it's a real project use case) without creating two instances of the palette manipulator, f.e.:

```php
PaletteManipulator::create()
    ->removeField('published', 'publish_legend')
    ->addField('published', 'title_legend', PaletteManipulator::POSITION_APPEND);
```

aschempp: I see. Does it ever make sense to have the same field in multiple legends? Like if you add it twice, should the palette manipulator ignore that or should the later win?

dmolineus: No, it doesn't make sense to add a field twice and probably causes afair some issues.It could happen when multiple extensions change the dca. 

Imho there's no need to handle such cases. ATM the palette manipulator has a low level implementation as it does not set any constraints. F.e. you could argue that adding a field which does not exist in the dca shouldn't be allowed as well. Such constraints does not exist so far and I think it's not the purpose of the manipulator.

asaage: Isn't that the case with tl_content singleSrc for example?
If the type is "Hyperlink" the legend is "Image link settings" - if the type is "Image" the legend is"Files and folders".
You might want to remove one but keep the other or am i mixing up paletes and legends now 😕 ?.

aschempp: That would be separate palettes then, which is a different case and totally supported.

leofeyer:
Merged in contao/contao@1e49b07ce00580138d50253644e9f67a9882ada1. Thank you @dmolineus.

As you have discussed, I have renamed `$parent` to `$legend` and made it a string, as you cannot have the same field in multiple field sets.

PR #1673, opened 2018-08-14
Fixed CORS headers not set correctly when rebuilding the search index

Toflar:
I noticed that rebuilding the search index does not work across domains anymore.
This PR fixes this issue.
The main reason being that `$this->connection->isConnected()` will always return `false` if nobody ever tried to connect before. `$this->connection->isConnected()` is lazy.


leofeyer:
Merged in contao/contao@9f077d44204cfdfd8a904f695e2d39169a822f2b. Thank you @Toflar.

PR #1693, opened 2019-06-09
Prevent message-stack from being consumed before reload

martinschumann:
Messages get lost e.g. in the backend user profile when purging session and cache data, when `Contao\Message::generate()` is called with it's underlying flash message in post submits with subsequent reload.

In Contao 3 excluding post submits from message consumption resides in `Contao\Message::generate()`.


aschempp:
Please create pull requests to `contao/contao` and not the split repositories.

leofeyer:
@martinschumann If you do not want to recreate the PR, I can adopt the changes directly in the correct branch and repo. Otherwise, please use `empty($_POST)` instead of `(\Environment::get('requestMethod') !== 'POST'`.

martinschumann:
I'll recreate PR. Even just to practice …

PR #1697, opened 2019-08-21
hotfix change chaching description

zonky2:
hotfix change chaching description
delete "seconds"


zonky2:
=> move https://github.com/contao/contao/pull/654

PR #1709, opened 2019-11-22
Fix 'undefined index' PHP notices

sebj54:
I implemented some unit tests to my Contao app and some of the tests fail because of the `Undefined Index` PHP notices.

Some of my changes are already implemented in newer versions of Contao.


aschempp:
It's a known limitation of the old Contao code that there are `Undefined Index` notices. I think we agreed not to change that years ago. Contao automatically disabled the errors messages on boot time.

Also, please only make pull requests to https://github.com/contao/contao, this is a read-only repository split.

Issue #1710, opened 2019-11-22
[BUG] BackendAccessVoter fails with fatal error when using `allow_if` security access rules - Might not happen in prod enviornment


AndreasA:
**Affected version(s)**
4.8.5 and probably others

**Description**  
The backend access voter uses `strncmp` in `\Contao\CoreBundle\Security\Voter\BackendAccessVoter::supports`

**How to reproduce**  
Create a security rule like:
```
- { path: "^/my/route", allow_if: '!is_granted("ROLE_NO_ROUTE_ACCESS")' }
```

Now the issue is that if the expression voter does not return true it will end up at with the `BackendAccessVoter` which will fail with a fatal error as there is no string provided in supports and voteOnAttribute but a `Symfony\Component\ExpressionLanguage\Expression`

The easiest solution is probably to just cast the attribute to string before using it in strncmp or check if it iis a string beforehand.

leofeyer:
I am closing this in favor of contao/contao#1013.

PR #1711, opened 2019-11-22
BUGFIX: BackendAccessVoter ensure string when checking for supported attribute

AndreasA:
Ensure that the e.g. ExpressionLanguage objects are ignored when checking for support in the BackendAccessVoter.

Fixes: #1710

### DIFF
--- a/src/Security/Voter/BackendAccessVoter.php
+++ b/src/Security/Voter/BackendAccessVoter.php
@@ -23,7 +23,7 @@ class BackendAccessVoter extends Voter
      */
     protected function supports($attribute, $subject): bool
     {
-        return 0 === strncmp($attribute, 'contao_user.', 12);
+        return is_string($attribute) && 0 === strncmp($attribute, 'contao_user.', 12);
     }
 
     /**


Toflar:
Thanks for contributing! LGTM, however, you should open the PR on our monorepository contao/contao. Could you also add a unit test, please? :)

aschempp:
I'm not sure this is a bug in Contao. Voter attributes should always be strings, the argument is even enforced in the latest Symfony version: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/Authorization/Voter/Voter.php#L57

AndreasA:
Hmm.. not according to this:
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/Authorization/Voter/ExpressionVoter.php

But maybe there is a bug in Symfony core that goes beyond just the PHPDoc now

AndreasA:
I have created a symfony bug report too: https://github.com/symfony/symfony/issues/34532

AndreasA:
I have created the corresponding PR https://github.com/contao/contao/pull/1013/files

PR #1712, opened 2019-11-27
Fix backend system font stack for IE 11

loilo:
The native font stack introduced to the Contao backend in 67b8e7d breaks in Internet Explorer 11, resulting in the browser default font to be loaded.

Contao backend with serif font in IE 11

The cause is the fonts being part of a `font` shorthand property (which seems to be [a known issue](https://booking.design/implementing-system-fonts-on-booking-com-a-lesson-learned-bdc984df627f#7134)).

The PR fixes this behavior by moving the font face out of the `font` shorthand property into its own `font-family`.


bytehead:
This is the wrong repository (read only), PR should be done at https://github.com/contao/contao :)

loilo:
Oh. I thought this only was the case with issues. Well that makes way more sense now I guess. 🙈

So should I just open up the PR over there?

bytehead:
No worries!
Yes please.

loilo:
Out of curiosity: Why? Is this a split-off for updating Composer?

bytehead:
Afaik to use the `contao/core-bundle` as a single bundle in your Symfony application instead of the whole `contao/contao` collection.

loilo:
Sounds reasonable. Alright, brb with the right PR in the right place. 😁

bytehead:
Base your pull request against the `4.8` branch at `contao/contao`.

PR #1717, opened 2020-09-30
Check for known options in serialized values of a widget

martinschumann:
Fixes #b1d573c which does not check serialized values for known options


leofeyer:
First of all, you have created your pull request in the wrong repository, as this is a read-only split repository. Second of all, there already is a pull request to fix the issue: https://github.com/contao/contao/pull/2360

martinschumann:
Excuse me, I'll study that more carefully.

PR #1718, opened 2020-10-17
Update main.css

BennyK80:
makes the progress bar animation smooth ;-)


ausi:
You have created your pull request in the wrong repository, as this is a read-only split repository. Please create the pull request in https://github.com/contao/contao

I can confirm that this change fixes a “jumping” effect in the animation, but I think we should change the duration of the animation to something like `0.3s` to keep a similar speed of the animation.

PR #1728, opened 2024-01-31
task | add foreignKey declaration for form and module fields

rorych:
As asked in the slack support channel, there seems to be no reason why the tl_content fields form and module should not have a relation/foreignKey declaration. For the use of "getRelated" or other logic following the "foreignKey" DCA standards there should be a declaration in my opinion. Hence the pull request.


leofeyer:
Thank you for your contribution.
  
Unfortunately, you have created your pull request in the wrong repository, as this is a read-only split repository. Please go to https://github.com/contao/contao and create your pull request there.

